PR #10, opened 2024-07-01
Include security check and coverage info in gh workflow

dimkouv:
Enable `gosec` for security-checks, enable `unparam` and display coverage info in a PR comment.

> e2e pipeline is kept ~40sec

### DIFF
--- a/internal/libs/testhelpers/ocr3runner.go
+++ b/internal/libs/testhelpers/ocr3runner.go
@@ -2,15 +2,16 @@ package testhelpers
 
 import (
 	"context"
+	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math/big"
 
 	"github.com/smartcontractkit/libocr/commontypes"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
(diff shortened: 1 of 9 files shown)


internal/libs/testhelpers/ocr3runner.go
```diff
@@ -170,7 +171,15 @@ func (r *OCR3Runner[RI]) selectLeader() ocr3types.ReportingPlugin[RI] {
 	if numNodes == 0 {
 		return nil
 	}
-	return r.nodes[rand.Intn(numNodes)]
+
+	idx, err := rand.Int(rand.Reader, big.NewInt(int64(numNodes)))
+	if err != nil {
+		panic(err)
+	}
+	if !idx.IsInt64() {
```
asoliman92, line 179: Is this really needed after using `big.NewInt(int64(numNodes))` above? This is something for a test maybe?

asoliman92 (approved):
Minor comment. Not a blocker.

internal/libs/testhelpers/ocr3runner.go
dimkouv: it's not really needed, i just added it as a defensive check, not a big deal imo

makramkd (approved):

PR #13, opened 2024-07-01
Execute plugin report generation.

winder:
Introduces logic to select messages to execute and build the execution report for transmission.

Migration of [ccip#1091](https://github.com/smartcontractkit/ccip/pull/1091).

### DIFF
--- a/execute/errors.go
+++ b/execute/errors.go
@@ -2,4 +2,4 @@ package execute
 
 import "errors"
 
-var errNothingExecuted = errors.New("no reports were executed")
+var errEmptyReport = errors.New("no messages can fit in the report")
--- a/execute/errors.go
+++ b/execute/errors.go
@@ -1,0 +1,5 @@
+package execute
+
+import "errors"
+
+var errNothingExecuted = errors.New("no reports were executed")
--- a/execute/factory.go
+++ b/execute/factory.go
@@ -47,9 +47,11 @@ func (p PluginFactory) NewReportingPlugin(
 	config ocr3types.ReportingPluginConfig,
 ) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
 	return NewPlugin(
-		context.Background(),
 		config,
 		cciptypes.ExecutePluginConfig{},
+		nil,
+		nil,
+		nil,
 		nil,
 	), ocr3types.ReportingPluginInfo{}, nil
 }
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -276,7 +276,7 @@ func buildSingleChainReportMaxSize(
 		return cciptypes.ExecutePluginReportSingleChain{},
 			0,
 			cciptypes.ExecutePluginCommitDataWithMessages{},
-			errNothingExecuted
+			errEmptyReport
 	}
 
 	report = markNewMessagesExecuted(finalReport, report)
@@ -421,7 +421,7 @@ func selectReport(
 			buildSingleChainReportMaxSize(ctx, lggr, hasher, tokenDataReader, encoder,
 				report, maxReportSizeBytes-accumulatedSize)
 		// No messages fit into the report, stop adding more reports.
-		if errors.Is(err, errNothingExecuted) {
+		if errors.Is(err, errEmptyReport) {
 			break
 		}
 		if err != nil {
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -317,7 +317,7 @@ func buildSingleChainReport(
 	tree, err := constructMerkleTree(ctx, hasher, report)
 	if err != nil {
 		return cciptypes.ExecutePluginReportSingleChain{}, 0,
-			fmt.Errorf("unable to construct merkle tree from messages: %w", err)
+			fmt.Errorf("unable to construct merkle tree from messages for report (%s): %w", report.MerkleRoot.String(), err)
 	}
 	numMsgs := len(report.Messages)
 
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -1,7 +1,9 @@
 package execute
 
 import (
+	"bytes"
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"slices"
@@ -19,6 +21,7 @@ import (
 
 // maxReportSizeBytes that should be returned as an execution report payload.
 const maxReportSizeBytes = 250_000
+const SentinelRoot = "0x289502ebf164ea87871e19cd9c8734016e954539fffc8c789ac4dadc7efcd933"
 
 // Plugin implements the main ocr3 plugin logic.
 type Plugin struct {
@@ -351,6 +354,16 @@ func buildSingleChainReport(
 			offchainTokenData = append(offchainTokenData, tokenData)
 			toExecute = append(toExecute, i)
 			msgInRoot = append(msgInRoot, msg)
+		}
+	}
+
+	hash := tree.Root()
+	if !bytes.Equal(hash[:], report.MerkleRoot[:]) {
+		// For testing purposees, we allow a sentinel root to be used.
+		if report.MerkleRoot.String() != SentinelRoot {
+			actualStr := "0x" + hex.EncodeToString(hash[:])
+			return cciptypes.ExecutePluginReportSingleChain{}, 0,
+				fmt.Errorf("merkle root mismatch: expected %s, got %s", report.MerkleRoot.String(), actualStr)
 		}
 	}
 
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -311,7 +311,7 @@ func buildSingleChainReport(
 	lggr.Debugw(
 		"constructing merkle tree",
 		"sourceChain", report.SourceChain,
-		"expected-root", report.MerkleRoot.String(),
+		"expectedRoot", report.MerkleRoot.String(),
 		"treeLeaves", len(report.Messages))
 
 	tree, err := constructMerkleTree(ctx, hasher, report)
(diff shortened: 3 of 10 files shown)


execute/plugin_test.go
```diff
@@ -166,3 +155,496 @@ func Test_getPendingExecutedReports(t *testing.T) {
...
+		},
+	})
+	return commitReport
+}
+
+func makeTestCommitReport(
+	numMessages, srcChain, firstSeqNum, block int, timestamp int64, executed []cciptypes.SeqNum,
+) cciptypes.ExecutePluginCommitDataWithMessages {
+	for _, e := range executed {
+		if e < cciptypes.SeqNum(firstSeqNum) || e > cciptypes.SeqNum(firstSeqNum+numMessages-1) {
```
makramkd, line 195: nit: extract `firstSeqNum + numMessages - 1` into a const/var since its used elsewhere in this func?

```golang
const lastSeqNum = firstSeqNum + numMessages - 1
```

execute/plugin_test.go
```diff
@@ -166,3 +155,496 @@ func Test_getPendingExecutedReports(t *testing.T) {
...
+			panic("executed message out of range")
```
makramkd, line 196: nit: I wonder if its nicer for test output to have this be in a `require.True` instead of doing a `panic`, e.g

```golang
inRange := e >= cciptypes.SeqNum(firstSeqNum) && e <= cciptypes.SeqNum(firstSeqNum+numMessages-1)
require.True(t, inRange)
```

execute/plugin_test.go
```diff
@@ -166,3 +155,496 @@ func Test_getPendingExecutedReports(t *testing.T) {
...
+}
+
+func breakCommitReport(
+	commitReport cciptypes.ExecutePluginCommitDataWithMessages,
+) cciptypes.ExecutePluginCommitDataWithMessages {
+	commitReport.Messages = append(commitReport.Messages, cciptypes.CCIPMsg{
+		CCIPMsgBaseDetails: cciptypes.CCIPMsgBaseDetails{
+			ID:          mustRandID(),
+			SourceChain: cciptypes.ChainSelector(1),
+			SeqNum:      cciptypes.SeqNum(999),
```
makramkd, line 184: nit: Would it make more sense if this isn't hardcoded and is passed in instead?

execute/plugin_test.go
```diff
@@ -166,3 +155,496 @@ func Test_getPendingExecutedReports(t *testing.T) {
...
+	for _, msg := range commitReport.Messages {
+		hash, err := hasher.Hash(context.Background(), msg)
+		require.NoError(t, err)
+		leafHashes = append(leafHashes, hash)
+	}
+	tree, err := merklemulti.NewTree(keccak, leafHashes)
+	require.NoError(t, err)
+	merkleRoot := tree.Root()
+
+	// Generate merkle root from exec report messages and proofj
```
makramkd, line 247:
```suggestion
	// Generate merkle root from exec report messages and proof
```

execute/plugin_test.go
```diff
@@ -166,3 +155,496 @@ func Test_getPendingExecutedReports(t *testing.T) {
...
+	for _, msg := range execReport.Messages {
+		hash, err := hasher.Hash(ctx, msg)
+		require.NoError(t, err)
+		leaves = append(leaves, hash)
+	}
+	proofCast := make([][32]byte, len(execReport.Proofs))
+	for i, p := range execReport.Proofs {
+		copy(proofCast[i][:], p[:32])
+	}
+	var proof merklemulti.Proof[[32]byte]
```
makramkd, line 259: Ah this is interesting, the generics won't work with the `cciptypes.Bytes32` type?

execute/plugin.go
```diff
@@ -199,6 +216,248 @@ func (p *Plugin) ObservationQuorum(outctx ocr3types.OutcomeContext, query types.
...
+
+	tree, err := constructMerkleTree(ctx, hasher, report)
+	if err != nil {
+		return cciptypes.ExecutePluginReportSingleChain{}, 0,
+			fmt.Errorf("unable to construct merkle tree from messages: %w", err)
+	}
+	lggr.Debugw(
+		"constructing merkle tree",
+		"sourceChain", report.SourceChain,
+		"treeLeaves", len(report.Messages))
```
makramkd, line 316: Nit: Maybe move this log above the `tree, err := ...` line? Or should this be `constructed merkle tree`?

Also, should we log the merkle root here for easy searching?

execute/plugin.go
```diff
@@ -199,6 +216,248 @@ func (p *Plugin) ObservationQuorum(outctx ocr3types.OutcomeContext, query types.
...
+	var msgInRoot []cciptypes.CCIPMsg
+	executedIdx := 0
+	for i := 0; i < numMsgs && len(toExecute) <= maxMessages; i++ {
+		seqNum := report.SequenceNumberRange.Start() + cciptypes.SeqNum(i)
+		// Skip messages which are already executed
+		if executedIdx < len(report.ExecutedMessages) && report.ExecutedMessages[executedIdx] == seqNum {
+			executedIdx++
+		} else {
+			msg := report.Messages[i]
+			tokenData, err := tokenDataReader.ReadTokenData(context.Background(), report.SourceChain, msg.SeqNum)
```
makramkd, line 331: In addition to the readiness of the token data, there are things like e.g gas limit that determine whether we can include a message in a batch. Do we have tickets for that?

execute/factory.go
```diff
@@ -47,10 +47,12 @@ func (p PluginFactory) NewReportingPlugin(
...
 ) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
 	return NewPlugin(
-		context.Background(),
 		config,
 		cciptypes.ExecutePluginConfig{},
 		nil,
+		nil,
+		nil,
+		nil,
 	), ocr3types.ReportingPluginInfo{}, nil
```
makramkd, line 56: I think you should at least specify the max report size here in the `Limits` field (declared in plugin.go as a const)

execute/errors.go
```diff
@@ -0,0 +1,5 @@
+package execute
+
+import "errors"
+
+var errNothingExecuted = errors.New("no reports were executed")
```
makramkd, line 5: nit: this name is kind of misleading, it seems to be returned when no messages fit into the report. If I see this in the log I'd think its some sort of transmission error or related. Perhaps `errEmptyReport`?

```golang
var errEmptyReport = errors.New("no messages can fit in the report")
```

execute/plugin.go
```diff
@@ -199,6 +216,248 @@ func (p *Plugin) ObservationQuorum(outctx ocr3types.OutcomeContext, query types.
...
+			fmt.Errorf("unable to build a single chain report (max): %w", err)
+	}
+
+	// return fully executed report
+	if encodedSize <= maxSizeBytes {
+		report = markNewMessagesExecuted(finalReport, report)
+		return finalReport, encodedSize, report, nil
+	}
+
+	var searchErr error
```
makramkd, line 251: Do we care about accumulating these `searchErr` instead of overwriting them? i.e using `multierr.Append` instead?

execute/plugin.go
```diff
@@ -2,42 +2,59 @@ package execute
...
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 
-	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/chainlink-ccip/internal/libs/slicelib"
 )
 
+// maxReportSizeBytes that should be returned as an execution report payload.
+const maxReportSizeBytes = 250_000
```
makramkd, line 21: I think this is too big if this report will be transmitted onchain, this is 250 kilobytes which is >> the max size of a non-empty ethereum transaction of 46kB (although this is unlikely, for it to be all nonzero bytes).

execute/plugin.go
winder: The errors aren't overwritten, processing is short circuited if there's an error and the first one is returned.

execute/plugin.go
winder: This value was selected because it's the same one v1 uses. Should I change it to 46_000?

execute/plugin_test.go
winder: I'll add a comment. It could probably just be an empty object. In this case it's breaking the commit report by adding an extra message to the report, the len(messages) != len(seqNumRange) check fails.

execute/plugin_test.go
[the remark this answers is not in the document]
winder: numMessages is the most important parameter, so a conversion is needed either way.

The panic message makes it pretty clear whats going on so I slightly prefer the terseness of the  current code.
```
	numRange := cciptypes.NewSeqNumRange(cciptypes.SeqNum(firstSeqNum), cciptypes.SeqNum(firstSeqNum+numMessages))
	for _, e := range executed {
		if !numRange.Contains(e) {
			panic("executed message out of range")
		}
	}
```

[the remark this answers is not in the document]
winder: edit: nevermind! The num range is used below as you also pointed out. The SeqNumRange object is needed either way.

execute/plugin_test.go
winder: Maybe. I went with a panic since this is only used when setting up the test cases.

execute/plugin.go
winder: Thanks, that one wasn't on my radar. [Ticket created.](https://smartcontract-it.atlassian.net/browse/CCIP-2684)

execute/factory.go
winder: Is this something in the HomeChain? I guess it could vary per destination.

makramkd: Maybe they'd vary by family for the reports, outcomes are fully off-chain though so you could set them here family-agnostically

makramkd (approved):

execute/plugin.go
winder: Updated the logging and realized that I wasn't comparing the computed root to the expected root.

PR #14, opened 2024-07-02
Report root staleness checks

dimkouv:
Check in transmission protocol if report seq nums match the on-chain state. If not reject report transmission.


commit/plugin_functions.go
```diff
@@ -585,6 +585,50 @@ func validateObservedGasPrices(gasPrices []cciptypes.GasPriceChain) error {
...
+		return true, nil
+	}
+
+	onchainSeqNums, err := reader.NextSeqNum(ctx, reportChains)
+	if err != nil {
+		return false, fmt.Errorf("get next sequence numbers: %w", err)
+	}
+
+	for i, onchainSeqNum := range onchainSeqNums {
+		if i >= len(reportChains) {
```
asoliman92, line 612: check before loop with `len(onchainSeqNums)`

commit/plugin_functions_test.go
```diff
@@ -1180,3 +1180,80 @@ func Test_gasPricesConsensus(t *testing.T) {
...
+		},
+		{
+			name: "one root is stale",
+			reportSeqNums: []cciptypes.SeqNumChain{
+				cciptypes.NewSeqNumChain(10, 100),
+				cciptypes.NewSeqNumChain(20, 200),
+			},
+			onchainSeqNums: []cciptypes.SeqNum{99, 200}, // <- 200 is already on chain
+			expValid:       false,
+			expErr:         false,
```
asoliman92, line 1210: expErr should be true here, right?

commit/plugin_e2e_test.go
```diff
@@ -324,6 +324,9 @@ func setupAllNodesReadAllChains(ctx context.Context, t *testing.T, lggr logger.L
 		n.ccipReader.On("NextSeqNum", ctx, []cciptypes.ChainSelector{chainA, chainB}).
 			Return([]cciptypes.SeqNum{10, 20}, nil)
 
+		// transmission phase root staleness check passes
+		n.ccipReader.On("NextSeqNum", ctx, []cciptypes.ChainSelector{chainB}).
+			Return([]cciptypes.SeqNum{20}, nil)
```
asoliman92, line 329: Shouldn't it be  `21` to be valid?
```suggestion
			Return([]cciptypes.SeqNum{21}, nil)
```

Also any reason for not adding chainA?

```
		n.ccipReader.On("NextSeqNum", ctx, []cciptypes.ChainSelector{chainA}).
			Return([]cciptypes.SeqNum{11}, nil)
```

commit/plugin_functions.go
```diff
@@ -585,6 +585,50 @@ func validateObservedGasPrices(gasPrices []cciptypes.GasPriceChain) error {
...
+	for _, mr := range report.MerkleRoots {
+		reportChains = append(reportChains, mr.ChainSel)
+		reportMinSeqNums[mr.ChainSel] = mr.SeqNumsRange.Start()
+	}
+
+	if len(reportChains) == 0 {
+		return true, nil
+	}
+
+	onchainSeqNums, err := reader.NextSeqNum(ctx, reportChains)
```
makramkd, line 606: Lets disambiguate and make it clear which sequence numbers we're talking about


```suggestion
	onchainNextExpectedSeqNums, err := reader.NextSeqNum(ctx, reportChains)
```

commit/plugin_functions.go
```diff
@@ -585,6 +585,50 @@ func validateObservedGasPrices(gasPrices []cciptypes.GasPriceChain) error {
 	return nil
 }
 
+// validateMerkleRootsState merkle roots seq nums validation by comparing with on-chain state.
+func validateMerkleRootsState(
+	ctx context.Context,
+	lggr logger.Logger,
+	report cciptypes.CommitPluginReport,
+	reader cciptypes.CCIPReader,
+) (bool, error) {
```
makramkd, line 594: Will this func gracefully fail (i.e return an error) if the node doesn't support the dest?

makramkd (commented):
Minor comments

commit/plugin_e2e_test.go
dimkouv: if 21 was already committed we wouldn't be able to post 21

dimkouv: the report of this test does not include chainA

commit/plugin_functions.go
dimkouv: I'll rename the interface method, contract returns the last committed seq num. And rename var to `onchainCommittedSeqNums`

commit/plugin_functions.go
dimkouv: if the node does not support the dest this is not called at all

dimkouv: caller provides a reader

commit/plugin_e2e_test.go
dimkouv: we have to rename the interface method name to ``LastCommittedSeqNums` not `NextSeqNums`. I'll do it as a followup

commit/plugin_functions.go
dimkouv: https://github.com/smartcontractkit/chainlink-common/pull/627

asoliman92 (approved):
LGTM

makramkd (approved):

PR #15, opened 2024-07-02
go.mod: match chainlink-common versions

makramkd:
The go.mod file has many versions that are different than chainlink-common's go.mod.

This is causing issues with unnecessary version bumps in the ccip repo when we import chainlink-ccip.

In order to rectify this, we need to match chainlink-common's versions exactly.

This PR does exactly that.


makramkd (commented):
Related PR: https://github.com/smartcontractkit/ccip/pull/1123

PR #16, opened 2024-07-02
Increase test delay

dimkouv:
Make the test less flaky until we re-write this test to not use `time.Sleep`


asoliman92 (approved):

asoliman92 (commented):
TBH I think it's okay to leave it with a Sleep here. This case will not happen in production environment, we don't poll on millisecond bases and even if it happens it's not it will stop completely shortly after the program using it finishes.

dimkouv:
It's a problem on my local machine, i have to re-run the test suite until it succeeds.

PR #17, opened 2024-07-02
Implement CommitPlugin with RMN OffChain Blessing

rstout:
Implements a new version of the Commit Plugin in /commit_rmn_ocb (ocb = off chain blessing) based on the [spec](https://github.com/smartcontractkit/ccip/pull/1043/files). This is basically only the OCR3 function implementations (e.g. Query, Observation, Outcome, etc). No tests have been written yet. I recommend starting with the README.


plugintypes/commit.go
```diff
@@ -11,11 +11,12 @@ import (
...
-	GasPrices   []cciptypes.GasPriceChain       `json:"gasPrices"`
-	TokenPrices []cciptypes.TokenPrice          `json:"tokenPrices"`
-	MaxSeqNums  []SeqNumChain                   `json:"maxSeqNums"`
-	FChain      map[cciptypes.ChainSelector]int `json:"fChain"`
+	NewMsgs     []cciptypes.RampMessageHeader `json:"newMsgs"`
+	GasPrices   []cciptypes.GasPriceChain     `json:"gasPrices"`
+	TokenPrices []cciptypes.TokenPrice        `json:"tokenPrices"`
+	MaxSeqNums  []SeqNumChain                 `json:"maxSeqNums"`
+	// TODO: Why is this a map and not a list?
+	FChain map[cciptypes.ChainSelector]int `json:"fChain"`
```
dimkouv, line 19: The other types are related to commit report and seemed better to be a slice (e.g. for determinism and matching onchain types).

FChain is not part of the report, determinism not needed anywhere so seemed easier to keep as a map.

commit_rmn_ocb/types.go
```diff
@@ -0,0 +1,348 @@
...
+	SeqNumRange cciptypes.SeqNumRange   `json:"seqNumRange"`
+}
+
+// Rmn is an interface for all RMN logic
+type Rmn interface {
+	// RequestOnRampMaxSeqNums returns the maximum sequence number that RMN finds on the OnRamp for each given chain
+	// (for the configured dest chain)
+	RequestOnRampMaxSeqNums(chains []cciptypes.ChainSelector) ([]plugintypes.SeqNumChain, error)
+	RequestSignedIntervals(chainRanges []ChainRange) ([]SignedMerkleRoot, error)
+	VerifySignedMerkleRoot(merkleRoot SignedMerkleRoot) error
```
makramkd, line 306: I believe these are out of date w/ the latest design doc, can we update?

commit_rmn_ocb/validate_observation.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		seenChains.Add(seqNumChain.ChainSel)
+	}
+
+	return nil
+}
+
+func validateFChain(fChain map[cciptypes.ChainSelector]int) error {
+	for _, f := range fChain {
+		if f < 0 {
```
makramkd, line 138: Should be `f <= 0` since `f` is never 0.

commit_rmn_ocb/types.go
```diff
@@ -0,0 +1,348 @@
...
+
+	return tokenPrices
+}
+
+type CommitPluginOutcomeType int
+
+const (
+	ReportIntervalsSelected CommitPluginOutcomeType = iota
+	ReportGenerated
+	ReportEmpty
```
makramkd, line 236: This doesn't seem to be used in the code, is that intentional?

commit_rmn_ocb/types.go
```diff
@@ -0,0 +1,348 @@
...
+	ReportNotYetTransmitted
+	ReportTransmitted
```
makramkd, line 238: These three:

* `ReportNotYetTransmitted`
* `ReportTransmitted`
* `ReportNotTransmitted`

seem to be so lexicographically similar I feel like its easy to mistake one for the other. How about changing `ReportNotYetTransmitted` to `ReportInFlight`?

commit_rmn_ocb/plugin.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := p.ccipReader.Close(ctx); err != nil {
+		return fmt.Errorf("close ccip reader: %w", err)
+	}
+
+	p.bgSyncCancelFunc()
+	p.bgSyncWG.Wait()
+	return nil
+}
+
+func (p *Plugin) decodeOutcome(outcome ocr3types.Outcome) (CommitPluginOutcome, CommitPluginState) {
```
makramkd, line 93: Feels like this should return an error instead of just returning `SelectingRangesForReport` if there is an error decoding

makramkd (commented):
Still combing through the PR, this is just my first pass

commit_rmn_ocb/types.go
```diff
@@ -0,0 +1,348 @@
...
+	RequestSignedIntervals(chainRanges []ChainRange) ([]SignedMerkleRoot, error)
+	VerifySignedMerkleRoot(merkleRoot SignedMerkleRoot) error
+	// ChainThreshold returns how many RMN signatures are required for the given chain
+	ChainThreshold(chain cciptypes.ChainSelector) uint
+}
+
+type OnChain interface {
+	// GetOnRampMaxSeqNums returns the maximum sequence number that this oracle finds on the OnRamp for each given chain
+	// (for the configured dest chain)
+	GetOnRampMaxSeqNums() ([]plugintypes.SeqNumChain, error)
```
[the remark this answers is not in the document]
rstout: `getExpectedNextSequenceNumber` is a method on the OffRamp, not the OnRamp right? I've renamed the OffRamp seq nums to `OffRampNextSeqNums` and kept `OnRampMaxSeqNums`, we can discuss changing the names further.

commit_rmn_ocb/types.go
```diff
@@ -0,0 +1,348 @@
...
+
+	// GetOffRampMaxSeqNums returns the maximum sequence number that this oracle finds on the OffRamp for each given
+	// chain (for the configured dest chain)
+	GetOffRampMaxSeqNums() ([]plugintypes.SeqNumChain, error)
```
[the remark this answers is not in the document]
rstout: I've renamed this to `OffRampNextSeqNums`, but happy to discuss clearer names

commit_rmn_ocb/types.go
```diff
@@ -0,0 +1,348 @@
...
+
+type CommitPluginOutcomeType int
+
+const (
+	ReportIntervalsSelected CommitPluginOutcomeType = iota
+	ReportGenerated
+	ReportEmpty
+	ReportNotYetTransmitted
+	ReportTransmitted
+	ReportNotTransmitted
```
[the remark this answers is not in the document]
rstout: What's the rationale for this? I'm not a fan of the extra verbosity.

commit_rmn_ocb/validate_observation.go
rstout: f can be 0 for chains that are not RMN enabled

commit_rmn_ocb/types.go
rstout: Good catch, it's now used

commit_rmn_ocb/types.go
rstout: Updated

commit_rmn_ocb/plugin.go
rstout: My logic is: if there's an issue decoding, just go to the "select interval" state instead of erroring-out

commitrmnocb/metrics.go
```diff
@@ -0,0 +1 @@
+package commitrmnocb
```
makramkd, line 1: Is the intention of this file to house some prometheus metrics?

commitrmnocb/observation.go
```diff
@@ -0,0 +1,170 @@
...
+
+	for i, msg := range msgs {
+		// Assert there are no sequence number gaps in msgs
+		if i > 0 {
+			if msg.Header.SequenceNumber != msgs[i-1].Header.SequenceNumber+1 {
+				return [32]byte{}, fmt.Errorf("found non-consecutive sequence numbers when computing merkle root, "+
+					"gap between sequence nums %d and %d, messages: %v", msgs[i-1].Header.SequenceNumber,
+					msg.Header.SequenceNumber, msgs)
+			}
+		}
```
makramkd, line 141: Nit: would this be useful as a separate helper that is unit tested?

commitrmnocb/observation.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return roots
+}
+
+// computeMerkleRoot Compute the merkle root of a list of messages
+func (p *Plugin) computeMerkleRoot(ctx context.Context, msgs []cciptypes.Message) (cciptypes.Bytes32, error) {
+	hashes := make([][32]byte, 0)
```
makramkd, line 130: Style: since you're just appending to it prefer `var hashes [][32]byte` see e.g https://github.com/uber-go/guide/blob/master/style.md#nil-is-a-valid-slice

commitrmnocb/observation.go
```diff
@@ -0,0 +1,170 @@
...
+				return [32]byte{}, fmt.Errorf("found non-consecutive sequence numbers when computing merkle root, "+
+					"gap between sequence nums %d and %d, messages: %v", msgs[i-1].Header.SequenceNumber,
+					msg.Header.SequenceNumber, msgs)
+			}
+		}
+
+		msgHash, err := p.msgHasher.Hash(ctx, msg)
+		if err != nil {
+			p.lggr.Warnw("failed to hash message", "msg", msg, "err", err)
+			return cciptypes.Bytes32{}, err
```
makramkd, line 146: Lets wrap errors where added context is useful, e.g in this case

```suggestion
			return cciptypes.Bytes32{}, fmt.Errorf("failed to hash message with id %s: %w", msg.ID.String(), err)
```

commitrmnocb/observation.go
```diff
@@ -0,0 +1,170 @@
...
+						ChainSel:     chainRange.ChainSel,
+						SeqNumsRange: chainRange.SeqNumRange,
+						MerkleRoot:   root,
+					}
+					roots = append(roots, merkleRoot)
+				}
+			}
+		}
+	}
+
```
makramkd, line 124: At this point we expect most of the time for len(roots) == len(ranges) right? Would we want to log some sort of summary log here? I think its OK that sometimes these lengths don't match but just wanna make sure.

commitrmnocb/observation.go
```diff
@@ -0,0 +1,170 @@
...
+			result[i] = plugintypes.SeqNumChain{ChainSel: sourceChains[i], SeqNum: offRampNextSeqNums[i]}
+		}
+
+		return result
+	}
+
+	return nil
+}
+
+// ObserveMerkleRoots Compute the merkle roots for the given sequence number ranges
```
makramkd, line 95: Style: Go comments start with the name of the exported symbol (in this case ObserveMerkleRoots) and continues in the normal tense as if that is part of the sentence. e.g in this case

```suggestion
// ObserveMerkleRoots computes the merkle roots for the given sequence number ranges
```

makramkd: Similarly for the rest of the comments in this file

commitrmnocb/plugin.go
```diff
@@ -0,0 +1,161 @@
...
+	ccipReader        reader.CCIP
+	readerSyncer      *plugincommon.BackgroundReaderSyncer
+	tokenPricesReader reader.TokenPrices
+	reportCodec       cciptypes.CommitPluginCodec
+	msgHasher         cciptypes.MessageHasher
+	lggr              logger.Logger
+	homeChain         reader.HomeChain
+
+	// TODO: add back
+	// reportingCfg ocr3types.ReportingPluginConfig
```
makramkd, line 37: Adding it back in this PR?

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,325 @@
...
+	for chain, fValues := range fChainValues {
+		f, _ := mostFrequentElem(fValues)
+		// TODO: uncomment when p.reportingCfg is added back
+		//if count < p.reportingCfg.F {
+		//	// TODO: metrics
+		//	p.lggr.Warnf("failed to reach consensus on fChain values for chain %d because no single f "+
+		//		"value was observed more than the expected %d times, found f value %d observed by only %d oracles, "+
+		//		"f values: %v",
+		//		chain, p.reportingCfg.F, f, count, fValues)
+		//}
```
makramkd, line 292: Lets add this back, was there an issue with passing through reportingPluginConfig?

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,325 @@
...
+	fDestChain int,
+) map[cciptypes.ChainSelector]cciptypes.SeqNum {
+	consensus := make(map[cciptypes.ChainSelector]cciptypes.SeqNum)
+
+	for chain, offRampMaxSeqNums := range offRampMaxSeqNumsByChain {
+		seqNum, count := mostFrequentElem(offRampMaxSeqNums)
+		if count <= fDestChain {
+			// TODO: metrics
+			p.lggr.Warnf("could not reach consensus on offRampMaxSeqNums for chain %d "+
+				"because we did not receive a sequence number that was observed by at least f (%d) oracles, "+
```
makramkd, line 267:
```suggestion
				"because we did not receive a sequence number that was observed by at least fChain (%d) oracles, "+
```

To make it clear this is the fChain and not big F value

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,325 @@
...
+	fChains map[cciptypes.ChainSelector]int,
+) map[cciptypes.ChainSelector]cciptypes.SeqNum {
+	consensus := make(map[cciptypes.ChainSelector]cciptypes.SeqNum)
+
+	for chain, onRampMaxSeqNums := range onRampMaxSeqNumsByChain {
+		if f, exists := fChains[chain]; exists {
+			if len(onRampMaxSeqNums) < 2*f+1 {
+				// TODO: metrics
+				p.lggr.Warnf("could not reach consensus on onRampMaxSeqNums for chain %d "+
+					"because we did not receive more than 2f+1 observed sequence numbers, 2f+1: %d, "+
```
makramkd, line 236:
```suggestion
					"because we did not receive more than 2fChain+1 (%d) observed sequence numbers, "+
```

Maybe this is clearer, matches the pattern of the log below

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,325 @@
...
+	consensus := make(map[cciptypes.ChainSelector]cciptypes.MerkleRootChain)
+
+	for chain, roots := range rootsByChain {
+		if f, exists := fChains[chain]; exists {
+			root, count := mostFrequentElem(roots)
+
+			if count <= f {
+				// TODO: metrics
+				p.lggr.Warnf("failed to reach consensus on a merkle root for chain %d "+
+					"because no single merkle root was observed more than the expected %d times, found merkle root %d "+
```
makramkd, line 207:
```suggestion
					"because no single merkle root was observed more than the expected fChain (%d) times, found merkle root %d "+
```

commitrmnocb/report.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/smartcontractkit/chainlink-ccip/commit"
+)
+
+func (p *Plugin) Reports(seqNr uint64, outcomeBytes ocr3types.Outcome) ([]ocr3types.ReportWithInfo[[]byte], error) {
+	outcome, err := DecodeCommitPluginOutcome(outcomeBytes)
+	if err != nil {
+		// TODO: metrics
+		p.lggr.Errorw("failed to decode Outcome", "outcomeBytes", outcomeBytes, "err", err)
+		return nil, fmt.Errorf("failed to decode Outcome: %w", err)
```
makramkd, line 19:
```suggestion
		return nil, fmt.Errorf("failed to decode Outcome (%s): %w", hex.EncodeToString(outcomeBytes), err)
```

Including the outcome data for debugging

commitrmnocb/report.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if outcome.OutcomeType != ReportGenerated {
+		return []ocr3types.ReportWithInfo[[]byte]{}, nil
+	}
```
makramkd, line 24: Worth adding a log for this case?

commitrmnocb/report.go
```diff
@@ -0,0 +1,84 @@
...
+
+	rep := cciptypes.NewCommitPluginReport(outcome.RootsToReport, outcome.TokenPrices, outcome.GasPrices)
+
+	encodedReport, err := p.reportCodec.Encode(context.Background(), rep)
+	if err != nil {
+		return nil, fmt.Errorf("encode commit plugin report: %w", err)
+	}
+
```
makramkd, line 32: A log here might be useful as well, of the pre-encoded report `rep`

commitrmnocb/types.go
```diff
@@ -0,0 +1,255 @@
...
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+type Query struct {
+	RmnOnRampMaxSeqNums []plugintypes.SeqNumChain
+	MerkleRoots         []cciptypes.MerkleRootChain
+}
+
+func (q Query) Encode() ([]byte, error) {
+	return json.Marshal(q)
```
makramkd, line 21: Is it important that the encoding of the Query be deterministic? If so, lets add some `sort.Slice`s here to make sure that the slices in the struct are sorted (perhaps according to chain selector, like the other sorts)

commitrmnocb/types.go
```diff
@@ -0,0 +1,255 @@
...
+	OutcomeType                     OutcomeType                 `json:"outcomeType"`
+	RangesSelectedForReport         []ChainRange                `json:"rangesSelectedForReport"`
+	RootsToReport                   []cciptypes.MerkleRootChain `json:"rootsToReport"`
+	OffRampNextSeqNums              []plugintypes.SeqNumChain   `json:"offRampNextSeqNums"`
+	TokenPrices                     []cciptypes.TokenPrice      `json:"tokenPrices"`
+	GasPrices                       []cciptypes.GasPriceChain   `json:"gasPrices"`
+	ReportTransmissionCheckAttempts uint                        `json:"reportTransmissionCheckAttempts"`
+}
+
+// Encode TODO: sort all lists here to ensure deterministic serialization
```
makramkd, line 209: I think we should do this in this PR so we don't run into nondeterminism issues down the line

pluginconfig/commit.go
```diff
@@ -22,6 +22,9 @@ type CommitPluginConfig struct {
 	// NewMsgScanBatchSize is the number of max new messages to scan, typically set to 256.
 	NewMsgScanBatchSize int `json:"newMsgScanBatchSize"`
 
+	// The maximum number of times to check if the previous report has been transmitted
+	MaxReportTransmissionCheckAttempts uint
```
makramkd, line 26: Is this something we want all plugin instances in the DON to have in common? If so, it should go in the OffchainConfig struct below

makramkd:
Looks good overall, mostly minor comments above 😄

commitrmnocb/metrics.go
rstout: Yes

commitrmnocb/observation.go
rstout: In the case where we fail to compute the merkle root for one or more chains, then len(roots) != len(ranges). Observe() already logs a summary that includes merkle roots, and this function logs merkle root failures, so logging a summary could be verbose here.

commitrmnocb/report.go
rstout: This is the happy path (e.g. in the round when intervals are chosen), so I think logging here would just be verbose.

commitrmnocb/report.go
rstout: We log the report in `ShouldTransmitAcceptedReport`, I think that's sufficient.

commitrmnocb/types.go
rstout: I'll add a TODO to ensure we do this when Query has fields

pluginconfig/commit.go
rstout: What's the difference between `CommitPluginConfig` and `CommitOffchainConfig`?

commitrmnocb/observation.go
winder: See `ConstructMerkleTree` in `execute/report/report.go`. Not a drop-in helper since it uses exec specific types.

commitrmnocb/observation.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			msgID := hex.EncodeToString(msg.Header.MessageID[:])
+			p.lggr.Warnw("failed to hash message", "msg", msg, "msg_id", msgID, "err", err)
+			return cciptypes.Bytes32{}, fmt.Errorf("failed to hash message with id %s: %w", msgID, err)
+		}
+
+		hashes = append(hashes, msgHash)
+	}
+
+	tree, err := merklemulti.NewTree(hashutil.NewKeccak(), hashes)
```
winder, line 153:
```suggestion
	// TODO: Do not hard code the hash function, it should be derived from the message hasher.
	tree, err := merklemulti.NewTree(hashutil.NewKeccak(), hashes)
```

commitrmnocb/observation_test.go
```diff
@@ -0,0 +1,140 @@
...
+func NewMessageHasher() *MessageHasher {
+	return &MessageHasher{}
+}
+
+func (m *MessageHasher) Hash(ctx context.Context, msg cciptypes.Message) (cciptypes.Bytes32, error) {
+	// simply return the msg id as bytes32
+	var b32 [32]byte
+	copy(b32[:], msg.Header.MessageID[:])
+	return b32, nil
+}
```
winder, line 129: duplicate of `internal/mocks/messagehasher.go`.

commitrmnocb/observation_test.go
```diff
@@ -0,0 +1,140 @@
...
+type BadMessageHasher struct{}
+
+func NewBadMessageHasher() *BadMessageHasher {
+	return &BadMessageHasher{}
+}
+
+// Always returns an error
+func (m *BadMessageHasher) Hash(ctx context.Context, msg cciptypes.Message) (cciptypes.Bytes32, error) {
+	return cciptypes.Bytes32{}, fmt.Errorf("failed to hash")
+}
```
winder, line 140: Could put this in the mock dir as well, I have something almost identical:
```
type badHasher struct{}

func (bh badHasher) Hash(context.Context, cciptypes.Message) (cciptypes.Bytes32, error) {
	return cciptypes.Bytes32{}, fmt.Errorf("bad hasher")
}
```

commitrmnocb/types.go
```diff
@@ -0,0 +1,270 @@
...
+
+	encodedOutcome, err := json.Marshal(o)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode Outcome: %w", err)
+	}
+
+	return encodedOutcome, nil
+}
+
+func DecodeCommitPluginOutcome(b []byte) (Outcome, error) {
```
winder, line 234: nit: Hopefully get rid of everything that says `CommitPlugin`
```suggestion
func DecodeOutcome(b []byte) (Outcome, error) {
```

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,316 @@
...
+		FChain:             fChains,
+	}
+
+	return consensusObs, nil
+}
+
+// Given a mapping from chains to a list of merkle roots, return a mapping from chains to a single consensus merkle
+// root. The consensus merkle root for a given chain is the merkle root with the most observations that was observed at
+// least fChain times.
+func (p *Plugin) merkleRootConsensus(
```
winder, line 186: There's a generic helper to manage counting observations that could be put in a reusable location: `execute/internal/validation/reports.go`

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,316 @@
...
+	return outcome
+}
+
+// checkForReportTransmission checks if the OffRamp has an updated set of max seq nums compared to the seq nums that
+// were observed when the most recent report was generated. If an update to these max seq sums is detected, it means
+// that the previous report has been transmitted, and we output ReportTransmitted to dictate that a new report
+// generation phase should begin. If no update is detected, and we've exhausted our check attempts, output
+// ReportTransmissionFailed to signify we stop checking for updates and start a new report generation phase. If no
+// update is detected, and we haven't exhausted our check attempts, output ReportInFlight to signify that we should
+// check again next round.
```
winder, line 122: Is it accurate to say that this is mandatory to ensure sequence numbers and nonces are ordered correctly?

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,316 @@
...
+) Outcome {
+	rangesToReport := make([]ChainRange, 0)
+
+	rmnOnRampMaxSeqNumsMap := make(map[cciptypes.ChainSelector]cciptypes.SeqNum)
+	for _, seqNumChain := range query.RmnOnRampMaxSeqNums {
+		rmnOnRampMaxSeqNumsMap[seqNumChain.ChainSel] = seqNumChain.SeqNum
+	}
+
+	observedOnRampMaxSeqNumsMap := consensusObservation.OnRampMaxSeqNums
+	observedOffRampNextSeqNumsMap := consensusObservation.OffRampNextSeqNums
```
winder, line 64: These are literally the same object in the `SelectingRangesForReport` observation. Is it some sort of placeholder or should you drop `OnRampMaxSeqNums`?

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,316 @@
...
+		p.lggr.Warnw("Unexpected state in Outcome", "state", nextState)
+		return outcome.Encode()
+	}
+
+	p.lggr.Infow("Commit Plugin Outcome", "outcome", outcome, "oid", p.nodeID)
+	return outcome.Encode()
+}
+
+// ReportRangesOutcome determines the sequence number ranges for each chain to build a report from in the next round
+func (p *Plugin) ReportRangesOutcome(
```
winder, line 52: nit: this is a pure function
```suggestion
func ReportRangesOutcome(
```

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,316 @@
...
+		OutcomeType:             ReportIntervalsSelected,
+		RangesSelectedForReport: rangesToReport,
+	}
+
+	return outcome
+}
+
+// Given a set of observed merkle roots, gas prices and token prices, and roots from RMN, construct a report
+// to transmit on-chain
+func (p *Plugin) buildReport(
```
winder, line 95: nit: this is a pure function
```suggestion
// buildReport takes observed merkle roots, gas prices and token prices, and roots from RMN, construct a report
// to transmit on-chain
func buildReport(
```

winder: I'll leave out any further comments like this. I'd be tempted to make most of the helpers in this file pure functions by surfacing the logger dependency (or logging errors outside the helper).

commitrmnocb/outcome.go
```diff
@@ -0,0 +1,316 @@
...
+
+	return Outcome{
+		OutcomeType:                     ReportInFlight,
+		OffRampNextSeqNums:              previousOutcome.OffRampNextSeqNums,
+		ReportTransmissionCheckAttempts: previousOutcome.ReportTransmissionCheckAttempts + 1,
+	}
+}
+
+// getConsensusObservation Combine the list of observations into a single consensus observation
+func (p *Plugin) getConsensusObservation(aos []types.AttributedObservation) (ConsensusObservation, error) {
```
winder, line 159: gonna steal this name, I called it something like "mergeObservations".

commitrmnocb/types.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	return tokenPrices
+}
+
+type OutcomeType int
+
+const (
+	ReportIntervalsSelected OutcomeType = iota + 1
+	ReportGenerated
```
winder, line 184: nit: the report doesn't generate until after the outcome. I'd also consider removing the `Report` prefix from all of these.
```suggestion
	ReportGeneration
```

commitrmnocb/report.go
```diff
@@ -0,0 +1,85 @@
...
+func (p *Plugin) Reports(seqNr uint64, outcomeBytes ocr3types.Outcome) ([]ocr3types.ReportWithInfo[[]byte], error) {
+	outcome, err := DecodeCommitPluginOutcome(outcomeBytes)
+	if err != nil {
+		// TODO: metrics
+		p.lggr.Errorw("failed to decode Outcome", "outcomeBytes", outcomeBytes, "err", err)
+		return nil, fmt.Errorf("failed to decode Outcome (%s): %w", hex.EncodeToString(outcomeBytes), err)
+	}
+
+	if outcome.OutcomeType != ReportGenerated {
```
winder, line 23:
```suggestion
	// Reports are only generated from "ReportGenerated" outcomes.
	if outcome.OutcomeType != ReportGenerated {
```

winder (commented):
Mostly nits and questions. Looks great overall.

I did have some confusion around the transient nature of states and the outcome type -- especially in the `Reports` function where it would have been nice to check for a `BuildingReport` state rather than a `ReportGenerated` outcome.

commit_rmn_ocb/types.go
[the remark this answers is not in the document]
winder: haha I left a nit for the exact opposite, get rid of `Report` :joy:

commit_rmn_ocb/types.go
```diff
@@ -0,0 +1,348 @@
...
+	case ReportTransmitted:
+		return SelectingRangesForReport
+	case ReportNotTransmitted:
+		return SelectingRangesForReport
+	default:
+		return SelectingRangesForReport
+	}
+}
+
+type CommitPluginState int
```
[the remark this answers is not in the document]
winder: I thought there was a linter that made that recommendation, but I'm not seeing it at the moment.

commitrmnocb/outcome.go
rstout: This is on purpose, it will need to be changed in the future. I'll add a comment.

commitrmnocb/outcome.go
rstout: What is mandatory?

commitrmnocb/observation.go
rstout: Let's unify `ConstructMerkleTree` in a future PR

commitrmnocb/observation.go
```diff
@@ -0,0 +1,171 @@
...
+					p.lggr.Warnw("call to computeMerkleRoot failed", "err", err)
+				} else {
+					merkleRoot := cciptypes.MerkleRootChain{
+						ChainSel:     chainRange.ChainSel,
+						SeqNumsRange: chainRange.SeqNumRange,
+						MerkleRoot:   root,
+					}
+					roots = append(roots, merkleRoot)
+				}
+			}
```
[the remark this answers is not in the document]
rstout: I intentionally didn't return early so that failure to compute a merkle root for one chain does not prevent computing the merkle root for other chains

commitrmnocb/observation.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		p.lggr.Warnw("call to supportedChains failed", "err", err)
+		return nil
+	}
+
+	for _, chainRange := range ranges {
+		if supportedChains.Contains(chainRange.ChainSel) {
+			msgs, err := p.ccipReader.MsgsBetweenSeqNums(ctx, chainRange.ChainSel, chainRange.SeqNumRange)
+			if err != nil {
+				p.lggr.Warnw("call to MsgsBetweenSeqNums failed", "err", err)
```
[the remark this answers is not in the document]
rstout: The convention I've been using is that "warn" is for issues that are unexpected errors, but the system can function fine in spite of them, and may recover on its own. "Error" means an assumption was broken, the system might now be in a broken state, and human intervention is needed. 

I'm happy to switch these "warns" to "errors" if that better reflects the existing convention in Go/CLL.

commitrmnocb/observation.go
```diff
@@ -0,0 +1,171 @@
...
+) (types.Observation, error) {
+	previousOutcome, nextState := p.decodeOutcome(outCtx.PreviousOutcome)
+
+	observation := Observation{}
+	switch nextState {
+	case SelectingRangesForReport:
+		offRampNextSeqNums := p.ObserveOffRampNextSeqNums(ctx)
+		observation = Observation{
+			OnRampMaxSeqNums:   offRampNextSeqNums,
+			OffRampNextSeqNums: offRampNextSeqNums,
```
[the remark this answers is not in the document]
rstout: This is on purpose, I'll add a comment.

commitrmnocb/outcome.go
winder: Waiting for confirmation that the previous report has completed transmission.

commitrmnocb/observation.go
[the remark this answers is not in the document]
winder: Nice, I didn't notice that it was tolerant of read errors. In that case I would switch `return` with `continue`.

makramkd (approved):

winder (approved):
I think most/all of the feedback I left can wait until a followup PR. Feel free to merge at your discretion.

commitrmnocb/observation.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			p.lggr.Warnw("call to NextSeqNum failed", "err", err)
+			return nil
+		}
+
+		if len(offRampNextSeqNums) != len(sourceChains) {
+			p.lggr.Warnf("call to NextSeqNum returned unexpected number of seq nums, got %d, expected %d",
+				len(offRampNextSeqNums), len(sourceChains))
+			return nil
+		}
```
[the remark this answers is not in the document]
rstout: We're tolerant to an unexpected number of next seq nums, though this error make sense as an "error" instead of a "warn".

>  Is it possible to avoid that dependency with a more tolerant check here?

Possibly, let's think about this in future PRs.

commitrmnocb/outcome.go
rstout: Not necessarily. The Commit Plugin always starts each round by reading the OnRamp and OffRamp state, so sane Seq Num ranges will be chosen, and in the case where an in-flight report lands after these seq nums have been chosen, I assume the OffRamp will reject a report generated with these obsolete seq num ranges.

PR #18, opened 2024-07-03
Commit plugin dynamic reader sync

dimkouv:


commit/plugin_functions.go
```diff
@@ -628,6 +630,40 @@ func validateMerkleRootsState(
...
+			select {
+			case <-ctx.Done():
+				wg.Done()
+				return
+			case <-ticker:
+				timeoutCtx, cf := context.WithTimeout(ctx, syncTimeout)
+				updated, err := reader.Sync(timeoutCtx)
+				if err != nil {
+					lggr.Errorw("error syncing reader", "err", err)
+					cf()
```
asoliman92, line 652: Why call `cf` if we want to continue?

commit/plugin_functions.go
```diff
@@ -628,6 +630,40 @@ func validateMerkleRootsState(
...
+					continue
+				}
+
+				if !updated {
+					lggr.Debug("no updates found after trying to sync")
+				} else {
+					lggr.Info("ccip reader sync success")
+				}
+				cf()
```
asoliman92, line 661: Same, why call `cf`, won't this just make one update and return?

commit/plugin_functions.go
dimkouv: i cancel the timeout ctx that i initiate in that same loop. a new ctx with timeout is created in next iteration. without it we have resource leak

asoliman92: ah understood, I misread it to be the cf for the `ctx` not the `timeoutCtx` 👍

asoliman92 (approved):

commit/plugin.go
```diff
@@ -62,6 +64,21 @@ func NewPlugin(
 		lggr:              lggr,
 		homeChain:         homeChain,
 	}
+
+	bgSyncCtx, bgSyncCf := context.WithCancel(context.Background())
+	p.bgSyncCancelFunc = bgSyncCf
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
```
makramkd, line 71: WaitGroup needs a `.Wait()` somewhere. In `Close` perhaps?

commit/plugin_functions.go
```diff
@@ -628,6 +630,40 @@ func validateMerkleRootsState(
...
+	lggr logger.Logger,
+	reader cciptypes.CCIPReader,
+	syncTimeout time.Duration,
+	ticker <-chan time.Time,
+) {
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				wg.Done()
```
makramkd, line 645: More idiomatic to put `defer wg.Done()` at the top of the func

commit/plugin.go
```diff
@@ -345,6 +362,8 @@ func (p *Plugin) Close() error {
 	if err := p.ccipReader.Close(ctx); err != nil {
 		return fmt.Errorf("close ccip reader: %w", err)
 	}
+
+	p.bgSyncCancelFunc()
```
makramkd, line 366: If you want to wait for the syncing goroutine to finish put a `wg.Wait()` here, otherwise you can remove the wg.

commit/plugin_functions.go
makramkd, line 652: Again minor but this can be more idiomatic if you call a func that does `defer cancel()` e.g

```golang
case <-ticker:
  tick(syncTimeout, reader)

func tick(ctx context.Context, syncTimeout time.Duration, reader reader.CCIPReader) {
  ctx, cancel := context.WithTimeout(ctx, syncTimeout)
  defer cancel()

  // .. rest of code here, return as usual, ctx will be canceled
}
```

makramkd (approved):

PR #20, opened 2024-07-04
Update reporting plugin factory constructors

makramkd:
We need to inject the required dependencies in order for the plugins to correctly function.

### DIFF
--- a/commit/factory.go
+++ b/commit/factory.go
@@ -8,6 +8,8 @@ import (
 
 	"google.golang.org/grpc"
 
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 	"github.com/smartcontractkit/chainlink-common/pkg/types/core"
 
@@ -35,7 +37,7 @@ func (p PluginFactoryConstructor) NewReportingPluginFactory(
 	keyValueStore core.KeyValueStore,
 	relayerSet core.RelayerSet,
 ) (core.OCR3ReportingPluginFactory, error) {
-	return NewPluginFactory(), nil
+	return nil, nil
 }
 
 func (p PluginFactoryConstructor) NewValidationService(ctx context.Context) (core.ValidationService, error) {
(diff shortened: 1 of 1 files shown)


pkg/crconsts/consts.go
makramkd, line 1: Renamed package to `consts` since its not just chain reader stuff.

commit/factory.go
```diff
@@ -35,42 +38,88 @@ func (p PluginFactoryConstructor) NewReportingPluginFactory(
 	keyValueStore core.KeyValueStore,
 	relayerSet core.RelayerSet,
 ) (core.OCR3ReportingPluginFactory, error) {
-	return NewPluginFactory(), nil
+	return nil, errors.New("unimplemented")
```
makramkd, line 41: We'll revisit this when the loopification work starts.

commit/factory.go
```diff
@@ -35,42 +38,88 @@ func (p PluginFactoryConstructor) NewReportingPluginFactory(
...
+			p.lggr,
+			p.homeChainReader,
+		), ocr3types.ReportingPluginInfo{
+			Name: "CCIPRoleCommit",
+			Limits: ocr3types.ReportingPluginLimits{
+				// No query for this commit implementation.
+				MaxQueryLength:       0,
+				MaxObservationLength: 20_000, // 20kB
+				MaxOutcomeLength:     10_000, // 10kB
+				MaxReportLength:      10_000, // 10kB
```
makramkd, line 119: These are ballpark, we'll revisit as we test.

commit/plugin.go
```diff
@@ -157,7 +157,7 @@ func (p *Plugin) Observation(
 	// If there's no previous outcome (first round ever), we only observe the latest committed sequence numbers.
 	// and on the next round we use those to look for messages.
 	if outctx.PreviousOutcome == nil {
-		p.lggr.Debugw("first round ever, can't observe new messages yet")
+		p.lggr.Infow("first round ever, can't observe new messages yet")
```
makramkd, line 160: Making these info because turning on debug logs in the CL node brings an avalanche of logs that are not relevant.

commit/factory.go
```diff
@@ -35,42 +38,88 @@ func (p PluginFactoryConstructor) NewReportingPluginFactory(
...
 }
 
-func (p PluginFactory) NewReportingPlugin(config ocr3types.ReportingPluginConfig,
+func (p *PluginFactory) NewReportingPlugin(config ocr3types.ReportingPluginConfig,
 ) (ocr3types.ReportingPlugin[[]byte], ocr3types.ReportingPluginInfo, error) {
-	// TODO: Get this from ocr config, it's the mapping of the oracleId index in the DON
-	var oracleIDToP2pID map[commontypes.OracleID]libocrtypes.PeerID
+	var oracleIDToP2pID = make(map[commontypes.OracleID]ragep2ptypes.PeerID)
+	for i, p2pID := range p.ocrConfig.Config.P2PIds {
+		oracleIDToP2pID[commontypes.OracleID(i)] = p2pID
```
dimkouv, line 83: rename `i` to `oracleID` ?

commit/plugin.go
asoliman92: but then it will always show even without turning on debug logs, right?

commit/factory.go
asoliman92: Will this be read from config in the future?

dimkouv (approved):
🚀 (approving with minor comments)

commit/plugin.go
makramkd: Log levels filter downards, so Debug means Debug and all levels above, Info means Info and all levels above, etc.

commit/factory.go
makramkd: No these will likely be constants, but might not be these values.

commit/factory.go
```diff
@@ -35,42 +38,88 @@ func (p PluginFactoryConstructor) NewReportingPluginFactory(
...
-		nil, // msgHasher
-		nil, // lggr
-		nil, // homeChain
-	), ocr3types.ReportingPluginInfo{}, nil
+			context.Background(),
+			config.OracleID,
+			oracleIDToP2pID,
+			pluginconfig.CommitPluginConfig{
+				DestChain:           p.ocrConfig.Config.ChainSelector,
+				NewMsgScanBatchSize: 256,
```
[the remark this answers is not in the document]
makramkd: Will make it a const for now, but we can put it in the OCR config as well once we have offchain config.

commit/plugin.go
```diff
@@ -167,7 +167,7 @@ func (p *Plugin) Observation(
 	if err != nil {
 		return types.Observation{}, fmt.Errorf("decode commit plugin previous outcome: %w", err)
 	}
-	p.lggr.Debugw("previous outcome decoded", "outcome", prevOutcome.String())
+	p.lggr.Infow("previous outcome decoded", "outcome", prevOutcome.String())
```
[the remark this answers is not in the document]
makramkd: I'm not 100% sure but doesn't hurt. I think depending on the logger it sometimes tries to json encode it.

internal/reader/ccip.go
```diff
@@ -343,8 +355,16 @@ func (r *CCIPChainReader) Sync(ctx context.Context) (bool, error) {
 		return false, fmt.Errorf("get onramps: %w", err)
 	}
 
+	r.lggr.Infow("got source chain configs", "onramps", func() []string {
+		var r []string
+		for _, scc := range sourceConfigs {
+			r = append(r, "0x"+hex.EncodeToString(scc.OnRamp))
```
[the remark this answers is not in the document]
makramkd: Will copy/paste.

internal/reader/ccip.go
```diff
@@ -343,8 +355,16 @@ func (r *CCIPChainReader) Sync(ctx context.Context) (bool, error) {
...
+		}
+		return r
+	}())
+
 	for chain, cfg := range sourceConfigs {
-		if cfg.OnRamp == "" {
+		if len(cfg.OnRamp) == 0 {
```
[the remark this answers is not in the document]
makramkd: I think this is more idiomatic, since both `nil` and the empty slice will have length zero.

commit/factory.go
```diff
@@ -2,19 +2,29 @@ package commit
...
+	ragep2ptypes "github.com/smartcontractkit/libocr/ragep2p/types"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+)
+
+const (
+	// defaultMsgScanBatchSize is the default batch size for sequence number range queries.
+	// Since 256 messages is the limit for a merkle root, this is set to 256.
+	// NOTE: maybe we can also set this in the OCR config, offchainConfig.
```
winder, line 26: sounds like a good constant for the crypto package

commit/plugin.go
```diff
@@ -349,7 +349,7 @@ func (p *Plugin) ShouldTransmitAcceptedReport(
 		return false, fmt.Errorf("validate merkle roots state: %w", err)
 	}
 
-	p.lggr.Debugw("transmitting report",
+	p.lggr.Infow("transmitting report",
```
winder, line 352: Making this one Info level seems fine.... not sure about the others. Making everything Info means there would be an avalanche of Info logs as well

winder (approved):
I think you might have gotten a little carried away with Debug -> Info. Other than that, LGTM.

commit/plugin.go
makramkd: We'll have to revisit these later, for now I think this is slightly better than turning on Debug and getting all the debug logs of the app.

commit/factory.go
```diff
@@ -2,19 +2,29 @@ package commit
...
+	defaultMsgScanBatchSize = 256
```
winder, line 27: Looks like there's already a constant for this
```suggestion
	defaultMsgScanBatchSize = merklemulti.MaxNumberTreeLeaves
```

makramkd: Fixed, just used that const directly.

PR #21, opened 2024-07-04
modify the getAllChainConfigs function call from smartcontract with pagination based logic

defistar:
feat: CCIP-2594 modify the getAllChainConfigs function call from smartcontract with pagination based logic


internal/reader/home_chain.go
```diff
@@ -111,19 +111,38 @@ func (r *homeChainPoller) poll() {
...
 func (r *homeChainPoller) fetchAndSetConfigs(ctx context.Context) error {
-	var chainConfigInfos []ChainConfigInfo
-	err := r.homeChainReader.GetLatestValue(
-		ctx, "CCIPConfig", "getAllChainConfigs", nil, &chainConfigInfos,
-	)
-	if err != nil {
-		return err
+	var allChainConfigInfos []ChainConfigInfo
+	pageIndex := uint64(0)
+	pageSize := uint64(500)
```
elatoskinas, line 116: suggestion: should we make this configurable through the function params, or separate it out as a `const` to another file / top of file?

internal/reader/home_chain.go
```diff
@@ -111,19 +111,38 @@ func (r *homeChainPoller) poll() {
...
+			ctx,
+			"CCIPConfig",
+			"getAllChainConfigs",
+			map[string]interface{}{
+				"pageIndex": pageIndex,
+				"pageSize":  pageSize,
+			},
+			&chainConfigInfos,
+		)
+		if err != nil {
```
elatoskinas, line 130: q: when do we get an error? Is it always guaranteed to return an empty array if I am accessing out of bounds pages (e.g. 10 pages beyond what is available)?

Should we add a comment that we always expect a result to be returned?

internal/reader/home_chain.go
```diff
@@ -111,19 +111,38 @@ func (r *homeChainPoller) poll() {
...
 	}
-	if len(chainConfigInfos) == 0 {
+
+	if len(allChainConfigInfos) == 0 {
 		// That's a legitimate case if there are no chain configs on chain yet
 		r.lggr.Warnw("no on chain configs found")
 		return nil
 	}
-	r.setState(convertOnChainConfigToHomeChainConfig(chainConfigInfos))
+	r.setState(convertOnChainConfigToHomeChainConfig(allChainConfigInfos))
```
elatoskinas, line 145: q: why are we skipping state sets when there are no configs? Isn't an empty array a valid state?

internal/reader/home_chain.go
defistar: will be moving to constant on top of same file

internal/reader/home_chain.go
defistar: it returns anempty array in absence of any data, will never throw errors
there could be scenarios due to rpc issues

internal/reader/home_chain.go
defistar: this was an existing approach, yeah in case if the contract function returns empty array
we can still set it but isnt it redundant execution where we can retain initial state as-is
unless there are entries earluer and for some reason the config is removed and state is stale
is this a valid scenario?

elatoskinas: It could save a null pointer reference though down the line - my thinking is that would be less error-prone

defistar: updated to set in case if results are empty or non-empty

internal/reader/home_chain.go
```diff
@@ -112,19 +115,43 @@ func (r *homeChainPoller) poll() {
...
+
+		if len(chainConfigInfos) == 0 {
+			break
+		}
+
+		allChainConfigInfos = append(allChainConfigInfos, chainConfigInfos...)
+		pageIndex++
+	}
+
+	if len(allChainConfigInfos) >= 0 {
```
elatoskinas, line 145: suggestion: the check is redundant - the length of an array will always be >= 0 - we can remove it

internal/reader/home_chain.go
```diff
@@ -16,7 +16,10 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 )
 
-//go:generate mockery --name HomeChain --output ./mocks/ --case underscore
```
elatoskinas, line 19: q: shouldn't we keep this? Are the mocks generated somewhere else?

internal/reader/home_chain.go
defistar: removed redundant check

internal/reader/home_chain.go
defistar: removed the obselete comment

elatoskinas (approved):

internal/reader/home_chain.go
```diff
@@ -115,24 +118,42 @@ func (r *homeChainPoller) poll() {
...
-		nil,
-		&chainConfigInfos,
-	)
-	if err != nil {
-		return err
+	var allChainConfigInfos []ChainConfigInfo
+	pageIndex := uint64(0)
+
+	for {
+		var chainConfigInfos []ChainConfigInfo
```
dimkouv, line 125: Can we add something like the following for extra safety?

```
const pageIndexHardLimit = 100

for {
   ...
   pageIndex++

   if pageIndex > pageIndexHardLimit {
      return fmt.Errorf("something is wrong!")
   }
}
```

internal/reader/home_chain.go
```diff
@@ -115,24 +118,42 @@ func (r *homeChainPoller) poll() {
...
+			consts.MethodNameGetAllChainConfigs,
+			primitives.Unconfirmed,
+			map[string]interface{}{
+				"pageIndex": pageIndex,
+				"pageSize":  PageSize,
+			},
+			&chainConfigInfos,
+		)
+		if err != nil {
+			return err
```
dimkouv, line 138: let's wrap that err `"get config index:1 pagesize:2: <err>"`

internal/reader/home_chain.go
```diff
@@ -19,7 +19,10 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
 )
 
-//go:generate mockery --name HomeChain --output ./mocks/ --case underscore
+const (
+	PageSize = uint64(500)
```
dimkouv, line 23: no need to export, also better to rename `defaultConfigPageSize`.
Isn't 500 too much?, I would expect something closer to `10`

internal/reader/home_chain.go
defistar: there was a comment added to use mockery command. that was not supposed to be here.it must be in readme doc.
so deleted that from code @dimkouv

internal/reader/home_chain.go
defistar: u[pdated with error wrapping

internal/reader/home_chain.go
defistar: u[pdated

internal/reader/home_chain.go
defistar: added extra safety with hardCap for pageIndex
its included in for loop and once for loop exits we check if the pageIndex > pageIndexHardLimit
log a warn message

but wouldnt this cap the max chains we can extract the config from?

@dimkouv

internal/reader/home_chain.go
dimkouv: I think we should keep the go:generate here.
Created the ticket CCIP-2901 that @winder will work on.

internal/reader/home_chain.go
dimkouv: We can add some big value for the limit.

internal/reader/home_chain.go
defistar: retained the comment back

execute/plugin_e2e_test.go
```diff
@@ -78,18 +78,28 @@ type nodeSetup struct {
 
 func setupHomeChainPoller(lggr logger.Logger, chainConfigInfos []reader.ChainConfigInfo) reader.HomeChain {
```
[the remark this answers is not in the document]
defistar: should we create a JIRA task for this to followup? @dimkouv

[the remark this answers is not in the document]
dimkouv: yes, create one pls

dimkouv (approved):

internal/reader/home_chain.go
elatoskinas: Why do we need a hard limit, as opposed to just giving a warning and fetching everything?

elatoskinas: I think it would be painful to debug if we don't return all pages and just emit a warning (as opposed to either returning all pages, or throwing an exception)

elatoskinas (approved):

PR #25, opened 2024-07-05
update chainlink-common

makramkd:
Update chainlink-common to use the family-agnostic types from https://github.com/smartcontractkit/chainlink-common/pull/639.

### DIFF
--- a/execute/plugin_test.go
+++ b/execute/plugin_test.go
@@ -309,15 +309,15 @@ func Test_selectReport(t *testing.T) {
 		{
 			name: "half report",
 			args: args{
-				maxReportSize: 2600,
+				maxReportSize: 2300,
 				reports: []plugintypes.ExecutePluginCommitDataWithMessages{
 					makeTestCommitReport(10, 1, 100, 999, 10101010101, nil),
 				},
 			},
 			expectedExecReports:   1,
 			expectedCommitReports: 1,
-			expectedExecThings:    []int{8},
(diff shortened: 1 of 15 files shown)


execute/plugin_test.go
```diff
@@ -315,8 +316,8 @@ func Test_selectReport(t *testing.T) {
 			},
 			expectedExecReports:   1,
 			expectedCommitReports: 1,
-			expectedExecThings:    []int{5},
-			lastReportExecuted:    []cciptypes.SeqNum{100, 101, 102, 103, 104},
+			expectedExecThings:    []int{8},
+			lastReportExecuted:    []cciptypes.SeqNum{100, 101, 102, 103, 104, 105, 106, 107},
```
makramkd, line 320: Not sure why this would change, is it because it marshals to less JSON bytes?

winder: Yes exactly, this is potentially going to be a bit annoying until things settle down.

Could you fix this by adjusting `maxReportSize` instead of adjusting which messages were executed? While this still tests the same edge cases as before, that wont always be the case. Also it would keep the name "half report" accurate. Looks like you figured out that approach for most of the later tests.

makramkd: Yeah, that makes sense. I've fixed these to only tweak the sizes when needed.

winder (approved):

dimkouv (approved):

PR #26, opened 2024-07-08
Exec OCR3Runner Test Boilerplate and Misc

winder:
This PR implements some boilerplate code to setup the OCR3Runner test for the exec plugin.

Some changes have been made in order for the plugin to run cleanly. Strictly speaking these 3 changes could be separate PRs. See also inline comments pointing out each of these changes:
* use home-chain for configuration
    * functions like "supportsDestChain" borrowed from commit.
    * validation types changed to match home chain types.
* allow a nil previous outcome
* basic implementations for ShouldAcceptAttestedReport and ShouldTransmitAcceptedReport

### DIFF
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -219,7 +219,7 @@ func (p *Plugin) ValidateObservation(
 		return fmt.Errorf("decode observation: %w", err)
 	}
 
-	supportedChains, err := p.supportedChainsForOracle(ao.Observer)
+	supportedChains, err := p.supportedChains(ao.Observer)
 	if err != nil {
 		return fmt.Errorf("error finding supported chains by node: %w", err)
 	}
@@ -630,11 +630,7 @@ func (p *Plugin) Close() error {
 	panic("implement me")
 }
 
-func (p *Plugin) supportedChains() (mapset.Set[cciptypes.ChainSelector], error) {
-	return p.supportedChainsForOracle(p.reportingCfg.OracleID)
-}
-
-func (p *Plugin) supportedChainsForOracle(id commontypes.OracleID) (mapset.Set[cciptypes.ChainSelector], error) {
+func (p *Plugin) supportedChains(id commontypes.OracleID) (mapset.Set[cciptypes.ChainSelector], error) {
 	p2pID, exists := p.oracleIDToP2pID[id]
 	if !exists {
 		return nil, fmt.Errorf("oracle ID %d not found in oracleIDToP2pID", p.reportingCfg.OracleID)
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -27,7 +27,6 @@ import (
 
 // maxReportSizeBytes that should be returned as an execution report payload.
 const maxReportSizeBytes = 250_000
-const SentinelRoot = "0x289502ebf164ea87871e19cd9c8734016e954539fffc8c789ac4dadc7efcd933"
 
 // Plugin implements the main ocr3 plugin logic.
 type Plugin struct {
@@ -345,12 +344,9 @@ func buildSingleChainReport(
 	// Verify merkle root.
 	hash := tree.Root()
 	if !bytes.Equal(hash[:], report.MerkleRoot[:]) {
-		// For testing purposees, we allow a sentinel root to be used.
-		if report.MerkleRoot.String() != SentinelRoot {
-			actualStr := "0x" + hex.EncodeToString(hash[:])
-			return cciptypes.ExecutePluginReportSingleChain{}, 0,
-				fmt.Errorf("merkle root mismatch: expected %s, got %s", report.MerkleRoot.String(), actualStr)
-		}
+		actualStr := "0x" + hex.EncodeToString(hash[:])
+		return cciptypes.ExecutePluginReportSingleChain{}, 0,
+			fmt.Errorf("merkle root mismatch: expected %s, got %s", report.MerkleRoot.String(), actualStr)
 	}
 
 	// Iterate sequence range and executed messages to select messages to execute.
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -455,6 +455,11 @@ func selectReport(
 	accumulatedSize := 0
 	var finalReports []cciptypes.ExecutePluginReportSingleChain
 	for reportIdx, report := range reports {
+		// Reports at the end may not have messages yet.
+		if len(report.Messages) == 0 {
+			break
+		}
+
 		execReport, encodedSize, updatedReport, err :=
 			buildSingleChainReportMaxSize(ctx, lggr, hasher, tokenDataReader, encoder,
 				report, maxReportSizeBytes-accumulatedSize)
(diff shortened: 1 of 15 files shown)


execute/plugin.go
```diff
@@ -137,24 +138,34 @@ func getPendingExecutedReports(
...
-	if err != nil {
-		return types.Observation{}, err
+	var err error
+	var previousOutcome plugintypes.ExecutePluginOutcome
+
+	if outctx.PreviousOutcome != nil {
+		previousOutcome, err = plugintypes.DecodeExecutePluginOutcome(outctx.PreviousOutcome)
+		if err != nil {
+			return types.Observation{}, err
+		}
```
winder, line 148: Allow nil previous outcome, it is expected for a cold start.

execute/plugin.go
```diff
@@ -137,24 +138,34 @@ func getPendingExecutedReports(
...
 	}
 
 	// Phase 1: Gather commit reports from the destination chain and determine which messages are required to build a
 	//          valid execution report.
-	ownConfig := p.cfg.ObserverInfo[p.reportingCfg.OracleID]
 	var groupedCommits plugintypes.ExecutePluginCommitObservations
-	if slices.Contains(ownConfig.Reads, p.cfg.DestChain) {
+	supportsDest, err := p.supportsDestChain()
```
winder, line 154: Use home chain provider to fetch configuration.

execute/plugin.go
```diff
@@ -325,9 +341,20 @@ func buildSingleChainReport(
 		return cciptypes.ExecutePluginReportSingleChain{}, 0,
 			fmt.Errorf("unable to construct merkle tree from messages for report (%s): %w", report.MerkleRoot.String(), err)
 	}
-	numMsgs := len(report.Messages)
+
+	// Verify merkle root.
```
winder, line 345: Move this check to earlier to be as early as possible.

execute/plugin.go
```diff
@@ -438,6 +455,11 @@ func selectReport(
 	accumulatedSize := 0
 	var finalReports []cciptypes.ExecutePluginReportSingleChain
 	for reportIdx, report := range reports {
+		// Reports at the end may not have messages yet.
+		if len(report.Messages) == 0 {
+			break
+		}
```
winder, line 461: Empty commit reports are expected due to the two phase report/message reading. Break out of processing rather than erroring.

execute/plugin.go
```diff
@@ -562,18 +589,72 @@ func (p *Plugin) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]ocr3types.R
 func (p *Plugin) ShouldAcceptAttestedReport(
 	ctx context.Context, u uint64, r ocr3types.ReportWithInfo[[]byte],
 ) (bool, error) {
-	panic("implement me")
```
winder, line 565: Basic implementations for ShouldAcceptAttestedReport and ShouldTransmitAcceptedReport

execute/plugin_functions.go
```diff
@@ -17,29 +17,20 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 
 	"github.com/smartcontractkit/chainlink-ccip/execute/internal/validation"
-	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
 	"github.com/smartcontractkit/chainlink-ccip/plugintypes"
 )
 
 // validateObserverReadingEligibility checks if the observer is eligible to observe the messages it observed.
 func validateObserverReadingEligibility(
```
winder, line 24: This function changes with the home chain provider.

execute/plugin.go
```diff
@@ -562,18 +589,68 @@ func (p *Plugin) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]ocr3types.R
...
+
+	return supportedChains, nil
+}
+
+func (p *Plugin) supportsDestChain() (bool, error) {
+	destChainConfig, err := p.homeChain.GetChainConfig(p.cfg.DestChain)
+	if err != nil {
+		return false, fmt.Errorf("get chain config: %w", err)
+	}
+	return destChainConfig.SupportedNodes.Contains(p.oracleIDToP2pID[p.reportingCfg.OracleID]), nil
```
0xnogo, line 652: Should we also check that `p.oracleIDToP2pID[p.reportingCfg.OracleID]` exists? Like it was done in `supportedChains`?

execute/plugin_e2e_test.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+func TestPlugin(t *testing.T) {
+	ctx := context.Background()
+	lggr := logger.Test(t)
+
+	nodeSetups := setupSimpleTest(ctx, t, lggr, 1, 2)
+	//runner := testhelpers.NewOCR3Runner(nodes, nodeIDs, o)
+
+	nodesSetup := nodeSetups
```
0xnogo, line 32: nit/ is it useful?

execute/plugin_e2e_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+func GetP2pIDs(ids ...int) map[commontypes.OracleID]libocrtypes.PeerID {
+	res := make(map[commontypes.OracleID]libocrtypes.PeerID)
+	for _, id := range ids {
+		res[commontypes.OracleID(id)] = libocrtypes.PeerID{byte(id)}
+	}
+	return res
+}
+
+var ()
```
0xnogo, line 216: nit/ Can we remove this line?

internal/mocks/inmem/ccipreader_inmem.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/plugintypes"
+)
+
+type MessagesWithMetadata struct {
+	cciptypes.Message
+	Executed    bool
+	Destination cciptypes.ChainSelector
+}
+
+type InMemoryCCIPReader struct {
```
0xnogo, line 20: Is it intended to be used concurrently?

execute/plugin.go
```diff
@@ -325,9 +341,20 @@ func buildSingleChainReport(
...
 			fmt.Errorf("unable to construct merkle tree from messages for report (%s): %w", report.MerkleRoot.String(), err)
 	}
-	numMsgs := len(report.Messages)
+
+	// Verify merkle root.
+	hash := tree.Root()
+	if !bytes.Equal(hash[:], report.MerkleRoot[:]) {
+		// For testing purposees, we allow a sentinel root to be used.
+		if report.MerkleRoot.String() != SentinelRoot {
+			actualStr := "0x" + hex.EncodeToString(hash[:])
```
dimkouv, line 350: you can also do `cciptypes.Bytes32(hash).String()`

execute/plugin.go
```diff
@@ -325,9 +341,20 @@ func buildSingleChainReport(
 		return cciptypes.ExecutePluginReportSingleChain{}, 0,
 			fmt.Errorf("unable to construct merkle tree from messages for report (%s): %w", report.MerkleRoot.String(), err)
 	}
-	numMsgs := len(report.Messages)
+
+	// Verify merkle root.
+	hash := tree.Root()
+	if !bytes.Equal(hash[:], report.MerkleRoot[:]) {
+		// For testing purposees, we allow a sentinel root to be used.
+		if report.MerkleRoot.String() != SentinelRoot {
```
dimkouv, line 349: I am not a fan of having testing code in the prod logic.
What I understand is that you want to do:
```
if computed root does not match report root:
   if report root is a test root:
      we are withing a test so all good
   else:
       error
```


If there is no other way to test this you could instead do:
```
if !p.testMode() && !bytes.Equal(hash[:], report.MerkleRoot[:]) {
    return error
}
```

and set testMode to true only in the test. That way the package does also not need to export `SentinelRoot`

execute/plugin.go
dimkouv: maybe better to do:

```
chains, err := p.supportedChains(p.reportingCfg.OracleID)
if err != nil {
	return false, fmt.Errorf("error getting supported chains: %w", err)
}
return chains.Contains(p.cfg.DestChain), nil
```

execute/plugin_e2e_test.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+	"github.com/smartcontractkit/chainlink-ccip/plugintypes"
+)
+
+func TestPlugin(t *testing.T) {
+	ctx := context.Background()
+	lggr := logger.Test(t)
+
+	nodeSetups := setupSimpleTest(ctx, t, lggr, 1, 2)
+	//runner := testhelpers.NewOCR3Runner(nodes, nodeIDs, o)
```
dimkouv, line 30: remove commented out line

execute/plugin_e2e_test.go
```diff
@@ -0,0 +1,216 @@
...
+	for _, n := range nodesSetup {
+		nodeIDs = append(nodeIDs, n.node.reportingCfg.OracleID)
+	}
+
+	//o, err := tc.initialOutcome.Encode()
+	//require.NoError(t, err)
+	runner := testhelpers.NewOCR3Runner(nodes, nodeIDs, nil)
+
+	res, err := runner.RunRound(ctx)
+	fmt.Println(res, err)
```
dimkouv, line 48: I believe you want to do some assertions instead of just printing `res, err`

dimkouv:
Did you find the `OCR3Runner` helpful? Did you catch any bugs using it? Any suggestions for improving it?

execute/plugin_e2e_test.go
winder: I'll remove this.

internal/mocks/inmem/ccipreader_inmem.go
winder: I don't think it needs to work concurrently at the moment

execute/plugin_e2e_test.go
winder: Yes, I'm adding these in a followup PR.

execute/plugin.go
winder: Removed the sentinel value. Removed some randomness from the test setup so that the root can be calculated directly.

0xnogo (approved):

PR #29, opened 2024-07-10
commit plugin - fix wrong assumption that nextSeqNums returns the committed seq nums

dimkouv:
There was a wrong assumption that `nextSeqNums` returns the committed seqNums.

### DIFF
--- a/commit/plugin_e2e_test.go
+++ b/commit/plugin_e2e_test.go
@@ -275,11 +275,11 @@ func setupAllNodesReadAllChains(ctx context.Context, t *testing.T, lggr logger.L
 
 		// all nodes observe the same sequence numbers lastCommittedSeqNumA for chainA and lastCommittedSeqNumB for chainB
 		n.ccipReader.On("NextSeqNum", ctx, []cciptypes.ChainSelector{chainA, chainB}).
-			Return([]cciptypes.SeqNum{lastCommittedSeqNumA, lastCommittedSeqNumB}, nil)
+			Return([]cciptypes.SeqNum{lastCommittedSeqNumA + 1, lastCommittedSeqNumB + 1}, nil)
 
 		// transmission phase root staleness check passes
 		n.ccipReader.On("NextSeqNum", ctx, []cciptypes.ChainSelector{chainB}).
-			Return([]cciptypes.SeqNum{lastCommittedSeqNumB}, nil)
+			Return([]cciptypes.SeqNum{lastCommittedSeqNumB + 1}, nil)
 	}
 	require.NoError(t, homeChain.Close())
 	return nodes
@@ -302,7 +302,7 @@ func setupNodesDoNotAgreeOnMsgs(ctx context.Context, t *testing.T, lggr logger.L
 		nodes = append(nodes, n)
 		// all nodes observe the same sequence numbers lastCommittedSeqNumA for chainA and lastCommittedSeqNumB for chainB
 		n.ccipReader.On("NextSeqNum", ctx, []cciptypes.ChainSelector{chainA, chainB}).
-			Return([]cciptypes.SeqNum{lastCommittedSeqNumA, lastCommittedSeqNumB}, nil)
+			Return([]cciptypes.SeqNum{lastCommittedSeqNumA + 1, lastCommittedSeqNumB + 1}, nil)
 
 		// then they fetch new msgs, there is nothing new on chainA
 		mockMsgsBetweenSeqNums(ctx, n.ccipReader, chainA, seqNumA, emptyMsgs)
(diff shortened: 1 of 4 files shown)


asoliman92 (approved):

asoliman92 (commented):
Maybe change `lastCommittedSeqNumB+1` and the like to use `nextSeqNumB` variables instead? Not a blocker though but will be cleaner ;)

commit/plugin_e2e_test.go
```diff
@@ -275,11 +275,11 @@ func setupAllNodesReadAllChains(ctx context.Context, t *testing.T, lggr logger.L
 
 		// all nodes observe the same sequence numbers lastCommittedSeqNumA for chainA and lastCommittedSeqNumB for chainB
 		n.ccipReader.On("NextSeqNum", ctx, []cciptypes.ChainSelector{chainA, chainB}).
-			Return([]cciptypes.SeqNum{lastCommittedSeqNumA, lastCommittedSeqNumB}, nil)
+			Return([]cciptypes.SeqNum{lastCommittedSeqNumA + 1, lastCommittedSeqNumB + 1}, nil)
```
asoliman92, line 278: Maybe change `lastCommittedSeqNumB+1` and the like to use `nextSeqNumB` variables instead?

dimkouv: imo it's fine, keeps the test more simple by having the "LastCommitedSeqNums" in the test cases

commit/plugin_functions.go
```diff
@@ -35,17 +35,18 @@ func observeLatestCommittedSeqNums(
 	sort.Slice(knownSourceChains, func(i, j int) bool { return knownSourceChains[i] < knownSourceChains[j] })
 	latestCommittedSeqNumsObservation := make([]plugintypes.SeqNumChain, 0)
 	if readableChains.Contains(destChain) {
-		lggr.Debugw("reading latest committed sequence from destination")
-		onChainLatestCommittedSeqNums, err := ccipReader.NextSeqNum(ctx, knownSourceChains)
+		lggr.Debugw("reading sequence numbers", "chains", knownSourceChains)
+		onChainNextSeqNums, err := ccipReader.NextSeqNum(ctx, knownSourceChains)
```
makramkd, line 39: It took me a good 5 mins to understand this line. Its not clear at all that this makes a call to the destination chain, I think we should make our names more clear

makramkd: With the multiramp contracts the naming has kind of been lost but I prefer making this `ExpectedNextSequenceNumbers` or if we want to shorten it `ExpectedNextSeqNums`.

Another name I'd be fine with, because it lines up with the contracts, is `OfframpMinSeqNrs` which makes it clear we are reading from the dest + makes it clear what it is reading (min seq nr being defined as the expected next seq nr for the future commit report).

commit/plugin_functions.go
```diff
@@ -35,17 +35,18 @@ func observeLatestCommittedSeqNums(
 	sort.Slice(knownSourceChains, func(i, j int) bool { return knownSourceChains[i] < knownSourceChains[j] })
 	latestCommittedSeqNumsObservation := make([]plugintypes.SeqNumChain, 0)
 	if readableChains.Contains(destChain) {
-		lggr.Debugw("reading latest committed sequence from destination")
-		onChainLatestCommittedSeqNums, err := ccipReader.NextSeqNum(ctx, knownSourceChains)
+		lggr.Debugw("reading sequence numbers", "chains", knownSourceChains)
```
makramkd, line 38: Lets include as much context into the log as possible, maybe add `destChain` as well into the fields?

```suggestion
		lggr.Debugw("reading offramp min seq nrs", "destChain", destChain, "knownSourceChains", knownSourceChains)
```

As for the offramp address, I think we can include that as part of the logger's `Name` so it'll always be logged, so no need to include it here probably.

commit/plugin_functions.go
```diff
@@ -35,17 +35,18 @@ func observeLatestCommittedSeqNums(
...
+		onChainNextSeqNums, err := ccipReader.NextSeqNum(ctx, knownSourceChains)
 		if err != nil {
 			return latestCommittedSeqNumsObservation, fmt.Errorf("get next seq nums: %w", err)
 		}
-		lggr.Debugw("observed latest committed sequence numbers on destination",
-			"latestCommittedSeqNumsObservation", onChainLatestCommittedSeqNums)
+		lggr.Debugw("observing latest committed sequence numbers", "onChainNextSeqNums", onChainNextSeqNums)
```
makramkd, line 43: Lets add dest chain to this log as well.

commit/plugin_functions.go
dimkouv: there's a todo to include destChainSelector in the interface. that way it will become cleaner, where it reads from

winder (approved):
Changes look fine. Would be nice to add a new unit test that would have caught this, but I don't see a way to do that.

PR #32, opened 2024-07-17
Fix on ramp address type and pass chains to getConfig

dimkouv:


internal/libs/typconv/address.go
```diff
@@ -0,0 +1,9 @@
+package typconv
+
+import (
+	"encoding/hex"
+)
+
+func AddressBytesToString(addr []byte) string {
```
asoliman92, line 7: This is Ethereum specific AFAIU. Maybe rename it to indicate this?

internal/reader/ccip.go
```diff
@@ -381,8 +386,8 @@ func (r *CCIPChainReader) getSourceChainsConfig(
 	mu := new(sync.Mutex)
 
 	eg := new(errgroup.Group)
-	for chainSel := range r.contractReaders {
-		chainSel := chainSel
+	for _, ch := range chains {
+		chainSel := ch
```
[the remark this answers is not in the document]
dimkouv: done

internal/libs/typconv/address.go
dimkouv: it's not evm specific, just hex encoding.

makramkd (approved):

asoliman92 (approved):

PR #33, opened 2024-07-17
Refactoring: use builder pattern for execute report building

winder:
Move report building functions into a report package and move them behind a builder pattern interface.

This PR has no functional changes. Unit tests are practically identical to the prior code and are passing.


execute/temp/token_data_reader.go
```diff
@@ -0,0 +1,15 @@
...
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+// TokenDataReader is an interface for reading extra token data from an async process.
+// TODO: Build a token data reading process.
+//
+//go:generate mockery --quiet --name TokenDataReader --output ../../internal/mocks --case=underscore
+type TokenDataReader interface {
+	ReadTokenData(ctx context.Context, srcChain cciptypes.ChainSelector, num cciptypes.SeqNum) ([][]byte, error)
+}
```
winder, line 15: Moved to avoid circular dependency. There is already ticket to move this to a more permanent location.

execute/plugin.go
```diff
@@ -446,60 +253,40 @@ func selectReport(
...
-			return nil, nil, fmt.Errorf("unable to build single chain report: %w", err)
+			return nil, nil, fmt.Errorf("unable to add report to builder: %w", err)
 		}
-		reports[reportIdx] = updatedReport
-		accumulatedSize += encodedSize
-		finalReports = append(finalReports, execReport)
-
-		// partially executed report detected, stop adding more reports.
-		// TODO: do not break if messages were intentionally skipped.
-		if len(updatedReport.Messages) != len(updatedReport.ExecutedMessages) {
```
asoliman92, line 482: Where's the early break if we exceeded the size?

asoliman92 (approved):
I like it ❤️

execute/temp/token_data_reader.go
makramkd: Putting this in `execute/types` seems fine to me.

makramkd (approved):

PR #34, opened 2024-07-18
bump chainlink-common

0xnogo:


internal/reader/ccip.go
```diff
@@ -412,6 +411,7 @@ func (r *CCIPChainReader) getSourceChainsConfig(
 				ctx,
 				consts.ContractNameOffRamp,
 				consts.MethodNameGetSourceChainConfig,
+				primitives.Finalized,
```
0xnogo, line 414: New input param introduced. It represents the level of confidence of the data asked

makramkd: Interesting. I don't think this particular call needs to be finalized though

internal/reader/home_chain.go
```diff
@@ -116,7 +117,12 @@ func (r *homeChainPoller) poll() {
 func (r *homeChainPoller) fetchAndSetConfigs(ctx context.Context) error {
 	var chainConfigInfos []ChainConfigInfo
 	err := r.homeChainReader.GetLatestValue(
-		ctx, consts.ContractNameCCIPConfig, consts.MethodNameGetAllChainConfigs, nil, &chainConfigInfos,
+		ctx,
+		consts.ContractNameCCIPConfig,
+		consts.MethodNameGetAllChainConfigs,
+		primitives.Finalized,
```
makramkd, line 123: Ditto here, what other options are there?

0xnogo: Finalized or Unconfirmed. Why would we want an unfinalized data?

internal/reader/ccip.go
makramkd: Doesn't need to be finalized.

internal/reader/home_chain.go
makramkd: Does not need to be finalized, can use Unconfirmed.

internal/reader/onchain_prices_reader.go
```diff
@@ -54,7 +55,14 @@ func (pr *OnchainTokenPricesReader) GetTokenPricesUSD(
 			if staticPrice, exists := pr.TokenPriceConfig.StaticPrices[token]; exists {
 				price.Set(&staticPrice)
 			} else {
-				if err := pr.ContractReader.GetLatestValue(ctx, contractName, functionName, token, price); err != nil {
+				if err :=
+					pr.ContractReader.GetLatestValue(
+						ctx,
+						contractName,
+						functionName,
+						primitives.Finalized,
```
makramkd, line 63: Does not need to be finalized.

makramkd (approved):

internal/reader/onchain_prices_reader.go
```diff
@@ -54,7 +55,14 @@ func (pr *OnchainTokenPricesReader) GetTokenPricesUSD(
 			if staticPrice, exists := pr.TokenPriceConfig.StaticPrices[token]; exists {
 				price.Set(&staticPrice)
 			} else {
-				if err := pr.ContractReader.GetLatestValue(ctx, contractName, functionName, token, price); err != nil {
+				if err :=
```
winder, line 58: style nit: I'm personally not a fan of "one liner" if conditions when they span multiple lines.

winder (approved):
Changes look fine. Just had a minor suggestion for you to consider.

asoliman92 (approved):

PR #35, opened 2024-07-18
Change SendRequestedEvent type in CCIP reader query

asoliman92:


internal/reader/ccip.go
```diff
@@ -239,27 +239,17 @@ func (r *CCIPChainReader) MsgsBetweenSeqNums(
 		return nil, err
 	}
 
+	type SendRequestedEvent struct {
```
makramkd, line 242: Lets add a TODO / create a ticket in the common epic to look into using modifiers instead of this.

internal/reader/ccip.go
```diff
@@ -239,27 +239,17 @@ func (r *CCIPChainReader) MsgsBetweenSeqNums(
...
-								Value:    seqNumRange.Start().String(),
-								Operator: primitives.Gte,
-							},
-							{
-								Value:    seqNumRange.End().String(),
-								Operator: primitives.Lte,
-							},
-						},
-					},
-				},
```
makramkd, line 262: If we're going to remove this filter lets do the filtering manually in the for loop below

internal/reader/ccip.go
ilija42: the easiest way would be to modify the ABI string to so that it matches just cciptypes.message

makramkd: cciptypes.Message is the "Any2Any" message type for CCIP so we'll likely always have to do some translation from whats onchain to whats in cciptypes.Message, but we can figure that our later

internal/reader/ccip.go
ilija42, line 262: did we try this with the filter? It should work right? The problem wasn't in the filter

PR #36, opened 2024-07-19
add ocr offchainConfig and chain config structs

makramkd:
We need to define the offchainConfig and chain config structs. The configuration has been organized so as to make maximal use of blue-green deployments. Things in `ChainConfig` should not be changed very often if at all, and will affect all DONs.

Things in OCR offchainConfig will be scoped to a particular plugin instance.

Related PR in ccip: https://github.com/smartcontractkit/ccip/pull/1264


pluginconfig/execute.go
```diff
@@ -3,27 +3,42 @@ package pluginconfig
...
+	commonconfig "github.com/smartcontractkit/chainlink-common/pkg/config"
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 )
 
 type ExecutePluginConfig struct {
 	// DestChain is the ccip destination chain configured for the execute DON.
 	DestChain cciptypes.ChainSelector `json:"destChain"`
 
-	// ObserverInfo is a map of oracle IDs to ObserverInfo.
-	ObserverInfo map[commontypes.OracleID]ObserverInfo `json:"observerInfo"`
```
makramkd, line 16: Was not being used, so 🪓

chainconfig/chainconfig.go
```diff
@@ -0,0 +1,28 @@
...
+	// written data-availability gas price on-chain before we write a new data-availability
+	// gas price.
+	// Note that this is only applicable for some chains, such as L2's.
+	DAGasPriceDeviationPPB cciptypes.BigInt `json:"daGasPriceDeviationPPB"`
+
+	// TODO: do we want something like finality tag enabled / finality depth here?
+	// FinalityDepth is the number of blocks on top of a block that need to be included
+	// before a block is considered finalized.
+	// If set to 0, finality tags will be used.
+	FinalityDepth uint64 `json:"finalityDepth"`
```
makramkd, line 27: @connorwstein in v1.5 we have SourceFinalityDepth and DestFinalityDepth defined in the offchainConfig (as well as "optimistic confirmations"). SourceFinalityDepth seems like it can be a chain config setting, whereas dest optimistic confirmations may be chain config level setting but not sure if we'd want a bit more flexibility on it?

pluginconfig/execute.go
```diff
@@ -3,27 +3,42 @@ package pluginconfig
...
 
-type ObserverInfo struct {
-	// Writer indicates that the node can contribute by sending reports to the destination chain.
-	// Being a Writer guarantees that the node can also read from the destination chain.
-	Writer bool `json:"writer"`
+type ExecuteReportingPluginConfig struct {
+	// DestOptimisticConfirmations is how many confirmations to wait for the dest chain event before we consider it
+	// confirmed (optimistically, need not be finalized).
+	// TODO: seems like a good candiate for chain config?
+	DestOptimisticConfirmations uint32 `json:"destOptimisticConfirmations"`
```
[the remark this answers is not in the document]
makramkd: Moved to chain config.

pluginconfig/execute.go
```diff
@@ -25,11 +22,53 @@ type ExecutePluginConfig struct {
 	MessageVisibilityInterval time.Duration `json:"messageVisibilityInterval"`
 }
 
-type ObserverInfo struct {
-	// Writer indicates that the node can contribute by sending reports to the destination chain.
-	// Being a Writer guarantees that the node can also read from the destination chain.
-	Writer bool `json:"writer"`
+type ExecuteOffchainConfig struct {
```
winder, line 25: Whats the difference between this and `ExecutePluginConfig`? There is some overlap in values.

makramkd: This will get posted on-chain and will inform what ends up in `ExecutePluginConfig`

winder: Could you add some comments to the struct to this effect?

pluginconfig/commit.go
```diff
@@ -27,20 +27,107 @@ type CommitPluginConfig struct {
...
 	if c.NewMsgScanBatchSize == 0 {
 		return fmt.Errorf("newMsgScanBatchSize not set")
 	}
 
+	return c.OffchainConfig.Validate()
+}
+
+// ArbitrumPriceSource is the source of the TOKEN/USD price data of a particular token
+// on Arbitrum.
+type ArbitrumPriceSource struct {
```
winder, line 49: Why do we need this?

winder (approved):
Changes seem fine. A lot of them aren't used yet, how big of an issue would it be to add them in as needed? I'd be tempted to throw in lots of TODO comments on the things which still need to be implemented.

pluginconfig/commit.go
makramkd: Can add some more detailed comments.

makramkd:
Agreed @winder I've created a few tickets for what needs to be used.

pluginconfig/execute.go
makramkd: I've ended up embedding this struct in the `ExecutePluginConfig` struct, which holds configuration not just from the offchain config but from the general OCR config.

pluginconfig/execute.go
```diff
@@ -1,35 +1,89 @@
...
-	ObserverInfo map[commontypes.OracleID]ObserverInfo `json:"observerInfo"`
-
 	// SyncTimeout is the timeout for syncing the commit plugin reader.
 	SyncTimeout time.Duration `json:"syncTimeout"`
 
 	// SyncFrequency is the frequency at which the commit plugin reader should sync.
 	SyncFrequency time.Duration `json:"syncFrequency"`
 
+	// OffchainConfig is the offchain config set for the exec DON.
+	OffchainConfig ExecuteOffchainConfig `json:"offchainConfig"`
```
winder, line 26: nice, I was wondering if they overlapped completely like this.

You could fully embed it as well, most off the shelf decoders will append changes rather than overwrite them. Not sure if our custom decoders work that way.
```suggestion
	// ExecuteOffchainConfig is the offchain config set for the exec DON.
	ExecuteOffchainConfig`
```

asoliman92 (approved):

0xnogo (approved):

PR #40, opened 2024-07-19
exec: Skip messages that do not fit

winder:
Three main changes:
1. Rewrite `selectReport` test to use the report builder, move to report builder package.
2. Update `buildSingleChainReport` to select reports by index rather than by a maximum number of messages to select.
3. Introduce `checkMesssage` and `verifyReport` functions to contain business logic to determine whether messages are ready to be executed and if the final report is valid.

### DIFF
--- a/execute/report/builder.go
+++ b/execute/report/builder.go
@@ -79,7 +79,7 @@ func (b *execReportBuilder) Add(
 		return commitReport, nil
 	}
 	if err != nil {
-		return commitReport, fmt.Errorf("unable to build single chain report: %w", err)
+		return commitReport, fmt.Errorf("unable to add a single chain report: %w", err)
 	}
 
 	b.execReports = append(b.execReports, execReport)
--- a/execute/report/builder.go
+++ b/execute/report/builder.go
@@ -72,7 +72,7 @@ type execReportBuilder struct {
 func (b *execReportBuilder) Add(
 	commitReport plugintypes.ExecutePluginCommitDataWithMessages,
 ) (plugintypes.ExecutePluginCommitDataWithMessages, error) {
-	execReport, updatedReport, err := b.buildSingleChainReportMaxSize(b.ctx, commitReport)
+	execReport, updatedReport, err := b.buildSingleChainReport(b.ctx, commitReport)
 
 	// No messages fit into the report, move to next report
 	if errors.Is(err, ErrEmptyReport) {
--- a/execute/report/builder.go
+++ b/execute/report/builder.go
@@ -41,6 +41,14 @@ func NewBuilder(
 	}
 }
 
+// validationMetadata contains all metadata needed to accumulate results across multiple reports and messages.
+type validationMetadata struct {
+	encodedSizeBytes uint64
+
+	// TODO: gas limit
+	//gas             uint64
+}
+
 type execReportBuilder struct {
 	ctx  context.Context // TODO: remove context from builder so that it can be pure?
 	lggr logger.Logger
@@ -55,9 +63,7 @@ type execReportBuilder struct {
 	maxGas             uint64
 
 	// State
-	reportSizeBytes uint64
-	// TODO: gas limit
-	//gas             uint64
+	accumulated validationMetadata
 
 	// Result
 	execReports []cciptypes.ExecutePluginReportSingleChain
@@ -66,11 +72,7 @@ type execReportBuilder struct {
 func (b *execReportBuilder) Add(
 	commitReport plugintypes.ExecutePluginCommitDataWithMessages,
 ) (plugintypes.ExecutePluginCommitDataWithMessages, error) {
-	// TODO: buildSingleChainReportMaxSize needs to be part of the builder in order to access state.
-	execReport, encodedSize, updatedReport, err :=
-		buildSingleChainReportMaxSize(b.ctx, b.lggr, b.hasher, b.tokenDataReader, b.encoder,
-			commitReport,
-			int(b.maxReportSizeBytes-b.reportSizeBytes))
+	execReport, updatedReport, err := b.buildSingleChainReportMaxSize(b.ctx, commitReport)
 
 	// No messages fit into the report, move to next report
 	if errors.Is(err, ErrEmptyReport) {
@@ -80,7 +82,6 @@ func (b *execReportBuilder) Add(
 		return commitReport, fmt.Errorf("unable to build single chain report: %w", err)
 	}
 
-	b.reportSizeBytes += uint64(encodedSize)
 	b.execReports = append(b.execReports, execReport)
 
 	return updatedReport, nil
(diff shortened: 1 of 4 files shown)


execute/report/report_test.go
```diff
@@ -199,24 +408,288 @@ func Test_buildSingleChainReport_Errors(t *testing.T) {
...
+				updatedMessage, err := builder.Add(report)
+				if err != nil && tt.wantErr != "" {
+					if strings.Contains(err.Error(), tt.wantErr) {
+						foundError = true
+						break
+					}
+				}
+				require.NoError(t, err)
+				updatedMessages = append(updatedMessages, updatedMessage)
+			}
```
winder, line 668: Except for this part that swaps `selectReports` with the new builder, this test is almost identical to the previous version. There are 2 new test cases to cover some "skip message that wont fit" cases.

execute/report/report_test.go
```diff
@@ -158,25 +386,6 @@ func Test_buildSingleChainReport_Errors(t *testing.T) {
...
-					Messages: []cciptypes.Message{
-						{Header: cciptypes.RampMessageHeader{
-							SourceChainSelector: 1234567,
-							SequenceNumber:      cciptypes.SeqNum(100),
-						}},
-					},
-				},
-				codec: badCodec{},
-			},
-		},
```
winder, line 179: This case will return when I test the new `checkMessage` and `verifyReport` methods.

execute/report/report.go
```diff
@@ -118,81 +121,144 @@ func buildSingleChainReport(
...
+	return true, validationMetadata{
+		encodedSizeBytes: uint64(len(encoded)),
+	}, nil
 }
 
-// buildSingleChainReportMaxSize generates the largest report which fits into maxSizeBytes.
+// buildSingleChainReport generates the largest report which fits into maxSizeBytes.
 // See buildSingleChainReport for more details about how a report is built.
-func buildSingleChainReportMaxSize(
+func (b *execReportBuilder) buildSingleChainReport(
```
winder, line 191: Changed to be a receiver method, so most of the arguments go away. Some functionality has been moved to the new `checkMessage` and `verifyReport` methods.

execute/report/report.go
```diff
@@ -118,81 +121,144 @@ func buildSingleChainReport(
...
+		return finalize(finalReport, report, meta)
 	}
 
-	var searchErr error
-	idx := sort.Search(len(report.Messages), func(mid int) bool {
-		if searchErr != nil {
-			return false
+	finalReport = cciptypes.ExecutePluginReportSingleChain{}
+	msgs := make(map[int]struct{})
+	for i := range report.Messages {
```
winder, line 224: @asoliman92  this is where `sort.Search` is replaced. The algorithm now adds messages in one at a time rather than attempting a binary search. I'm still not entirely happy with either approach.

execute/report/report.go
```diff
@@ -118,81 +121,144 @@ func buildSingleChainReport(
...
+	// TODO: get rid of the nolint when the error is used
+) (messageStatus, error) { // nolint this will use the error eventually
+	if slices.Contains(execReport.ExecutedMessages, execReport.Messages[idx].Header.SequenceNumber) {
+		return AlreadyExecuted, nil
+	}
+
+	return ReadyToExecute, nil
+}
+
+func (b *execReportBuilder) verifyReport(
```
winder, line 161: Notice `validationMetadata`, this type holds the encoded size and will also hold the gas limit and perhaps the fee boost ratio. It is computed per report, and accumulated as part of the builder.

execute/report/report.go
```diff
@@ -28,12 +28,16 @@ func buildSingleChainReport(
 	lggr logger.Logger,
 	hasher cciptypes.MessageHasher,
 	tokenDataReader types.TokenDataReader,
-	encoder cciptypes.ExecutePluginCodec,
 	report plugintypes.ExecutePluginCommitDataWithMessages,
-	maxMessages int,
-) (cciptypes.ExecutePluginReportSingleChain, int, error) {
-	if maxMessages == 0 {
-		maxMessages = len(report.Messages)
+	messages map[int]struct{},
```
winder, line 32: Switched from taking the first N messages, to taking specific messages. The "executed" logic will be moved into `checkMessage`.

execute/report/builder.go
```diff
@@ -66,21 +72,16 @@ type execReportBuilder struct {
 func (b *execReportBuilder) Add(
 	commitReport plugintypes.ExecutePluginCommitDataWithMessages,
 ) (plugintypes.ExecutePluginCommitDataWithMessages, error) {
-	// TODO: buildSingleChainReportMaxSize needs to be part of the builder in order to access state.
-	execReport, encodedSize, updatedReport, err :=
-		buildSingleChainReportMaxSize(b.ctx, b.lggr, b.hasher, b.tokenDataReader, b.encoder,
-			commitReport,
-			int(b.maxReportSizeBytes-b.reportSizeBytes))
+	execReport, updatedReport, err := b.buildSingleChainReport(b.ctx, commitReport)
```
winder, line 76: Notice that the accumulator is gone, it's now baked into `buildSingleChainReport` and stored as part of an accumulator object.

execute/report/report.go
0xnogo: nit/ Both the internal and the interface `buildSingleChainReport` methods are named the same.

What do you think of `buildSingleChainReportHelper`?

winder: Good idea, renamed.

execute/report/report.go
```diff
@@ -118,81 +121,144 @@ func buildSingleChainReport(
...
+				plugintypes.ExecutePluginCommitDataWithMessages{},
+				fmt.Errorf("unable to verify report: %w", err)
+		} else if validReport {
 			finalReport = finalReport2
-			encodedSize = encodedSize2
-			return false // not full
+			meta = meta2
+		} else {
+			// this message didn't work, continue to the next one
+			delete(msgs, i)
```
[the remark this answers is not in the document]
0xnogo: It is not as this is not dealing with `nonce` but msg id

execute/report/report.go
```diff
@@ -118,81 +121,144 @@ func buildSingleChainReport(
...
+		commitReport plugintypes.ExecutePluginCommitDataWithMessages,
+		meta validationMetadata,
+	) (cciptypes.ExecutePluginReportSingleChain, plugintypes.ExecutePluginCommitDataWithMessages, error) {
+		b.accumulated.encodedSizeBytes += meta.encodedSizeBytes
+		commitReport = markNewMessagesExecuted(execReport, commitReport)
+		return execReport, commitReport, nil
+	}
+	// Attempt to include all messages in the report.
+	finalReport, err :=
+		buildSingleChainReport(b.ctx, b.lggr, b.hasher, b.tokenDataReader, report, nil)
```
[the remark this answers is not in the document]
winder: Good idea, added a log for now.

0xnogo (approved):

PR #46, opened 2024-07-25
Bind onramps in exec plugin background

dimkouv:


execute/plugin.go
```diff
@@ -56,11 +62,13 @@ func NewPlugin(
 	lggr logger.Logger,
 ) *Plugin {
 	lastReportTS := &atomic.Int64{}
-	lastReportTS.Store(time.Now().Add(-cfg.MessageVisibilityInterval).UnixMilli())
+	// lastReportTS.Store(time.Now().Add(-cfg.MessageVisibilityInterval).UnixMilli())
+	// TODO: Temporary while figuring out the best way to inject it in tests.
+	lastReportTS.Store(time.Unix(0, 0).UnixMilli())
```
makramkd, line 67: q: can't we just set this field in the struct for the test?

execute/plugin.go
```diff
@@ -444,5 +479,19 @@ func (p *Plugin) supportsDestChain() (bool, error) {
...
+	}
+	return configuredValue
+}
+
+func syncTimeout(configuredValue time.Duration) time.Duration {
+	if configuredValue.Milliseconds() == 0 {
+		return 3 * time.Second
+	}
+	return configuredValue
+}
```
makramkd, line 494: Worth making these common or would the defaults be different?

execute/plugin.go
```diff
@@ -72,6 +80,21 @@ func NewPlugin(
...
+	p.bgSyncWG = &sync.WaitGroup{}
+	p.bgSyncWG.Add(1)
+	plugincommon.BackgroundReaderSync(
+		bgSyncCtx,
+		p.bgSyncWG,
+		lggr,
+		ccipReader,
+		syncTimeout(cfg.SyncTimeout),
+		time.NewTicker(syncFrequency(p.cfg.SyncFrequency)).C,
+	)
```
makramkd, line 95: idea: maybe this background reader syncer could be an object instead and the bg sync cancel / waitgroup could be implementation details instead? It'd have a `Close` method and a `Start` method similar to the `ServiceCtx`?

makramkd (approved):
LGTM, just some potential improvements

execute/plugin.go
dimkouv: could be different, but atm they should be the same, since they serve the same purpose

execute/plugin.go
dimkouv: done

internal/plugincommon/ccipreader.go
```diff
@@ -0,0 +1,121 @@
...
+	reader reader.CCIP,
+	syncTimeout time.Duration,
+	syncFrequency time.Duration,
+) *BackgroundReaderSyncer {
+
+	return &BackgroundReaderSyncer{
+		lggr:          lggr,
+		reader:        reader,
+		syncTimeout:   syncTimeout,
+		syncFrequency: syncFrequency,
```
0xnogo, line 36: nit/ Do we need to add some validation here? (i.e. syncTimeout > syncFrequency)?

Same for loggr and reader not nil?

internal/plugincommon/ccipreader.go
```diff
@@ -0,0 +1,121 @@
...
+	ctx context.Context,
+	wg *sync.WaitGroup,
+	lggr logger.Logger,
+	reader reader.CCIP,
+	syncTimeout time.Duration,
+	ticker <-chan time.Time,
+) {
+	go func() {
+		defer wg.Done()
+
```
0xnogo, line 85: nit/ Do we need `defer ticker.Stop()` to ensure the ticker is stopped and resources are freed?

0xnogo (approved):

internal/plugincommon/ccipreader.go
dimkouv: Yes, makes sense.
Just that the ticker should be managed by the caller that provided it to this function.

PR #50, opened 2024-07-29
exec: use checkMessage for isExecuted and preparing token data.

winder:
Move logic for checking if a message is executed and preparing token data into the new checkMessage function.

Add tests for checkMessage and verifyReport.

### DIFF
--- a/execute/report/report.go
+++ b/execute/report/report.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"slices"
+	"strings"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 
@@ -22,8 +23,7 @@ import (
 // The hasher and encoding codec are provided as arguments to allow for chain-specific formats to be used.
 //
 // The messages argument indicates which messages should be included in the report. If messages is empty
-// all messages will be included. This allows the caller to create smaller reports if needed. Executed messages
-// are skipped automatically.
+// all messages will be included. This allows the caller to create smaller reports if needed.
 func buildSingleChainReportHelper(
 	ctx context.Context,
 	lggr logger.Logger,
@@ -39,6 +39,12 @@ func buildSingleChainReportHelper(
 		for i := 0; i < len(report.Messages); i++ {
 			messages[i] = struct{}{}
 		}
+	}
+
+	numMsg := len(report.Messages)
+	if len(report.TokenData) != numMsg {
+		return cciptypes.ExecutePluginReportSingleChain{},
+			fmt.Errorf("token data length mismatch: got %d, expected %d", len(report.TokenData), numMsg)
 	}
 
 	lggr.Infow(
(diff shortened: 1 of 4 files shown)


plugintypes/execute.go
```diff
@@ -31,9 +31,8 @@ type ExecutePluginCommitData struct {
 	// ExecutedMessages are the messages in this report that have already been executed.
 	ExecutedMessages []cciptypes.SeqNum `json:"executedMessages"`
 
-	// TODO: cache for token data.
 	// TokenData for each message.
-	//TokenData [][][]byte `json:"-"`
+	TokenData [][][]byte `json:"-"`
```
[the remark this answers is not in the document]
makramkd: I thought it would have to be serialized so that we can do a consensus check?

execute/report/report.go
```diff
@@ -41,6 +38,12 @@ func buildSingleChainReportHelper(
 		}
 	}
 
+	numMsg := len(report.Messages)
+	if len(report.TokenData) != numMsg {
```
makramkd, line 42: Just double checking - for tokens that have no offchain token data, it'll just be an empty byte slice?

execute/report/report.go
```diff
@@ -22,13 +21,11 @@ import (
...
-// all messages will be included. This allows the caller to create smaller reports if needed. Executed messages
-// are skipped automatically.
+// all messages will be included. This allows the caller to create smaller reports if needed.
 func buildSingleChainReportHelper(
 	ctx context.Context,
 	lggr logger.Logger,
 	hasher cciptypes.MessageHasher,
-	tokenDataReader types.TokenDataReader,
 	report plugintypes.ExecutePluginCommitData,
 	messages map[int]struct{},
```
makramkd, line 30: I wonder if its worth renaming this to match the call sites (`readyMessages`)

execute/report/report.go
```diff
@@ -22,13 +21,11 @@ import (
 // The hasher and encoding codec are provided as arguments to allow for chain-specific formats to be used.
 //
 // The messages argument indicates which messages should be included in the report. If messages is empty
-// all messages will be included. This allows the caller to create smaller reports if needed. Executed messages
-// are skipped automatically.
+// all messages will be included. This allows the caller to create smaller reports if needed.
```
makramkd, line 24: Is it worth mentioning here that it is expected that messages have already been checked for execution validity? i.e this method is "blindly" proving and encoding a report.

makramkd (commented):
LGTM overall just some small comments and one question re: token data inclusion in the observation

plugintypes/execute.go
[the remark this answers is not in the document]
winder: Any messages selected for the final execution report get moved to another object along with the merkle proofs, so those ones will have consensus during the outcome phase.

execute/report/report.go
winder: Yes, it also pads the slice in case messages are skipped and/or checked out of order for some reason. Added a couple more test cases for this.

execute/report/report.go
```diff
@@ -149,14 +129,62 @@ const (
...
+	if slices.Contains(execReport.ExecutedMessages, msg.Header.SequenceNumber) {
+		b.lggr.Infow(
+			"message already executed",
+			"messageID", msg.Header.MessageID,
+			"sourceChain", execReport.SourceChain,
+			"seqNum", msg.Header.SequenceNumber)
+		return execReport, AlreadyExecuted, nil
+	}
+
+	// Check if token data iss ready.
```
makramkd, line 152:
```suggestion
	// Check if token data is ready.
```

We're not in the ISS ;-)

execute/report/report.go
```diff
@@ -149,14 +129,62 @@ const (
...
+	tokenData, err := b.tokenDataReader.ReadTokenData(ctx, execReport.SourceChain, msg.Header.SequenceNumber)
```
makramkd, line 153: Lets have a nil check on the `b.tokenDataReader` since its still nil by construction and has caused panics in the integration tests.

0xnogo (approved):

PR #51, opened 2024-07-30
Mockery update & porcelain checking.

winder:
Switch from `//go:generate mockery` to `.mockery.yaml`.

Introduce new CI step to ensure all mocks are fully up to date.


.mockery.yaml
```diff
@@ -0,0 +1,16 @@
+with-expecter: true
+filename: "{{.InterfaceName | snakecase}}.go"
+dir: "internal/mocks"
+outpkg: "mocks"
```
makramkd, line 4: This puts all mocks in the same directory by default right? Is this the current behavior or is it genned in the local directory + `/mocks` path?

Makefile
```diff
@@ -1,12 +1,21 @@
+TEST_COUNT ?= 10
+COVERAGE_FILE ?= coverage.out
 
-ensure_go_1_21:
+build: ensure_go_version
+	go build -v ./...
+
+generate: ensure_go_version
```
makramkd, line 7: Recommend we install the correct mockery version (v2.43.0 seems like)

.github/workflows/codegen.yml
```diff
@@ -0,0 +1,42 @@
...
+          go-version: ${{ matrix.go-version }}
+      - name: Install Mockery
+        run: go install github.com/vektra/mockery/v2@v2.43.0
+      - name: Display Go version
+        run: go version
+      - name: Generate mocks
+        run: make generate
+      - name: test git
+        run: git status
+      - name: ensure no changes
```
0xnogo, line 33: nice to have/ We could even include a go mod tidy and check that we have no difference.

.github/workflows/codegen.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        go-version: ['1.21']
+    defaults:
+      run:
+        working-directory: .
+    steps:
+      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+      - name: Setup Go ${{ matrix.go-version }}
```
0xnogo, line 21: Should we also remove all the generated file committed to make sure we have a diff? It might be useful to detect files to clean up.

0xnogo: https://github.com/smartcontractkit/ccip/blob/451984ad0469c7d685e305dba0a0d94eb0c9a053/.github/workflows/ci-core.yml#L426

.mockery.yaml
winder: In this PR I kept all locations the same as before. The default behavior was crazy, and put things in something like "mocks/github.com/smartcontractkit/chainlink-ccip/internal/reader". That didn't even work because of the internal package. Something like this might be good: `dir: mocks/{{ replaceAll .InterfaceDirRelative "internal" "internal_" }}`.

winder: I went ahead and made this change, now there are no overrides to the default locations when mocks are generated.

Makefile
winder: Responding above.

.github/workflows/codegen.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Install Mockery
+        run: go install github.com/vektra/mockery/v2@v2.43.0
```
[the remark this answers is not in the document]
winder: Sure, I can add this to the Makefile. We're safe either way because the mockery version is in the generated file, but maybe this is more usable.

.github/workflows/codegen.yml
winder: Nice :+1:

.github/workflows/codegen.yml
winder: Done. :+1:

Makefile
```diff
@@ -1,12 +1,22 @@
...
+	go install github.com/vektra/mockery/v2@v2.43.0
+	mockery
+
+test: ensure_go_version
+	go test -race -fullpath -shuffle on -count $(TEST_COUNT) -coverprofile=$(COVERAGE_FILE) ./...
+
+lint: ensure_go_version
+	golangci-lint run -c .golangci.yml
```
makramkd, line 15: I recommend making this

```suggestion
	golangci-lint run --tests=true --verbose --max-issues-per-linter 0 --max-same-issues 0
```

So that the lint run can catch everything in one go.

.github/workflows/ccip-ocr3-build-lint-test.yml
```diff
@@ -24,14 +24,14 @@ jobs:
       - name: Display Go version
         run: go version
       - name: Build
-        run: go build -v ./...
+        run: make
       - name: Install linter
         run: |
           curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.59.0
       - name: Run linter
         run: golangci-lint run -c .golangci.yml
```
makramkd, line 32: minor: can use `make lint` here

makramkd (approved):

Makefile
winder: Nice. Updated .golangci.yml to include these 3 options.

PR #53, opened 2024-08-01
feat: CCIP-2495 ConfigPagination for CCIPConfig

defistar:
feat: CCIP-2594 modify the getAllChainConfigs function call from smartcontract with pagination based logic


internal/reader/home_chain.go
```diff
@@ -114,24 +123,46 @@ func (r *homeChainPoller) poll() {
...
-		primitives.Unconfirmed,
-		nil,
-		&chainConfigInfos,
-	)
-	if err != nil {
-		return err
+	var allChainConfigInfos []ChainConfigInfo
+	pageIndex := uint64(0)
+
+	for pageIndex < pageIndexHardLimit {
```
elatoskinas, line 129: I believe based on discussion we should not have a hard cap anymore. We can keep fetching all pages until there are no more configs left. Because our page size is 100, we will always be fetching at most 1 page

internal/reader/home_chain.go
```diff
@@ -114,24 +123,46 @@ func (r *homeChainPoller) poll() {
...
+				"pageIndex": pageIndex,
+				"pageSize":  defaultConfigPageSize,
+			},
+			&chainConfigInfos,
+		)
+		if err != nil {
+			return fmt.Errorf("get config index:%d pagesize:%d: %w", pageIndex, defaultConfigPageSize, err)
+		}
+
+		if len(chainConfigInfos) == 0 {
```
elatoskinas, line 146: suggestion: you can optimize this loop by checking `len(chainConfigInfos) < pageSize` - in which case you can break, because if the number of returned entries is < pageSize there will be no more pages.

defistar: updayed

internal/reader/home_chain.go
```diff
@@ -19,6 +19,15 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
 )
 
+const (
+	defaultConfigPageSize = uint64(100)
+	// pageIndexHardLimit is the maximum number of pages that the poller will fetch
+	// this is set as 1 so that we can get 100 chainConfigs with in 1 rpc call
+	// this can be increased once total chains reach 100 or more.
+	pageIndexHardLimit = 1
```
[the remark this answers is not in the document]
defistar: remove hardCap for pageIndex

elatoskinas (approved):

internal/reader/home_chain_test.go
```diff
@@ -175,7 +175,7 @@ func Test_PollingWorking(t *testing.T) {
 		}
 	}
 	// called at least 2 times, one for start and one for the first tick
```
asoliman92, line 177: Please change comment

internal/reader/home_chain_test.go
```diff
@@ -175,7 +175,7 @@ func Test_PollingWorking(t *testing.T) {
...
-	require.GreaterOrEqual(t, callCount, 2)
+	require.GreaterOrEqual(t, callCount, 1)
```
asoliman92, line 178: Actually this will still be 2.

asoliman92: the number of calls should still be 2 as it will poll as soon as it's started and then again when it starts polling.
If it's not then something is not correct in the polling itself.

defistar: yeap, updated to 2

asoliman92 (approved):
LGTM

PR #55, opened 2024-08-01
Exec related fixes

dimkouv:
1) fix contract reader query limit
2) use msgVisibilityInterval and stop tracking last report timestamp

### DIFF
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -166,9 +166,7 @@ func (p *Plugin) Observation(
 		}
 	}
 
-	msgVisibilityInterval := p.cfg.MessageVisibilityInterval
-	fetchFrom := time.Now().Add(-msgVisibilityInterval)
-
+	fetchFrom := time.Now().Add(-p.cfg.MessageVisibilityInterval).UTC()
 	p.lggr.Infow("decoded previous outcome", "previousOutcome", previousOutcome)
 
 	// Phase 1: Gather commit reports from the destination chain and determine which messages are required to build a
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"sort"
-	"sync/atomic"
 	"time"
 
 	mapset "github.com/deckarep/golang-set/v2"
@@ -43,7 +42,6 @@ type Plugin struct {
 
 	oracleIDToP2pID map[commontypes.OracleID]libocrtypes.PeerID
 	tokenDataReader types2.TokenDataReader
-	lastReportTS    *atomic.Int64
 	lggr            logger.Logger
 }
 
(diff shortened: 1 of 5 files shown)


execute/plugin.go
```diff
@@ -172,6 +166,9 @@ func (p *Plugin) Observation(
 		}
 	}
 
+	msgVisibilityInterval := p.cfg.MessageVisibilityInterval
+	fetchFrom := time.Now().Add(-msgVisibilityInterval)
```
makramkd, line 170: I'm not 100% sure about this but does this time need to be in UTC? We're comparing this to block timestamps eventually right?

execute/plugin_functions.go
```diff
@@ -154,6 +154,31 @@ func filterOutExecutedMessages(
...
+				executedSeqNumsForReport.Add(seqNum)
+			}
+		}
+		report.ExecutedMessages = executedSeqNumsForReport.ToSlice()
+		sort.Slice(report.ExecutedMessages, func(i, j int) bool {
+			return report.ExecutedMessages[i] < report.ExecutedMessages[j]
+		})
+	}
+	// ----------------------------------------
+
```
winder, line 181: We need a test case for this. I'm not sure what this does differently than the original logic

execute/plugin.go
makramkd, line 170: Can probably inline this into the call site, unless we're using for logging somewhere

internal/reader/ccip.go
```diff
@@ -161,7 +161,6 @@ func (r *CCIPChainReader) CommitReportsGTETimestamp(
 		},
 		query.LimitAndSort{
 			SortBy: []query.SortBy{query.NewSortByTimestamp(query.Asc)},
-			Limit:  query.Limit{Count: uint64(limit)},
```
makramkd, line 164: Is this limit applied somewhere in the filtering code below?

execute/plugin_functions.go
dimkouv: not needed eventually, removing it

dimkouv: was fine before

winder (approved):
lgtm

internal/reader/ccip.go
```diff
@@ -235,7 +234,10 @@ func (r *CCIPChainReader) CommitReportsGTETimestamp(
 		})
 	}
 
-	return reports, nil
+	if len(reports) < limit {
+		return reports, nil
+	}
+	return reports[:limit], nil
```
winder, line 240: If we aren't limiting in the query, should we remove the limit option and return everything?
With the current logic if we hit the limit with normal volume, I think we'd never get to some messages until things start timing out.

dimkouv: we'll fix those queries using proper contractReader args soon

0xnogo (approved):

PR #56, opened 2024-08-01
Limit exec report based on gas estimate limits

winder:
Introduce `gas.EstimateProvider` interface for calculating whether estimated gas requirements should prevent an exec report from being created. This PR wires this interface into the plugin via standard factory methods, but hard codes it to the `evm.EstimateProvider` implementation. This implementation will be moved to a separate repo in the future.

Use `gas.EstimateProvider` as part of the `verifyReport` check which is executed dynamically with different message sets to find the largest allowable report.

The new factory parameter is initialized in the ccip repo with this PR: https://github.com/smartcontractkit/ccip/pull/1274

### DIFF
--- a/execute/internal/gas/evm/gas_helpers.go
+++ b/execute/internal/gas/evm/gas_helpers.go
@@ -1,0 +1,81 @@
+// Package evm provides an EVM implementation to the gas.EstimateProvider interface.
+// TODO: Move this package into the EVM repo, chainlink-ccip should be chain agnostic.
+package evm
+
+import (
+	"math"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+const (
+	EvmAddressLengthBytes           = 20
+	EvmWordBytes                    = 32
+	CalldataGasPerByte              = 16
+	TokenAdminRegistryWarmupCost    = 2_500
+	TokenAdminRegistryPoolLookupGas = 100 + // WARM_ACCESS_COST TokenAdminRegistry
+		700 + // CALL cost for TokenAdminRegistry
+		2_100 // COLD_SLOAD_COST loading the pool address
+	SupportsInterfaceCheck = 2600 + // because the receiver will be untouched initially
+		30_000*3 // supportsInterface of ERC165Checker library performs 3 static-calls of 30k gas each
+	PerTokenOverheadGas = TokenAdminRegistryPoolLookupGas +
+		SupportsInterfaceCheck +
+		200_000 + // releaseOrMint using callWithExactGas
+		50_000 // transfer using callWithExactGas
+	RateLimiterOverheadGas = 2_100 + // COLD_SLOAD_COST for accessing token bucket
+		5_000 // SSTORE_RESET_GAS for updating & decreasing token bucket
+	ConstantMessagePartBytes            = 10 * 32 // A message consists of 10 abi encoded fields 32B each (after encoding)
+	ExecutionStateProcessingOverheadGas = 2_100 + // COLD_SLOAD_COST for first reading the state
+		20_000 + // SSTORE_SET_GAS for writing from 0 (untouched) to non-zero (in-progress)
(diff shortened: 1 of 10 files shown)


execute/internal/gas/evm/gas_helpers.go
```diff
@@ -0,0 +1,81 @@
+// Package evm provides an EVM implementation to the gas.EstimateProvider interface.
+// TODO: Move this package into the EVM repo, chainlink-ccip should be chain agnostic.
```
makramkd, line 2: Lets have a ticket for this once you merge these changes 🙏

execute/internal/gas/evm/gas_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	return (EvmAddressLengthBytes + EvmWordBytes) * numTokens
+}
+
+// CalculateMessageMaxGas computes the maximum gas overhead for a message.
+func (gp EstimateProvider) CalculateMessageMaxGas(msg ccipocr3.Message) uint64 {
+	numTokens := len(msg.TokenAmounts)
+	var data []byte = msg.Data
+	dataLength := len(data)
+
+	// TODO: parse max gas from ExtraArgs.
```
makramkd, line 57: To get this to work in the integration tests, for now I think we can duplicate the implementation here: https://github.com/smartcontractkit/chainlink/blob/6ab3eb5b67739ff88d3c4cf8ea125fd8273bc2b1/core/capabilities/ccip/ccipevm/helpers.go#L9-L13

makramkd: Ah dang actually this uses the ABI to decode. We definitely don't wanna import that here

makramkd: Maybe we can do a super hacky way for now, something like:

```golang
extraArgsNoTag := msg.ExtraArgs[4:]
// gas limit is always the first 32 bytes of ExtraArgs for EVM
gasLimitBytes := extraArgsNoTag[:32]
// convert bytes to big
gasLimit := new(big.Int).SetBytes(gasLimitBytes)
```

Maybe can pull this into a helper func and run a few test cases through it

winder: I was going to open a ticket and punt on that part until this function has access to go-ethereum.

makramkd: That's a good point - maybe implement the EVM version of this interface in the ccipevm package in chainlink? like e.g https://github.com/smartcontractkit/chainlink/blob/6ab3eb5b67739ff88d3c4cf8ea125fd8273bc2b1/core/capabilities/ccip/ccipevm/commitcodec.go#L21

winder: My understanding is that the ExtraArgs param is abi encoded, so it should be implemented somewhere that go-ethereum can be used, so probably not the ccip / chainlink repos.

makramkd: go-ethereum can (at least currently) be used in ccip and chainlink

makramkd (approved):

0xnogo (approved):

execute/internal/gas/evm/gas_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+// TODO: Move this package into the EVM repo, chainlink-ccip should be chain agnostic.
+package evm
+
+import (
+	"math"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+const (
```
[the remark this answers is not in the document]
winder: Yes, there is risk of drift. The current logic is [taken from here](https://github.com/smartcontractkit/ccip/blob/ccip-develop/core/services/ocr2/plugins/ccip/ccipexec/gashelpers.go), which are only accurate for 1.5. Hopefully v2 is similar.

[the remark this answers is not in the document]
0xnogo: It would probably make sense to check it with onchain team when contracts are defined and audited.

PR #58, opened 2024-08-07
Move execute types to exectypes package.

winder:
To avoid needing `ExecutePlugin` prefixes on all exec plugin types, move types from `plugintypes` package to a new `exectypes` package.

There are no functional changes in this PR.

These changes are built ontop of #56


execute/exectypes/commit_data.go
```diff
@@ -0,0 +1,32 @@
+package exectypes
```
makramkd, line 1: I'm indifferent but why not just `types` and have an import alias to make it `exectypes`? I'm not crazy about it because it kinda makes imports harder (my IDE doesn't do auto-aliasing and there are a ton of packages called `types` already)

makramkd (approved):

execute/exectypes/commit_data.go
winder: Thats exactly why I wanted exec in the package name. There are so many `types` packages already and the aliases are often slightly different.

asoliman92 (approved):

mateusz-sekara (approved):

PR #60, opened 2024-08-12
execute: add token data reader to constructor

makramkd:
Token data reader was always being set to nil and is required to get through the reporting plugin phases successfully.

Add an argument to the factory to pass it in.

### DIFF
--- a/execute/factory.go
+++ b/execute/factory.go
@@ -16,6 +16,7 @@ import (
 	"github.com/smartcontractkit/libocr/commontypes"
 	ragep2ptypes "github.com/smartcontractkit/libocr/ragep2p/types"
 
+	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
 	"github.com/smartcontractkit/chainlink-ccip/execute/internal/gas"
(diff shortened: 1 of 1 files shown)


execute/factory.go
```diff
@@ -115,7 +119,7 @@ func (p PluginFactory) NewReportingPlugin(
 			p.execCodec,
 			p.msgHasher,
 			p.homeChainReader,
-			nil, // TODO: token data reader
+			p.tokenDataReader,
```
asoliman92, line 122: estimate provider is before data reader in the above code.

execute/factory.go
```diff
@@ -115,7 +119,7 @@ func (p PluginFactory) NewReportingPlugin(
...
 			p.estimateProvider,
```
asoliman92, line 123: estimate provider is before data reader in the above code.

makramkd: Fixed

asoliman92 (approved):

0xnogo (approved):

rstout (approved):

PR #63, opened 2024-08-13
exec plugin: Introduce a State Machine

winder:
### What

Introduce a state machine to manage execution across multiple rounds of OCR.

### Why

To help manage complexity when a 3rd round of OCR is required.

### DIFF
--- a/execute/exectypes/outcome.go
+++ b/execute/exectypes/outcome.go
@@ -7,8 +7,47 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 )
 
+type PluginState string
+
+const (
+	// Unknown is the zero value, this allows a "Next" state transition for uninitialized values (i.e. the first round).
+	Unknown PluginState = ""
+
+	// GetCommitReports is the first step, it is used to select commit reports from the destination chain.
+	GetCommitReports PluginState = "GetCommitReports"
+
+	// GetMessages is the second step, given a set of commit reports it fetches the associated messages.
+	GetMessages PluginState = "GetMessages"
+
+	// Filter is the final step, any additional destination data is collected to complete the execution report.
+	Filter PluginState = "Filter"
+)
+
+// Next returns the next state for the plugin. The Unknown state is used to transition from uninitialized values.
+func (p PluginState) Next() PluginState {
+	switch p {
+	case GetCommitReports:
+		return GetMessages
+
+	case GetMessages:
+		// TODO: go to Filter after GetMessages
+		return GetCommitReports
+
+	case Unknown:
+		fallthrough
+	case Filter:
+		return GetCommitReports
+
+	default:
+		panic("unexpected execute plugin state")
+	}
+}
+
 // Outcome is the outcome of the ExecutePlugin.
 type Outcome struct {
+	// State that the outcome was generated for.
+	State PluginState
+
 	// PendingCommitReports are the oldest reports with pending commits. The slice is
 	// sorted from oldest to newest.
 	PendingCommitReports []CommitData `json:"commitReports"`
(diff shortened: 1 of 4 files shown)


makramkd:
Can we have a diagram with the states and the state transitions?

winder:
> Can we have a diagram with the states and the state transitions?

In this PR
```mermaid
flowchart TD
    UNKNOWN[Initial State]
    A[Get Commit Reports] 
    B[Get Report Messages]
    C[Filter]
    UNKNOWN --> A
    A -->|no report| B
    B -->|report + restart| A
```

In the next PR
```mermaid
flowchart TD
    UNKNOWN[Initial State]
    A[Get Commit Reports] 
    B[Get Report Messages]
    C[Filter]
    UNKNOWN --> A
    A -->|no report| B
    B -->|no report| C
    C -->|report + restart| A
```

execute/exectypes/outcome.go
```diff
@@ -7,8 +7,40 @@ import (
...
+func (p PluginState) Next() PluginState {
+	switch p {
+	case GetCommitReports:
+		return GetMessages
+
+	case GetMessages:
+		// TODO: go to Filter after GetMessages
+		return GetCommitReports
+
+	case Unknown:
```
asoliman92, line 29: `// Unknown is the default value indicating there was no previous state.`

If there was no previous state, should it start getting new messages? Are we going to start the plugin state with Unknown? If we expect to start with it, why fallthrough?

winder: Basically, I chose to do it this way so that an uninitialized previous outcome (i.e. the first round after startup) can transition to the "Next" state with no special handling. So you can get to the first state from `Unknown` or `Filter` (or `GetMessages` in this PR, since I haven't implemented `Filter` yet).

execute/exectypes/outcome.go
```diff
@@ -7,8 +7,40 @@ import (
...
 )
 
+type PluginState int
+
+const (
+	// Unknown is the default value indicating there was no previous state.
+	Unknown PluginState = iota
+	GetCommitReports
+	GetMessages
+	Filter
```
makramkd, line 17: First state is self explanatory but these might need some comments, especially `Filter`

makramkd: Or maybe call it `FilterValidMessages`?

execute/exectypes/outcome.go
```diff
@@ -7,8 +7,40 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 )
 
+type PluginState int
```
makramkd, line 10: I wonder if it would be useful to have a `MarshalJSON` on this so that the outcomes that are JSON encoded are still easily understandable (due to a human-readable string rather than a number). I think similar logic applies to the commit states

winder: Maybe. I would personally prefer if these had a more efficient encoding rather than attempting to make them human readable.

In this case the states would be 1->2->3, so still pretty human readable.

execute/exectypes/outcome.go
winder: Added comments :+1:

winder:
> @winder you can include those mermaid diagrams in a README.md file, it should render natively on GitHub IIRC.

Good idea. I'll create a plugin-level README with this diagram and other information in a separate PR.

execute/exectypes/outcome.go
makramkd: I may be unlike others when it comes to this but even just 1, 2 and 3 is already a cognitive jump because I have to continuously remind myself what each number means.

Having debugged the v2 plugins quite a bit I'd rather we optimize for debuggability to start and then come back and make encoding tighter if needed.

winder: OK. Changed this type to a string.

makramkd: Also worth noting that we have states for commit as well, with their own integers, and states in many other systems which have their own integers, etc. Having all of this in mind I feel like the actual string encoding is just better and it doesn't really have that much overhead to outcome encoding size

winder: The phases could also be named "ReadDest", "ReadSource" and "ReadDestAgain"

makramkd: Less descriptive I feel, I kinda like the current names better, but not sure what other folks think. @asoliman92?

execute/exectypes/outcome_test.go
```diff
@@ -0,0 +1,53 @@
...
+			want: GetMessages,
+		},
+		{
+			name: "Phase 2 to 1",
+			p:    GetMessages,
+			want: GetCommitReports,
+		},
+		{
+			name:    "panic",
+			p:       PluginState("ElToroLoco"),
```
makramkd, line 33: 🤠

makramkd (approved):

asoliman92 (approved):

PR #66, opened 2024-08-14
.github/workflows: fix integration test workflow

makramkd:
### What

Update the integration test workflow to also work on merges into ccip-develop.

### Why

The integration test workflow was failing on ccip-develop due to the `github.event.pull_request.head.sha` not always being available.

### DIFF
--- a/.github/workflows/ccip-integration-test.yml
+++ b/.github/workflows/ccip-integration-test.yml
@@ -32,10 +32,20 @@ jobs:
           cd ccip
           git fetch
           git checkout ccip-develop
+      - name: Get the correct commit SHA via GitHub API
+        id: get_sha
+        run: |
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            COMMIT_SHA=${{ github.event.pull_request.head.sha }}
+          else
+            COMMIT_SHA=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+              "https://api.github.com/repos/${{ github.repository }}/commits/${{ github.ref }}" | jq -r .sha)
+          fi
+          echo "::set-output name=sha::$COMMIT_SHA"
       - name: Update chainlink-ccip dependency in ccip
         run: |
           cd ccip
-          go get github.com/smartcontractkit/chainlink-ccip@${{ github.event.pull_request.head.sha }}
+          go get github.com/smartcontractkit/chainlink-ccip@${{ steps.get_sha.outputs.sha }}
           make gomodtidy
       - name: Setup Postgres
         uses: ./.github/actions/setup-postgres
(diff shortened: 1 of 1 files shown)


.github/workflows/ccip-integration-test.yml
```diff
@@ -32,10 +32,20 @@ jobs:
           cd ccip
           git fetch
           git checkout ccip-develop
```
winder, line 34: Why not use the `actions/checkout` action for this?

.github/workflows/ccip-integration-test.yml
```diff
@@ -32,10 +32,20 @@ jobs:
...
+      - name: Get the correct commit SHA via GitHub API
+        id: get_sha
+        run: |
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            COMMIT_SHA=${{ github.event.pull_request.head.sha }}
+          else
+            COMMIT_SHA=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+              "https://api.github.com/repos/${{ github.repository }}/commits/${{ github.ref }}" | jq -r .sha)
```
winder, line 42: Does providing the branch name work?
```suggestion
            COMMIT_SHA=ccip-develop
```

.github/workflows/ccip-integration-test.yml
makramkd: I think we may be able to use actions/checkout for the ccip checkout but not the chainlink-ccip checkout because it does a shallow clone which messes with the git SHAs

makramkd: Oops ignore what I said above, I misinterpreted the context of the comment.

.github/workflows/ccip-integration-test.yml
makramkd: I guess this works so long as we only ever run this workflow on merges to ccip-develop and pull requests, and not just branch pushes. If we wanna keep it like that its fine with me 😄

makramkd: Ah actually this might have to be `latest` but will test

winder: Good point, maybe something like this?
```
COMMIT_SHA=${{github.GITHUB_REF_NAME}}
```

winder: ref name might work for PRs and Merges

makramkd: Tried it out here and doesn't work unfortunately: https://github.com/smartcontractkit/chainlink-ccip/actions/runs/10457686236/job/28957870251

.github/workflows/ccip-integration-test.yml
makramkd: Switched to actions/checkout

winder (approved):

rstout (approved):

PR #67, opened 2024-08-14
[CCIP-2958] Token price reader implementation

asoliman92:
Use Aggregator interface in price reader and get it ready for multi contract binding. Multi contract binding as well as fixing the tests will be in the next PR once we merge new chain reader changes that use binding instead of name.


internal/reader/onchain_prices_reader.go
```diff
@@ -18,55 +24,60 @@ type TokenPrices interface {
...
+					consts.MethodNameGetLatestRoundData,
+					primitives.Finalized,
+					nil,
+					latestRoundData,
+					//boundContract,
+				); err != nil {
+				return fmt.Errorf("failed to get token price for %s: %w", token, err)
 			}
-			prices[idx] = price
+			prices[idx] = latestRoundData.Answer
```
makramkd, line 80: There's a couple of things we need to check before we return this price; we need to make sure that the price is in USD with 18 decimals, per 1e18 of the smallest token denomination. See https://github.com/smartcontractkit/ccip/blob/89519db20808d4210c1f9025d1355fd0b0076585/contracts/src/v0.8/ccip/PriceRegistry.sol#L152-L157 for details.

internal/reader/onchain_prices_reader.go
```diff
@@ -84,5 +89,57 @@ func (pr *OnchainTokenPricesReader) GetTokenPricesUSD(
...
 }
 
+func (pr *OnchainTokenPricesReader) getRawTokenPrice(ctx context.Context, token types.Account) (*big.Int, error) {
+	var latestRoundData *LatestRoundData
+	if err :=
+		pr.ContractReader.GetLatestValue(
+			ctx,
+			consts.ContractNamePriceAggregator,
+			consts.MethodNameGetLatestRoundData,
+			primitives.Unconfirmed,
```
asoliman92, line 99: Is it okay to use this in production to get prices?

internal/reader/onchain_prices_reader.go
```diff
@@ -84,5 +89,57 @@ func (pr *OnchainTokenPricesReader) GetTokenPricesUSD(
...
+			nil,
+			&latestRoundData,
+			//boundContract,
```
makramkd, line 102: How simple would it be to fix this call once the breaking changes are in? We need the contract address to call. Seems to me it should be fine since we have all the addresses in the PriceSource field.

internal/reader/onchain_prices_reader.go
```diff
@@ -18,55 +23,55 @@ type TokenPrices interface {
...
+			//	Address: pr.PriceSources[token].AggregatorAddress,
+			//	Name: consts.ContractNamePriceAggregator,
+			//}
+			rawTokenPrice, err := pr.getRawTokenPrice(ctx, token)
+			if err != nil {
+				return fmt.Errorf("failed to get token price for %s: %w", token, err)
+			}
+			decimals, err := pr.getTokenDecimals(ctx, token)
+			if err != nil {
+				return fmt.Errorf("failed to get decimals for %s: %w", token, err)
```
makramkd, line 71: I wonder if you can use the BatchGetLatestValue API to do these calls as a batch call, it should be more efficient.

internal/reader/onchain_prices_reader.go
```diff
@@ -84,5 +89,57 @@ func (pr *OnchainTokenPricesReader) GetTokenPricesUSD(
...
+}
+
+// Input price is USD per full token, with 18 decimal precision
+// Result price is USD per 1e18 of smallest token denomination, with 18 decimal precision
+// Examples:
+//
+//	1 USDC = 1.00 USD per full token, each full token is 1e6 units -> 1 * 1e18 * 1e18 / 1e6 = 1e30
+//	1 ETH = 2,000 USD per full token, each full token is 1e18 units -> 2000 * 1e18 * 1e18 / 1e18 = 2_000e18
+//	1 LINK = 5.00 USD per full token, each full token is 1e18 units -> 5 * 1e18 * 1e18 / 1e18 = 5e18
+func calculateUsdPer1e18TokenAmount(price *big.Int, decimals uint8) *big.Int {
```
makramkd, line 139: Hmm the token price isn't always to 18 decimals though, that depends on the feed.

makramkd: You need to normalize the price to 18 decimals before you call this function, see https://github.com/smartcontractkit/ccip/blob/5e1b6d179ca8428b52a025597381c32edecf37a7/core/services/ocr2/plugins/ccip/internal/pricegetter/evm.go#L175-L184

internal/reader/onchain_prices_reader.go
asoliman92: Should be pretty straight forward, I even prepared the `boundContract` before the call as you can see.

internal/reader/onchain_prices_reader.go
asoliman92: But they're different contracts for each token, What's the efficiency you mean here?

internal/reader/onchain_prices_reader.go
asoliman92: That is what the function is doing. It also shows in the tests, it's normalizing to 18 decimals depending on the decimals sent as a parameter.

pluginconfig/commit.go
```diff
@@ -96,6 +96,14 @@ type CommitOffchainConfig struct {
 	// Note that the token address is that on the remote chain.
 	PriceSources map[types.Account]ArbitrumPriceSource `json:"priceSources"`
 
+	// TokenDecimals is a map of token decimals for each token.
+	// As **not** each node supports both the
+	// 1. Token price feed chain (where we get the token price in USD)
+	// 2. Destination chain (where we can get the token decimals from PriceRegistry).
+	// So to be able to calculate the effective price we need both token decimals.
+	// This is why we need to store the token decimals in the config.
+	TokenDecimals map[types.Account]uint8 `json:"decimals"`
```
makramkd, line 105: Can we add a validation for this in `Validate` that each token in `PriceSources` has a decimal in `TokenDecimals`?

internal/reader/onchain_prices_reader_test.go
```diff
@@ -27,39 +29,26 @@ var (
 )
 
 func TestOnchainTokenPricesReader_GetTokenPricesUSD(t *testing.T) {
+	t.Skip("Skipping until we have a full price reader implementation")
```
makramkd, line 32: Should we unskip?

internal/reader/onchain_prices_reader.go
```diff
@@ -18,55 +23,58 @@ type TokenPrices interface {
...
 	return &OnchainTokenPricesReader{
-		TokenPriceConfig: tokenPriceConfig,
-		ContractReader:   contractReader,
+		ContractReader: contractReader,
+		PriceSources:   priceSources,
+		TokenDecimals:  tokenDecimals,
 	}
 }
 
+type LatestRoundData struct {
```
makramkd, line 45: Can you reference the solidity function in a comment above this struct that returns this information? Make sure to use a permalink

internal/reader/onchain_prices_reader.go
```diff
@@ -18,55 +23,58 @@ type TokenPrices interface {
...
+			if err != nil {
+				return fmt.Errorf("failed to get token price for %s: %w", token, err)
+			}
+			decimals, ok := pr.TokenDecimals[token]
+			if !ok {
+				return fmt.Errorf("failed to get decimals for %s: %w", token, err)
 			}
-			prices[idx] = price
+
+			prices[idx] = calculateUsdPer1e18TokenAmount(rawTokenPrice, decimals)
```
makramkd, line 77: You still need to normalize the price to 18 decimals prior to piping it in here

internal/reader/onchain_prices_reader_test.go
asoliman92: In the next PR when we have BoundContracts

internal/reader/onchain_prices_reader_test.go
```diff
@@ -90,20 +78,64 @@ func TestOnchainTokenPricesReader_GetTokenPricesUSD(t *testing.T) {
...
-	for _, acc := range errorAccounts {
-		acc := acc
-		reader.On(
-			"GetLatestValue", mock.Anything, "PriceAggregator", "getTokenPrice", mock.Anything, acc, mock.Anything,
-		).Return(fmt.Errorf("error"))
-	}
-	for acc, price := range mockPrices {
-		acc := acc
+	println(errorAccounts)
+	println(priceSources)
```
makramkd, line 128: Can remove?

internal/reader/onchain_prices_reader.go
```diff
@@ -84,5 +94,67 @@ func (pr *OnchainTokenPricesReader) GetTokenPricesUSD(
...
+func (pr *OnchainTokenPricesReader) getRawTokenPriceE18Normalized(ctx context.Context, token types.Account) (*big.Int, error) {
+	var latestRoundData *LatestRoundData
+	if err :=
+		pr.ContractReader.GetLatestValue(
+			ctx,
+			consts.ContractNamePriceAggregator,
+			consts.MethodNameGetLatestRoundData,
+			primitives.Unconfirmed,
+			nil,
+			&latestRoundData,
```
makramkd, line 124: Is this correct? We're passing in a pointer to a pointer?

internal/reader/onchain_prices_reader.go
```diff
@@ -84,5 +94,67 @@ func (pr *OnchainTokenPricesReader) GetTokenPricesUSD(
...
+func (pr *OnchainTokenPricesReader) getFeedDecimals(ctx context.Context, token types.Account) (*uint8, error) {
+	var decimals *uint8
+	if err :=
+		pr.ContractReader.GetLatestValue(
+			ctx,
+			consts.ContractNamePriceAggregator,
+			consts.MethodNameGetDecimals,
+			primitives.Unconfirmed,
+			nil,
+			&decimals,
```
makramkd, line 106: Ditto here

pluginconfig/commit.go
```diff
@@ -121,6 +129,12 @@ func (c CommitOffchainConfig) Validate() error {
 			c.TokenPriceBatchWriteFrequency, c.TokenPriceChainSelector)
 	}
 
+	for token := range c.PriceSources {
+		if _, exists := c.TokenDecimals[token]; !exists {
+			return fmt.Errorf("missing TokenDecimals for account: %s", token)
```
makramkd, line 134:
```suggestion
			return fmt.Errorf("missing TokenDecimals for token: %s", token)
```

internal/reader/onchain_prices_reader_test.go
makramkd: Not sure I understand the reasoning behind that?

pluginconfig/commit.go
```diff
@@ -121,6 +129,12 @@ func (c CommitOffchainConfig) Validate() error {
...
+		}
+	}
+
```
makramkd, line 137: One more validation to add here is to validation each `ArbitrumPriceSource` passed in:

```golang
for token, aps := range c.PriceSources {
  if err := aps.Validate(); err != nil {
    return fmt.Errorf("invalid arbitrum price source for token %s: %w", token, err)
  }
}
```

internal/reader/onchain_prices_reader.go
asoliman92: Yes, it's passing integration tests.

internal/reader/onchain_prices_reader.go
asoliman92: It was passing integration tests without issues. I fixed it to be a non pointer anyways.

internal/reader/onchain_prices_reader_test.go
```diff
@@ -7,62 +7,62 @@ import (
...
-	ArbPrice = big.NewInt(1)
+	EthPrice   = big.NewInt(1).Mul(big.NewInt(7), big.NewInt(1e18))
+	OpPrice    = big.NewInt(1).Mul(big.NewInt(6), big.NewInt(1e18))
+	ArbPrice   = big.NewInt(1).Mul(big.NewInt(5), big.NewInt(1e18))
+	OnlyPrice  = big.NewInt(1).Mul(big.NewInt(5), big.NewInt(1e18))
+	Decimals18 = uint8(18)
 )
 
 func TestOnchainTokenPricesReader_GetTokenPricesUSD(t *testing.T) {
+	//t.Skip("Skipping until we have a full price reader implementation")
```
makramkd, line 35: Can delete comment

pluginconfig/commit.go
```diff
@@ -99,6 +99,14 @@ type CommitOffchainConfig struct {
 	// Note that the token address is that on the remote chain.
 	PriceSources map[types.Account]ArbitrumPriceSource `json:"priceSources"`
 
+	// TokenDecimals is a map of token decimals for each token.
+	// As **not** each node supports both the
```
makramkd, line 103:
```suggestion
	// As **not necessarily** each node supports both the
```

pluginconfig/commit.go
```diff
@@ -99,6 +99,14 @@ type CommitOffchainConfig struct {
...
+	// 1. Token price feed chain (where we get the token price in USD)
+	// 2. Destination chain (where we can get the token decimals from PriceRegistry).
+	// So to be able to calculate the effective price we need both token decimals.
```
makramkd, line 106:
```suggestion
	// So to be able to calculate the effective price we need both token decimals and feed decimals.
```

pluginconfig/commit.go
```diff
@@ -121,6 +129,16 @@ func (c CommitOffchainConfig) Validate() error {
...
 	}
 
+	for token, arbSource := range c.PriceSources {
+		if err := arbSource.Validate(); err != nil {
+			return fmt.Errorf("invalid arbitrum price source for token %s: %w", token, err)
+		}
+
+		if _, exists := c.TokenDecimals[token]; !exists {
+			return fmt.Errorf("missing TokenDecimals for token: %s", token)
+		}
```
makramkd, line 139: Lets check that the token decimal is > 0

makramkd (commented):
almost there - once these are addressed things LGTM

makramkd (approved):

mateusz-sekara (approved):

PR #69, opened 2024-08-19
exec: Implement filter state and nonce filtering business logic.

winder:
Primary changes:
1. Move report generation into `Filter` state.
2. Add `Nonces` function to `reader.CCIP` interface.
3. Observe sender nonces during `Filter` state.
4. Consider sender nonces during message filtering.
5. Feature flag for checking message nonces.


execute/report/builder.go
```diff
@@ -65,16 +68,22 @@ type execReportBuilder struct {
...
 	// State
 	accumulated validationMetadata
+	// expectedNonce is used to track nonces for multiple messages from the same sender.
+	expectedNonce map[cciptypes.ChainSelector]map[string]uint64
 
 	// Result
 	execReports []cciptypes.ExecutePluginReportSingleChain
+
+	// TODO: remove temporary feature flagging
+	nonceCheckingEnabled bool // defaults to disabled for backwards compatibility.
```
winder, line 86: Feature flagged until the ccip reader interface is implemented.

execute/exectypes/observation.go
```diff
@@ -8,6 +8,7 @@ import (
 
 type CommitObservations map[cciptypes.ChainSelector][]CommitData
 type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]cciptypes.Message
+type NonceObservations map[cciptypes.ChainSelector]map[string]uint64
```
makramkd, line 11: What's the `string` part in the second map? Sender address?

execute/plugin.go
```diff
@@ -168,6 +169,7 @@ func (p *Plugin) Observation(
 	}
 
 	state := previousOutcome.State.Next()
+	p.lggr.Debugw("Execute plugin performing observation.", "state", state)
```
makramkd, line 172:
```suggestion
	p.lggr.Debugw("Execute plugin performing observation", "state", state)
```

Don't need the period at the end

execute/plugin.go
```diff
@@ -215,15 +217,16 @@ func (p *Plugin) Observation(
 
 				// Read messages for each range.
 				for _, seqRange := range ranges {
-					msgs, err := p.ccipReader.MsgsBetweenSeqNums(ctx, selector, seqRange)
+					// TODO: check if srcChain is supported.
+					msgs, err := p.ccipReader.MsgsBetweenSeqNums(ctx, srcChain, seqRange)
```
makramkd, line 221: It seems like MsgsBetweenSeqNums will return an error if the chain is not supported. I don't think we'd want to error out from the whole observation in that case.

Maybe we need to define a specific error type in the ccip reader, e.g `ErrChainNotSupported` so that we can do `errors.Is(err, reader.ErrChainNotSupported)` and just continue in that case.

execute/plugin.go
```diff
@@ -426,17 +454,31 @@ func (p *Plugin) Outcome(
 			commitReports[i].Messages = report.Messages
 		}
 
+		outcome := exectypes.NewOutcome(state, commitReports, cciptypes.ExecutePluginReport{})
+		if outcome.IsEmpty() {
+			return nil, nil
```
makramkd, line 459: Useful to have a `Debug` log here indicating an empty outcome?

Actually should this ever happen? Should we stay in the `GetCommitReports` state until we actually have a non-empty set of commit reports to execute?

execute/report/report.go
```diff
@@ -191,19 +200,68 @@ func (b *execReportBuilder) checkMessage(
...
+	if msg.Header.Nonce != 0 && b.nonceCheckingEnabled {
+		// Sequenced messages have non-zero nonces.
+
+		if _, ok := b.sendersNonce[execReport.SourceChain]; !ok {
+			b.lggr.Errorw("Skipping message - nonces not available for chain",
+				"messageID", msg.Header.MessageID,
+				"sourceChain", execReport.SourceChain,
+				"seqNum", msg.Header.SequenceNumber,
+			)
+			return execReport, MissingNoncesForChain, nil
```
makramkd, line 222: Would it be useful to `fmt.Errorf` the `SourceChain` here for the `MissingNoncesForChain` error? Where would this error should up?

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22.5
 
 require (
 	github.com/deckarep/golang-set/v2 v2.6.0
-	github.com/smartcontractkit/chainlink-common v0.1.7-0.20240712162033-89bd3351ce6e
+	github.com/smartcontractkit/chainlink-common v0.2.1-0.20240717132349-ee5af9b79834
```
makramkd, line 7: Why is this bump being done?

internal/mocks/ccipreader.go
```diff
@@ -48,6 +48,15 @@ func (r CCIPReader) NextSeqNum(ctx context.Context, chains []cciptypes.ChainSele
 	return args.Get(0).([]cciptypes.SeqNum), args.Error(1)
 }
 
+func (r CCIPReader) Nonces(
+	ctx context.Context,
+	source, dest cciptypes.ChainSelector,
+	addresses []string,
+) (map[string]uint64, error) {
```
makramkd, line 55: Is this file being autogenerated via mockery? I don't see the expected `DO NOT EDIT` on top of this file thats usually inserted by mockery

internal/reader/ccip.go
```diff
@@ -57,6 +58,13 @@ type CCIP interface {
 	// TODO: if destination was a parameter, this could be a capability reused across plugin instances.
 	NextSeqNum(ctx context.Context, chains []cciptypes.ChainSelector) (seqNum []cciptypes.SeqNum, err error)
 
+	// Nonces fetches all nonces for the provided selector/address pairs. Addresses are the hex encoded raw address.
```
makramkd, line 61: Hmm the hex encoding here may be problematic since thats just an EVM thing. Other chains use e.g base58

execute/plugin.go
winder: Yes, errors are returned a bit too aggressively right now. This one in particular.

I like the `ErrChainNotSupported` idea.

go.mod
winder: To match the version used in `ccip`. Not strictly necessary for this PR.

execute/report/report.go
winder: Just the logs. The only message state currently used used is to check if the message is ready.

internal/mocks/ccipreader.go
winder: It is not, I'll switch it over.

internal/reader/ccip.go
winder: This should be fine because it's only used as a UID. Internally address are stored as `cciptypes.Bytes` and we would use the `String()` function for a string representation of those bytes. Chain specific encoding is not important until we start building merkle roots, that happens elsewhere.

execute/plugin.go
winder: Do you know what happens when there are errors or we form consensus on a nil outcome?

If OCR3 sets `previousOutcome` to that nil outcome then everything will work as you describe because the plugin currently treats an empty `Outcome` as `Unknown`, which transitions to `GetCommitReports`.

execute/report/report.go
winder: I'll add the message state to the structured log, that should be easier to key off of than the message text.

execute/plugin.go
makramkd: Created a ticket: https://smartcontract-it.atlassian.net/browse/CCIP-3071

execute/plugin.go
makramkd: It should set it to `nil` is my understanding

go.mod
makramkd: Oh interesting

makramkd: Did not know we're behind. If it matches ccip then we should be ok but things might go haywire if we match ccip but not chainlink 😛

commitrmnocb/observation_test.go
```diff
@@ -223,10 +225,12 @@ func Test_ObserveOffRampNextSeqNums(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			ctx := context.Background()
 			var nodeID commontypes.OracleID = 1
-			reader := mocks.NewCCIPReader()
-			reader.On(
-				"NextSeqNum", ctx, tc.knownSourceChains,
-			).Return(tc.nextSeqNums, tc.nextSeqNumsError)
+			reader := reader_mock.NewMockCCIP(t)
+			if tc.supportsDestChain && tc.supportsDestChainError == nil && tc.knownSourceChainsError == nil {
```
winder, line 229: @rstout I switched our manually written CCIP reader mock with a generated version which automatically calls `AssertExpectations`. Several of these tests needed to be updated to account for the stricter assertions.

execute/plugin.go
```diff
@@ -240,11 +243,35 @@ func (p *Plugin) Observation(
...
+		// TODO: add in nonces, other data comes from previous outcome.
+		nonceRequestArgs := make(map[cciptypes.ChainSelector]map[string]struct{})
+
+		// Collect unique senders.
+		for _, commitReport := range previousOutcome.Report.ChainReports {
+			if _, ok := nonceRequestArgs[commitReport.SourceChainSelector]; !ok {
+				nonceRequestArgs[commitReport.SourceChainSelector] = make(map[string]struct{})
+			}
+			for _, msg := range commitReport.Messages {
+				nonceRequestArgs[commitReport.SourceChainSelector][msg.Sender.String()] = struct{}{}
```
makramkd, line 258: I think this should use `typeconv.AddressBytesToString` so that the address is encoded in the chain-family-native string encoding. `String()` here always returns hex.

makramkd (approved):

asoliman92 (approved):
LGTM

execute/plugin.go
winder: You're right, we'll want it in the native format when looking up the value from the reader.

PR #70, opened 2024-08-19
pluginconfigs: axe unnecessary config types

makramkd:
### What

Axe unnecessary config types.

### Why

They're confusing and unnecessary

Fixed tests here: https://github.com/smartcontractkit/ccip/pull/1325


pluginconfig/commit.go
```diff
@@ -104,9 +72,25 @@ type CommitOffchainConfig struct {
 	// This will typically be an arbitrum testnet/mainnet chain depending on
 	// the deployment.
 	TokenPriceChainSelector uint64 `json:"tokenPriceChainSelector"`
+
+	// The maximum number of times to check if the previous report has been transmitted
+	MaxReportTransmissionCheckAttempts uint `json:"maxReportTransmissionCheckAttempts"`
+
+	// SyncTimeout is the timeout for syncing the commit plugin reader.
+	SyncTimeout time.Duration `json:"syncTimeout"`
```
asoliman92, line 80: If we don't use defaults for these let's add them validation for them.

pluginconfig/commit.go
```diff
@@ -104,9 +72,25 @@ type CommitOffchainConfig struct {
 	// This will typically be an arbitrum testnet/mainnet chain depending on
 	// the deployment.
 	TokenPriceChainSelector uint64 `json:"tokenPriceChainSelector"`
+
+	// The maximum number of times to check if the previous report has been transmitted
+	MaxReportTransmissionCheckAttempts uint `json:"maxReportTransmissionCheckAttempts"`
+
+	// SyncTimeout is the timeout for syncing the commit plugin reader.
```
asoliman92, line 79:
```suggestion
	// SyncTimeout is the timeout for syncing the commit plugin ccip reader.
```

pluginconfig/commit.go
```diff
@@ -104,9 +72,25 @@ type CommitOffchainConfig struct {
...
+	SyncTimeout time.Duration `json:"syncTimeout"`
+
+	// SyncFrequency is the frequency at which the commit plugin reader should sync.
```
asoliman92, line 82:
```suggestion
	// SyncFrequency is the frequency at which the commit plugin ccip reader should sync.
```

pluginconfig/commit.go
asoliman92: we have multiple readers and better clarify here.

pluginconfig/commit_test.go
```diff
@@ -211,10 +162,20 @@ func TestCommitOffchainConfig_Validate(t *testing.T) {
...
+		{
+			"invalid, new msg scan batch size not provided",
+			fields{
+				RemoteGasPriceBatchWriteFrequency: *commonconfig.MustNewDuration(1),
+				TokenPriceBatchWriteFrequency:     *commonconfig.MustNewDuration(1),
+				PriceSources:                      map[types.Account]ArbitrumPriceSource{},
+			},
+			true,
+		},
 	}
```
asoliman92, line 174: Add tests for sync durations and timeouts?

pluginconfig/commit.go
makramkd: We have defaults for this that are applied in the plugin constructors. I'm wondering if there should be a standard place where defaults are applied. In the plugin constructor seems fine to me at the moment though.

pluginconfig/commit_test.go
makramkd: Sync duration and timeout have defaults that are applied in the plugin constructors, so I think its OK if they're zero.

Although there could be an argument for not having these defaults since its less visible...

rstout (approved):

pluginconfig/commit_test.go
makramkd: Then will rip out the defaults and enforce that they're specified in the offchain config

makramkd:
Will re-implement in another branch

PR #74, opened 2024-08-22
Populate token prices in commitocb plugin

asoliman92:
Select token prices to add to report. We update token prices based on 2 criteria.

1. Either the last update is more than the frequency from the offchain config.
2. Token price on price registry deviated from feed price per parts in billions PPB.

Steps to achieve:
* Each node will observe current observation timestamp
* In outcome phase, calculate the median for timestamps and implement the logic to check the time against last update to update all tokens or check each token for deviation.
* Update the `lastPricesUpdate` in outcome if we're updating all tokens or keep it as is from previous outcome.

Thanks @dimkouv for the pseudo code
```
observation = {
   newTokenPrices: {
      tokenAddr: price
   }
   currentTokenPrices: {
      tokenAddr: (price, updatedAt)
   }
   ts: time.Now()
}

outcome = {
   come to consensus on current token prices
   come to consensus on new token prices
   include token prices that match the condition (too old or deviated)
}
```


commitrmnocb/observation.go
```diff
@@ -45,10 +47,12 @@ func (p *Plugin) Observation(
 
 	case BuildingReport:
 		observation = Observation{
-			MerkleRoots: p.observer.ObserveMerkleRoots(ctx, previousOutcome.RangesSelectedForReport),
-			GasPrices:   p.observer.ObserveGasPrices(ctx),
-			TokenPrices: p.observer.ObserveTokenPrices(ctx),
-			FChain:      p.observer.ObserveFChain(),
+			MerkleRoots:               p.observer.ObserveMerkleRoots(ctx, previousOutcome.RangesSelectedForReport),
```
asoliman92, line 50: observing timestamp, rest are same

commitrmnocb/types.go
```diff
@@ -35,12 +37,16 @@ func NewCommitQuery(rmnOnRampMaxSeqNums []plugintypes.SeqNumChain, merkleRoots [
...
-	TokenPrices        []cciptypes.TokenPrice          `json:"tokenPrices"`
-	OnRampMaxSeqNums   []plugintypes.SeqNumChain       `json:"onRampMaxSeqNums"`
-	OffRampNextSeqNums []plugintypes.SeqNumChain       `json:"offRampNextSeqNums"`
-	FChain             map[cciptypes.ChainSelector]int `json:"fChain"`
+	MerkleRoots []cciptypes.MerkleRootChain `json:"merkleRoots"`
+	GasPrices   []cciptypes.GasPriceChain   `json:"gasPrices"`
+	// Prices for tokens from the feed on the feed chain
+	FeedTokenPrices []cciptypes.TokenPrice `json:"feedTokenPrices"`
+	// Prices for tokens from the PriceRegistry on the dest chain
+	PriceRegistryTokenUpdates map[types.Account]sharedtypes.NumericalUpdate `json:"priceRegistryTokenPrices"`
```
asoliman92, line 45: Thinking about ditching the `NumericalUpdate` and have it TokenPrice without the timestamp info. We already rely on observation timestamps and no need for the last update info. WDYT? @makramkd

commitrmnocb/outcome.go
```diff
@@ -333,3 +452,55 @@ func counts[T comparable](elems []T) map[T]int {
...
+// Deviates checks if x1 and x2 deviates based on the provided ppb (parts per billion)
+// ppb is calculated based on the smaller value of the two
+// e.g, if x1 > x2, deviation_parts_per_billion = ((x1 - x2) / x2) * 1e9
+func Deviates(x1, x2 *big.Int, ppb int64) bool {
+	// if x1 == 0 or x2 == 0, deviates if x2 != x1, to avoid the relative division by 0 error
+	if x1.BitLen() == 0 || x2.BitLen() == 0 {
+		return x1.Cmp(x2) != 0
+	}
+	diff := big.NewInt(0).Sub(x1, x2) // diff = x1-x2
+	diff.Mul(diff, big.NewInt(1e9))   // diff = diff * 1e9
```
rstout, line 480: Should 1e9 be a const?

rstout (approved):

winder (commented):
Still reviewing, didn't finish today. Pushing current feedback.

internal/reader/onchain_prices_reader.go
```diff
@@ -17,28 +17,25 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-type TokenPrices interface {
-	// GetTokenPricesUSD returns the prices of the provided tokens in USD.
+type PriceReader interface {
+	// GetTokenPricesUSD returns the prices of the provided tokens in USD normalized to e18.
```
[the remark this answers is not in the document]
asoliman92: 1USD = 1e18, added it in the comment.

commitrmnocb/outcome.go
asoliman92: I think it's okay within this small function to be like that.

commitrmnocb/outcome.go
```diff
@@ -108,11 +123,23 @@ func buildReport(
...
 
+	var lastPricesUpdate time.Time
+	tokenPrices := consensusObservation.TokenPricesArray()
+
+	// If we're updating all token prices, set the last prices update to the observed timestamp,
+	// otherwise keep it as the last update time.
+	if len(offchainCfg.TokenInfo) == len(tokenPrices) {
+		lastPricesUpdate = consensusObservation.Timestamp
+	} else {
+		lastPricesUpdate = lastUpdate
```
winder, line 134: Is this field copied through in the outcomes of every state? Seems like in many cases `previousOutcome.LastPriceUpdate` could be nil.

commitrmnocb/outcome.go
```diff
@@ -167,31 +194,123 @@ func getConsensusObservation(
...
 		OnRampMaxSeqNums:   onRampMaxSeqNumsConsensus(lggr, aggObs.OnRampMaxSeqNums, fChains),
 		OffRampNextSeqNums: offRampMaxSeqNumsConsensus(lggr, aggObs.OffRampNextSeqNums, fDestChain),
 		FChain:             fChains,
+		Timestamp:          timestampConsensus,
 	}
 
 	return consensusObs, nil
 }
 
+// Checks which tokens need to be updated based on the observed token prices and the price registry updates
```
winder, line 236: godoc format
```suggestion
// selectTokens checks which tokens need to be updated based on the observed token prices and the price registry updates
```

commitrmnocb/outcome.go
```diff
@@ -167,31 +194,123 @@ func getConsensusObservation(
...
+	medianizedFeedTokenPrices map[types.Account]cciptypes.BigInt,
+	medianizedRegistryPrices map[types.Account]cciptypes.BigInt,
+	consensusTimestamp time.Time,
+	lastPricesUpdate time.Time,
+	updateFrequency commonconfig.Duration,
+	tokenInfo map[types.Account]pluginconfig.TokenInfo,
+) map[types.Account]cciptypes.BigInt {
+	tokenPrices := make(map[types.Account]cciptypes.BigInt)
+	// if the time since the last update is greater than the update frequency, update all tokens
+	if consensusTimestamp.Sub(lastPricesUpdate) > updateFrequency.Duration() {
```
winder, line 250: nit: use `time.After`
```suggestion
	nextPricesUpdate := lastPricesUpdate.Add(updateFrequency.Duration())
	if consensusTimestamp.After(nextPricesUpdate) {
```

commitrmnocb/outcome.go
```diff
@@ -333,3 +452,55 @@ func counts[T comparable](elems []T) map[T]int {
...
+	}
+	return diff.CmpAbs(big.NewInt(ppb)) > 0 // abs(diff) > ppb
+}
+
+// MedianBigInt returns the middle number after sorting the provided numbers.
+// nil is returned if the provided slice is empty.
+// If length of the provided slice is even, the right-hand-side value of the middle 2 numbers is returned.
+// The objective of this function is to always pick within the range of values reported
+// by honest nodes when we have 2f+1 values.
+func MedianBigInt(vals []cciptypes.BigInt) cciptypes.BigInt {
```
winder, line 495: Nit: could you use generics to combine `MedianBigInt` and `MedianTimestamp`?

> // If length of the provided slice is even, the right-hand-side value of the middle 2 numbers is returned.
I don't see this logic, you seem to be returning `valsCopy[len(valsCopy)/2]` without checking.

commitrmnocb/types.go
```diff
@@ -35,12 +36,16 @@ func NewCommitQuery(rmnOnRampMaxSeqNums []plugintypes.SeqNumChain, merkleRoots [
...
-	TokenPrices        []cciptypes.TokenPrice          `json:"tokenPrices"`
-	OnRampMaxSeqNums   []plugintypes.SeqNumChain       `json:"onRampMaxSeqNums"`
-	OffRampNextSeqNums []plugintypes.SeqNumChain       `json:"offRampNextSeqNums"`
-	FChain             map[cciptypes.ChainSelector]int `json:"fChain"`
+	MerkleRoots []cciptypes.MerkleRootChain `json:"merkleRoots"`
+	GasPrices   []cciptypes.GasPriceChain   `json:"gasPrices"`
+	// Prices for tokens from the feed on the feed chain
+	FeedTokenPrices []cciptypes.TokenPrice `json:"feedTokenPrices"`
+	// Prices for tokens from the PriceRegistry on the dest chain
+	PriceRegistryTokenUpdates []cciptypes.TokenPrice          `json:"priceRegistryTokenPrices"`
```
winder, line 44: nit: godoc format
```suggestion
	// FeedTokenPrices for tokens from the feed on the feed chain
	FeedTokenPrices []cciptypes.TokenPrice `json:"feedTokenPrices"`
	// PriceRegistryTokenUpdates for tokens from the PriceRegistry on the dest chain
	PriceRegistryTokenUpdates []cciptypes.TokenPrice          `json:"priceRegistryTokenPrices"`
```

commitrmnocb/validate_observation.go
```diff
@@ -120,7 +120,7 @@ func validateObservedOffRampMaxSeqNums(
 	}
 
 	if !supportsDestChain {
-		return fmt.Errorf("observer %d does not support dest chain, but has observed %d offRampMaxSeqNums",
+		return fmt.Errorf("observer.go %d does not support dest chain, but has observed %d offRampMaxSeqNums",
```
winder, line 123: ??

internal/mocks/chain_support.go
```diff
@@ -34,3 +34,9 @@ func (c ChainSupport) SupportsDestChain(oracleID commontypes.OracleID) (bool, er
 	args := c.Called(oracleID)
 	return args.Get(0).(bool), args.Error(1)
 }
+
+// SupportsChain returns true if the given oracle supports the given chain, returns false otherwise
+func (c ChainSupport) SupportsChain(oracleID commontypes.OracleID, selector cciptypes.ChainSelector) (bool, error) {
+	args := c.Called(oracleID, selector)
+	return args.Get(0).(bool), args.Error(1)
+}
```
winder, line 42: This file could be generated as well.

commitrmnocb/outcome.go
asoliman92: I tried making it generic but BigInt don't have `>` and Timestamp don't have `Cmp` AFAIR. 

> I don't see this logic, you seem to be returning valsCopy[len(valsCopy)/2] without checking.

Good catch. This is how it's implemented now in ocr2, should we keep it as is better?

winder: Following the ocr2 pattern should be fine.

commitrmnocb/outcome.go
```diff
@@ -171,21 +173,24 @@ func checkForReportTransmission(
 
 	if offRampUpdated {
 		return Outcome{
-			OutcomeType: ReportTransmitted,
+			OutcomeType:      ReportTransmitted,
+			LastPricesUpdate: previousOutcome.LastPricesUpdate,
```
winder, line 177: Instead of having this this common logic repeated in each of these, maybe raised to outside of the switch?
```
switch nextState {
    ...
}

// copy cache to next outcome if it isn't updated.
if (outcome.LastPricesUpdate == time.Time{}) {
    outcome.LastPricesUpdate = previousOutcome.LastPricesUpdate
}

p.lggr.Infow("Commit Plugin Outcome", "outcome", outcome, "oid", p.nodeID)
return outcome.Encode()
```

commitrmnocb/outcome_test.go
```diff
@@ -0,0 +1,196 @@
...
+				tt.registryPrices,
+				tt.consensusTimestamp,
+				tt.lastPricesUpdate,
+				tt.updateFrequency,
+				tt.tokenInfo)
+			assert.Equal(t, tt.expectedOutput, result)
+		})
+	}
+}
+func TestDeviates(t *testing.T) {
```
winder, line 101: Move to utils.

winder (changes requested):
I'm not sure this approach will work reliably. It seems like in some normal conditions (i.e. a newly provisioned plugin or maybe even a simple restart) plugin outcomes would be stalled until a price update.

How frequently do we perform price updates? I thought it was much less frequent than every round, so this could be a problem.

makramkd (commented):
Not done reviewing but here's a few comments to start

commitrmnocb/outcome.go
```diff
@@ -108,11 +125,23 @@ func buildReport(
 		outcomeType = ReportEmpty
 	}
 
+	var lastPricesUpdate time.Time
+	tokenPrices := consensusObservation.TokenPricesArray()
+
+	// If we're updating all token prices, set the last prices update to the observed timestamp,
+	// otherwise keep it as the last update time.
+	if len(offchainCfg.TokenInfo) == len(tokenPrices) {
```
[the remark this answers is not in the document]
asoliman92: At this step we already filtered what tokens we want to update on the registry, not everytime we'll update all token prices, this should happen when the time period to update all has passed. Otherwise it will only update a subset of tokens (which deviated from the threshold)

commitrmnocb/types.go
```diff
@@ -196,6 +221,7 @@ type Outcome struct {
 	TokenPrices                     []cciptypes.TokenPrice      `json:"tokenPrices"`
 	GasPrices                       []cciptypes.GasPriceChain   `json:"gasPrices"`
 	ReportTransmissionCheckAttempts uint                        `json:"reportTransmissionCheckAttempts"`
+	LastPricesUpdate                time.Time                   `json:"lastPricesUpdate"`
```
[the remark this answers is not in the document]
asoliman92: Yes.

PR #75, opened 2024-08-22
bump libocr; add context

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2887

Requires:
- https://github.com/smartcontractkit/chainlink-common/pull/490

Supports:
- https://github.com/smartcontractkit/chainlink/pull/14011

core ref: 045c76f80d7249d572353ae904acc640b0597eef


commit/merkleroot/observation.go
```diff
@@ -29,9 +31,12 @@ import (
...
 )
 
-func (w *Processor) ObservationQuorum(_ ocr3types.OutcomeContext, _ types.Query) (ocr3types.Quorum, error) {
+func (w *Processor) ObservationQuorum(
+	_ context.Context, _ ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation,
+) (bool, error) {
 	// Across all chains we require at least 2F+1 observations.
-	return ocr3types.QuorumTwoFPlusOne, nil
+	return quorumhelper.ObservationCountReachesObservationQuorum(
+		quorumhelper.QuorumTwoFPlusOne, w.reportingCfg.N, w.reportingCfg.F, aos), nil
```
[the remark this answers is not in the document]
jmank88: Total number of nodes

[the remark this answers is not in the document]
jmank88: It is not actually used in the implementation for the `QuorumTwoFPlusOne` case, but we aren't supposed to depend on that as a caller, so passing for completeness and forward compatability.

execute/plugin.go
```diff
@@ -243,15 +249,17 @@ func (p *Plugin) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]ocr3types.R
 	}
 
 	// TODO: this function should be pure, a context should not be needed.
```
winder, line 251:
```suggestion
```
You can also remove a similar TODO in `execute/report/builder.go`

commit/merkleroot/observation.go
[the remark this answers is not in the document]
winder: Makes sense, thanks.

internal/plugincommon/ccipreader.go
```diff
@@ -0,0 +1,130 @@
+package plugincommon
```
winder, line 1: Delete this file, it's no longer used.

internal/plugincommon/discovery/processor_test.go
```diff
@@ -6,13 +6,12 @@ import (
 	"testing"
 
 	mapset "github.com/deckarep/golang-set/v2"
+	"github.com/smartcontractkit/libocr/commontypes"
+	ragep2ptypes "github.com/smartcontractkit/libocr/ragep2p/types"
```
winder, line 10: nit: these got moved into the internal dep section

commit/factory.go
```diff
@@ -90,7 +90,7 @@ func NewPluginFactory(
 	}
 }
 
-func (p *PluginFactory) NewReportingPlugin(config ocr3types.ReportingPluginConfig,
+func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types.ReportingPluginConfig,
```
winder, line 93: What is this context? Is there a proper lifecycle context now?

This is called by `chainlink` code, so the full integration test will break as a result.

pkg/reader/usdc_reader.go
```diff
@@ -74,7 +74,7 @@ func (m MessageSentEvent) unpackID() (eventID, error) {
 	return result, nil
 }
 
-func NewUSDCMessageReader(
+func NewUSDCMessageReader(ctx context.Context,
 	lggr logger.Logger,
```
winder, line 78:
```suggestion
func NewUSDCMessageReader(
	ctx context.Context,
	lggr logger.Logger,
```

execute/plugin.go
winder: Nice, thanks

commit/plugin.go
```diff
@@ -176,9 +176,12 @@ func (p *Plugin) Query(ctx context.Context, outCtx ocr3types.OutcomeContext) (ty
...
 }
 
-func (p *Plugin) ObservationQuorum(_ ocr3types.OutcomeContext, _ types.Query) (ocr3types.Quorum, error) {
+func (p *Plugin) ObservationQuorum(
+	ctx context.Context, _ ocr3types.OutcomeContext, _ types.Query, aos []types.AttributedObservation,
+) (bool, error) {
 	// Across all chains we require at least 2F+1 observations.
-	return ocr3types.QuorumTwoFPlusOne, nil
+	return quorumhelper.ObservationCountReachesObservationQuorum(
+		quorumhelper.QuorumFPlusOne, p.reportingCfg.N, p.reportingCfg.F, aos), nil
```
makramkd, line 184:
```suggestion
	return quorumhelper.ObservationCountReachesObservationQuorum(
		quorumhelper.QuorumTwoFPlusOne, p.reportingCfg.N, p.reportingCfg.F, aos), nil
```

makramkd (changes requested):

execute/tokendata/usdc/http_test.go
```diff
@@ -64,7 +65,7 @@ func Test_NewHTTPClient_New(t *testing.T) {
 func Test_HTTPClient_Get(t *testing.T) {
 	tt := []struct {
 		name               string
-		getTs              func() *httptest.Server
+		getTs              func(t *testing.T) *httptest.Server
```
makramkd, line 68: Why this change? I don't see the `testing.T` used anywhere in these funcs.

go.mod
```diff
@@ -5,8 +5,8 @@ go 1.22.5
 require (
 	github.com/deckarep/golang-set/v2 v2.6.0
 	github.com/smartcontractkit/chain-selectors v1.0.23
-	github.com/smartcontractkit/chainlink-common v0.2.3-0.20241008175210-167715aa8613
-	github.com/smartcontractkit/libocr v0.0.0-20240717100443-f6226e09bee7
+	github.com/smartcontractkit/chainlink-common v0.3.0
```
makramkd, line 8: Noice, semver!

commit/plugin.go
jmank88: Ah, good catch

execute/tokendata/usdc/http_test.go
jmank88: The `default timeout` case uses a `require.NoError`

makramkd: Oh I see, they use the top level `t *testing.T` in the test func. I usually rename nested `testing.T` to `tt` but thats just preference.

makramkd (approved):

makramkd:
Integration test will fail until https://github.com/smartcontractkit/chainlink/pull/14011 is merged

execute/tokendata/usdc/http_test.go
jmank88: Hmmm I think using a different name could be error prone, because the original `t` is still in scope to be referenced accidentally. This happens with ctx and err often as well.

jmank88:
> Integration test will fail until [smartcontractkit/chainlink#14011](https://github.com/smartcontractkit/chainlink/pull/14011) is merged

Is there a way to configure it to run against latest core like we do in the non-evm repos?
image

PR #77, opened 2024-08-23
Initialize nonce manager.

winder:
Add bindings for the NonceManager and finish implementing the CCIPReader.

### DIFF
--- a/internal/reader/ccip.go
+++ b/internal/reader/ccip.go
@@ -551,7 +551,7 @@ func (r *CCIPChainReader) bindNonceManager(ctx context.Context) error {
 	// If the contract not binded -> binds to the new address
 	if err := r.contractReaders[r.destChain].Bind(ctx, []types.BoundContract{
 		{
-			Address: typeconv.AddressBytesToString(staticConfig.nonceManager, uint64(r.destChain)),
+			Address: typeconv.AddressBytesToString(staticConfig.NonceManager, uint64(r.destChain)),
 			Name:    consts.ContractNameNonceManager,
 		},
 	}); err != nil {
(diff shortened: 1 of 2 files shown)


internal/reader/ccip.go
```diff
@@ -476,10 +479,52 @@ func (r *CCIPChainReader) Sync(ctx context.Context) (bool, error) {
...
+		return fmt.Errorf("get onramps: %w", err)
+	}
+
+	if _, ok := r.contractReaders[r.destChain]; !ok {
+		r.lggr.Debugw("skipping nonce manager, dest chain not configured for this deployment",
+			"destChain", r.destChain)
+		return nil
+	}
+
+	// Bind the onRamp contract address to the reader.
```
asoliman92, line 501:
```suggestion
	// Bind the nonceManager contract address to the reader.
```

internal/reader/ccip.go
```diff
@@ -557,5 +602,33 @@ func (r *CCIPChainReader) validateWriterExistence(chains ...cciptypes.ChainSelec
...
+		map[string]any{},
+		&resp,
+	)
+	if err != nil {
+		return offrampStaticChainConfig{}, fmt.Errorf("failed to get source chain config: %w", err)
+	}
+	return resp, nil
+}
+
+type offrampStaticChainConfig struct {
```
asoliman92, line 626: Can you put link to the offramp type? Will make it easier to find it.

winder: Yeah, I'll add a comment, what did you have in mind? A comment? I copied the pattern used for `type sourceChainConfig struct {`, so I'll update that as well.

makramkd (commented):
You can write an integration test to test that this works in the ccipreader_test.go file in chainlink/ccip. See `TestCCIPReader_CommitReportsGTETimestamp` as an example.

internal/reader/ccip.go
```diff
@@ -523,10 +567,52 @@ func (r *CCIPChainReader) Sync(ctx context.Context) (bool, error) {
...
+			Name:    consts.ContractNameNonceManager,
+		},
+	}); err != nil {
+		return fmt.Errorf("bind nonce manager: %w", err)
+	}
+
+	return nil
+}
+
+func (r *CCIPChainReader) Sync(ctx context.Context) (bool, error) {
```
makramkd, line 605: @dimkouv why the `(bool, error)` return signature? I feel like `error` is enough?

internal/reader/ccip.go
makramkd: A permalink to the solidity struct would be good enough IMO

makramkd (approved):

PR #79, opened 2024-08-27
internal/reader: check if source chain is enabled

makramkd:
Ticket: https://smartcontract-it.atlassian.net/browse/CCIP-3083

Check that the source chain set on the offramp is enabled prior to reading any messages sent from that source.

### DIFF
--- a/internal/reader/ccip.go
+++ b/internal/reader/ccip.go
@@ -466,6 +466,11 @@ func (r *CCIPChainReader) Sync(ctx context.Context) (bool, error) {
 			return false, fmt.Errorf("onRamp address not found for chain %d", chain)
 		}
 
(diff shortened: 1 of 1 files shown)


internal/reader/ccip.go
```diff
@@ -533,8 +538,9 @@ func (r *CCIPChainReader) getSourceChainsConfig(
 }
 
 type sourceChainConfig struct {
-	OnRamp   []byte `json:"onRamp"`
-	MinSeqNr uint64 `json:"minSeqNr"`
+	IsEnabled bool
```
dimkouv, line 541: Looks good, name matches contract struct.

dimkouv (approved):

PR #81, opened 2024-08-27
internal/reader: add GetExpectedNextSequenceNumber method

makramkd:
Test in core: https://github.com/smartcontractkit/ccip/pull/1373

The commit plugin currently builds merkle roots of size one because it only reads offramp sequence numbers. Add this so it can read onramp sequence numbers as well

### DIFF
--- a/internal/reader/ccip.go
+++ b/internal/reader/ccip.go
@@ -57,7 +57,7 @@ type CCIP interface {
 	// in the onramp.
 	GetExpectedNextSequenceNumber(
 		ctx context.Context,
-		sourceChainSelector cciptypes.ChainSelector,
+		sourceChainSelector, destChainSelector cciptypes.ChainSelector,
 	) (cciptypes.SeqNum, error)
 
 	// NextSeqNum reads the destination chain.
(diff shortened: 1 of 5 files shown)


internal/reader/ccip.go
```diff
@@ -53,6 +53,13 @@ type CCIP interface {
 		seqNumRange cciptypes.SeqNumRange,
 	) ([]cciptypes.Message, error)
 
+	// GetExpectedNextSequenceNumber returns the next sequence number to be used
+	// in the onramp.
+	GetExpectedNextSequenceNumber(
+		ctx context.Context,
+		sourceChainSelector cciptypes.ChainSelector,
```
winder, line 60: I thought we were going to stop using the implicit destination:

```suggestion
		source, dest cciptypes.ChainSelector,
```

makramkd: Good catch, done.

dimkouv (approved):

internal/reader/ccip.go
```diff
@@ -387,6 +394,41 @@ func (r *CCIPChainReader) MsgsBetweenSeqNums(
...
+// GetExpectedNextSequenceNumber implements CCIP.
+func (r *CCIPChainReader) GetExpectedNextSequenceNumber(
+	ctx context.Context,
+	sourceChainSelector, destChainSelector cciptypes.ChainSelector) (cciptypes.SeqNum, error) {
+	if destChainSelector != r.destChain {
+		return 0, fmt.Errorf("expected destination chain %d, got %d", r.destChain, destChainSelector)
+	}
+
+	if err := r.validateReaderExistence(sourceChainSelector); err != nil {
+		return 0, err
```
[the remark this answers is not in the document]
makramkd: We have a ticket for this I want to tackle next: https://smartcontract-it.atlassian.net/browse/CCIP-3071

PR #86, opened 2024-08-28
Commit plugin - Fix non-deterministic outcome on BuildingReport state

dimkouv:
Fix non-deterministic outcome on BuildingReport state.

### DIFF
--- a/commit/outcome_test.go
+++ b/commit/outcome_test.go
@@ -1,0 +1,44 @@
+package commit
+
+import (
+	"testing"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_buildReport(t *testing.T) {
+	t.Run("determinism check", func(t *testing.T) {
+		const rounds = 50
+
+		obs := ConsensusObservation{
+			MerkleRoots: map[cciptypes.ChainSelector]cciptypes.MerkleRootChain{
+				cciptypes.ChainSelector(1): {
+					ChainSel:     1,
+					SeqNumsRange: cciptypes.NewSeqNumRange(10, 20),
+					MerkleRoot:   cciptypes.Bytes32{1},
+				},
+				cciptypes.ChainSelector(2): {
+					ChainSel:     2,
+					SeqNumsRange: cciptypes.NewSeqNumRange(20, 30),
+					MerkleRoot:   cciptypes.Bytes32{2},
+				},
+			},
+			GasPrices: map[cciptypes.ChainSelector]cciptypes.BigInt{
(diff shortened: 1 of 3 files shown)


commit/outcome_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"testing"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_buildReport(t *testing.T) {
+	t.Run("determinism check", func(t *testing.T) {
```
makramkd, line 12: I feel like all these determinism checks would be good for fuzz tests, we can explore that in the future

commit/types.go
```diff
@@ -159,8 +160,12 @@ type ConsensusObservation struct {
...
 func (co ConsensusObservation) GasPricesArray() []cciptypes.GasPriceChain {
+	chains := maps.Keys(co.GasPrices)
+	sort.Slice(chains, func(i, j int) bool { return chains[i] < chains[j] })
+
 	gasPrices := make([]cciptypes.GasPriceChain, 0, len(co.GasPrices))
-	for chain, gasPrice := range co.GasPrices {
+	for _, chain := range chains {
+		gasPrice := co.GasPrices[chain]
 		gasPrices = append(gasPrices, cciptypes.NewGasPriceChain(gasPrice.Int, chain))
 	}
```
makramkd, line 170: small nit: I think you'd save one line of code and be consistent with the other sorts if you sort after the original loop?

```golang
for chain, gasPrice := range co.GasPrices {
  gasPrices = append(gasPrices, cciptypes.NewGasPriceChain(gasPrice.Int, chain))
}
sort.Slice(gasPrices, func() { ... })
```

commit/types.go
```diff
@@ -169,8 +174,12 @@ func (co ConsensusObservation) GasPricesArray() []cciptypes.GasPriceChain {
...
 // TokenPricesArray returns a list of token prices
 func (co ConsensusObservation) TokenPricesArray() []cciptypes.TokenPrice {
+	tokenIDs := maps.Keys(co.TokenPrices)
+	sort.Slice(tokenIDs, func(i, j int) bool { return tokenIDs[i] < tokenIDs[j] })
+
 	tokenPrices := make([]cciptypes.TokenPrice, 0, len(co.TokenPrices))
-	for tokenID, tokenPrice := range co.TokenPrices {
+	for _, tokenID := range tokenIDs {
+		tokenPrice := co.TokenPrices[tokenID]
 		tokenPrices = append(tokenPrices, cciptypes.NewTokenPrice(tokenID, tokenPrice.Int))
```
makramkd, line 183: nit: similarly here, can add the sort after the loop?

makramkd (approved):

PR #93, opened 2024-08-30
Fix issue with fChain consensus and catch determinism edge cases

dimkouv:
1. Consensus rule `that was observed at least f times, where f is the F of the DON` was not respected due to a missing `continue`. Which could lead to a consensus fChain containing some value only reported by one node.
2. Catch some determinism edge cases where two or more different groups of the same size observe different values.

### DIFF
--- a/commit/outcome_test.go
+++ b/commit/outcome_test.go
@@ -52,24 +52,22 @@ func Test_fChainConsensus(t *testing.T) {
 		cciptypes.ChainSelector(1): {5, 5, 5, 5, 5},
 		cciptypes.ChainSelector(2): {5, 5, 5, 3, 5},
 		cciptypes.ChainSelector(3): {5, 5},             // not enough observations, must be observed at least f times.
-		cciptypes.ChainSelector(4): {5, 3, 5, 3, 5, 3}, // both values appear at least f times
+		cciptypes.ChainSelector(4): {5, 3, 5, 3, 5, 3}, // both values appear at least f times, no consensus
 	}
 	fChainFinal := fChainConsensus(lggr, f, fChainValues)
 
 	assert.Equal(t, map[cciptypes.ChainSelector]int{
 		cciptypes.ChainSelector(1): 5,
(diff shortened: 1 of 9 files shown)


commit/outcome_test.go
```diff
@@ -42,3 +44,81 @@ func Test_buildReport(t *testing.T) {
...
+		cciptypes.ChainSelector(2): {5, 5, 5, 3, 5},
+		cciptypes.ChainSelector(3): {5, 5},             // not enough observations, must be observed at least f times.
+		cciptypes.ChainSelector(4): {5, 3, 5, 3, 5, 3}, // both values appear at least f times
+	}
+	fChainFinal := fChainConsensus(lggr, f, fChainValues)
+
+	assert.Equal(t, map[cciptypes.ChainSelector]int{
+		cciptypes.ChainSelector(1): 5,
+		cciptypes.ChainSelector(2): 5,
+		cciptypes.ChainSelector(4): 3,
```
winder, line 62: Why pick 3 instead of 5?

winder: Isn't this case impossible since the threshold is more than half of the committe?

winder: IMO this should be an error. There is no consensus about what the true value is.

winder: We have 2 implementations for how to merge observations together. If you were to use [MinObservationFilter](https://github.com/smartcontractkit/chainlink-ccip/blob/ccip-develop/execute/internal/validation/reports.go) I think it would return that there are 2 valid values. Since fChain is a single number that would be an error.

winder (changes requested):
Not sure this is the right behavior. A split consensus is the same as no consensus.

commit/outcome_test.go
dimkouv: So seems better to treat this case as "not reaching consensus". I pushed a fix

winder (approved):

PR #94, opened 2024-09-02
Small refactoring of commit plugin logic

dimkouv:
- Making commit plugin logic less error prone.
- Include duration for Observation and Outcome
- Small function fixes


commit/outcome.go
```diff
@@ -22,39 +23,42 @@ import (
 func (p *Plugin) Outcome(
 	outCtx ocr3types.OutcomeContext, query types.Query, aos []types.AttributedObservation,
 ) (ocr3types.Outcome, error) {
+	tStart := time.Now()
```
asoliman92, line 26: What's the rationale behind logging the outcome? Do you think this process will take long times that we need to follow?

the "duration"  in the log is not very clear in a "Sending Outcome" as well, 
maybe use "duration for getting Outcome" or move the logic to the `getOutcome` function with separate logging clarifying that this is for the get not send.

makramkd (approved):

PR #98, opened 2024-09-03
PluginProcessors [CCIP-3147]

asoliman92:
Add PluginProcessor interface to create multiple Processors under OCR plugin.

The idea is to separate the logic of different types of observations and outcomes into separate processors.
This makes it easier to manage and test the logic of each type of observation/outcome without affecting each others.
Some of them will implement state machines (e.g. merkleroot), others might implement simpler logic. (e.g. token)
Also makes running them in parallel more streamlined.
The OCR plugin becomes a coordinator/collector of these SubPlugins.

Example Pseudo code:
```
	OCRPlugin {
	  nodeID
	  merkleProcessor
	  tokenProcessor
	  feeProcessor
	  ...
	}

	OCRPlugin.Observer {
	  mObs := merkleProcessor.Observer
	  tObs := tokenProcessor.Observer
	  fObs := feeProcessor.Observer
	  return Observation{mObs, tObs, fObs}
	}
	OCRPlugin.Validate {
	  mObs := merkleProcessor.Validate
	  tObs := tokenProcessor.Validate
	  fObs := feeProcessor.Validate
	  check errors for each
	  return nil
	}
	OCRPlugin.Outcome {
	  mOut := merkleProcessor.Outcome
	  tOut := tokenProcessor.Outcome
	  fOut := feeProcessor.Outcome
	  return Outcome{mOut, tOut, fOut}
	}

	OCRPlugin.Report {
		return Report{mOut.X, tOut.Y, fOut.Z}
	}
```

Notice all PluginProcessor interface functions are using `prevOutcome` instead of `outCtx`.
We're interested in the prevOutcome, and it makes it easier to have all decoding on the top level (OCR plugin),
otherwise there might be cyclic dependencies or just complicating the code more.


commit/gasworker/observation.go
```diff
@@ -0,0 +1,15 @@
+package gasworker
```
makramkd, line 1: Lets not call this package `gasworker`, I think `fees` is more general (not EVM specific) and the `Worker` aspect is just a type inside this package.

commit/gasworker/observation.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+)
+
+func (p *GasWorker) Observation(
+	ctx context.Context,
+) GasObservation {
+	//tStart := time.Now()
+	//observation, nextState := p.getObservation(ctx, previousOutcome)
+	//p.lggr.Infow("Sending MerkleObs",
+	//	"observation", observation, "nextState", nextState, "observationDuration", time.Since(tStart))
```
makramkd, line 13: Lets delete this since its not really relevant

commit/gasworker/observation.go
```diff
@@ -0,0 +1,15 @@
+package gasworker
+
+import (
+	"context"
+)
+
+func (p *GasWorker) Observation(
+	ctx context.Context,
```
makramkd, line 8: Lets make this match libocr as much as possible for consistency. Needs to take in previous outcome (outctx) at least and maybe `types.Query`

commit/gasworker/outcome.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+)
+
+func (p *GasWorker) Outcome(
+	outCtx ocr3types.OutcomeContext, observations []GasObservation,
+) GasOutcome {
+	//tStart := time.Now()
+	//outcome, nextState := p.getOutcome(outCtx, observations, previousOutcome)
+	//p.lggr.Infow("Sending MerkleOutcome",
+	//	"outcome", outcome, "oid", p.nodeID, "nextState", nextState, "outcomeDuration", time.Since(tStart))
```
makramkd, line 13: Remove

commit/types.go
```diff
@@ -58,176 +57,22 @@ func DecodeCommitPluginObservation(encodedObservation []byte) (Observation, erro
...
+	TokensOutcome tokenworker.TokenPricesOutcome `json:"tokensOutcome"`
+	GasOutcome    gasworker.GasOutcome           `json:"gasOutcome"`
 }
 
 // Encode encodes an Outcome deterministically
 func (o Outcome) Encode() ([]byte, error) {
 
 	// Sort all lists to ensure deterministic serialization
-	o.sort()
+	o.MerkleOutcome.Sort()
```
dimkouv, line 70: I believe both of this are not required anymore, we have deterministic outcomes prior to encoding.

commit/gasworker/outcome.go
```diff
@@ -0,0 +1,15 @@
+package gasworker
+
+import (
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+)
+
+func (p *GasWorker) Outcome(
+	outCtx ocr3types.OutcomeContext, observations []GasObservation,
```
makramkd, line 8: Feels like this may need to be `AttributedObservation` to provide the most possible info to the sub-plugins, i.e maybe the oracle ID is required in order to log some information

commit/gasworker/types.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (obs GasObservation) Validate() error {
+	// validate feed
+	// validate registry
+	// validate timestamps?
+	return nil
+}
+
+func (o GasOutcome) Sort() {
```
makramkd, line 28: Whats this for?

commit/gasworker/worker.go
```diff
@@ -0,0 +1,4 @@
+package gasworker
+
+type GasWorker struct {
```
makramkd, line 3: Lets just call this `Worker`

commit/merkleworker/observation.go
makramkd, line 1: Package should be `merkleroots` and the worker type should just be `Worker`, to get `merkleroot.Worker`

commit/merkleworker/observation.go
```diff
@@ -0,0 +1,242 @@
...
+	// ObserveOffRampNextSeqNums observes the next sequence numbers for each source chain from the OffRamp
+	ObserveOffRampNextSeqNums(ctx context.Context) []plugintypes.SeqNumChain
+
+	// ObserveMerkleRoots computes the merkle roots for the given sequence number ranges
+	ObserveMerkleRoots(ctx context.Context, ranges []plugintypes.ChainRange) []cciptypes.MerkleRootChain
+
+	ObserveFChain() map[cciptypes.ChainSelector]int
+}
+
+type MerkleObserver struct {
```
makramkd, line 81: `Observer`

commit/merkleworker/types.go
```diff
@@ -0,0 +1,183 @@
...
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"github.com/smartcontractkit/chainlink-ccip/plugintypes"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+// TODO: Do we need this?
```
makramkd, line 13: Yes

commit/merkleworker/validate_observation.go
```diff
@@ -0,0 +1,179 @@
...
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	"github.com/smartcontractkit/libocr/commontypes"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
+	"github.com/smartcontractkit/chainlink-ccip/plugintypes"
+)
+
+func (obs MerkleObservation) Validate(
```
makramkd, line 21: Just `Observation` is enough. In general, things in Go read as `packagename.SymbolName` and there's no need to repeat package name information in the symbol name

commit/merkleworker/worker.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
+	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+	"github.com/smartcontractkit/chainlink-ccip/shared"
+)
+
+type MerkleWorker struct {
```
makramkd, line 20: `Worker`

commit/observation.go
```diff
@@ -28,210 +17,28 @@ func (p *Plugin) ObservationQuorum(_ ocr3types.OutcomeContext, _ types.Query) (o
...
-	p.lggr.Infow("Sending Observation",
-		"observation", observation, "nextState", nextState, "observationDuration", time.Since(tStart))
-	return observation.Encode()
-}
+	previousOutcome := p.decodeOutcome(outCtx.PreviousOutcome)
+	fChain := p.ObserveFChain()
+	//TODO: Send fchain instead of computing it inside
+	merkleObs := p.merkleWorker.Observation(ctx, previousOutcome.MerkleOutcome)
+	tokenPricesObs := p.tokenWorker.Observation(ctx)
+	gasObs := p.gasWorker.Observation(ctx)
```
makramkd, line 25: Need to pass in previous outcome

commit/tokenworker/observation.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+)
+
+func (p *TokenWorker) Observation(
+	ctx context.Context,
+) TokenPricesObservation {
+	//tStart := time.Now()
+	//observation, nextState := p.getObservation(ctx, previousOutcome)
+	//p.lggr.Infow("Sending MerkleObs",
+	//	"observation", observation, "nextState", nextState, "observationDuration", time.Since(tStart))
```
makramkd, line 13: Remove

makramkd (changes requested):
This is a good start but I think we need to improve unit-testability. One idea is to have a generic interface that is implemented by all sub-plugins:

```golang
type SubPlugin[QueryType any, ObservationType any, OutcomeType any] interface {
  // Will be needed by merkleroots subplugin
  Query(ctx context.Context, outctx OutcomeContext) (QueryType, error)
  Observation(ctx context.Context, outctx OutcomeContext, query QueryType) (ObservationType, error)
  ValidateObservation(outctx OutcomeContext, query QueryType, ao types.AttributedObservation) error
  Outcome(outctx OutcomeContext, query QueryType, aos []types.AttributedObservation) (OutcomeType, error)
}
```

Then calls to this can easily be mocked in unit tests, and would help us unit-test the "combiner" functions used in all the OCR phases.

commit/gasworker/observation.go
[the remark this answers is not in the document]
makramkd: I feel like `Service` and `Worker` names suggest background processes that are always running, but I'm not too against the service naming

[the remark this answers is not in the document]
dimkouv: `Handler` ?

`obs = p.TokenPricesHandler.Observation()`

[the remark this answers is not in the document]
asoliman92: No strong opinion. I like `Worker` because it's something I give it some work and collect back results to aggregate. but have no issue changing it.

commit/gasworker/outcome.go
asoliman92: The main reason I didn't do it like that is because we'll have n workers decoding it, instead I decode it once in the plugin and send only the required info to the worker.

asoliman92: I think it's just abstracting away something that doesn't really need to be abstracted IMO.

makramkd: Doesn't have to be encoded, it can be something like this:

```golang
type AttributedObservation[ObservationType any] struct {
  OracleID uint8
  Observation ObservationType
}
```

commit/merkleworker/types.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+type MerkleObservation struct {
+	MerkleRoots        []cciptypes.MerkleRootChain     `json:"merkleRoots"`
+	OnRampMaxSeqNums   []plugintypes.SeqNumChain       `json:"onRampMaxSeqNums"`
+	OffRampNextSeqNums []plugintypes.SeqNumChain       `json:"offRampNextSeqNums"`
+	FChain             map[cciptypes.ChainSelector]int `json:"fChain"`
+}
+
+func (obs MerkleObservation) Encode() ([]byte, error) {
```
winder, line 30: @makramkd @dimkouv what do you think about getting rid of every `Encode()` function in the plugins? Go allows us easily use reflection to do this in one place rather than sprinkling it throughout the types.

commit/gasworker/observation.go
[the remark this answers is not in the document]
winder: I agree with @makramkd, let's name this ssomething like `Handler`, `Processor` or `Provider`. The name `Worker` makes me think that this is the task processing component of an [async worker pool](https://gobyexample.com/worker-pools).

commit/merkleroot/observation.go
```diff
@@ -8,61 +8,68 @@ import (
...
+			FChain:             w.observer.ObserveFChain(),
 		}, nextState
 	case BuildingReport:
 		return Observation{
-			MerkleRoots: p.observer.ObserveMerkleRoots(ctx, previousOutcome.RangesSelectedForReport),
-			GasPrices:   p.observer.ObserveGasPrices(ctx),
-			TokenPrices: p.observer.ObserveTokenPrices(ctx),
-			FChain:      p.observer.ObserveFChain(),
+			MerkleRoots: w.observer.ObserveMerkleRoots(ctx, previousOutcome.RangesSelectedForReport),
+			FChain:      w.observer.ObserveFChain(),
```
asoliman92, line 64: Leaving this for now. Will iterate on having Fchain sent to the workers later.

mocks/cl-common/chainreader/chain_reader.go
```diff
@@ -5,9 +5,10 @@ package types
 import (
 	context "context"
 
+	mock "github.com/stretchr/testify/mock"
+
 	query "github.com/smartcontractkit/chainlink-common/pkg/types/query"
 	primitives "github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
-	mock "github.com/stretchr/testify/mock"
```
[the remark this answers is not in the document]
asoliman92: I ran the command to fix imports and it did this :/

commit/types.go
asoliman92: We don't currently sort the outcomes before returning it. We can do so but leaving this here is a good safe guard. WDYT?

shared/sub_plugin.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/smartcontractkit/libocr/commontypes"
+)
+
+type AttributedObservation[ObservationType any] struct {
+	OracleID    commontypes.OracleID
+	Observation ObservationType
+}
+
+type SubPlugin[QueryType any, ObservationType any, OutcomeType any] interface {
```
asoliman92, line 14: Using `prevOutcome` instead of `outCtx`. We're mainly interested in the `prevOutcome` and it makes it easier to have all decoding on the top level, otherwise there might be cyclic dependencies or just complicating the code more.

commit/merkleroot/observation.go
```diff
@@ -74,24 +81,20 @@ type Observer interface {
...
 
-	ObserveTokenPrices(ctx context.Context) []cciptypes.TokenPrice
-
-	ObserveGasPrices(ctx context.Context) []cciptypes.GasPriceChain
-
 	ObserveFChain() map[cciptypes.ChainSelector]int
 }
 
-type ObserverImpl struct {
+type MerkleObserver struct {
```
asoliman92, line 87: Didn't change the way it works in merkleroot in this PR. Did least invasive way. We can later iterate and maybe remove Observer as is completely if it's not needed.

commit/plugin_e2e_test.go
```diff
@@ -9,15 +9,18 @@ import (
 	"time"
```
asoliman92, line 9: Changed current tests to use multiple outcomes (specifically `merkleroot.Outcome`). No new tests added. Maybe later we'll refactor this to have some mocks and make sure it's aggregating correctly without testing the underlying processors as already each processor will have their own unit tests.

shared/chain_support.go
```diff
@@ -1,4 +1,4 @@
-package commit
+package shared
```
asoliman92, line 1: Moving it to shared so that we can use without cyclic dependencies. Just changing fields to make them publicly available. No added or updated functionality.

commit/merkleroot/processor.go
```diff
@@ -0,0 +1,79 @@
...
+	lggr         logger.Logger
+	observer     Observer
+	readerSyncer *plugincommon.BackgroundReaderSyncer
+	ccipReader   reader.CCIP
+	reportingCfg ocr3types.ReportingPluginConfig
+	chainSupport shared.ChainSupport
+}
+
+// NewMerkleWorker creates a new Processor
+func NewMerkleWorker(
```
makramkd, line 32: I think just `NewProcessor` is sufficient, no need for the `Merkle`. It'll be called like `merkleroot.NewProcessor(...)`

commit/merkleroot/processor.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
+	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+	"github.com/smartcontractkit/chainlink-ccip/shared"
+)
+
+type Processor struct {
```
makramkd, line 20: Go doc comment on this explaining its purpose would be useful

commit/plugin.go
```diff
@@ -29,8 +41,10 @@ type Plugin struct {
 	lggr              logger.Logger
 	homeChain         reader.HomeChain
 	reportingCfg      ocr3types.ReportingPluginConfig
-	chainSupport      ChainSupport
-	observer          Observer
+	chainSupport      shared.ChainSupport
+	merkleWorker      merkleroot.Processor
+	tokenWorker       token.Processor
+	gasWorker         fees.Processor
```
makramkd, line 47: Do we want to have the concrete types here or the interfaces? I thought the interfaces would make for easier testing of the root plugin logic.

commit/report.go
```diff
@@ -18,16 +20,15 @@ func (p *Plugin) Reports(seqNr uint64, outcomeBytes ocr3types.Outcome) ([]ocr3ty
...
-	if outcome.OutcomeType != ReportGenerated {
-		return []ocr3types.ReportWithInfo[[]byte]{}, nil
-	}
-
-	rep := cciptypes.NewCommitPluginReport(outcome.RootsToReport, outcome.TokenPrices, outcome.GasPrices)
+	rep := cciptypes.NewCommitPluginReport(
+		outcome.MerkleOutcome.RootsToReport,
+		outcome.TokensOutcome.TokenPrices,
+		outcome.FeesOutcome.GasPrices,
+	)
```
makramkd, line 27: I think this will create a potentially empty report no?

In order to stay consistent at least with the previous code and before we add new logic, lets check

```golang
// TODO: add logic for gas/token price reporting, they should be reported even
// if merkle roots are empty if there is something to report.
if outcome.MerkleOutcome.OutcomeType != merkleroots.ReportGenerated {
  return nil, nil // nothing to report yet
}
```

commit/report.go
```diff
@@ -38,7 +39,7 @@ func (p *Plugin) ShouldAcceptAttestedReport(
 ) (bool, error) {
 	decodedReport, err := p.reportCodec.Decode(ctx, r.Report)
 	if err != nil {
-		return false, fmt.Errorf("decode commit plugin report: %w", err)
+		return false, fmt.Errorf("decode merkleroot plugin report: %w", err)
```
makramkd, line 42: This is still in the root plugin right?

commit/report.go
```diff
@@ -64,10 +65,10 @@ func (p *Plugin) ShouldTransmitAcceptedReport(
 
 	decodedReport, err := p.reportCodec.Decode(ctx, r.Report)
 	if err != nil {
-		return false, fmt.Errorf("decode commit plugin report: %w", err)
+		return false, fmt.Errorf("decode merkleroot plugin report: %w", err)
```
makramkd, line 68: Root plugin

commit/plugin.go
```diff
@@ -29,8 +42,10 @@ type Plugin struct {
...
```
makramkd, line 48:
```suggestion
	merkleProcessor      merkleroot.Processor
	tokenProcessor       token.Processor
	gasProcessor         fees.Processor
```

commit/report.go
asoliman92: yes, accidental replace all that happened when refactoring.

commit/fees/types.go
```diff
@@ -0,0 +1,20 @@
...
+	GasPrices []cciptypes.GasPriceChain `json:"gasPrices"`
+}
+
+type Observation struct {
+	GasPrices []cciptypes.GasPriceChain `json:"gasPrices"`
+}
+
+func (o Outcome) Sort() {
+	// TODO: implement
+}
```
makramkd, line 20: I feel like we should just enforce that these outcomes that are produced are always correctly sorted in tests.

commit/plugin.go
```diff
@@ -3,12 +3,21 @@ package commit
...
 	"time"
 
 	"github.com/smartcontractkit/libocr/commontypes"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 	libocrtypes "github.com/smartcontractkit/libocr/ragep2p/types"
 
+	"github.com/smartcontractkit/chainlink-ccip/commit/fees"
+	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot"
+	merkleworker "github.com/smartcontractkit/chainlink-ccip/commit/merkleroot"
```
makramkd, line 16: Remove import alias.

shared/chain_support.go
makramkd, line 1: Since this is only used in the commit plugin, wouldn't it be more apt to place it in commit/shared?

shared/sub_plugin.go
makramkd: Please add these kinds of comments to the code so that they're persisted rather than a GH comment which is probably never read after a PR is merged.

commit/fees/types.go
asoliman92: Having outcome implement it makes it clear that we need to sort them. I agree we need to test it as well. I'm going to use the sort inside the Outcome function in each of the sub plugins

shared/chain_support.go
asoliman92: I think it can be used later with exec plugin as well. It has nothing to do with commit only logic. Leaving it in shared makes it easier to discover and use if needed in other places.

dimkouv (commented):
i suggest the following renames

```
fees -> chainfee
token -> tokenprice
```
then we get 
```
chainfee.Processor
tokenprice.Processor
```

commit/merkleroot/processor.go
```diff
@@ -0,0 +1,79 @@
...
+	nodeID       commontypes.OracleID
+	cfg          pluginconfig.CommitPluginConfig
+	lggr         logger.Logger
+	observer     Observer
+	readerSyncer *plugincommon.BackgroundReaderSyncer
+	ccipReader   reader.CCIP
```
dimkouv, line 26: I believe it's better if we move the plugin methods in this file.
Then we have:
```
merkleroot/
   processor.go
   processor_test.go
   types.go
   types_test.go
```

Right now I find it hard to navigate through the different files.

commit/plugin.go
```diff
@@ -18,6 +25,10 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 )
 
+type MerkleObservation = shared.AttributedObservation[merkleroot.Observation]
```
dimkouv, line 28: let's not use just `Merkle`, `MerkleRoots` is better.

commit/plugin.go
```diff
@@ -18,6 +25,10 @@ import (
...
+type TokenObservation = shared.AttributedObservation[token.Observation]
```
dimkouv, line 29: Same: TokenPrices and ChainFee

commit/plugin.go
```diff
@@ -29,8 +40,13 @@ type Plugin struct {
 	lggr              logger.Logger
 	homeChain         reader.HomeChain
 	reportingCfg      ocr3types.ReportingPluginConfig
-	chainSupport      ChainSupport
-	observer          Observer
+	chainSupport      shared.ChainSupport
+	//merkleProcessor   merkleroot.Processor
+	//tokenProcessor    token.Processor
+	//feeProcessor      fees.Processor
```
dimkouv, line 46: Forgot to delete this lines?

commit/merkleroot/outcome.go
```diff
@@ -20,40 +19,47 @@ import (
...
-			p.lggr, p.cfg.MaxReportTransmissionCheckAttempts, previousOutcome, consensusObservation), nextState
+		outcome = checkForReportTransmission(
+			w.lggr, w.cfg.MaxReportTransmissionCheckAttempts, previousOutcome, consensusObservation)
 	default:
-		p.lggr.Warnw("Unexpected next state in Outcome", "state", nextState)
+		w.lggr.Warnw("Unexpected next state in Outcome", "state", nextState)
 		return Outcome{}, nextState
 	}
+
+	outcome.Sort()
```
dimkouv, line 61: Why is this refactoring necessary?

commit/types.go
dimkouv: Determinism is guaranteed. Encoding should not be mixed with determinism.

Imagine `Json Encoding` re-ordering your arrays, not nice :D

dimkouv:
General recommendation for future similar PRs, let's first review/merge a small PR with the general approach / overall design.  Then do the migration of existing code into the new architecture in different PRs.

commit/merkleroot/outcome.go
asoliman92: We need to make sure the lists in the outcome are sorted to be deterministic. The sort was being used in the Encoding step for the Outcome, I moved it inside the Outcome method itself so that the Encode just encodes whatever comes.

commit/fees/types.go
makramkd: Don't think I agree with that, test enforcement is always better than this kind of implicit contract ("plz implement")

commit/merkleroot/processor.go
asoliman92: I think it's more readable having different files for observation, outcome, report,..etc. I can change the other processors to follow the same scheme so that navigating files is easier. @makramkd @winder WDYT? In all cases I don't think this needs to be done in this PR. We want to unblock the work on tokens and fees for the time being.

shared/chain_support.go
```diff
@@ -29,37 +29,37 @@ type ChainSupport interface {
...
-	lggr            logger.Logger
-	homeChain       reader.HomeChain
-	oracleIDToP2pID map[commontypes.OracleID]libocrtypes.PeerID
-	nodeID          commontypes.OracleID
-	destChain       cciptypes.ChainSelector
+	Lggr            logger.Logger
+	HomeChain       reader.HomeChain
+	OracleIDToP2pID map[commontypes.OracleID]libocrtypes.PeerID
+	NodeID          commontypes.OracleID
+	DestChain       cciptypes.ChainSelector
```
makramkd, line 36: Why were these made public?

shared/chain_support.go
```diff
@@ -29,37 +29,37 @@ type ChainSupport interface {
...
 
 type CCIPChainSupport struct {
-	lggr            logger.Logger
-	homeChain       reader.HomeChain
-	oracleIDToP2pID map[commontypes.OracleID]libocrtypes.PeerID
-	nodeID          commontypes.OracleID
-	destChain       cciptypes.ChainSelector
+	Lggr            logger.Logger
+	HomeChain       reader.HomeChain
+	OracleIDToP2pID map[commontypes.OracleID]libocrtypes.PeerID
```
makramkd, line 34:
```suggestion
	OracleIDToP2PID map[commontypes.OracleID]libocrtypes.PeerID
```

commit/types.go
```diff
@@ -58,173 +49,14 @@ func DecodeCommitPluginObservation(encodedObservation []byte) (Observation, erro
...
+	TokensOutcome tokenprice.Outcome `json:"tokensOutcome"`
+	FeesOutcome   chainfee.Outcome   `json:"gasOutcome"`
 }
 
 // Encode encodes an Outcome deterministically
 func (o Outcome) Encode() ([]byte, error) {
-
-	// Sort all lists to ensure deterministic serialization
-	o.sort()
-
```
makramkd, line 227: Feels error prone to remove this, it will ensure that the outcome is always deterministic

commit/tokenprice/types.go
```diff
@@ -0,0 +1,27 @@
...
+	FeedTokenPrices []cciptypes.TokenPrice `json:"feedTokenPrices"`
+	// PriceRegistryTokenUpdates for tokens from the PriceRegistry on the dest chain
+	PriceRegistryTokenUpdates []cciptypes.TokenPrice `json:"priceRegistryTokenPrices"`
+	// Observation time
+	Timestamp time.Time
+}
+
+func (o Outcome) Sort() {
+	// TODO: implement
+}
```
makramkd, line 27: I don't think this function signature makes sense - is it supposed to mutate the `Outcome` object? If so it needs to be a `o *Outcome` receiver (pointer receiver) rather than value receiver. 

That said I think its less error prone to have a `NewOutcome` method that always guarantees a sorted and deterministic outcome rather than this approach.

shared/chain_support.go
asoliman92: It moved to another package

makramkd: Don't think that explains it. These only need to be exported if they're used by external objects that have access to this object. I don't think thats the case? If it is, something is very wrong.

commit/merkleroot/processor.go
makramkd: Different strokes for different folks, I like the locality as long as the file is < 1,000 lines.

commit/tokenprice/types.go
asoliman92: The `NewOutcome` approach is good as well 👍

asoliman92: Anyways let's not change the current approach. Will be better to make in another PR. I'm also going to revert the part where I removed the sort from the encode part.
Created ticket: CCIP-3166

commit/fees/types.go
asoliman92: I don't see the conflict. What's the problem with having both?

commit/merkleroot/processor.go
asoliman92: I like locality if it's < 1000 lines, this one will be above the 1K lines and the tests specifically will be hard to navigate as it will be much more than 1K lines.

commit/merkleworker/types.go
dimkouv: Sounds good, we also want to replace this with proto soon

shared/chain_support.go
asoliman92: Yes. I didn't change how it was being used in commit. Added `New` function for it now.

shared/sub_plugin.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/smartcontractkit/libocr/commontypes"
+)
+
+type AttributedObservation[ObservationType any] struct {
+	OracleID    commontypes.OracleID
+	Observation ObservationType
+}
+
+// SubPlugin This interface is to create multiple SubPlugins/Processors under OCR plugin.
```
dimkouv, line 14: Let's rename subplugin to `PluginProcessor` and get rid of the ambiguity  of SubPlugins/Processors

commit/merkleroot/observation.go
```diff
@@ -74,18 +81,14 @@ type Observer interface {
...
-
 	ObserveFChain() map[cciptypes.ChainSelector]int
 }
 
 type ObserverImpl struct {
 	lggr         logger.Logger
 	homeChain    reader.HomeChain
 	nodeID       commontypes.OracleID
-	chainSupport ChainSupport
+	chainSupport shared.ChainSupport
```
dimkouv, line 91: `shared` is a very generic package name. Let's use something more go idiomatic

commit/merkleroot/outcome.go
```diff
@@ -20,40 +19,46 @@ import (
...
-		p.lggr.Warnw("Get consensus observation failed, empty outcome", "error", err)
+		w.lggr.Warnw("Get consensus observation failed, empty outcome", "error", err)
 		return Outcome{}, nextState
 	}
 
+	var outcome Outcome
 	switch nextState {
 	case SelectingRangesForReport:
-		return reportRangesOutcome(commitQuery, consensusObservation), nextState
+		outcome = reportRangesOutcome(query, consensusObservation)
```
dimkouv, line 50: In my previous comment I meant, why it was necessary to replace `return reportRangesOutcome` with `outcome = reportRangesOutcome`

commit/plugin_e2e_test.go
```diff
@@ -76,36 +79,40 @@ func TestPlugin_E2E_AllNodesAgree(t *testing.T) {
...
+				{ChainSel: sourceChain2, SeqNum: 20},
 			},
 		},
-		OffRampNextSeqNums: []plugintypes.SeqNumChain{
-			{ChainSel: sourceChain1, SeqNum: 10},
-			{ChainSel: sourceChain2, SeqNum: 20},
-		},
-		TokenPrices: make([]ccipocr3.TokenPrice, 0),
-		GasPrices:   make([]ccipocr3.GasPriceChain, 0),
+		//TokenPrices: make([]ccipocr3.TokenPrice, 0),
```
dimkouv, line 110: No need this commented out code

commit/report.go
```diff
@@ -18,16 +20,20 @@ func (p *Plugin) Reports(seqNr uint64, outcomeBytes ocr3types.Outcome) ([]ocr3ty
...
+	rep := cciptypes.NewCommitPluginReport(
+		outcome.MerkleRootOutcome.RootsToReport,
+		outcome.TokenPriceOutcome.TokenPrices,
+		outcome.ChainFeeOutcome.GasPrices,
+	)
 
 	encodedReport, err := p.reportCodec.Encode(context.Background(), rep)
 	if err != nil {
-		return nil, fmt.Errorf("encode commit plugin report: %w", err)
+		return nil, fmt.Errorf("encode merkleroot plugin report: %w", err)
```
dimkouv, line 36: isn't it still a commit plugin report?

commit/tokenprice/types.go
```diff
@@ -0,0 +1,27 @@
...
+type Observation struct {
+	// FeedTokenPrices for tokens from the feeds on the feed chain
+	FeedTokenPrices []cciptypes.TokenPrice `json:"feedTokenPrices"`
+	// PriceRegistryTokenUpdates for tokens from the PriceRegistry on the dest chain
+	PriceRegistryTokenUpdates []cciptypes.TokenPrice `json:"priceRegistryTokenPrices"`
+	// Observation time
+	Timestamp time.Time
+}
+
+func (o *Outcome) Sort() {
```
dimkouv, line 25: +1 for removing Sort()

dimkouv: some outcome might not even be possible to be sorted, e.g. if it's a map

commit/validate_observation.go
```diff
@@ -1,142 +1,65 @@
...
+	merkleObs := MerkleRootObservation{
+		OracleID:    ao.Observer,
+		Observation: obs.MerkleObs,
 	}
 
-	supportsDestChain, err := p.chainSupport.SupportsDestChain(ao.Observer)
+	err = p.merkleProcessor.ValidateObservation(prevOutcome.MerkleRootOutcome, merkleroot.Query{}, merkleObs)
 	if err != nil {
-		return fmt.Errorf("call to supportsDestChain failed: %w", err)
+		return fmt.Errorf("failed to validate MerkleObs: %w", err)
```
dimkouv, line 40: let's use proper names in logs/errs `validate merkle roots observation: %w`

shared/sub_plugin.go
```diff
@@ -0,0 +1,61 @@
...
+// The idea is to separate the logic of different types of observations and outcomes into separate processors.
+// This makes it easier to manage and test the logic of each type of observation/outcome without affecting each others
+// Some of them will implement state machines (e.g. merkleroot), others might implement simpler logic. (e.g. token)
```
dimkouv, line 17: (e.g. token prices)

shared/sub_plugin.go
```diff
@@ -0,0 +1,61 @@
...
+// The OCR plugin becomes a coordinator/collector of these SubPlugins.
+// Example Pseudo code:
+//
+//	OCRPlugin {
+//	  nodeID
+//	  merkleProcessor
```
dimkouv, line 23: renames
`merkleRootsProcessor, tokenPricesProcessor, chainFeeProcessor`

dimkouv: Also rename `OCRPLugin.Observer` to `OCRPlugin.Observation` and probably this example is enough, no need for validate and outcome

dimkouv (changes requested):
Final comments

commit/merkleroot/outcome.go
asoliman92: Preparing for the sorting to come at the end before returning.  reverting it back for now.

commit/tokenprice/types.go
makramkd: I'm not sure I understand, what do you mean by current approach? It seems like this PR changes what was being done in `main`.

commit/tokenprice/types.go
asoliman92: Removed.

commit/merkleroot/observation.go
asoliman92: `utils`? :D

commit/tokenprice/types.go
asoliman92: Not really. `main` already uses sort function in the encode. I just copy pasted it.

commit/plugin.go
```diff
@@ -85,8 +102,147 @@ func NewPlugin(
...
+		feeObservations = append(feeObservations,
+			ChainFeeObservation{
+				OracleID:    ao.Observer,
+				Observation: obs.GasObs,
+			},
+		)
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(3)
```
makramkd, line 201: This code below feels like premature optimization. `Outcome` should be pure and "instant" so this concurrency probably adds more overhead than it gets rid of; benchmarking would be needed to see if thats the case.

commit/plugin.go
```diff
@@ -85,8 +102,147 @@ func NewPlugin(
...
+			prevOutcome.MerkleRootOutcome,
+			merkleroot.Query{},
+			merkleObservations,
+		)
+		if err != nil {
+			p.lggr.Errorw("failed to get merkle outcome", "err", err)
+		}
+
+		*mOut = outcome
+	}(&merkleOutcome)
```
makramkd, line 219: Although this is technically safe because the pointer is only accessed by a single goroutine its unidiomatic and I don't really like it.

commit/plugin.go
```diff
@@ -85,8 +102,147 @@ func NewPlugin(
...
+	go func(mOut *merkleroot.Outcome) {
+		defer wg.Done()
+		outcome, err := p.merkleProcessor.Outcome(
+			prevOutcome.MerkleRootOutcome,
+			merkleroot.Query{},
+			merkleObservations,
+		)
+		if err != nil {
+			p.lggr.Errorw("failed to get merkle outcome", "err", err)
+		}
```
makramkd, line 216: Should return early if there's an error, `outcome` is undefined if `err != nil`

commit/plugin.go
```diff
@@ -85,8 +102,147 @@ func NewPlugin(
...
+	go func(gOut *chainfee.Outcome) {
+		defer wg.Done()
+		outcome, err := p.feeProcessor.Outcome(prevOutcome.ChainFeeOutcome, chainfee.Query{}, feeObservations)
+		if err != nil {
+			p.lggr.Errorw("failed to get gas prices outcome", "err", err)
+		}
+		*gOut = outcome
+	}(&gasOutcome)
+
+	wg.Wait()
```
makramkd, line 239: If you want to insist on using a concurrent approach here lets use errorgroup instead, since it seems like none of the errors above are checked outside of the goroutines themselves.

I wonder if the outcome should be "all or nothing" i.e if one of the processors has an error, should we error out of the outcome completely? The reasoning being that there's no way to know only from the outcome if one of the outcomes is empty due to an error or if its truly empty.

commit/plugin.go
makramkd: Same for other goroutines

commit/tokenprice/types.go
makramkd: Right, I see what you mean. I think next step would be to make sure that Outcome objects are deterministic by construction, i.e it is impossible that an Outcome object exists that isn't "sorted" (or similar)

commit/merkleroot/outcome.go
```diff
@@ -20,25 +19,28 @@ import (
 // - chooses the seq num ranges for the next round
 // - builds a report
 // - checks for the transmission of a previous report
-func (p *Plugin) Outcome(
-	outCtx ocr3types.OutcomeContext, query types.Query, aos []types.AttributedObservation,
-) (ocr3types.Outcome, error) {
+func (w *Processor) Outcome(
```
winder, line 22: nit: could change `w` back to `p`

commit/plugin.go
```diff
@@ -85,8 +101,130 @@ func NewPlugin(
...
+	outCtx ocr3types.OutcomeContext, query types.Query, aos []types.AttributedObservation,
+) (ocr3types.Outcome, error) {
+	prevOutcome := p.decodeOutcome(outCtx.PreviousOutcome)
+
+	var merkleObservations []MerkleRootObservation
+	var tokensObservations []TokenPricesObservation
+	var feeObservations []ChainFeeObservation
+
+	for _, ao := range aos {
+		obs, err := DecodeCommitPluginObservation(ao.Observation)
```
winder, line 172: A function to map the decoded attributed observations to the decoded ones might be handy. I have a `decodedAttributedObservation` type in commit that should use `shared.AttributedObservation`.

commit/merkleroot/processor.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
+	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+	"github.com/smartcontractkit/chainlink-ccip/shared"
+)
+
+// Processor is the processor responsible for
+// reading next messages and building merkle roots for them,
+// It's setup to use RMN to query which messages to include in the merkle root and ensures
+// the newly built merkle roots are the same as RMN roots.
```
winder, line 23:
```suggestion
// the merkle roots are the same as RMN roots.
```

commit/merkleroot/processor.go
winder: I don't have a strong preference, but usually lean towards keeping an interface implementation all in one file. That said execute is huge and I was planning to split it up like this at some point.

Either way I agree that it should be done in a separate PR.

dimkouv (approved):

commit/merkleroot/processor.go
```diff
@@ -0,0 +1,83 @@
...
+	observer     Observer
+	readerSyncer *plugincommon.BackgroundReaderSyncer
+	ccipReader   reader.CCIP
+	reportingCfg ocr3types.ReportingPluginConfig
+	chainSupport shared.ChainSupport
+}
+
+// NewProcessor creates a new Processor
+func NewProcessor(
+	nodeID commontypes.OracleID,
```
winder, line 37: Is this unused, maybe it can be removed?
```suggestion
	oracleID commontypes.OracleID,
```

commit/merkleroot/processor.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := w.readerSyncer.Close(); err != nil {
+		w.lggr.Errorw("error closing reader syncer", "err", err)
+	}
+
+	if err := w.ccipReader.Close(ctx); err != nil {
+		return fmt.Errorf("close ccip reader: %w", err)
+	}
+
+	return nil
+}
```
winder, line 81: These resources are already managed in the primary plugin. Also, is close ever called?

commit/merkleroot/types.go
```diff
@@ -0,0 +1,156 @@
...
+// Sort all fields of the given Outcome
+func (o *Outcome) Sort() {
+	sort.Slice(o.RangesSelectedForReport, func(i, j int) bool {
+		return o.RangesSelectedForReport[i].ChainSel < o.RangesSelectedForReport[j].ChainSel
+	})
+	sort.Slice(o.RootsToReport, func(i, j int) bool {
+		return o.RootsToReport[i].ChainSel < o.RootsToReport[j].ChainSel
+	})
+	sort.Slice(o.OffRampNextSeqNums, func(i, j int) bool {
+		return o.OffRampNextSeqNums[i].ChainSel < o.OffRampNextSeqNums[j].ChainSel
```
winder, line 127: Are we guaranteed to only have a single item per `ChainSel` in each of these slices? I'm pretty sure the answer is yes. If it isn't then this logic is wrong and it should consider `ChainSel` and the other value.

commit/merkleroot/validate_observation.go
```diff
@@ -0,0 +1,180 @@
...
+
+	seenChains := mapset.NewSet[cciptypes.ChainSelector]()
+	for _, root := range merkleRoots {
+		if !observerSupportedChains.Contains(root.ChainSel) {
+			return fmt.Errorf("found merkle root for chain %d, but this chain is not supported by Observer %d",
+				root.ChainSel, observer)
+		}
+
+		if seenChains.Contains(root.ChainSel) {
+			return fmt.Errorf("duplicate merkle root for chain %d", root.ChainSel)
```
winder, line 70: Ah yes, there it is. One ChainSel per slice. I think we should figure out the encoding so that we can use maps.

commit/merkleroot/validate_observation.go
```diff
@@ -0,0 +1,180 @@
...
+	if len(merkleRoots) == 0 {
+		return nil
+	}
+
+	seenChains := mapset.NewSet[cciptypes.ChainSelector]()
+	for _, root := range merkleRoots {
+		if !observerSupportedChains.Contains(root.ChainSel) {
+			return fmt.Errorf("found merkle root for chain %d, but this chain is not supported by Observer %d",
+				root.ChainSel, observer)
+		}
```
winder, line 67: Looks like a complete error message can be created by removing this block and adding a new one outside of the loop:
```
unsupportedChains := observerSupportedChains.Difference(seenChains)
if len(unsupportedChains) > 0 {
    return fmt.Errorf("found merkle root for chains unsupported by oracleID %d: %s", observer, unsupportedChains)
}
```

commit/merkleroot/validate_observation.go
```diff
@@ -0,0 +1,180 @@
...
+		if seenChains.Contains(root.ChainSel) {
+			return fmt.Errorf("duplicate merkle root for chain %d", root.ChainSel)
+		}
+		seenChains.Add(root.ChainSel)
+	}
+
+	return nil
+}
+
+func validateObservedOnRampMaxSeqNums(
```
winder, line 78: Same suggestions as above.

commit/merkleroot/validate_observation.go
```diff
@@ -0,0 +1,180 @@
...
+		if seenChains.Contains(seqNumChain.ChainSel) {
+			return fmt.Errorf("duplicate onRampMaxSeqNum for chain %d", seqNumChain.ChainSel)
+		}
+		seenChains.Add(seqNumChain.ChainSel)
+	}
+
+	return nil
+}
+
+func validateObservedOffRampMaxSeqNums(
```
winder, line 104: Same suggestions as above.

commit/plugin.go
```diff
@@ -85,8 +101,130 @@ func NewPlugin(
...
+	// Across all chains we require at least 2F+1 observations.
+	return ocr3types.QuorumTwoFPlusOne, nil
+}
+
+func (p *Plugin) Observation(
+	ctx context.Context, outCtx ocr3types.OutcomeContext, _ types.Query,
+) (types.Observation, error) {
+	prevOutcome := p.decodeOutcome(outCtx.PreviousOutcome)
+	fChain := p.ObserveFChain()
+	//TODO: Send fchain instead of computing it inside
```
winder, line 124: I think this is what you meant?
```suggestion
	//TODO: Move fchain to a new processor instead of computing it inside MerkleProcessor
```

commit/plugin.go
```diff
@@ -29,8 +40,10 @@ type Plugin struct {
 	lggr              logger.Logger
 	homeChain         reader.HomeChain
 	reportingCfg      ocr3types.ReportingPluginConfig
-	chainSupport      ChainSupport
-	observer          Observer
+	chainSupport      shared.ChainSupport
+	merkleProcessor   shared.PluginProcessor[merkleroot.Query, merkleroot.Observation, merkleroot.Outcome]
+	tokenProcessor    shared.PluginProcessor[tokenprice.Query, tokenprice.Observation, tokenprice.Outcome]
+	feeProcessor      shared.PluginProcessor[chainfee.Query, chainfee.Observation, chainfee.Outcome]
```
winder, line 46: You're not using these generic types so `shared.PluginProcessor` should define them as `any` for a simpler interface.

If you don't need the types, you can now use the generic processor:
```suggestion
	merkleProcessor   shared.PluginProcessor
	tokenProcessor    shared.PluginProcessor
	feeProcessor      shared.PluginProcessor
```

If you do need the types in some cases, specify the concrete type:
```suggestion
	merkleProcessor   merkleroot.Processor // because we need to call ObserveFChain()'
	tokenProcessor    shared.PluginProcessor
	feeProcessor      shared.PluginProcessor
```

commit/plugin.go
```diff
@@ -85,8 +101,130 @@ func NewPlugin(
...
+	}
+	tokenPricesObs, err := p.tokenProcessor.Observation(ctx, prevOutcome.TokenPriceOutcome, tokenprice.Query{})
+	if err != nil {
+		//log error
+		p.lggr.Errorw("failed to get token prices", "err", err)
+	}
+	gasObs, err := p.feeProcessor.Observation(ctx, prevOutcome.ChainFeeOutcome, chainfee.Query{})
+	if err != nil {
+		p.lggr.Errorw("failed to get gas prices", "err", err)
+	}
```
winder, line 137: Except for the deterministic ordering, this could be a fully decoupled generic SubPlugin aggregator.

Replace:
```
	merkleProcessor   shared.PluginProcessor
	tokenProcessor    shared.PluginProcessor
	feeProcessor      shared.PluginProcessor
```

With:
```
subPlugins map[string]shared.SubPlugin Observation function doesn't change when you add another SubPlugin.
```

This loop becomes:
```
var observations map[string]any
for k, processor := range subPlugins {
    // I suppose args need a little more thought. You could have shared k/v's
    // Not sure about the query.
    observations[k] = processor.Observation(ctx, prevOutcome, query)
}
return Encode(observations)
```

The neat part here is since you're using polymorphism, the

commit/plugin.go
```diff
@@ -105,18 +243,18 @@ func (p *Plugin) Close() error {
 	return nil
 }
 
-func (p *Plugin) decodeOutcome(outcome ocr3types.Outcome) (Outcome, State) {
+func (p *Plugin) decodeOutcome(outcome ocr3types.Outcome) Outcome {
```
winder, line 246: Might as well merge this function with `DecodeOutcome` now that the special part has been removed.

commit/types.go
```diff
@@ -27,20 +27,11 @@ func DecodeCommitPluginQuery(encodedQuery []byte) (Query, error) {
...
-	MerkleRoots        []cciptypes.MerkleRootChain     `json:"merkleRoots"`
-	GasPrices          []cciptypes.GasPriceChain       `json:"gasPrices"`
-	TokenPrices        []cciptypes.TokenPrice          `json:"tokenPrices"`
-	OnRampMaxSeqNums   []plugintypes.SeqNumChain       `json:"onRampMaxSeqNums"`
-	OffRampNextSeqNums []plugintypes.SeqNumChain       `json:"offRampNextSeqNums"`
-	FChain             map[cciptypes.ChainSelector]int `json:"fChain"`
+	MerkleObs merkleroot.Observation          `json:"merkleObs"`
+	TokenObs  tokenprice.Observation          `json:"tokenObs"`
+	GasObs    chainfee.Observation            `json:"gasObs"`
+	FChain    map[cciptypes.ChainSelector]int `json:"fChain"`
```
winder, line 34: The next logical step is to have this top level plugin fully generic, I suppose this would end up as `map[string]any`.

shared/chain_support.go
```diff
@@ -31,11 +31,27 @@ type ChainSupport interface {
...
+) CCIPChainSupport {
+	return CCIPChainSupport{
+		lggr:            lggr,
+		homeChain:       homeChain,
+		oracleIDToP2PID: oracleIDToP2PID,
+		nodeID:          nodeID,
+		destChain:       destChain,
+	}
+}
+
```
winder, line 54: Maybe this thing could be passed into the plugin instead of the home chain? And/or expanded to support all the subconsensus stuff.

shared/chain_support.go
winder: Would be nice to share this across both plugins :+1:

commit/plugin.go
asoliman92: I thought about this earlier and didn't go this way to make it simpler, different types will make it more complex once you start having conditions or other mapping for different arguments.

shared/plugin_processor.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/smartcontractkit/libocr/commontypes"
+)
+
+type AttributedObservation[ObservationType any] struct {
+	OracleID    commontypes.OracleID
+	Observation ObservationType
+}
+
+// PluginProcessor This interface is to create multiple SubPlugins/Processors under OCR plugin.
```
winder, line 14: Nit: the names again! Lets use SubPlugin or Processor, not both.

commit/plugin.go
asoliman92: I think we'll need the types when we start testing. and using the plugin makes the one mock enough to use. WDYT?

commit/merkleroot/processor.go
asoliman92: good catch. I think we can remove it from here. Do I need to initialize them as references in processor in this case?

commit/types.go
asoliman92: I like the idea

shared/plugin_processor.go
```diff
@@ -0,0 +1,45 @@
...
+//	  chainFeeProcessor
+//	  ...
+//	}
+//
+//	OCRPlugin.Observation {
+//	  mObs := merkleProcessor.Observation
+//	  tObs := tokenProcessor.Observation
+//	  fObs := feeProcessor.Observation
+//	  return Observation{mObs, tObs, fObs}
+//	}
```
winder, line 35: Godoc supports code blocks.

[Documentation](https://go.dev/doc/comment).
* [relevant section](https://go.dev/doc/comment#code): A code block is a span of indented or blank lines not starting with a bullet list marker or numbered list marker.
```suggestion
//		 type OCRPlugin struct {
//	 	  merkleRootsProcessor
//		   tokenPriceProcessor
//	 	   chainFeeProcessor
//		 }
//
//      // Observation excludes error handling for brevity.
// 	 func (p *OCRPlugin) Observation() ocrtype.Observation {
//	 	 	 return aggregatedObservations{
//            merkleRoots: p.merkleRootsProcessor.Observe(...)
//            tokenPrices: p.tokenPriceProcessor.Observe(...)
//            chainFees: p.chainFeeProcessor.Observe(...)
//          }
//	 	 }
```

commit/merkleroot/processor.go
winder: Hmm, we should consider that a little further. They're now being used concurrently as well. Are they all thread safe?

commit/plugin.go
winder: We don't need to do it in this PR, but I think its a natural progression of this work.

commit/merkleroot/processor.go
asoliman92: We removed the concurrency for simplicity now.

commit/merkleroot/validate_observation.go
```diff
@@ -0,0 +1,180 @@
...
+func validateObservedMerkleRoots(
+	merkleRoots []cciptypes.MerkleRootChain,
+	observer commontypes.OracleID,
+	observerSupportedChains mapset.Set[cciptypes.ChainSelector],
+) error {
+	if len(merkleRoots) == 0 {
+		return nil
+	}
+
+	seenChains := mapset.NewSet[cciptypes.ChainSelector]()
```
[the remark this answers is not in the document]
asoliman92: I'm okay using the `NewThreadUnsafeSet`. It's just usually very alarming for people when they read it and the performance gain won't be much considering the number of selectors we have.

commit/merkleroot/validate_observation.go
asoliman92: Added to ticket CCIP-3166 to fix later.

commit/merkleroot/validate_observation.go
asoliman92: Added to ticket CCIP-3166 to fix later.

commit/merkleroot/validate_observation.go
asoliman92: Added to ticket CCIP-3166 to fix later.

commit/plugin.go
asoliman92: It's called in multiple places so far. Nicer having it this way instead of copying the logic of empty outcome and the error checking 3 times IMO.

shared/chain_support.go
asoliman92: Might worth checking the feasability of this 👍

winder (approved):

commit/plugin.go
asoliman92: We've discussed on a call and decided to keep as is for testability and we'll revisit this if we can make it more generic later.

commit/merkleroot/processor.go
makramkd: We could add `Close` to the interface and call it from the root plugin

makramkd (approved):

winder (approved):

PR #101, opened 2024-09-05
Commit plugin RMN business logic

dimkouv:
### In this PR we Initialize the RMN Client interface and start using it in the commit plugin.

1. All the RMN Communication / retrying / caching logic is abstract for the plugin, behind a simple to use and test interface:
```
	ComputeSignatures(
		ctx context.Context,
		destChain DestChainInfo,
		requestedUpdates []FixedDestLaneUpdateRequest,
	) (*NodeSignatures, error)
```

2. When signature computation fails with a timeout we move to a new implicit state to retry RMN signatures.
In this state OCR nodes send an empty Observation and Resubmit the previous outcome.
That way in the next `Query` phase we can retry RMN signatures computation.

3. There is a feature flag `RMNEnabled` to easily enable/disable this feature.

## Follow-up PRs:
- The implementation of this interface and .proto types generation.
- OCR3Runner integration tests

## Open Questions
- Do we want to retry the OCR round only on RMN timeout or in any error case?


commit/merkleroot/rmn/types.go
```diff
@@ -0,0 +1,36 @@
+package rmn
+
+import "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+type NodeSignatures struct {
```
makramkd, line 5: Wouldn't it just be better to use the proto types directly? Most of these types are identical to whats in the protos right?

commit/merkleroot/outcome.go
```diff
@@ -46,9 +46,14 @@ func (w *Processor) getOutcome(
 
 	switch nextState {
 	case SelectingRangesForReport:
-		return reportRangesOutcome(commitQuery, consensusObservation), nextState
+		return reportRangesOutcome(q, consensusObservation), nextState
 	case BuildingReport:
-		return buildReport(commitQuery, consensusObservation, previousOutcome), nextState
+		if q.RetryRMNSignatures {
+			// We want to retry getting the RMN signatures on the exact same outcome we had before.
+			// The current observations should all be empty.
```
makramkd, line 53: Does it make sense to assert this at all? Maybe in `ValidateObservation`, if `q.Retry` is true, the observation should be empty?

commit/merkleroot/rmn/client.go
```diff
@@ -0,0 +1,19 @@
...
+	"errors"
+)
+
+// ErrTimeout is returned when the signature computation times out.
+var ErrTimeout = errors.New("signature computation timeout")
+
+// Client contains the methods required by the plugin to interact with the RMN nodes.
+type Client interface {
+	// ComputeSignatures computes and returns the signatures for the provided lane updates.
+	ComputeSignatures(
```
makramkd, line 14: Does this correspond to the `ReportSignatureRequest` in the RMN design doc?

IMO: in order to make things consistent I'd rather we have consistent naming across design docs and code. That way when RMN maintainers read our code its pretty clear what these functions are doing (and vice versa, if it ever comes to us reading RMN code)

commit/merkleroot/rmn/types.go
makramkd: I dislike these custom types because I'm not sure there's much benefit. You'll have to manually edit them if/when the protos change, and for the `Client` interface, you can just use the proto type there as well directly and all will be fine when mocking

makramkd: Another argument: you have to translate from this type to the proto anyway, before marshaling the protos. So this just adds an unnecessary step to the process

commit/merkleroot/types.go
```diff
@@ -5,20 +5,14 @@ import (
...
-	MerkleRoots         []cciptypes.MerkleRootChain
-}
-
-func NewCommitQuery(rmnOnRampMaxSeqNums []plugintypes.SeqNumChain, merkleRoots []cciptypes.MerkleRootChain) Query {
-	return Query{
-		RmnOnRampMaxSeqNums: rmnOnRampMaxSeqNums,
-		MerkleRoots:         merkleRoots,
-	}
+	RetryRMNSignatures bool
+	RMNSignatures      *rmn.NodeSignatures
```
makramkd, line 15: Another advantage of using the protos - we can just plug in `ReportingPluginQuery` here directly, which has signatures and `FixedDestLaneUpdate` objects.

commit/merkleroot/query.go
```diff
@@ -0,0 +1,61 @@
...
+	offRampAddress, err := w.ccipReader.GetContractAddress(consts.ContractNameOffRamp, w.cfg.DestChain)
+	if err != nil {
+		return Query{}, fmt.Errorf("get offRamp contract address: %w", err)
+	}
+	dstChainInfo := rmn.DestChainInfo{
+		Chain:          w.cfg.DestChain,
+		OffRampAddress: offRampAddress,
+	}
+
+	reqUpdates := make([]rmn.FixedDestLaneUpdateRequest, 0, len(prevOutcome.RangesSelectedForReport))
```
winder, line 31: Could this be a map so that uniqueness is clear?
```suggestion
	reqUpdates := make(map[ChainSelector]rmn.FixedDestLaneUpdateRequest, 0, len(prevOutcome.RangesSelectedForReport))
```

commit/merkleroot/rmn/types.go
dimkouv: Yeah agree, can replace them while working on
> The implementation of this interface and .proto types generation.

commit/merkleroot/outcome.go
dimkouv: Makes sense, yeah

commit/merkleroot/rmn/client.go
```diff
@@ -0,0 +1,19 @@
...
+		ctx context.Context,
+		destChain DestChainInfo,
+		requestedUpdates []FixedDestLaneUpdateRequest,
```
winder, line 17: Mentioned above, would be great to express uniqueness by using a map here:
```suggestion
		requestedUpdates map[ChainSelector]FixedDestLaneUpdateRequest,
```

commit/merkleroot/rmn/types.go
```diff
@@ -0,0 +1,36 @@
...
+	// Signatures are the ECDSA signatures for the lane updates for each node.
+	// NOTE: A signature[i] corresponds to the whole LaneUpdates slice and NOT LaneUpdates[i].
+	Signatures  []ECDSASignature
+	LaneUpdates []FixedDestLaneUpdate
+}
+
+type ECDSASignature struct {
+	R []byte
+	S []byte
+}
```
winder, line 15: Please use a stdlib, perhaps [crypto/ecdsa](https://pkg.go.dev/crypto/ecdsa)?

winder (commented):
Some initial feedback.

commit/merkleroot/rmn/client.go
dimkouv: Not really since it encapsulates both `ObservationReq` and `ReportSignatureReq`. But I renamed it a bit to make it more clear.

commit/merkleroot/rmn/types.go
winder: Where are these protos?
1. I agree that we should avoid copying types.
2. What about the [crypto/ecdsa](https://pkg.go.dev/crypto/ecdsa) package?

commit/merkleroot/rmn/client.go
winder: You're describing [domain driven design](https://en.wikipedia.org/wiki/Domain-driven_design). I haven't read the books but have friends who swear by it.

commit/merkleroot/outcome.go
```diff
@@ -46,9 +46,14 @@ func (w *Processor) getOutcome(
...
+			return previousOutcome, BuildingReport
+		}
```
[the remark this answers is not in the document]
dimkouv: > How many rounds might we stay in the RetryRMNSignatures state?

The way it works right now we retry forever until we get the signatures.
There are some improvements we can do here though.

commit/merkleroot/query.go
dimkouv: It's similar to the shared proto types, so better keep it consistent with rmn types. Also the types will be replaced with the auto-generated ones.

commit/merkleroot/rmn/client.go
dimkouv: Replied above

commit/merkleroot/rmn/client.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"context"
+	"errors"
+)
+
+// ErrTimeout is returned when the signature computation times out.
+var ErrTimeout = errors.New("signature computation timeout")
+
+// Client contains the methods required by the plugin to interact with the RMN nodes.
+type Client interface {
```
makramkd, line 12: Alright I see, so this Client encapsulates both ObservationRequest and ReportSignatureRequest? In that case can we make it clear how this interface's functions/behavior map to what is in the RMN DD?

commit/merkleroot/outcome.go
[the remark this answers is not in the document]
winder: IIRC the "wait for transmission" logic would only retry for 5 rounds. Something like that is needed so that the system would eventually recover from errors.

commit/merkleroot/rmn/types.go
winder: I guess if they're opaque types it doesn't really matter. They just get passed along to the contracts or will we verify them offchain?

commit/merkleroot/query.go
winder: I'm not suggesting you modify `FixedDestLaneUpdateRequest`.

It needs to be re-encapsulated in an `ObservationRequest` anyway, so this slice needs some processing either way:
```
message ObservationRequest {
  LaneDest lane_dest = 1; // could be implied
  repeated FixedDestLaneUpdateRequest fixed_dest_lane_update_requests = 2;
}
```

Putting it in a map avoids a category of bugs so it seems worthwhile.

commit/merkleroot/outcome.go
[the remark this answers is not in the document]
dimkouv: But still, I believe it will end up requesting signatures, just for bigger intervals.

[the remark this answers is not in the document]
dimkouv: What we can do is move forward only with the ones we got enough sigs

commit/merkleroot/rmn/types.go
dimkouv: We can verify them offchain for extra safety.

commit/merkleroot/rmn/client.go
dimkouv: added a comment explaining this

commit/merkleroot/types.go
winder: Are you expecting these to change? Looks like the report tyeps changed on RMNRemote.sol in the latest batch of contract changes

image

commit/merkleroot/query.go
dimkouv: Oh I see, yeah but maybe in the future we actually want to pass multiple ranges for the same chain since RMN supports it, so IMO we shouldn't treat it as unique.

commit/merkleroot/types.go
dimkouv: I don't really expect them to change

commit/merkleroot/rmn/types.go
makramkd: Yeah we will have an offchain version of the RMNRemote.verify function (here: https://github.com/smartcontractkit/chainlink/blob/70223b96e9862c02e6779cf37cc5d358766b0ea2/contracts/src/v0.8/ccip/rmn/RMNRemote.sol#L111-L123)

pluginconfig/commit.go
```diff
@@ -33,6 +33,13 @@ type CommitPluginConfig struct {
 
 	// OffchainConfig is the offchain config set for the commit DON.
 	OffchainConfig CommitOffchainConfig `json:"offchainConfig"`
+
+	// RMNEnabled is a flag to enable/disable RMN signature verification.
+	RMNEnabled bool `json:"rmnEnabled"`
+
+	// RMNSignaturesTimeout is the timeout for RMN signature verification.
+	// Typically set to `MaxQueryDuration - e`, where e some small duration.
+	RMNSignaturesTimeout time.Duration `json:"rmnSignaturesTimeout"`
```
makramkd, line 42: Would it be simpler to set the `e` here instead? Cuz the `ctx` for query will already be a `context.WithTimeout(MaxQueryDuration)`, so you can do something like:

```golang
// in query 
queryDeadline := ctx.Deadline()
// set earlier deadline for sig fetching
sigDeadline := queryDeadline.Sub(p.offchainConfig.RMNQueryE)
ctx, cancel := context.WithDeadline(ctx, sigDeadline)
defer cancel()
// use ctx...
```

makramkd (approved):
LGTM overall lets merge and iterate

winder (approved):

pluginconfig/commit.go
dimkouv: Yeah maybe safer also. I'll do it next.

PR #106, opened 2024-09-06
RMN Autogenerated types

dimkouv:
- Autogenerate RMN related types from a `.proto` file.
- CI checks if generated go file is outdated.
- Replace existing types with proto generated types.
- Ensure coverage report is not affected by auto-generated files.


.github/workflows/codegen.yml
```diff
@@ -24,9 +24,17 @@ jobs:
           go-version: ${{ matrix.go-version }}
       - name: Display Go version
         run: go version
-      - name: Re-Generate mocks
+      - name: Install protoc
+        run: |
+          PROTOC_ZIP=protoc-28.0-linux-x86_64.zip
```
mateusz-sekara, line 29: Wouldn't it be better to install protoc on the Makefile level to have less logic here?
I've checked how we do it in chainlink-common and chainlink repos and it seems that we have protoc installed and executed on the Makefile level. I guess it's better because you have one place that can be used both for CI and local setup. Also, it would be more aligned with other repos. WDYT?

https://github.com/smartcontractkit/chainlink-common/blob/2ff0f9628f4d5957d52061507971c7820080caee/script/install-protoc.sh#L16
https://github.com/smartcontractkit/chainlink-common/blob/2ff0f9628f4d5957d52061507971c7820080caee/Makefile

https://github.com/smartcontractkit/ccip/blob/95ffd86c84773069654251ee17999eff5fa13e5f/GNUmakefile#L150

dimkouv: agree, on it

asoliman92 (dismissed):

mateusz-sekara (dismissed):

asoliman92 (approved):

PR #107, opened 2024-09-06
Token price processor - Observation [CCIP-2958]

asoliman92:
**TokenPrice Observation phase**

* Add GetFeeQuoterTokenUpdates in onchain_prices_reader.go
* Observe feedPrices, feeQuoterUpdates, fDestChain and current timestamp (UTC)

Note: We're observing Fchain, although we already observe it in merkleroot processor. This is because it's more straight forward if each processor takes care of what they need. For example the token price processor don't need all the Fs for all chains, just the destination fChain. Also reduces complexity of the PluginProcessor design as sending an extra parameter will require changing the current interface. If we see that it's affecting performance in the future then we can change the interface.


commit/tokenprice/processor.go
```diff
@@ -47,6 +87,97 @@ func (w *Processor) ValidateObservation(
 	return nil
 }
 
+func (p *Processor) ObserveFDestChain() (*int, error) {
```
makramkd, line 90: This is a really wacky function signature, why return a pointer to an int?

commit/tokenprice/processor.go
```diff
@@ -47,6 +87,97 @@ func (w *Processor) ValidateObservation(
...
+	tokensToQuery := maps.Keys(p.cfg.OffchainConfig.TokenInfo)
+	p.lggr.Infow("observing feed token prices")
+	tokenPrices, err := p.tokenPriceReader.GetTokenFeedPricesUSD(ctx, tokensToQuery)
+	if err != nil {
+		p.lggr.Errorw("call to GetTokenFeedPricesUSD failed", "err", err)
+		return []cciptypes.TokenPrice{}
+	}
+
+	// Continue if we couldn't fetch all prices
+	if len(tokenPrices) != len(tokensToQuery) {
```
makramkd, line 133: The comment and the line after don't line up. What does "continue" mean in this context?

commit/tokenprice/processor.go
```diff
@@ -3,42 +3,82 @@ package tokenprice
...
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	"github.com/smartcontractkit/libocr/commontypes"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 
-	"github.com/smartcontractkit/chainlink-ccip/shared"
+	"golang.org/x/exp/maps"
 )
 
 type Processor struct {
```
makramkd, line 23: This type doesn't need to be exported, can be private. See https://github.com/smartcontractkit/chainlink-ccip/pull/108#discussion_r1749967753 for more info

commit/tokenprice/types.go
```diff
@@ -14,10 +17,23 @@ type Outcome struct {
...
 
 type Observation struct {
-	// FeedTokenPrices for tokens from the feeds on the feed chain
-	FeedTokenPrices []cciptypes.TokenPrice `json:"feedTokenPrices"`
-	// PriceRegistryTokenUpdates for tokens from the PriceRegistry on the dest chain
-	PriceRegistryTokenUpdates []cciptypes.TokenPrice `json:"priceRegistryTokenPrices"`
-	// Observation time
-	Timestamp time.Time
+	FeedTokenPrices       []cciptypes.TokenPrice            `json:"feedTokenPrices"`
+	FeeQuoterTokenUpdates map[types.Account]NumericalUpdate `json:"FeeQuoterTokenUpdates"`
```
makramkd, line 21:
```suggestion
	FeeQuoterTokenUpdates map[types.Account]NumericalUpdate `json:"feeQuoterTokenUpdates"`
```

commit/tokenprice/types.go
```diff
@@ -14,10 +17,23 @@ type Outcome struct {
...
+	// ObserveFeedTokenPrices returns the latest token prices from the feed chain
+	ObserveFeedTokenPrices(ctx context.Context) []cciptypes.TokenPrice
+
+	// ObserveFeeQuoterTokenUpdates returns the latest token prices from the FeeQuoter on the dest chain
+	ObserveFeeQuoterTokenUpdates(ctx context.Context) map[types.Account]NumericalUpdate
+
+	ObserveFDestChain() (*int, error)
+}
+
+type NumericalUpdate struct {
```
makramkd, line 36: name suggestion:

```suggestion
type TimestampedBig struct {
```

internal/reader/onchain_prices_reader.go
```diff
@@ -4,40 +4,57 @@ import (
...
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
 
 	"golang.org/x/sync/errgroup"
 )
 
-type TokenPrices interface {
-	// GetTokenPricesUSD returns the prices of the provided tokens in USD.
+type PriceReader interface {
+	// GetTokenFeedPricesUSD returns the prices of the provided tokens in USD normalized to e18. 1USD=1e18USD.
```
makramkd, line 23: > 1USD=1e18USD

This doesn't really make much sense as an equation - maybe just copy/paste the comment in the FeeQuoter.sol contract?

internal/reader/onchain_prices_reader.go
```diff
@@ -4,40 +4,57 @@ import (
...
+	GetFeeQuoterTokenUpdates(
+		ctx context.Context,
+		tokens []ocr2types.Account,
+	) (map[ocr2types.Account]NumericalUpdate, error)
+}
+
+// NumericalUpdate On-chain struct `TimestampedPackedUint224`:
+// nolint:lll
+// https://github.com/smartcontractkit/chainlink/blob/2d77ff4623d0a0032533c89f32fc251400382455/contracts/src/v0.8/ccip/libraries/Internal.sol#L43-L47
+type NumericalUpdate struct {
```
makramkd, line 37: This type is duplicated across different files/packages

asoliman92: Yes, it's going to be cyclic dependency if I have a new package for the shared types and use it in the internal/reader and commit as commit uses internal/reader.

commit/tokenprice/processor.go
makramkd: Yeah, return 0. If an error is returned, the value should never be used as its undefined.

internal/reader/onchain_prices_reader.go
makramkd: I don't think that's true. There is no cycle if you have the following organization:
* shared
* internal/reader imports shared
* commit imports internal/reader and shared
To get a cycle you need to have e.g commit import internal/reader which imports commit.

asoliman92: fixing.

commit/tokenprice/processor.go
asoliman92: Not sure I get what you want here. I use it to do `maps.Keys(..)` inside the functions.

makramkd (approved):

makramkd (approved):

PR #108, opened 2024-09-06
RMN Client logic implementation

dimkouv:
Implementation of RMN Client according to the RMN Specification.

- The client is capable of performing the retries as described in the doc.
- The implementation is e2e tested along with the retrying logic.

There are some TODOs that will be covered in follow-up PRs.

The RMN communication logic is a bit complex since it's async and contains some special retrying according to some timers. Please review this pieces properly.

Thank you.


commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+		rmnRemoteAddress:                        rmnRemoteAddress,
+		rmnHomeConfigDigest:                     rmnHomeConfigDigest,
+		minObservers:                            minObservers,
+		minSigners:                              minSigners,
+		observationsInitialRequestTimerDuration: observationsInitialRequestTimerDuration,
+		reportsInitialRequestTimerDuration:      reportsInitialRequestTimerDuration,
+	}
+}
+
+func (c *PBClient) ComputeReportSignatures(
```
winder, line 93:
```suggestion
// ComputeReportSignatures sends a request to each rmn node to handle requests and build a signature.
func (c *PBClient) ComputeReportSignatures(
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+
+	// For each lane update request send an observation request to at most 'minObservers' number of rmn nodes.
+	for sourceChain, lur := range lursPerChain {
+		requestedNodes[sourceChain] = mapset.NewSet[uint32]()
+
+		// At this point we assume at least minObservers RMN nodes are available for each source chain.
+		for nodeID := range lur.rmnNodes.Iter() {
+			if requestedNodes[sourceChain].Cardinality() >= c.minObservers {
+				break
+			}
```
winder, line 156: Why at most? What if you get a bad observation from one of them?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	return nil
+}
+
+type RawRmnClient interface {
+	Send(rmnNodeID uint32, request []byte) error
+	// Recv returns a channel which can be used to listen on for
+	// responses by all RMN nodes. This is expected to be monitored
+	// by the plugin in order to get RMN responses.
+	Recv() <-chan RawRmnResponse
+}
```
winder, line 524: Does send and Recv need to be separated? I think the code would be simpler if you pass the response channel into the send function. Alternatively you could make the channel in the send function and return it for each send, but then you'd have to read each channel.

winder (commented):
Some initial feedback

commit/merkleroot/rmn/client.go
dimkouv: There is a retrying logic, this was written according to the RMN design doc.

commit/merkleroot/rmn/client.go
dimkouv: This interface was given to me as is, but I believe we can make such changes.
@makramkd @gtklocker ?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	Signatures []*rmnpb.EcdsaSignature
+	// LaneUpdates are the lane updates for each source chain that we got the signatures for.
 	// NOTE: A signature[i] corresponds to the whole LaneUpdates slice and NOT LaneUpdates[i].
-	Signatures  []rmnpb.EcdsaSignature
-	LaneUpdates []rmnpb.FixedDestLaneUpdate
+	LaneUpdates []*rmnpb.FixedDestLaneUpdate
+}
+
+// PBClient is the base RMN Client implementation.
+type PBClient struct {
```
makramkd, line 55: Just a basic naming thing but `PBClient` is not a great name, the RMN client will always use protobuf. I don't think this type really needs to be exported either so it can just be called `client`? The constructor can actually return the private pointer `*client` or the interface type `Client` and still be usable. See top answer [here](https://stackoverflow.com/questions/37952354/golang-public-method-to-private-struct-is-this-have-any-use-case)

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	lggr                                    logger.Logger
+	rawRmnClient                            RawRmnClient
+	rmnNodes                                []RMNNodeInfo
+	rmnRemoteAddress                        []byte
+	rmnHomeConfigDigest                     []byte
```
makramkd, line 60: I think we can make this `[32]byte` safely or even use `cciptypes.Bytes32` to get some convenience JSON/string encoding for free

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
-	LaneUpdates []rmnpb.FixedDestLaneUpdate
+	LaneUpdates []*rmnpb.FixedDestLaneUpdate
+}
+
+// PBClient is the base RMN Client implementation.
+type PBClient struct {
+	lggr                                    logger.Logger
+	rawRmnClient                            RawRmnClient
+	rmnNodes                                []RMNNodeInfo
+	rmnRemoteAddress                        []byte
```
makramkd, line 59: Ditto here, `cciptypes.Bytes`?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
-	Signatures  []rmnpb.EcdsaSignature
-	LaneUpdates []rmnpb.FixedDestLaneUpdate
+	LaneUpdates []*rmnpb.FixedDestLaneUpdate
+}
+
+// PBClient is the base RMN Client implementation.
+type PBClient struct {
+	lggr                                    logger.Logger
+	rawRmnClient                            RawRmnClient
+	rmnNodes                                []RMNNodeInfo
```
makramkd, line 58: It is critical that these nodes are in the same order in this slice as the RMN home config right? Maybe add a comment about that?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	rmnHomeConfigDigest                     []byte
+	minObservers                            int
+	minSigners                              int
+	observationsInitialRequestTimerDuration time.Duration
+	reportsInitialRequestTimerDuration      time.Duration
+}
+
+// RMNNodeInfo contains the information about an RMN node.
+type RMNNodeInfo struct {
+	ID                    uint32 // ID is the index of this node in the RMN config
```
makramkd, line 69: nit:
```suggestion
    // ID is the index of this node in the RMN config
	ID                    uint32 
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	SupportedSourceChains mapset.Set[uint64]
```
makramkd, line 70: Are these selectors or EVM chain IDs? If selectors maybe can use `cciptypes.ChainSelector`?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+		for _, node := range c.rmnNodes {
+			if node.SupportedSourceChains.Contains(updateReq.LaneSource.SourceChainSelector) {
+				lursPerChain[updateReq.LaneSource.SourceChainSelector].rmnNodes.Add(node.ID)
+			}
+		}
+	}
+
+	// Filter out the lane update requests for chains without enough RMN nodes supporting them.
+	for chain, l := range lursPerChain {
+		if l.rmnNodes.Cardinality() < c.minObservers {
```
makramkd, line 125: This should only happen on misconfigs right?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+			delete(lursPerChain, chain)
+		}
+	}
+
+	if len(lursPerChain) == 0 {
+		return nil, errors.New("no source chains with enough RMN nodes, nothing to do")
```
makramkd: Same here, only on misconfigs right?

Might be useful to log the exact misconfiguration. For example, I think this would only happen if:

* RMN is enabled, AND
* `minObservers` is not set correctly on ALL chains

Is this correct? If so might be useful to include in the error (which will show up in a log) making it easy to diagnose the situation.

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+) ([]rmnSignedObservationWithMeta, error) {
+	requestedNodes := make(map[uint64]mapset.Set[uint32])                   // sourceChain -> requested rmnNodeIDs
+	requestsPerNode := make(map[uint32][]*rmnpb.FixedDestLaneUpdateRequest) // grouped requests for each node
+
+	// For each lane update request send an observation request to at most 'minObservers' number of rmn nodes.
+	for sourceChain, lur := range lursPerChain {
+		requestedNodes[sourceChain] = mapset.NewSet[uint32]()
+
+		// At this point we assume at least minObservers RMN nodes are available for each source chain.
+		for nodeID := range lur.rmnNodes.Iter() {
```
makramkd: I think we should shuffle this node list prior to iterating and choosing the RMN node to send to right?

commit/merkleroot/rmn/common.go
```diff
@@ -0,0 +1,44 @@
...
+	mapset "github.com/deckarep/golang-set/v2"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn/rmnpb"
+)
+
+func (c *PBClient) marshalAndSend(req *rmnpb.Request, nodeID uint32) error {
+	reqBytes, err := proto.Marshal(req)
+	if err != nil {
+		return fmt.Errorf("proto marshal: %w", err)
```
makramkd, line 15: We use protobuf in many parts of the stack, would be useful to provide more context:

```suggestion
		return fmt.Errorf("proto marshal RMN request: %w", err)
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+		perm := rand.Perm(len(s))
+		for _, i := range perm {
+			ch <- s[i]
+		}
+	}()
+	return ch
+}
+
+func newRequestID() uint64 {
+	now := uint64(time.Now().UTC().UnixNano())
```
makramkd, line 573: q: do we need `UnixNano` here? Why not just `Unix` (which is seconds from the epoch) and you'd be able to cast it to `uint32` safely? (will last us till 2106 😄 )

commit/merkleroot/rmn/client.go
makramkd: I think you can use `randomIter`, seems like you use it somewhere else just not here.

commit/merkleroot/rmn/client.go
makramkd: The raw client interface is not set in stone, but I think we need to actually implement this client to see what is possible. We can come back and change it if we see that passing in a chan is doable and easy on the implementation side

makramkd: I believe that the thinking behind `Recv` is you can call it "anytime" (i.e before you even send any requests). It can be considered the channel that ALL RMN comms arrive at and will have the same lifetime as the RMN raw client itself.

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	}
+}
+
+func (c *PBClient) validateSignedObservationResponse(
+	rmnNodeID uint32,
+	lurs map[uint64]lurWithMeta,
+	signedObs *rmnpb.SignedObservation,
+	destChain *rmnpb.LaneDest,
+) error {
+	if signedObs.Observation.LaneDest != destChain {
```
makramkd: Is this comparison correct? Seems to be comparing pointers which may not work as expected?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+		return fmt.Errorf("unexpected lane dest %v", signedObs.Observation.LaneDest)
+	}
+
+	if !bytes.Equal(signedObs.Observation.RmnHomeContractConfigDigest, c.rmnHomeConfigDigest) {
+		return fmt.Errorf("unexpected rmn home contract config digest %v",
```
makramkd, line 312: Can use `%x` here for the format verb

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+
+		if !lur.rmnNodes.Contains(rmnNodeID) {
+			return fmt.Errorf("rmn node %d not expected to read chain %d",
+				rmnNodeID, signedObsLu.LaneSource.SourceChainSelector)
+		}
+
+		if lur.lur.ClosedInterval != signedObsLu.ClosedInterval {
+			return fmt.Errorf("unexpected closed interval %v", signedObsLu.ClosedInterval)
+		}
+		if lur.lur.LaneSource != signedObsLu.LaneSource {
```
makramkd: Ditto about these comparisons, I _think_ these are pointers, are these comparisons legit or just pointer equality (addresses)?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	// e.g.
+	// The following nodes support the following chains and minObservers=2:
+	// node1: [1]	node2:[1,2,3]	node3:[1,2,3]
+	//
+	// node1: getObservations(1)          ->             never_responds
+	// node2: getObservations(1,2,3)      -> responds_ok
+	// node3: getObservations(2,3)        ->     responds_ok
+	// node3: getObservations(1)          ->                         responds_ok (after timeout we made a new request)
+
+	sigObservations := make([]*rmnpb.AttributedSignedObservation, 0)
```
makramkd, line 365: style nit: nil is a valid slice (see https://github.com/uber-go/guide/blob/master/style.md#nil-is-a-valid-slice)

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +33,533 @@ type Client interface {
...
+	ch := make(chan T)
+	go func() {
+		defer close(ch)
+		perm := rand.Perm(len(s))
+		for _, i := range perm {
+			ch <- s[i]
+		}
+	}()
+	return ch
+}
```
makramkd, line 570: This function seems overly complex. Doesn't it just shuffle an array? Why create channels and goroutines? Why not just use `rand.Perm` directly in the loop and access that particular index in the slice?

```golang
for _, idx := range rand.Perm(len(s)) {
  // check if rmnNode[idx] has already been used, if so, skip
  // else, use
}
```

commit/merkleroot/rmn/client.go
winder: It seems strange because you might have enough observations w/o retries if you query all available rmn nodes.

commit/merkleroot/rmn/client.go
dimkouv: Not required, there is a `NodeIDX` field in RMNNodeInfo.

commit/merkleroot/rmn/client.go
dimkouv: Yeah. When not enough rmn nodes configured for a chain.

commit/merkleroot/rmn/client.go
dimkouv: I thought Iter() would iterate over the elements randomly because it's a set, but it doesn't, I am curious if mapset really persists ordering.

Changed it with randomIter(rmnNodes.ToSlice())

commit/merkleroot/rmn/client.go
winder: The more I think about it, I the less sense it makes to have concurrency in the client at all. Follow the pattern used in the go http client, this code can focus on communication. Since communication is localized to a single function you don't need anything more complicated than a go routine to send the requests and put the response in a go routine, and an inline range to process results.

```
respChan := make(chan RawRMNResponse)
go func() {
    for id, req := range requests {
        resp, err <- Send(ctx, id, req)  // the ctx is also missing here.
        if err != nil {
            cloe(respChan) // or the err group thingie
        }
    }
}

var results []RawRMNResponse
for result := range respChan {
    results = append(results, result)
}

aggregate(results)
```

commit/merkleroot/rmn/client.go
winder: There's also a section of effective go dedicated to this idea. They call it Generality: https://go.dev/doc/effective_go#generality

commit/merkleroot/rmn/client.go
winder: Didn't realize that was in the uber guide recommended it this way
```suggestion
	var sigObservations []*rmnpb.AttributedSignedObservation
```

commit/merkleroot/rmn/client.go
dimkouv: RMN Team wanted to avoid that.

commit/merkleroot/rmn/client.go
dimkouv: Atm, this interface did fit nice in the approach and the testing. We can revisit in a follow-up

commit/merkleroot/rmn/client.go
dimkouv: agree

commit/merkleroot/rmn/client.go
dimkouv: used `cciptypes.Bytes` to match pb type `[]byte`

commit/merkleroot/rmn/client.go
dimkouv: y

commit/merkleroot/rmn/client.go
dimkouv: not sure i follow, what's the suggestion, dmed

commit/merkleroot/rmn/client.go
dimkouv: selectors, ok

commit/merkleroot/rmn/client.go
dimkouv: done

commit/merkleroot/rmn/client.go
dimkouv: Replaced with Random

commit/merkleroot/rmn/client.go
dimkouv: y

commit/merkleroot/rmn/client.go
```diff
@@ -1,14 +1,36 @@
...
 
 	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn/rmnpb"
 )
 
-// ErrTimeout is returned when the signature computation times out.
-var ErrTimeout = errors.New("signature computation timeout")
+// TODO: testing - more unit tests and some code cleanup.
+// TODO: use the populated Query / sig verification / etc..
+// TODO: rmn config watcher
+
```
dimkouv, line 25: This TODOs will come in follow-up PRs

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+// RMNNodeInfo contains the information about an RMN node.
+type RMNNodeInfo struct {
+	// ID is the index of this node in the RMN config
+	ID                    uint32
+	SupportedSourceChains mapset.Set[cciptypes.ChainSelector]
+	IsSigner              bool
+}
+
+// Newclient creates a new RMN Client to be used by the plugin.
+func Newclient(
```
makramkd, line 78:
```suggestion
func NewClient(
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+	minObservers                            int
+	minSigners                              int
+	observationsInitialRequestTimerDuration time.Duration
+	reportsInitialRequestTimerDuration      time.Duration
+}
+
+// RMNNodeInfo contains the information about an RMN node.
+type RMNNodeInfo struct {
+	// ID is the index of this node in the RMN config
+	ID                    uint32
```
makramkd, line 72: Very nitty: would be nice to have a type alias of this `type RMNNodeID = uint32` and can use that in all the types below to be a bit more expressive.

commit/merkleroot/rmn/client_test.go
```diff
@@ -0,0 +1,376 @@
...
+	lggr           logger.Logger
+	ctx            context.Context
+	name           string
+	t              *testing.T
+	rmnClient      *client
+	rawRmnClient   *mockRawRmnClient
+	updateRequests []*rmnpb.FixedDestLaneUpdateRequest
+}
+
+func TestPBClientComputeReportSignatures(t *testing.T) {
```
makramkd, line 44:
```suggestion
func TestClient_ComputeReportSignatures(t *testing.T) {
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+	c.lggr.Infow("Waiting for observation requests", "requestIDs", requestIDs.String())
+
+	respChan := c.rawRmnClient.Recv()
+
+	finishedRequestIDs := mapset.NewSet[uint64]()
+	rmnObservationResponses := make([]rmnSignedObservationWithMeta, 0)
+	merkleRootsCount := make(map[uint64]map[string]int) // sourceChain -> merkleRoot -> count
+
+	initialObservationRequestTimer := time.NewTimer(c.observationsInitialRequestTimerDuration)
+	defer initialObservationRequestTimer.Stop()
```
makramkd, line 238: I haven't read the tests yet but this might be a way to avoid instantiating timers in your test completely to test this function:

```golang
func (c *client) listenForRmnObservationResponses(
	ctx context.Context,
	destChain *rmnpb.LaneDest,
	requestIDs mapset.Set[uint64],
	lursPerChain map[uint64]updateRequestWithMeta,
	requestedNodes map[uint64]mapset.Set[uint32],
    timerChan chan time.Time, // NEW arg
)
```

The idea here being you inject the timer channel into the listenForRmnObservationResponses func instead of creating the timer inside. Then, in the tests, you have full control over both `respChan` and `timerChan` and can write to whichever one you like to trigger the appropriate action. It should make for less flakey (or hopefully not flakey at all) tests.

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+	LaneUpdates []*rmnpb.FixedDestLaneUpdate
+}
+
+// client is the base RMN Client implementation.
+type client struct {
+	lggr                                    logger.Logger
+	rawRmnClient                            RawRmnClient
+	rmnNodes                                []RMNNodeInfo
+	rmnRemoteAddress                        cciptypes.Bytes
+	rmnHomeConfigDigest                     cciptypes.Bytes
```
makramkd, line 62: I believe would always be a `Bytes32`, but Kostis can confirm.

```suggestion
	rmnHomeConfigDigest                     cciptypes.Bytes32
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+			parsedResp, err := c.parseResponse(&resp, requestIDs, finishedRequestIDs)
+			if err != nil {
+				c.lggr.Warnw("failed to parse RMN observation response", "err", err)
+				continue
+			}
+			finishedRequestIDs.Add(parsedResp.RequestId)
+
+			signedObs := parsedResp.GetSignedObservation()
+			if signedObs == nil {
+				c.lggr.Infof("RMN node returned an unexpected type of response: %+v", parsedResp.Response)
```
makramkd, line 251:
```suggestion
				c.lggr.Warnf("RMN node returned an unexpected type of response: %x", parsedResp.Response)
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+
+				rmnObservationResponses = append(rmnObservationResponses, rmnSignedObservationWithMeta{
+					RMNNodeID:         resp.RMNNodeID,
+					SignedObservation: signedObs,
+				})
+				for _, lu := range signedObs.Observation.FixedDestLaneUpdates {
+					if _, exists := merkleRootsCount[lu.LaneSource.SourceChainSelector]; !exists {
+						merkleRootsCount[lu.LaneSource.SourceChainSelector] = make(map[string]int)
+					}
+					merkleRootsCount[lu.LaneSource.SourceChainSelector][cciptypes.Bytes(lu.Root).String()]++
```
makramkd, line 270: Ah right the protobufs don't have an explicit Bytes32 type... wonder if we should check that len(lu.Root) == 32?

makramkd: Not super future proof if we ever get > 32 byte hashes, but should be fine for now

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+			// We got all the responses we were waiting for.
+			if finishedRequestIDs.Equal(requestIDs) {
+				c.lggr.Infof("all observation requests were finished")
+				return rmnObservationResponses, nil
+			}
+
+			// We got sufficient number of observation responses with matching roots for every source chain.
+			allChainsHaveEnoughResponses := true
+			for chainRequest := range lursPerChain {
+				var merkleRoot cciptypes.Bytes
```
makramkd, line 283:
```suggestion
				var merkleRoot cciptypes.Bytes32
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+				for merkleRootStr, count := range merkleRootsCount[chainRequest] {
+					if count >= c.minObservers {
+						merkleRoot, err = cciptypes.NewBytesFromString(merkleRootStr)
```
makramkd, line 286:
```suggestion
						merkleRoot, err = cciptypes.NewBytes32FromString(merkleRootStr)
```

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+	lursPerChain map[uint64]updateRequestWithMeta,
+	requestedNodes map[uint64]mapset.Set[uint32],
+) ([]rmnSignedObservationWithMeta, error) {
+	c.lggr.Infow("Waiting for observation requests", "requestIDs", requestIDs.String())
+
+	respChan := c.rawRmnClient.Recv()
+
+	finishedRequestIDs := mapset.NewSet[uint64]()
+	rmnObservationResponses := make([]rmnSignedObservationWithMeta, 0)
+	merkleRootsCount := make(map[uint64]map[string]int) // sourceChain -> merkleRoot -> count
```
makramkd, line 235: nit: looks like we do merkle root bytes => string => bytes, maybe just keep it as `Bytes32` in this map?

commit/merkleroot/rmn/client.go
```diff
@@ -19,13 +41,595 @@ type Client interface {
...
+			}
+			newRequestIDs := c.sendObservationRequests(destChain, requestsPerNode)
+			requestIDs = requestIDs.Union(newRequestIDs)
+		case <-ctx.Done():
+			return nil, ErrTimeout
+		}
+	}
+}
+
+func (c *client) validateSignedObservationResponse(
```
makramkd, line 327: I believe this func can be free (i.e not part of `c *client`)

makramkd (approved):
LGTM overall, most of these are nits

commit/merkleroot/rmn/client.go
dimkouv: Sounds good for unit testing of this functions, but for the e2e tests it's not giving any benefit I believe.
I will apply this pattern while unit testing later.

commit/merkleroot/rmn/client.go
dimkouv: Reason I didn't use `cciptypes.Bytes32` was to be aligned with proto generated types, so I don't have to convert each time I want to use it.

PR #109, opened 2024-09-06
Reuse MinObservation helper in commit.

winder:
Remove duplicate observation counting code.

### DIFF
--- a/commit/merkleroot/outcome_test.go
+++ b/commit/merkleroot/outcome_test.go
@@ -3,10 +3,8 @@ package merkleroot
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 )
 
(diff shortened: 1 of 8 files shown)


shared/consensus.go
```diff
@@ -0,0 +1,45 @@
...
+// GetConsensusMap takes a mapping from chains to a list of items,
+// return a mapping from chains to a single consensus item.
+// The consensus item for a given chain is the item with the
+// most observations that was observed at least fChain times.
+func GetConsensusMap[T any](
+	lggr logger.Logger,
+	objectName string,
+	itemsByChain map[cciptypes.ChainSelector][]T,
+	minObs map[cciptypes.ChainSelector]int,
+) map[cciptypes.ChainSelector]T {
```
makramkd, line 19: A test for this would be nice

commit/merkleroot/outcome_test.go
```diff
@@ -36,78 +34,3 @@ func Test_buildReport(t *testing.T) {
 		}
 	})
 }
-
-func Test_fChainConsensus(t *testing.T) {
```
makramkd, line 40: Maybe we can copy/paste this test to the shared folder?

execute/plugin_functions.go
```diff
@@ -227,9 +227,9 @@ func mergeMessageObservations(
 	aos []decodedAttributedObservation, fChain map[cciptypes.ChainSelector]int,
 ) (exectypes.MessageObservations, error) {
 	// Create a validator for each chain
-	validators := make(map[cciptypes.ChainSelector]validation.MinObservationFilter[cciptypes.Message])
```
makramkd, line 230: If this is moved to `filter`, can it be axed from the `validation` package in exec?

winder: Thanks, I missed that. Yes, that was the intention.

makramkd: Ah you're right, I don't know why I expected the files to show up as deleted, they'd be just moved.

makramkd (approved):

commit/merkleroot/outcome_test.go
winder: done.

PR #111, opened 2024-09-09
wip: contract discovery via observations

winder:


pkg/consts/errors.go
```diff
@@ -0,0 +1,8 @@
+package consts
+
+import "errors"
+
+var (
+	ErrContractReaderNotFound = errors.New("contract reader not found")
+	ErrContractWriterNotFound = errors.New("contract writer not found")
+)
```
winder, line 8: These needed to move so that I could use them outside of the internal package.

shared/discovery/processor.go
```diff
@@ -0,0 +1,74 @@
...
+
+	onrampAddrs := make(map[ccipocr3.ChainSelector][][]byte)
+	for _, ao := range aos {
+		for chain, addr := range ao.Observation.OnRamp {
+			onrampAddrs[chain] = append(onrampAddrs[chain], addr)
+		}
+	}
+
+	onrampConsensus := shared.GetConsensusMap(cdp.lggr, "onramp", onrampAddrs, fChain)
+	cdp.reader.Sync(context.Background(), onrampConsensus)
```
makramkd, line 71: I know this is still WIP but since outcome is pure any network calls must be avoided (Sync being one of them).

We can still update the ccip reader in `Observation` though, by checking the previous outcome.

internal/mocks/inmem/ccipreader_inmem.go
```diff
@@ -122,14 +122,20 @@ func (r InMemoryCCIPReader) GasPrices(
 	panic("implement me")
 }
```
dimkouv, line 123: And then in the plugin for consensus we want the `(name, chain, address)` pairs observed by at least `F` followers?

shared/discovery/processor.go
```diff
@@ -0,0 +1,96 @@
...
+
+	onrampAddrs := make(map[ccipocr3.ChainSelector][][]byte)
+	for _, ao := range aos {
+		for chain, addr := range ao.Observation.OnRamp {
+			onrampAddrs[chain] = append(onrampAddrs[chain], addr)
+		}
+	}
+
+	contracts := make(map[string]map[ccipocr3.ChainSelector][]byte)
+	contracts[consts.ContractNameOnRamp] = shared.GetConsensusMap(cdp.lggr, "onramp", onrampAddrs, fChain)
```
winder, line 90: @dimkouv fChain consensus is computed with `shared.GetConsensusMap` here. I'll initialize this discovery processor in both plugins  and call observe/outcome.

PR #114, opened 2024-09-10
Commit plugin - RMN signature verifications

dimkouv:
### Most important steps of the process:

1. OCR3 Leader verifies the received RMN signed observations using `ed25519` with the rmn node offchain public key.
2. OCR3 Leader verifies the received signed reports using the `RMNCrypto` package, the underlying implementation uses `secpk256k1` and abi encoding similar to `RMNRemote.sol` contract. (implementation will be added in chainlink repo)`
3. If the signature verification fails the response of the RMN node is simply dropped, and the rmn node id is logged.
4. Each follower then receives the signatures through the OCR3 query.
5. Each follower validates the signatures similar to the way leader did in step 2 in the observation phase.
6. If the signature verification fails the Observation stops with an error.
7. If signature verification succeeds the signatures are added from the Query (no consensus required) in the Outcome phase while building the outcome and eventually they get in the report.


commit/merkleroot/observation.go
```diff
@@ -33,13 +36,79 @@ func (w *Processor) Observation(
...
+	}
+
+	laneUpdates, err := rmn.NewLaneUpdatesFromPB(q.RMNSignatures.LaneUpdates)
+	if err != nil {
+		return fmt.Errorf("failed to convert lane updates from protobuf: %w", err)
+	}
+
+	rmnReport := cciptypes.RMNReport{
+		ReportVersion:               w.rmnConfig.Home.RmnReportVersion,
+		DestChainID:                 cciptypes.NewBigIntFromInt64(int64(ch.EvmChainID)),
```
makramkd, line 98: I think this is going to problematic and not future proof, e.g solana does not have a chain ID.

makramkd: cc @gtklocker

commit/merkleroot/outcome.go
```diff
@@ -125,12 +127,31 @@ func buildReport(
...
+			sigs = parsedSigs
+		}
+	}
+	sort.Slice(sigs, func(i, j int) bool {
+		riStr := sigs[i].R.String()
+		rjStr := sigs[j].R.String()
+		if riStr != rjStr {
+			return riStr < rjStr
+		}
+		return sigs[i].S.String() < sigs[j].S.String()
```
makramkd, line 147: Do we want to do a lexicographical sort? I thought the sort had to be on the recovered pubkey / address.

makramkd: c.f https://github.com/smartcontractkit/chainlink/blob/a2514d725f66e2c4a66d9bd0dd211a617fa32367/contracts/src/v0.8/ccip/rmn/RMNRemote.sol#L101

commit/merkleroot/rmn/client.go
```diff
@@ -541,12 +541,56 @@ func (c *client) listenForRmnReportSignatures(
...
+				RmnHomeContractConfigDigest: c.rmnCfg.Home.ConfigDigest,
+				LaneUpdates:                 laneUpdates,
+			}
+
+			sig, err := NewECDSASigFromPB(reportSig.Signature)
+			if err != nil {
+				return nil, fmt.Errorf("failed to convert signature from protobuf: %w", err)
+			}
+
+			err = c.rmnCrypto.VerifyReportSignatures(
```
makramkd, line 579: Is there a reason we verify report signatures twice? This code runs in query (I think) and we seem to also do it in observation

commit/merkleroot/rmn/client_test.go
```diff
@@ -48,21 +50,41 @@ func TestClient_ComputeReportSignatures(t *testing.T) {
 		resChan := make(chan RawRmnResponse, 200)
 		rawRmnClient := newMockRawRmnClient(resChan)
 
+		const numNodes = 4
+		rmnNodes := make([]RMNNodeInfo, numNodes)
+		for i := 0; i < numNodes; i++ {
+			publicKey, _, err := ed25519.GenerateKey(strings.NewReader(strings.Repeat("deterministic", 1000)))
```
makramkd, line 56: Haha, does it need to be repeated 1,000 times?

makramkd: That's definitely an interesting detail

commit/merkleroot/rmn/config.go
```diff
@@ -0,0 +1,37 @@
...
+// Config contains the RMN configuration required by the plugin and the RMN client in a single struct.
+type Config struct {
+	Home   RMNHomeConfig
+	Remote RMNRemoteConfig
+}
+
+type RMNHomeConfig struct {
+	RmnNodes         []RMNNodeInfo
+	ConfigDigest     cciptypes.Bytes32
+	RmnReportVersion string // e.g. "RMN_V1_6_ANY2EVM_REPORT"
```
makramkd, line 19: Is this actually fetchable from the contract?

commit/merkleroot/rmn/crypto.go
```diff
@@ -0,0 +1,51 @@
...
+	verifier ED25519Verifier,
+) error {
+	observationBytes, err := proto.Marshal(signedObs.GetObservation())
+	if err != nil {
+		return fmt.Errorf("failed to marshal observation: %w", err)
+	}
+
+	observationBytesSha256 := sha256.Sum256(observationBytes)
+	msg := append([]byte(rmnNode.SignObservationPrefix), observationBytesSha256[:]...)
+	msgSha256 := sha256.Sum256(msg)
```
makramkd, line 43: Is this preimage defined somewhere in the design doc?

commit/merkleroot/rmn/crypto_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+var (
+	empty32ByteArr [32]byte
+	empty20ByteArr [20]byte
+)
+
+func Test_verifyObservationSignature(t *testing.T) {
+	offchainPublicKey := "0x137d557b609823f3f8f841265d5797e0fe97892b496764af326fd62a82d881aa"
+
+	expSig := "edc308a23ce4aabf7518ad38778447e407b89b13011f7abc10e7641a51728e9ec6dd49a014ca238d8ba5d91fe9aff115d" +
```
makramkd, line 22: Is this a shared test case w/ RMN? If so lets add a comment indicating that.

commit/merkleroot/rmn/crypto_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// After we update one byte in the signature, the signature verification should fail
+	signedObs.Signature[0] = signedObs.Signature[0] + 1
+	err = verifyObservationSignature(rmnNode, signedObs, NewED25519Verifier())
+	assert.Error(t, err)
+}
+
+func Test_verifyObservationSignature2(t *testing.T) {
+	offchainPublicKey := "b1f20030f19ace75552a59125500d58d2d43305af537cb680da126e9a01d4100"
+	expSig := "e6e7eaab2f66f8cabbf98ed301827992d096c795d036322146671b5acb18c54f8e7945f0847a0d5da5218a2cb4dd0c471f" +
```
makramkd, line 67: Ditto here re: shared test case.

commit/merkleroot/rmn/crypto_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn/rmnpb"
+)
+
+var (
+	empty32ByteArr [32]byte
+	empty20ByteArr [20]byte
+)
+
+func Test_verifyObservationSignature(t *testing.T) {
```
makramkd, line 19: I think you can re-organize these test cases a bit better, maybe put the required values in an `args` struct and have two different test cases? The actual code that is running seems to be the same for both.

makramkd (commented):
LGTM overall, some questions

commit/merkleroot/outcome.go
dimkouv: nice catch

commit/merkleroot/rmn/client.go
dimkouv: Because we might have a malicious leader that sends wrong sigs.
cc @gtklocker correct me if i'm wrong

commit/merkleroot/rmn/client_test.go
dimkouv: Should be way less, I'll find and use the minimum

commit/merkleroot/rmn/config.go
dimkouv: not sure yet, haven't looked much into configs, I just assume I have it.. somehow.
cc @0xnogo

commit/merkleroot/rmn/crypto.go
dimkouv: I don't think so. cc @gtklocker

commit/merkleroot/observation.go
```diff
@@ -33,13 +36,79 @@ func (w *Processor) Observation(
...
+	}
+
+	sigs, err := rmn.NewECDSASigsFromPB(q.RMNSignatures.Signatures)
+	if err != nil {
+		return fmt.Errorf("failed to convert signatures from protobuf: %w", err)
+	}
+
+	signerAddresses := make([]cciptypes.Bytes, 0, len(sigs))
+	for _, rmnNode := range w.rmnConfig.Home.RmnNodes {
+		signerAddresses = append(signerAddresses, rmnNode.SignReportsAddress)
```
0xnogo, line 88: Does the sigs ordering matter here?

commit/merkleroot/rmn/config.go
0xnogo: I'm not sure yet. The only place where it could be in the RMNHome is in the `bytes offchainConfig`. But we need to clarify what's going to go there.

commit/merkleroot/rmn/config.go
```diff
@@ -0,0 +1,37 @@
...
+	"crypto/ed25519"
+
+	mapset "github.com/deckarep/golang-set/v2"
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+// Config contains the RMN configuration required by the plugin and the RMN client in a single struct.
+type Config struct {
+	Home   RMNHomeConfig
+	Remote RMNRemoteConfig
```
0xnogo, line 13: A split with a struct for Home config and another one for Remote config might help the way we are fetching the data. But `RMNNodeInfo` is actually having data from both (like SignReportsAddress). So we might have to have 2 type of RmnNodeInfo struct: one for Home and one for Remote.

commit/merkleroot/rmn/crypto.go
dimkouv: it's in the .proto

commit/merkleroot/outcome.go
dimkouv: fixed and added a test

commit/merkleroot/rmn/client_test.go
dimkouv: Fixed, added a 32char string

commit/merkleroot/rmn/config.go
dimkouv: It's a constant in the contract

dimkouv: https://github.com/smartcontractkit/chainlink/blob/c83c68735bdee6bbd8510733b7415797cd08ecbd/contracts/src/v0.8/ccip/rmn/RMNRemote.sol#L9

commit/merkleroot/rmn/crypto_test.go
dimkouv: added a comment

commit/merkleroot/rmn/crypto_test.go
dimkouv: added a comment

commit/merkleroot/rmn/crypto_test.go
dimkouv: done

commit/merkleroot/observation.go
dimkouv: No, at this point we expect to have the signatures ordered, by signer address.

The implementation of:
```
VerifyReportSignatures(
		ctx context.Context,
		sigs []RMNECDSASignature,
		report RMNReport,
		signerAddresses []Bytes,
	) error
```

Can cross-check if they are actually ordered.

commit/merkleroot/rmn/config.go
dimkouv: ah interesting, i believe we can merge, like this. and apply relevant config changes in a follow-up

commit/merkleroot/outcome.go
```diff
@@ -125,12 +127,23 @@ func buildReport(
...
 	sort.Slice(roots, func(i, j int) bool { return roots[i].ChainSel < roots[j].ChainSel })
 
-	// TODO: use q.RMNSignatures in the generated outcome and eventually report. - Blocked by onchain work.
+	sigs := make([]cciptypes.RMNECDSASignature, 0)
+	if q.RMNSignatures == nil {
+		lggr.Error("RMNSignatures are nil")
+	} else {
+		parsedSigs, err := rmn.NewECDSASigsFromPB(q.RMNSignatures.Signatures)
+		if err != nil {
+			lggr.Errorw("Failed to parse RMN signatures", "error", err)
```
makramkd, line 136: q: if all nodes fail to parse sigs, outcome will have an empty sigs array, but what's going to happen in that case? No commit report posted onchain? Or will the commit report just fail to verify onchain because RMN sigs are empty?

commit/merkleroot/observation.go
makramkd: Followed up on this offline but we're thinking of moving RMN report creation behind an interface. Not blocking this PR.

makramkd (approved):

PR #115, opened 2024-09-10
Token price processor outcome

asoliman92:
Implement Outcome phase of token price processor


commit/tokenprice/observation.go
```diff
@@ -0,0 +1,110 @@
+package tokenprice
```
asoliman92, line 1: Moving observation logic from processor.go, no need to review.

commit/tokenprice/outcome.go
```diff
@@ -0,0 +1,151 @@
...
+	obs ConsensusObservation,
+) []cciptypes.TokenPrice {
+	var tokenPrices []cciptypes.TokenPrice
+	cfg := p.cfg.OffchainConfig
+	tokenInfo := cfg.TokenInfo
+
+	for token, feedPrice := range obs.FeedTokenPrices {
+		lastUpdate, exists := obs.FeeQuoterTokenUpdates[token]
+		if !exists {
+			p.lggr.Warnf("could not find fee quoter update for token %s", token)
```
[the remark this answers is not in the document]
asoliman92: Yes 👌🏼

commit/tokenprice/outcome.go
```diff
@@ -0,0 +1,151 @@
...
+		return a < b
+	})
+
+	twoFPlus1 := 2*fDestChain + 1
+
+	feedPricesConsensus := shared.GetConsensusMapAggregator(
+		p.lggr,
+		"FeedTokenPrices",
+		aggObs.FeedTokenPrices,
+		twoFPlus1,
```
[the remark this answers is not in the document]
asoliman92: Good catch!

commit/tokenprice/observation.go
```diff
@@ -0,0 +1,110 @@
...
+	fDestChain, ok := fChain[p.cfg.DestChain]
+	if !ok {
+		return 0, fmt.Errorf("fChain does not have dest chain")
+	}
+
+	return fDestChain, nil
+}
+
+func (p *processor) ObserveFeedTokenPrices(ctx context.Context) []cciptypes.TokenPrice {
+	supportedChains, err := p.chainSupport.SupportedChains(p.oracleID)
```
mateusz-sekara, line 34: Food for thought, I see this pattern a lot across other processors, shouldn't we extract that then, let's say to something like this and call that from the upper layer (probably plugin) instead of doing these checks in every underlying processor? 

```go
type Processor interface {
   IsChainSupported() (bool, error)
}
```

```go
func (p *plugin) Observer() {
    if !p.TokenProcessor.IsChainSupported() {
       return sth
    }
    if !p.MerkleRootProcessor.IsChainSupported() {
       return sth
    }

}```

File: commit/tokenprice/observation.go [hunk 5]
```diff
@@ -0,0 +1,110 @@
...
+
+	if p.tokenPriceReader == nil {
+		p.lggr.Errorw("no token price reader available")
+		return []cciptypes.TokenPrice{}
+	}
+
+	tokensToQuery := maps.Keys(p.cfg.OffchainConfig.TokenInfo)
+	//sort tokens to query to ensure deterministic order
+	sort.Slice(tokensToQuery, func(i, j int) bool { return tokensToQuery[i] < tokensToQuery[j] })
+	p.lggr.Infow("observing feed token prices")
```
mateusz-sekara, line 54: Maybe a bit premature, but I guess logging tokensToQuery in here might be useful, probably there won't be many tokens here so impact of logging that shouldn't be noticeable

commit/tokenprice/observation.go
```diff
@@ -0,0 +1,110 @@
...
+	tokenPrices, err := p.tokenPriceReader.GetTokenFeedPricesUSD(ctx, tokensToQuery)
+	if err != nil {
+		p.lggr.Errorw("call to GetTokenFeedPricesUSD failed", "err", err)
+		return []cciptypes.TokenPrice{}
+	}
+
+	// If we couldn't fetch all prices log and return only the ones we could fetch
+	if len(tokenPrices) != len(tokensToQuery) {
+		p.lggr.Errorw("token prices length mismatch", "got", len(tokenPrices), "want", len(tokensToQuery))
```
mateusz-sekara, line 63: From past experience, seeing the diff in the log would be helpful. Otherwise, this log doesn't bring any debugging value, the only information you have is that lengths are different but it's still very hard to investigate which token/tokens cause problems

commit/tokenprice/observation.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		p.lggr.Warnw("call to SupportsDestChain failed", "err", err)
+		return map[types.Account]shared.TimestampedBig{}
+	}
+	if !supportsDestChain {
+		p.lggr.Debugw("oracle does not support price registry observation", "oracleID", p.oracleID)
+		return map[types.Account]shared.TimestampedBig{}
+	}
+
+	if p.tokenPriceReader == nil {
```
mateusz-sekara, line 85: Seems like this call should be first, because it's cheaper than checking supportedDestChains and could immediately terminate the function if `tokenPriceReader` is nil, right?

mateusz-sekara: How about thread-safety here? I guess it could be changed by other thread? I mean another thread can try to set this field in parallel

commit/tokenprice/outcome.go
```diff
@@ -0,0 +1,151 @@
...
+	if len(aggObs.FDestChain) < p.bigF {
+		return ConsensusObservation{},
+			fmt.Errorf("not enough observations for fDestChain")
+	}
+
+	fDestChain := shared.Median(aggObs.FDestChain, func(a, b int) bool {
+		return a < b
+	})
+
+	twoFPlus1 := 2*fDestChain + 1
```
mateusz-sekara, line 30: nit, but this is also something that I see a lot in different places, do you think it's worth exporting to enum or function? 

```go
twoFPlus1 = function (int) int
```

Don't we have a similar concept somewhere in the libocr? Could that be reused or sth?

mateusz-sekara (commented):
Do we have some kind of integration / e2e tests that we can enrich with cases covering token prices?

commit/tokenprice/observation.go
asoliman92: good call. no need to worry about thread safety. the reader is initialized and injected into the plugin beforehand.

commit/tokenprice/outcome.go
asoliman92: I like the idea, let me try it out. 👍

commit/tokenprice/observation.go
asoliman92: Going to postpone this to another PR for now.

commit/tokenprice/observation.go
asoliman92: Done. Thanks!

commit/tokenprice/observation.go
```diff
@@ -0,0 +1,102 @@
...
+		return []cciptypes.TokenPrice{}
+	}
+
+	supportedChains, err := p.chainSupport.SupportedChains(p.oracleID)
+	if err != nil {
+		p.lggr.Warnw("call to SupportedChains failed", "err", err)
+		return []cciptypes.TokenPrice{}
+	}
+
+	if !supportedChains.Contains(cciptypes.ChainSelector(p.cfg.OffchainConfig.TokenPriceChainSelector)) {
```
winder, line 37: You can change the `OffchainConfig` type to `cciptypes.ChainSelector`

commit/tokenprice/outcome.go
```diff
@@ -0,0 +1,169 @@
...
+	if len(aggObs.FChain) < p.bigF {
+		return ConsensusObservation{},
+			fmt.Errorf("not enough observations for FChain")
+	}
+
+	fMin := make(map[cciptypes.ChainSelector]int)
+	for chain := range aggObs.FChain {
+		fMin[chain] = p.bigF
+	}
+	fChains := shared.GetConsensusMap(p.lggr, "fChain", aggObs.FChain, fMin)
```
makramkd, line 30: Lets add comments for the f usage on all of these calls; consensus on the fChain map uses the role DON F value because all nodes can observe the home chain.

commit/tokenprice/outcome.go
```diff
@@ -0,0 +1,169 @@
...
+// getConsensusObservation Combine the list of observations into a single consensus observation
+func (p *processor) getConsensusObservation(
+	aos []shared.AttributedObservation[Observation],
+) (ConsensusObservation, error) {
+	aggObs := aggregateObservations(aos)
+
+	if len(aggObs.FChain) < p.bigF {
+		return ConsensusObservation{},
+			fmt.Errorf("not enough observations for FChain")
+	}
```
winder, line 24: Maybe this could be part of `GetConsensusMap`?

makramkd (approved):
LGTM overall

commit/tokenprice/observation_test.go
```diff
@@ -86,15 +80,15 @@ func Test_Observation(t *testing.T) {
...
 					homeChain:        homeChain,
 					cfg:              defaultCfg,
+					bigF:             f,
 				}
 			},
 			expObs: Observation{
 				FeedTokenPrices:       feedTokenPrices,
 				FeeQuoterTokenUpdates: feeQuoterTokenUpdates,
-				FDestChain:            fDestChain,
+				FChain:                fChains,
```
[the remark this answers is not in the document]
asoliman92: Seems like it. We need the feedChain F as well in this case. For simplicity I duplicated the logic. Once the feature is complete and have integration tests we can revisit.

commit/tokenprice/outcome.go
```diff
@@ -0,0 +1,169 @@
...
+			tokenPrices = append(tokenPrices, cciptypes.TokenPrice{
+				TokenID: token,
+				Price:   cciptypes.NewBigInt(feedPrice.Price.Int),
+			})
+		} else if shared.Deviates(feedPrice.Price.Int, lastUpdate.Value.Int, ti.DeviationPPB.Int64()) {
+			tokenPrices = append(tokenPrices, cciptypes.TokenPrice{
+				TokenID: token,
+				Price:   cciptypes.NewBigInt(feedPrice.Price.Int),
+			})
+		}
```
winder, line 123: nit: make it clear that both cases do the same thing
```suggestion
		shouldUpdate := obs.Timestamp.After(nextUpdateTime) || shared.Deviates(feedPrice.Price.Int, lastUpdate.Value.Int, ti.DeviationPPB.Int64())
		if shouldUpdate {
			tokenPrices = append(tokenPrices, cciptypes.TokenPrice{
				TokenID: token,
				Price:   cciptypes.NewBigInt(feedPrice.Price.Int),
			})
		}
```

commit/tokenprice/types.go
```diff
@@ -21,16 +21,32 @@ type Outcome struct {
...
 	Timestamp             time.Time                               `json:"timestamp"`
 }
 
+// AggregateObservation is the aggregation of a list of observations
+type AggregateObservation struct {
+	FeedTokenPrices       map[types.Account][]cciptypes.TokenPrice
+	FeeQuoterTokenUpdates map[types.Account][]shared.TimestampedBig
+	FChain                map[cciptypes.ChainSelector][]int `json:"fChain"`
+	Timestamps            []time.Time
+}
```
winder, line 34: I know it screws up all of the logic but....
```suggestion
type AggregateObservation []ConsensusObservation
```

commit/tokenprice/outcome.go
asoliman92: Done

shared/calc.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+	diff := big.NewInt(0).Sub(x1, x2) // diff = x1-x2
+	diff.Mul(diff, big.NewInt(1e9))   // diff = diff * 1e9
+	// dividing by the smaller value gives consistent ppb regardless of input order, and supports >100% deviation.
+	if x1.Cmp(x2) > 0 {
+		diff.Div(diff, x2)
+	} else {
+		diff.Div(diff, x1)
+	}
+	return diff.CmpAbs(big.NewInt(ppb)) > 0 // abs(diff) > ppb
```
winder, line 23: Still not sure what it does, but maybe this is slightly clearer? Still not sure what it means by >100% deviation.
```suggestion
	// ensure x1 > x2
	if x1.Cmp(x2) < 0 {
		x1, x2 = x2, x1
	}
	diff := big.NewInt(0).Sub(x1, x2) // diff = x1-x2
	diff.Mul(diff, big.NewInt(1e9))   // diff = diff * 1e9
	// dividing by the smaller value gives consistent ppb regardless of input order, and supports >100% deviation.
	diff.Div(diff, x2)
	return diff.Cmp(big.NewInt(ppb)) > 0 // diff > ppb
```

PR #117, opened 2024-09-12
CCIP-2219 USDC Support - Stubbing interfaces and new flow 

mateusz-sekara:
Going top-down to fit USDC/CCTP support in ocr3

Please focus on the interfaces and connecting components together, implementation details are not that critical at this stage. 
* `tokenData` details are now part of the Observation step - we don't want to call any 3rd party API during the `Outcome` phase
* TokenDataReader is replaced with TokenDataObserver, it's used during Observation. Goal here is to support multiple TokenDataObservers (per token) with one go
* `ExecuteOffchainConfig` is enriched with the `TokenDataObservers` config under `tokenDataObservers`, please take a look at tests to understand the new schema 

High-level config would be the following (skipping irrelevant details from the `ExecuteOffchainConfig`):
```json
{
  "tokenDataObservers": [
    {
      "type": "usdc-cctp",
      "version": "1.0",
      "tokens": {
        "1": {
          "sourceTokenAddress": "0xabc",
          "sourceMessageTransmitterAddress": "0xefg"
        }
      },
      "attestationAPI": "http://localhost:8080",
      "attestationAPITimeout": "1s",
      "attestationAPIIntervalMilliseconds": "500ms"
    },
    {
      "type": "my-fancy-token",
      "version": "1.0",
      "tokens": {
        "2": {
          "sourceTokenAddress": "0xabc"
        }
      },
      "myFancyTokenAPI": "http://localhost:8080"
    }
  ]
}
```

Around 250 bytes per observer entry so it should be around 4k gas per processor (assuming 16 gas per byte and a single token per processor), it gives us roughly 4k entries to hit 12M gas (+ of course overhead of the existing config)

Changes in Chainlink repo https://github.com/smartcontractkit/chainlink/pull/14418

### DIFF
--- a/execute/exectypes/commit_data.go
+++ b/execute/exectypes/commit_data.go
@@ -26,5 +26,5 @@ type CommitData struct {
 	ExecutedMessages []cciptypes.SeqNum `json:"executedMessages"`
 
 	// TokenData for each message.
-	MessageTokensData []MessageTokenData `json:"messageTokenData"`
+	MessageTokenData []MessageTokenData `json:"messageTokenData"`
 }
--- a/execute/exectypes/commit_data.go
+++ b/execute/exectypes/commit_data.go
@@ -25,8 +25,6 @@ type CommitData struct {
 	// ExecutedMessages are the messages in this report that have already been executed.
 	ExecutedMessages []cciptypes.SeqNum `json:"executedMessages"`
 
-	// The following values are cached for validation algorithms, serialization is not required for consensus.
-
 	// TokenData for each message.
-	TokenData [][][]byte `json:"-"`
+	MessageTokensData []MessageTokensData `json:"messageTokenData"`
 }
--- a/execute/exectypes/commit_data.go
+++ b/execute/exectypes/commit_data.go
@@ -26,5 +26,5 @@ type CommitData struct {
 	ExecutedMessages []cciptypes.SeqNum `json:"executedMessages"`
 
 	// TokenData for each message.
-	MessageTokensData []MessageTokensData `json:"messageTokenData"`
+	MessageTokensData []MessageTokenData `json:"messageTokenData"`
 }
--- a/execute/exectypes/observation.go
+++ b/execute/exectypes/observation.go
@@ -18,6 +18,16 @@ type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]ccipty
 // must be encoding according to the destination chain requirements with typeconv.AddressBytesToString.
 type NonceObservations map[cciptypes.ChainSelector]map[string]uint64
 
+// TokenDataObservations contain token data for messages organized by source chain selector and sequence number.
+// There could be multiple tokens per a single message, so MessageTokensData is a slice of TokenData.
+// TokenDataObservations are populated during the Observation phase and depends on previously fetched
+// MessageObservations details and the `tokenDataProcessors` configured in the ExecuteOffchainConfig.
+// Content of the MessageTokensData is determined by the TokenDataProcessor implementations.
+//   - if Message doesn't have any tokens, TokenData slice will be empty.
+//   - if Message has tokens, but these tokens don't require any special treatment,
+//     TokenData slice will contain empty TokenData objects.
+//   - if Message has tokens and these tokens require additional processing defined in ExecuteOffchainConfig,
+//     specific TokenDataProcessor will be used to populate the TokenData slice.
 type TokenDataObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]MessageTokensData
 
 type MessageTokensData struct {
@@ -50,10 +60,20 @@ func (mtd MessageTokensData) ToByteSlice() [][]byte {
 	return out
 }
 
+// TokenData is the token data for a single token in a message.
+// It contains the token data and a flag indicating if the data is ready.
 type TokenData struct {
 	Ready bool   `json:"ready"`
-	Error error  `json:"error"`
 	Data  []byte `json:"data"`
+	Error error
+}
+
+func NewEmptyTokenData() TokenData {
+	return TokenData{
+		Ready: false,
+		Error: nil,
+		Data:  nil,
+	}
 }
 
 func (td TokenData) IsReady() bool {
@@ -84,7 +104,12 @@ type Observation struct {
 }
 
 // NewObservation constructs a Observation object.
-func NewObservation(commitReports CommitObservations, messages MessageObservations, tokenData TokenDataObservations, nonces NonceObservations) Observation {
+func NewObservation(
+	commitReports CommitObservations,
+	messages MessageObservations,
+	tokenData TokenDataObservations,
+	nonces NonceObservations,
+) Observation {
 	return Observation{
 		CommitReports: commitReports,
 		Messages:      messages,
--- a/execute/exectypes/observation.go
+++ b/execute/exectypes/observation.go
@@ -20,14 +20,14 @@ type NonceObservations map[cciptypes.ChainSelector]map[string]uint64
 
 // TokenDataObservations contain token data for messages organized by source chain selector and sequence number.
 // There could be multiple tokens per a single message, so MessageTokenData is a slice of TokenData.
-// TokenDataObservations are populated during the Observation phase and depends on previously fetched
-// MessageObservations details and the `tokenDataProcessors` configured in the ExecuteOffchainConfig.
-// Content of the MessageTokenData is determined by the TokenDataProcessor implementations.
+// TokenDataObservations are populated during the Observation phase and depend on previously fetched
+// MessageObservations details and the `tokenDataObservers` configured in the ExecuteOffchainConfig.
+// Content of the MessageTokenData is determined by the tokendata.TokenDataObserver implementations.
 //   - if Message doesn't have any tokens, TokenData slice will be empty.
 //   - if Message has tokens, but these tokens don't require any special treatment,
 //     TokenData slice will contain empty TokenData objects.
 //   - if Message has tokens and these tokens require additional processing defined in ExecuteOffchainConfig,
-//     specific TokenDataProcessor will be used to populate the TokenData slice.
+//     specific tokendata.TokenDataObserver will be used to populate the TokenData slice.
 type TokenDataObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]MessageTokenData
 
 type MessageTokenData struct {
@@ -65,7 +65,9 @@ func (mtd MessageTokenData) ToByteSlice() [][]byte {
 type TokenData struct {
 	Ready bool   `json:"ready"`
 	Data  []byte `json:"data"`
-	Error error
+	// Error is used only for internal processing, we don't want nodes to gossip about the
+	// errors they see during processing
+	Error error `json:"-"`
 }
 
 func NewEmptyTokenData() TokenData {
--- a/execute/exectypes/observation.go
+++ b/execute/exectypes/observation.go
@@ -18,21 +18,47 @@ type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]ccipty
 // must be encoding according to the destination chain requirements with typeconv.AddressBytesToString.
 type NonceObservations map[cciptypes.ChainSelector]map[string]uint64
 
-type TokenDataObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]TokenData
+type TokenDataObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]MessageTokensData
 
-type TokenData struct {
-	TokenDataStatus TokenDataStatus
-	Error           error
-	TokenData       [][]byte
+type MessageTokensData struct {
+	TokenData []TokenData
 }
 
-type TokenDataStatus uint8
+func (mtd MessageTokensData) IsReady() bool {
+	for _, td := range mtd.TokenData {
+		if !td.IsReady() {
+			return false
+		}
+	}
+	return true
+}
 
-const (
-	TokenDataNotNeeded TokenDataStatus = iota + 1
-	TokenDataFetchError
-	TokenDataNotAvailable
-)
+func (mtd MessageTokensData) Error() error {
+	for _, td := range mtd.TokenData {
+		if td.Error != nil {
+			return td.Error
+		}
+	}
+	return nil
+}
+
+func (mtd MessageTokensData) ToByteSlice() [][]byte {
+	out := make([][]byte, len(mtd.TokenData))
+	for i, td := range mtd.TokenData {
+		out[i] = td.Data
+	}
+	return out
+}
+
+type TokenData struct {
+	Ready bool   `json:"ready"`
+	Error error  `json:"error"`
+	Data  []byte `json:"data"`
+}
+
+func (td TokenData) IsReady() bool {
+	return td.Ready
+}
 
 // Observation is the observation of the ExecutePlugin.
 // TODO: revisit observation types. The maps used here are more space efficient and easier to work
--- a/execute/exectypes/observation.go
+++ b/execute/exectypes/observation.go
@@ -18,6 +18,22 @@ type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]ccipty
 // must be encoding according to the destination chain requirements with typeconv.AddressBytesToString.
 type NonceObservations map[cciptypes.ChainSelector]map[string]uint64
 
+type TokenDataObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]TokenData
+
+type TokenData struct {
+	TokenDataStatus TokenDataStatus
+	Error           error
+	TokenData       [][]byte
+}
+
+type TokenDataStatus uint8
+
+const (
+	TokenDataNotNeeded TokenDataStatus = iota + 1
+	TokenDataFetchError
+	TokenDataNotAvailable
+)
+
 // Observation is the observation of the ExecutePlugin.
 // TODO: revisit observation types. The maps used here are more space efficient and easier to work
 // with but require more transformations compared to the on-chain representations.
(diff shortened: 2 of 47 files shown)


execute/exectypes/observation.go
```diff
@@ -18,6 +18,68 @@ type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]ccipty
 // must be encoding according to the destination chain requirements with typeconv.AddressBytesToString.
 type NonceObservations map[cciptypes.ChainSelector]map[string]uint64
 
+// TokenDataObservations contain token data for messages organized by source chain selector and sequence number.
+// There could be multiple tokens per a single message, so MessageTokensData is a slice of TokenData.
+// TokenDataObservations are populated during the Observation phase and depends on previously fetched
+// MessageObservations details and the `tokenDataProcessors` configured in the ExecuteOffchainConfig.
+// Content of the MessageTokensData is determined by the TokenDataProcessor implementations.
+//   - if Message doesn't have any tokens, TokenData slice will be empty.
```
winder, line 26: Suppose ChainSelector `chain` and SeqNum `seq` is a message that does not have token data. Is there no `MessageTokenData` object in the map? In that case `tokenObs[chain][seq].IsReady()` is false even though we're not waiting for data.

mateusz-sekara: `MessageTokenData` has to be always there, that's gonna be a requirement here. Therefore client doesn't have to handle nils or other weird edge cases. We move this responsibility to the processor to init everything properly, even messages that don't have tokens. In this case, it should be populated with an empty `MessageTokensData` which is always `IsReady=true` in that case 

```go
func (mtd MessageTokensData) IsReady() bool {
	for _, td := range mtd.TokenData {
		if !td.IsReady() {
			return false
		}
	}
	return true
}
```

mateusz-sekara: I think comment explains that, no? `TokenData slice will be empty` and for empty TokenData we return `IsReady=true`

execute/tokendata/processor.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
+)
+
+type TokenDataProcessor interface {
```
makramkd, line 11: naming nit: we've been using the term "processor" (in this codebase) to reference an object that implements the processor interface defined here: https://github.com/smartcontractkit/chainlink-ccip/blob/30c877439f7a797edc0aaa6909433bb929082e10/shared/plugin_processor.go#L39-L44. Maybe `TokenDataHandler`?

makramkd: Or perhaps `TokenDataObserver` and the method is called `ObserveTokenData`

mateusz-sekara: `TokenDataObserver` sounds good to me

makramkd (commented):
Overall looks good, just have a few suggestions and some comments on how versioning will be handled in the future will be useful. Also, since this breaks the build (due to a constructor signature change) lets have a PR in chainlink ready to do any fixes and make sure the integration test passes.

mateusz-sekara:
> Overall looks good, just have a few suggestions and some comments on how versioning will be handled in the future will be useful. Also, since this breaks the build (due to a constructor signature change) lets have a PR in chainlink ready to do any fixes and make sure the integration test passes.

@makramkd PR with fixes in the chainlink repo is linked in the description of that PR :P

execute/exectypes/observation.go
```diff
@@ -18,6 +18,70 @@ type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]ccipty
...
+//   - if Message has tokens, but these tokens don't require any special treatment,
+//     TokenData slice will contain empty TokenData objects.
+//   - if Message has tokens and these tokens require additional processing defined in ExecuteOffchainConfig,
+//     specific tokendata.TokenDataObserver will be used to populate the TokenData slice.
+type TokenDataObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]MessageTokenData
+
+type MessageTokenData struct {
+	TokenData []TokenData
+}
```
winder, line 35: Why is there a slice of token data per message?

winder: Is it a slice of 1 so that you can distinguish between not waiting for token data and waiting for token data?

execute/plugin.go
```diff
@@ -413,13 +423,13 @@ func (p *Plugin) Outcome(
 			context.Background(),
 			p.lggr,
 			p.msgHasher,
-			p.tokenDataReader,
 			p.reportCodec,
 			p.estimateProvider,
 			observation.Nonces,
```
winder, line 428: Maybe Nonces should be added to `selectReport` instead of being part of the this constructor.

execute/plugin_functions.go
```diff
@@ -305,6 +305,17 @@ func mergeCommitObservations(
 	return results, nil
 }
 
+func mergeTokenObservations(
```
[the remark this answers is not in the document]
winder: FYI: I have some very minor refactoring to replace `decodedAttributedObservation` with `shared.AttributedObservation[exectypes.Observation]`

https://github.com/smartcontractkit/chainlink-ccip/pull/111/files#diff-b6dd42deec4e0c93c7fb650e56db36ca55f0ea0e83dfacea1b419d9070cba1f3R224

pluginconfig/execute.go
```diff
@@ -78,6 +81,128 @@ func (e ExecuteOffchainConfig) Validate() error {
...
+	//    "customCirlceAPI": "http://cirle.com/gohere",
+	//    "yetAnotherAPI": "http://cirle.com/anotherone",
+	//    "customCircleAPITimeout": "1s",
+	//    "yetAnotherAPITimeout": "500ms"
+	//  }
+	//]
+	// Having version in that JSONisn't expensive, but it could reduce the risk of breaking the observers in the future.
+	Version string `json:"version"`
+
+	*USDCCCTPObserverConfig
```
winder, line 136: No need for a pointer here, it's already a slice

pluginconfig/execute.go
```diff
@@ -78,6 +81,128 @@ func (e ExecuteOffchainConfig) Validate() error {
...
+		}
+	default:
+		return errors.New("unknown token data observer type")
+	}
+	return nil
+}
+
+// Validate checks that the observer's config is semantically correct - fields are set correctly
+// depending on the config's type
+func (t TokenDataObserverConfig) Validate() error {
```
winder, line 154: Any reason not to combine this with `WellFormed`?

execute/exectypes/commit_data.go
```diff
@@ -25,8 +25,6 @@ type CommitData struct {
 	// ExecutedMessages are the messages in this report that have already been executed.
 	ExecutedMessages []cciptypes.SeqNum `json:"executedMessages"`
 
-	// The following values are cached for validation algorithms, serialization is not required for consensus.
-
 	// TokenData for each message.
-	TokenData [][][]byte `json:"-"`
+	MessageTokenData []MessageTokenData `json:"messageTokenData"`
```
dimkouv, line 29: Making sure that `type CommitData struct {` are not the data expected to be in a commit report.

Because e.g. this comment indicates the opposite.
```
// CommitObservations contain the commit plugin report data organized by the source chain selector.
type CommitObservations map[cciptypes.ChainSelector][]CommitData
```

pluginconfig/execute.go
```diff
@@ -78,6 +81,128 @@ func (e ExecuteOffchainConfig) Validate() error {
...
+const USDCCCTPHandlerType = "usdc-cctp"
+
+type USDCCCTPObserverConfig struct {
+	Tokens                 map[int]USDCCCTPTokenConfig `json:"tokens"`
+	AttestationAPI         string                      `json:"attestationAPI"`
+	AttestationAPITimeout  *commonconfig.Duration      `json:"attestationAPITimeout"`
+	AttestationAPIInterval *commonconfig.Duration      `json:"attestationAPIInterval"`
+}
+
+func (p USDCCCTPObserverConfig) Validate() error {
```
winder, line 175: Check these as well?
```suggestion
func (p USDCCCTPObserverConfig) Validate() error {
	if p.AttestationAPITimeout == nil {
		return errors.New("AttestationAPITimeout not set")
	}
	if p.AttestationAPIInterval == nil {
		return errors.New("AttestationAPIInterval not set")
	}
```

execute/exectypes/observation.go
```diff
@@ -32,17 +96,26 @@ type Observation struct {
...
+	// TokenData are determined during the second phase of execute.
+	// It contains the token data for the messages identified in the same stage as Messages
+	TokenData TokenDataObservations `json:"tokenDataObservations"`
+
 	// Nonces are determined during the third phase of execute.
 	// It contains the nonces of senders who are being considered for the final report.
 	Nonces NonceObservations `json:"nonces"`
 }
 
 // NewObservation constructs a Observation object.
```
dimkouv, line 108: typo `an Observation object`

execute/tokendata/processor.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
+)
+
+type TokenDataObserver interface {
```
dimkouv, line 11: Would be nice to have some doc in the interface and maybe the method.

execute/tokendata/processor.go
```diff
@@ -0,0 +1,64 @@
...
+// Goal of that is to support multiple token observers supporting different tokens (e.g. CCTP, MyFancyToken etc)
+type CompositeTokenDataObserver struct {
+	Observers []TokenDataObserver
+}
+
+func (t *CompositeTokenDataObserver) Observe(
+	ctx context.Context,
+	messages exectypes.MessageObservations,
+) (exectypes.TokenDataObservations, error) {
+	// Dummy implementation, don't focus on that
```
dimkouv, line 28: Probably you want to remove that comment before merging

execute/plugin.go
```diff
@@ -243,8 +244,12 @@ func (p *Plugin) Observation(
 			groupedCommits[report.SourceChain] = append(groupedCommits[report.SourceChain], report)
 		}
 
-		// TODO: Fire off messages for an attestation check service.
-		return exectypes.NewObservation(groupedCommits, messages, nil).Encode()
+		tkData, err1 := p.tokenDataObserver.Observe(ctx, messages)
+		if err1 != nil {
+			return types.Observation{}, fmt.Errorf("unable to process token data %w", err)
```
dimkouv, line 249: Logging `err` instead of `err1`

execute/plugin.go
```diff
@@ -243,8 +244,12 @@ func (p *Plugin) Observation(
 			groupedCommits[report.SourceChain] = append(groupedCommits[report.SourceChain], report)
 		}
 
-		// TODO: Fire off messages for an attestation check service.
-		return exectypes.NewObservation(groupedCommits, messages, nil).Encode()
+		tkData, err1 := p.tokenDataObserver.Observe(ctx, messages)
```
dimkouv, line 247: Is it going to be synchronous?

e.g. If Circle's API is slow, it affects our observation duration?

execute/plugin.go
```diff
@@ -400,8 +405,13 @@ func (p *Plugin) Outcome(
 				if msg, ok := observation.Messages[report.SourceChain][i]; ok {
 					report.Messages = append(report.Messages, msg)
 				}
+
```
dimkouv, line 408: Is the `i` correct here? (I think so)

```
for i, report := range commitReports {
			report.Messages = nil
			for i := report.SequenceNumberRange.Start(); i <= report.SequenceNumberRange.End(); i++ {
```

I believe second loop should use `j` to prevent using the `i` from the outer loop.

pluginconfig/execute.go
```diff
@@ -78,6 +81,128 @@ func (e ExecuteOffchainConfig) Validate() error {
...
+	return nil
+}
+
+// nolint:lll // CCTP link
+type USDCCCTPTokenConfig struct {
+	// SourceTokenAddress is the address of the USDC ECR-20 token on the source chain
+	SourceTokenAddress string `json:"sourceTokenAddress"`
+	// SourceMessageTransmitterAddr is the address of the CCTP MessageTransmitter address on the source chain
+	// https://github.com/circlefin/evm-cctp-contracts/blob/adb2a382b09ea574f4d18d8af5b6706e8ed9b8f2/src/MessageTransmitter.sol
+	SourceMessageTransmitterAddr string `json:"sourceMessageTransmitterAddress"`
```
winder, line 196: General comment not specific to this PR - we really need a proper address type to abstract the different ways we need to encode addresses.

pluginconfig/execute.go
mateusz-sekara: I wanted to follow the existing pattern for the configs. Validate is a semantical check - verifying fields values. However, you can think of WellFormed as a syntactical check called during deserialization. We would not need that in an ideal world, because the JSON unmarshaller would cover it.

execute/exectypes/observation.go
mateusz-sekara: lol, haven't touched that, but I can fix :P

execute/tokendata/processor.go
mateusz-sekara: I am intentionally, splitting that into smaller PRs. This one focuses on high-level interfaces and code organization. Followup one with implementation will cover these

execute/plugin.go
mateusz-sekara: It's gonna be implementation details, for the plugin perspective we behave like it's synchronous, plugin shouldn't know that. Under the hood we will figure out background processing etc

execute/tokendata/processor.go
mateusz-sekara: I will do as followup, I'm going top-down here, so I wanted to start with interfaces. I think it would clarify once I have an implementation ready, then I will add missing docs

execute/exectypes/commit_data.go
mateusz-sekara: No, it's not, it's just the aggregate for all the information necessary to build an exec report. Maybe naming is not perfect, but I don't want to change that as part of my PR.

execute/plugin_functions.go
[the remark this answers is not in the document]
mateusz-sekara: I haven't added TODOs in the all places, because I'm gonna be implementing logic in a followup PR anyway :P

pluginconfig/execute.go
winder: Oops, I misread the code and thought it was defined as `type USDCCCTPObserverConfig []TokenConfig`.

I see what you mean, but you could do the same thing with. Now the dev doesn't have to worry about whether the pointer type is important
```diff
-		if t.USDCCCTPObserverConfig == nil {
+		if (t.USDCCCTPObserverConfig == USDCCCTPObserverConfig{}) {
```

pluginconfig/execute.go
winder: Assuming multiple embedded types in the future, the json Unmarshaller would still allow multiple of them to be initialized.

mateusz-sekara: We only verify here if someone didn't create a JSON monster that contains all possible structs or something with mixed fields

execute/exectypes/observation.go
mateusz-sekara: One message can carry multiple tokens that's why we have `[]TokenData`,
```
message seqNr[1234]
     ETH --> [0]TokenData == []bytes{}
     USDC --> [1]TokenData == []bytes{"attestation"}
     GHO --> [2]TokenData == []bytes{}
```

Probably it would be better to rely on the map here, but that will clarify once I start the real implementation. Right now moving it as it is in v1

execute/exectypes/observation.go
```diff
@@ -18,6 +18,70 @@ type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]ccipty
...
+func (mtd MessageTokenData) IsReady() bool {
+	for _, td := range mtd.TokenData {
+		if !td.IsReady() {
+			return false
+		}
+	}
+	return true
+}
+
+func (mtd MessageTokenData) Error() error {
```
asoliman92, line 46: Maybe use error group? Might be helpful for debugging.

mateusz-sekara: It's just a dummy implementation for now. Followup PR will cover implementation details, everything should clarify then ;)

pluginconfig/execute.go
mateusz-sekara: I've made these to be values instead of pointers

pluginconfig/execute.go
mateusz-sekara: I can't make it work as you posted, probably because `USDCCCTPObserverConfig` contains a map and I can't just use `==` to compare that with empty state. It would require comparing all the fields or using `reflect.DeepEqual()`

IMO, a pointer is a more obvious choice if something is optional, isn't it?

execute/exectypes/observation.go
winder: Makes sense! I guess there could be multiple messages for the same token too so a map wouldn't necessarily help much either

pluginconfig/execute.go
winder: Does 0 need to be an allowable value? IIRC these are hardcoded in the ocr2 plugin

pluginconfig/execute.go
mateusz-sekara: > Assuming multiple embedded types in the future, the json Unmarshaller would still allow multiple of them to be initialized.

We do have `DecodeExecuteOffchainConfig(encodedExecuteOffchainConfig []byte) (ExecuteOffchainConfig, error)` which is an entry point for deserialization and it's responsible for the JSON unmarshalling. This is also the place in which I call `WellFormed()` function. We can make this logic part of `Validate` but I feel it's better suited for `DecodeExecuteOffchainConfig` where we verify JSON against the expected schema. Therefore `DecodeExecuteOffchainConfig` always returns a syntactically valid object, meaning it would return an error if you try creating a single  `TokenDataObserverConfig` that contains configs for both USDC and MyFancyToken

pluginconfig/execute.go
mateusz-sekara: The short answer, "I don't know" :D Long answer, it's gonna be part of the follow-up PR. We have hardcoded values in the v1 and I don't know how to handle defaults yet. Once I backport usdc logic it should become more obvious on how to handle that.

pluginconfig/execute.go
```diff
@@ -78,6 +81,128 @@ func (e ExecuteOffchainConfig) Validate() error {
...
+	//  {
+	//    "type": "usdc-cctp",
+	//    "version": "2.0",
+	//    "customCirlceAPI": "http://cirle.com/gohere",
+	//    "yetAnotherAPI": "http://cirle.com/anotherone",
+	//    "customCircleAPITimeout": "1s",
+	//    "yetAnotherAPITimeout": "500ms"
+	//  }
+	//]
+	// Having version in that JSONisn't expensive, but it could reduce the risk of breaking the observers in the future.
```
makramkd, line 133:
```suggestion
	// Having version in that JSON isn't expensive, but it could reduce the risk of breaking the observers in the future.
```

execute/report/report.go
```diff
@@ -35,6 +34,11 @@ func buildSingleChainReportHelper(
 			readyMessages = make(map[int]struct{})
 		}
 		for i := 0; i < len(report.Messages); i++ {
+			if !report.MessageTokenData[i].IsReady() {
```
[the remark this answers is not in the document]
mateusz-sekara: Actually, it's redundant, we check `IsReady` in the `checkMessage` function

makramkd (approved):

winder (approved):

PR #119, opened 2024-09-13
RMNHome integration in the Home chain

0xnogo:
Proposed approach:
* RMNHome config will be fetched in the home chain reader
* RMNHome can have multiple config (buffer ring size = 2) so we storing the configs in a map[configDigest]RMNHomeConfig. It is going to be the responsibility of the client to get the correct the config depending on the Remote config it is using (a pointer to the configDigest is present in the remote config)

Changes:
* update of the `home_chain.go`
* seperation Home and Remote config and modification on the `config.go` to reflect the onchain data structures
* update of the references.


internal/reader/home_chain.go
```diff
@@ -39,6 +40,36 @@ type HomeChain interface {
 	services.Service
 }
 
+//go:generate mockery --name RMNConfigFetcher --output ./mocks/ --case underscore
```
dimkouv, line 43: This was replaced with `.mockery.yaml`

internal/reader/home_chain.go
```diff
@@ -39,6 +40,36 @@ type HomeChain interface {
...
+type RMNConfigFetcher interface {
+	GetRMNNodesInfo() ([]RMNNodeInfo, error)
+	GetRMNHomeConfigDigest() (cciptypes.Bytes, error)
+	// TODO: We could return the value for a selector, depending on how it's used
```
dimkouv, line 47: we can figure this out later, should be ok now

internal/reader/home_chain.go
```diff
@@ -39,6 +40,36 @@ type HomeChain interface {
...
+	GetMinObservers() (map[cciptypes.ChainSelector]int, error)
+	GetMinSigners() (map[cciptypes.ChainSelector]int, error)
+}
+
```
dimkouv, line 51: My concern is that probably this should not be part of `home_chain.go`. Since we are also querying other chains for `RMNRemote` config.

makramkd:
Why a new interface and not just methods on home chain reader?

internal/reader/home_chain.go
makramkd: In my mind they're totally separate flows, there is config that must be fetched from RMNHome and config from RMNRemote

makramkd: Anything that should be fetched from home can be part of the home chain reader. Anything not related to home chain would be part of the CCIP reader or we can create an RMN remote reader interface.

0xnogo: I agree with this. My plan has shifted since this PR.
In short, we will use the home chain reader for the RMNHome and build another reader for RMNRemote. I'm not quite sure how to deal with the fact that not all the nodes can read RMNRemote but I'll deal with it in another PR.

0xnogo: I also logically all the data coming from Home and Remote in different struct to avoid having to operate on the same struct by 2 different processes

commit/merkleroot/rmn/client.go
```diff
@@ -104,19 +105,21 @@ func (c *client) ComputeReportSignatures(
...
-	// Filter out the lane update requests for chains without enough RMN nodes supporting them.
-	for chain, l := range updatesPerChain {
-		if l.rmnNodes.Cardinality() < c.rmnCfg.Remote.MinObservers {
-			c.lggr.Warnw("not enough RMN nodes for chain",
-				"chain", chain, "minObservers", c.rmnCfg.Remote.MinObservers)
-			delete(updatesPerChain, chain)
+		// Filter out the lane update requests for chains without enough RMN nodes supporting them.
+		for chain, l := range updatesPerChain {
+			if l.rmnNodes.Cardinality() <
+				int(c.rmnCfg.Home.MinObservers[cciptypes.ChainSelector(updateReq.LaneSource.SourceChainSelector)]) {
```
asoliman92, line 117: I see similar access repeated multiple times with different variables. Maybe add a method to `cfg.Home` that takes a uint64 and returns the int instead.

internal/reader/home_chain.go
```diff
@@ -322,6 +413,67 @@ func convertOnChainConfigToHomeChainConfig(
...
+				SignObservationsPublicKey: &pubKey,
+				SupportedSourceChains:     mapset.NewSet[cciptypes.ChainSelector](),
+			}
+		}
+
+		minObservers := make(map[cciptypes.ChainSelector]uint64)
+
+		for _, chain := range config.SourceChains {
+			minObservers[chain.ChainSelector] = chain.MinObservers
+			for j := 0; j < 256; j++ {
```
asoliman92, line 444: what's the 256?

internal/reader/home_chain.go
```diff
@@ -26,6 +30,7 @@ const (
 
 //go:generate mockery --name HomeChain --output ./mocks/ --case underscore
```
makramkd, line 31: This should be deleted, we don't use go:generate for mockery anymore

internal/reader/home_chain.go
```diff
@@ -36,6 +41,17 @@ type HomeChain interface {
 	GetFChain() (map[cciptypes.ChainSelector]int, error)
 	// GetOCRConfigs Gets the OCR3Configs for a given donID and pluginType
 	GetOCRConfigs(ctx context.Context, donID uint32, pluginType uint8) ([]OCR3ConfigWithMeta, error)
+
+	// RMNHome specific methods
```
makramkd, line 45: Minor improvement that is self-documenting:

```golang
type RMNHome interface {
    // GetRMNNodesInfo gets the RMNHomeNodeInfo for the given configDigest
	GetRMNNodesInfo(configDigest cciptypes.Bytes32) ([]rmn.RMNHomeNodeInfo, error)
	// IsRMNHomeConfigDigestSet checks if the configDigest is set in the RMNHome contract
	IsRMNHomeConfigDigestSet(configDigest cciptypes.Bytes32) (bool, error)
	// GetMinObservers gets the minimum number of observers required for each chain in the given configDigest
	GetMinObservers(configDigest cciptypes.Bytes32) (map[cciptypes.ChainSelector]uint64, error)
	// GetOffChainConfig gets the offchain config for the given configDigest
	GetOffChainConfig(configDigest cciptypes.Bytes32) (cciptypes.Bytes, error)
}

type CCIPHome interface {
// CCIPHome methods here
}

type HomeChain interface {
  CCIPHome
  RMNHome
  services.Service
}
```

internal/reader/home_chain.go
```diff
@@ -36,6 +41,17 @@ type HomeChain interface {
...
+	// GetRMNNodesInfo gets the RMNHomeNodeInfo for the given configDigest
+	GetRMNNodesInfo(configDigest cciptypes.Bytes32) ([]rmn.RMNHomeNodeInfo, error)
```
makramkd, line 47: Not sure I like the RMN import here, I feel like it will lead to import cycles at some point. Maybe move `RMNHomeNodeInfo` to a `types` subpackage within `rmn` to reduce the likelihood?

internal/reader/home_chain.go
```diff
@@ -58,11 +74,13 @@ type homeChainPoller struct {
 	lggr            logger.Logger
 	mutex           *sync.RWMutex
 	state           state
+	rmnHomeConfig   map[cciptypes.Bytes32]rmn.RMNHomeConfig
 	failedPolls     uint
 	// TODO: currently unused but will be passed into GetLatestValue
 	// once the chainlink-common breaking change comes in
 	// (https://github.com/smartcontractkit/chainlink-common/pull/603).
```
makramkd, line 81: This comment is stale, should be removed.

internal/reader/home_chain.go
```diff
@@ -175,6 +202,45 @@ func (r *homeChainPoller) setState(chainConfigs map[cciptypes.ChainSelector]Chai
...
 
+func (r *homeChainPoller) fetchAndSetRMNConfig(ctx context.Context) error {
+	var versionedConfigWithDigests []VersionedConfigWithDigest
+	err := r.homeChainReader.GetLatestValue(
+		ctx,
+		r.rmnHomeBoundContract.ReadIdentifier(consts.MethodNameGetVersionedConfigsWithDigests),
+		primitives.Unconfirmed,
+		map[string]interface{}{
+			"offset": 0,
+			"limit":  2, // TODO: fetch CONFIG_RING_BUFFER_SIZE
```
makramkd, line 213: I placed a comment on the contract PR but this API seems convoluted to me

internal/reader/home_chain.go
```diff
@@ -175,6 +202,45 @@ func (r *homeChainPoller) setState(chainConfigs map[cciptypes.ChainSelector]Chai
...
+	// TODO: fetch CONFIG_RING_BUFFER_SIZE and compare with len(versionedConfigWithDigests)
+	if len(versionedConfigWithDigests) > 2 {
+		r.lggr.Errorw("more than 2 RMNHomeConfigs found", "numConfigs", len(versionedConfigWithDigests), "requestedLimit", 2)
+		return fmt.Errorf("more than 2 RMNHomeConfigs found")
+	}
+
+	r.setRMNHomeState(convertOnChainConfigToRMNHomeChainConfig(r.lggr, versionedConfigWithDigests))
+
+	if len(versionedConfigWithDigests) == 0 {
+		// That's a legitimate case if there are no chain configs on chain yet
```
makramkd, line 230:
```suggestion
		// That's a legitimate case if there are no rmn configs on chain yet
```

internal/reader/home_chain.go
makramkd: Max RMN committee size, should be pulled into a const

internal/reader/home_chain.go
```diff
@@ -322,6 +413,67 @@ func convertOnChainConfigToHomeChainConfig(
...
+			OffchainConfig: config.OffchainConfig,
+		}
+	}
+	return rmnHomeConfigs
+}
+
+// IsNodeObserver checks if a node is an observer for the given source chain
+func isNodeObserver(sourceChain SourceChain, nodeIndex int) bool {
+	if nodeIndex >= 256 {
+		return false // uint256 can only represent up to 256 nodes
```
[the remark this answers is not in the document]
makramkd: `uint256` as a bitmap, this comment is incomplete

internal/reader/home_chain.go
```diff
@@ -370,4 +522,36 @@ type OCR3ConfigWithMeta struct {
 	ConfigDigest [32]byte   `json:"configDigest"`
 }
 
+// VersionedConfigWithDigest mirrors RMNHome.sol's VersionedConfigWithDigest struct
```
makramkd, line 525: Lets organize things better and move this to a rmn_types.go or similar file

internal/reader/home_chain.go
```diff
@@ -370,4 +522,36 @@ type OCR3ConfigWithMeta struct {
...
+	OffchainConfig cciptypes.Bytes `json:"offchainConfig"`
+}
+
+// Node mirrors RMNHome.sol's Node struct
+type Node struct {
+	PeerID            cciptypes.Bytes32 `json:"peerId"`
+	OffchainPublicKey cciptypes.Bytes32 `json:"offchainPublicKey"`
+}
+
+// SourceChain mirrors RMNHome.sol's SourceChain struct
```
asoliman92, line 550: Bette link to the contract as well here.

internal/reader/home_chain.go
```diff
@@ -322,6 +413,67 @@ func convertOnChainConfigToHomeChainConfig(
...
+			MinObservers:   minObservers,
+			ConfigDigest:   versionedConfigWithDigest.ConfigDigest,
+			OffchainConfig: config.OffchainConfig,
+		}
+	}
+	return rmnHomeConfigs
+}
+
+// IsNodeObserver checks if a node is an observer for the given source chain
+func isNodeObserver(sourceChain SourceChain, nodeIndex int) bool {
```
makramkd, line 462: Lets add unit tests for this

internal/reader/home_chain_test.go
```diff
@@ -172,12 +192,30 @@ func Test_PollingWorking(t *testing.T) {
 			callCount++
 		}
 	}
-	// called at least 2 times, one for start and one for the first tick
-	require.GreaterOrEqual(t, callCount, 2)
-
+	// called at least 4 times, one for start and one for the first tick for each contract
```
asoliman92, line 195: Maybe explain that the 4 are for 2*2 for home and rmn?

internal/reader/home_chain_test.go
```diff
@@ -172,12 +192,30 @@ func Test_PollingWorking(t *testing.T) {
...
+	require.GreaterOrEqual(t, callCount, 4)
 	configs, err := configPoller.GetAllChainConfigs()
 	require.NoError(t, err)
 	require.Equal(t, homeChainConfig, configs)
+
+	for _, configs := range rmnHomeOnChainConfigs {
+		rmnNodes, err := configPoller.GetRMNNodesInfo(configs.ConfigDigest)
+		require.NoError(t, err)
+		require.NotNil(t, rmnNodes)
+		fmt.Println("GetRMNNodesInfo", rmnNodes)
```
asoliman92, line 205: remove print

internal/reader/home_chain.go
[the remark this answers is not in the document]
0xnogo: This [bitmap](https://github.com/smartcontractkit/ccip/blob/e6e26ad31eef625faf68806a2b4f0549bc89b15c/contracts/src/v0.8/ccip/RMNHome.sol#L24) can only represent the observer status of up to 256 nodes.

In the given struct, observerNodesBitmap is used as a bitmap where each bit represents a node. The i-th bit being 1 means the i-th node is an observer.

Example: 101: 
* node 0 is an obs
* node 1 is not
* node 2 is an obs

I have improved the method with more checks and tests.

internal/reader/home_chain.go
0xnogo: Based on our last meeting, the API is going to be changed. I'll update the code in an upcoming PR when changes are ready.

PR #123, opened 2024-09-17
Use DiscoverContracts and bindContractReader.

winder:
Replace `bindOfframp`, `bindOnramps` and `bindNonceManager` with `DiscoverContracts` and the new `Sync` function.

The offramp needs to be bound first, so that now happens during `CCIPReader` initialization.


pkg/reader/ccip_test.go
```diff
@@ -58,9 +59,6 @@ func TestCCIPChainReader_getSourceChainsConfig(t *testing.T) {
 		}, nil, chainC, offrampAddress,
 	)
 
-	sourceCRs[chainA].On("Bind", mock.Anything, mock.Anything).Return(nil)
-	sourceCRs[chainB].On("Bind", mock.Anything, mock.Anything).Return(nil)
-	destCR.On("Bind", mock.Anything, mock.Anything).Return(nil)
```
winder, line 63: Technically, `destCR.Bind` was the only one which needed to move (The offramp is now bound during`newCCIPChainReaderInternal`). Moved the rest up for consistency and switched to the `.EXPECT()` for better readability.

pkg/reader/ccip.go
```diff
@@ -57,16 +57,14 @@ func newCCIPChainReaderInternal(
...
-		if err := reader.Sync(context.Background(), contracts); err != nil {
-			lggr.Infow("failed to sync contracts", "err", err)
-		}
-	*/
+	contracts := ContractAddresses{
+		consts.ContractNameOffRamp: {
+			destChain: offrampAddress,
+		},
+	}
+	if err := reader.Sync(context.Background(), contracts); err != nil {
```
makramkd, line 65: Seems like the 2nd arg is unused, not sure if the offramp is getting bound?

pkg/reader/ccip.go
```diff
@@ -565,10 +561,25 @@ func (r *ccipChainReader) bindReaderContract(
 	return nil
 }
 
-// newSync goes through the input contracts and binds them to the contract reader.
-//
-//nolint:unused // it will be used soon.
+func (r *ccipChainReader) Sync(ctx context.Context, _ ContractAddresses) error {
+	contracts, err := r.DiscoverContracts(ctx, r.destChain)
```
makramkd, line 565: I don't think `DiscoverContracts` will work if the offramp is not bound first, that's probably causing the int test to fail

winder: aaaah. ty!

makramkd (approved):

PR #129, opened 2024-09-18
Contract Discovery Processor.

winder:
Implements a new discovery processor to observe contract addresses and bind the results.

### DIFF
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -497,6 +497,13 @@ func (p *Plugin) Outcome(
 	}
 
 	if outcome.IsEmpty() {
+		p.lggr.Warnw(
+			fmt.Sprintf("[oracle %d] exec outcome: empty outcome", p.reportingCfg.OracleID),
+			"oracle", p.reportingCfg.OracleID,
+			"execPluginState", state)
+		if p.initialized {
+			return exectypes.Outcome{State: exectypes.Initialized}.Encode()
+		}
 		return nil, nil
 	}
 
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -9,8 +9,6 @@ import (
 	mapset "github.com/deckarep/golang-set/v2"
 	"golang.org/x/exp/maps"
 
-	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery"
-	dt "github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery/discoverytypes"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 	"github.com/smartcontractkit/libocr/commontypes"
@@ -24,6 +22,8 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/execute/tokendata"
 	typeconv "github.com/smartcontractkit/chainlink-ccip/internal/libs/typeconv"
 	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery"
+	dt "github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery/discoverytypes"
 	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
 	readerpkg "github.com/smartcontractkit/chainlink-ccip/pkg/reader"
 	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -51,7 +51,7 @@ type Plugin struct {
 	lggr              logger.Logger
 
 	// state
-	initialized bool
+	contractsInitialized bool
 }
 
 func NewPlugin(
@@ -192,7 +192,7 @@ func (p *Plugin) Observation(
 			p.lggr.Errorw("failed to discover contracts", "err", err)
 		}
 
-		if !p.initialized {
+		if !p.contractsInitialized {
 			return exectypes.Observation{Contracts: discoveryObs}.Encode()
 		}
 	}
@@ -416,7 +416,7 @@ func (p *Plugin) Outcome(
 		if err != nil {
 			return nil, fmt.Errorf("unable to process outcome of discovery processor: %w", err)
 		}
-		p.initialized = true
+		p.contractsInitialized = true
 	}
 
 	fChain, err := p.homeChain.GetFChain()
@@ -501,7 +501,7 @@ func (p *Plugin) Outcome(
 			fmt.Sprintf("[oracle %d] exec outcome: empty outcome", p.reportingCfg.OracleID),
 			"oracle", p.reportingCfg.OracleID,
 			"execPluginState", state)
-		if p.initialized {
+		if p.contractsInitialized {
 			return exectypes.Outcome{State: exectypes.Initialized}.Encode()
 		}
 		return nil, nil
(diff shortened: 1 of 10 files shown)


internal/plugincommon/discovery/processor.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/pkg/reader"
+)
+
+// Ensure that PluginProcessor is implemented.
+var _ plugincommon.PluginProcessor[dt.Query, dt.Observation, dt.Outcome] = &ContractDiscoveryProcessor{}
+
+// ContractDiscoveryProcessor is a plugin processor for discovering contracts.
+type ContractDiscoveryProcessor struct {
+	lggr      logger.Logger
+	reader    *reader.CCIPReader
```
makramkd, line 23: Why a pointer to an interface?

internal/plugincommon/discovery/processor.go
```diff
@@ -0,0 +1,120 @@
...
+func (cdp *ContractDiscoveryProcessor) Query(_ context.Context, _ dt.Outcome) (dt.Query, error) {
+	return nil, nil
+}
+
+// Observation reads contract addresses from the destination chain.
+// In the future this should be extended to omit observations unless one of the nodes requests addresses.
+func (cdp *ContractDiscoveryProcessor) Observation(
+	ctx context.Context, _ dt.Outcome, _ dt.Query,
+) (dt.Observation, error) {
+	contracts, err := (*cdp.reader).DiscoverContracts(ctx, cdp.dest)
```
makramkd, line 55: If you remove the pointer-to-interface this can be simplified to


```suggestion
	contracts, err := cdp.reader.DiscoverContracts(ctx, cdp.dest)
```

execute/plugin.go
```diff
@@ -190,7 +216,7 @@ func (p *Plugin) Observation(
 			}
 
 			// TODO: truncate grouped to a maximum observation size?
-			return exectypes.NewObservation(groupedCommits, nil, nil, nil).Encode()
+			return exectypes.NewObservation(groupedCommits, nil, nil, nil, discoveryObs).Encode()
```
makramkd, line 219: Suggestion for future PRs: maybe have a special constructor for the observation that is expected to be returned in each state, e.g `NewGetCommitReportsObservation` which would only take two args in this case `groupedCommits` and `discoveryObs` to avoid this `nil, nil, nil` dance which is not super readable. Alternatively we could instantiate the struct directly `Observation{CommitReports: groupedCommits, DiscoveryObs: discoveryObs}`

internal/plugincommon/discovery/processor.go
winder: We need to make sure that when the reader is updated here it is also updated in other parts of the plugin.

It works without the pointer because the shallow copies mean each slice refers to the same array, but will that always be the case?

makramkd: The implementation is a pointer receiver so it should work without the pointer. However, I kind of see what you're saying about the shared aspect of it.

The way the reader is used right now, its basically a singleton. The way things are designed right now, I don't see this being the case. Feels like a change in this would cause a super subtle bug though, gotta think

makramkd: Correct me if I'm wrong on this but I think the type assertion on the `ccipChainReader` type helps us avoid this problem.

```golang
// Interface compliance check - on the pointer type
var _ CCIPReader = (*ccipChainReader)(nil)
// Interface compliance check - on the value type - FAILS
var _ CCIPReader = ccipChainReader{}
```

So as long as:

* CCIP reader is a singleton
* interface compliance is on the pointer receiver

It should be safe to have this as just the interface rather than pointer to interface. We can put a comment on the compliance check to ensure that this cannot change and what the implications are.

makramkd: Maybe there's a test we can write to ensure that the pointer type implements the interface - not sure though, seems out of scope of the Go testing facilities since its part of the type system.

winder: There are examples in the stdlib of tricks to avoid accidentally copying things that shouldn't be copied.

strings.Builder
```
// A Builder is used to efficiently build a string using [Builder.Write] methods.
// It minimizes memory copying. The zero value is ready to use.
// Do not copy a non-zero Builder.
type Builder struct {
	addr *Builder // of receiver, to detect copies by value
	buf  []byte
}
```

io.File
```
// File represents an open file descriptor.
type File struct {
	*file // os specific
}
```

Maybe we should adopt one of these

makramkd: Or maybe leverage `sync.Mutex` not being copyable:

```golang
type nonCopyable struct {
    _ sync.Mutex
}

type ccipChainReader struct {
  nonCopyable
// .. rest
}
```

internal/plugincommon/discovery/processor.go
```diff
@@ -0,0 +1,120 @@
...
+// Ensure that PluginProcessor is implemented.
+var _ plugincommon.PluginProcessor[dt.Query, dt.Observation, dt.Outcome] = &ContractDiscoveryProcessor{}
+
+// ContractDiscoveryProcessor is a plugin processor for discovering contracts.
+type ContractDiscoveryProcessor struct {
+	lggr      logger.Logger
+	reader    *reader.CCIPReader
+	homechain reader.HomeChain
+	dest      cciptypes.ChainSelector
+	bigF      int
```
makramkd, line 26: nit: this is now called `fRoleDON` in the contracts, I think its a more descriptive name :-)

execute/plugin.go
```diff
@@ -372,13 +398,34 @@ func (p *Plugin) Outcome(
...
+			discoveryAos[i] = plugincommon.AttributedObservation[dt.Observation]{
+				OracleID:    decodedAos[i].OracleID,
+				Observation: decodedAos[i].Observation.Contracts,
+			}
+		}
+		_, err = p.discovery.Outcome(dt.Outcome{}, dt.Query{}, discoveryAos)
+		if err != nil {
+			return nil, fmt.Errorf("unable to process outcome of discovery processor: %w", err)
+		}
+		p.initialized = true
```
asoliman92, line 419: Should we initialize once it's reported?

internal/plugincommon/discovery/discoverytypes/types.go
```diff
@@ -0,0 +1,21 @@
+package discoverytypes
+
+import "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+// Outcome isn't needed for this processor.
+type Outcome struct {
+	// TODO: some sort of request flag to avoid including this every time.
```
asoliman92, line 7: Isn't this what the initialized state do?

internal/plugincommon/discovery/discoverytypes/types.go
```diff
@@ -0,0 +1,21 @@
+package discoverytypes
+
+import "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+// Outcome isn't needed for this processor.
```
asoliman92, line 5: wrong comment?

execute/plugin.go
asoliman92, line 419: Need to check that the Outcome actually did sync contracts before initializing here?

internal/plugincommon/discovery/discoverytypes/types.go
winder: This is referring to the type. I'm not including the contracts in the final outcome.

execute/plugin.go
winder: The idea was that no error means that everything has initialized.

asoliman92 (approved):

makramkd (approved):

PR #130, opened 2024-09-18
CCIP-2219 - Phase 2 - stubbing parts of USDC/CCTP processing 

mateusz-sekara:
Continuing my top-down journey with USDC implementation. The scope of this PR is the following:
* add the tests showing contracts between components and adjusting interfaces based on that
* wiring all components together, USDCReader and AttestationClients are still stubbed, but higher-order components know how to aggregate data from different interfaces
* adding defaults to the `USDCCCTPObserverConfig`

Next PRs would include:
* USDC reader implementations
* AttestationClient implementation
* Plugin level tests with USDC tokens (exec)
* Integration test verifying end-to-end scenarios


execute/tokendata/observer.go
```diff
@@ -2,42 +2,145 @@ package tokendata
...
 // Goal of that is to support multiple token observers supporting different tokens (e.g. CCTP, MyFancyToken etc)
 type CompositeTokenDataObserver struct {
-	Observers []TokenDataObserver
+	observers []TokenDataObserver
+}
+
+// nolint early-return
+func NewConfigBasedCompositeObservers(config []pluginconfig.TokenDataObserverConfig) (*CompositeTokenDataObserver, error) {
+	observers := make([]TokenDataObserver, len(config))
+	for i, c := range config {
```
makramkd, line 44: Is there an assumption here that `config` is de-duped, i.e we don't have something like `[usdcConfig, usdcConfig]`?

execute/tokendata/observer.go
```diff
@@ -2,42 +2,145 @@ package tokendata
...
+	for chainSelector, observation := range observations {
+		tokenObservation[chainSelector] = make(map[cciptypes.SeqNum]exectypes.MessageTokenData)
+
+		for seq, message := range observation {
+			tokenData := make([]exectypes.TokenData, len(message.TokenAmounts))
+			for i, token := range message.TokenAmounts {
+				if !t.IsTokenSupported(chainSelector, token) {
+					// It means that none of the registers support that,
+					// we assume token doesn't require additional processing and skip it
+					tokenData[i] = exectypes.NewNoopTokenData()
```
makramkd, line 108: I thought we have to use `exectypes. NotSupportedTokenData` according to the doc comment above:

```golang
// * if token is not supported, TokenDataObserver must set the token data to exectypes.NotSupportedTokenData
```

execute/tokendata/observer.go
```diff
@@ -2,42 +2,145 @@ package tokendata
...
+			}
+			tokenObservation[chainSelector][seq] = exectypes.MessageTokenData{TokenData: tokenData}
+		}
+	}
+	return tokenObservation
+}
+
+func merge(
+	base exectypes.TokenDataObservations,
+	from exectypes.TokenDataObservations,
```
makramkd, line 123: Feels like this could do with a test

execute/tokendata/usdc/usdc.go
```diff
@@ -4,43 +4,168 @@ import (
...
+	"golang.org/x/exp/maps"
+
 	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
+	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
 	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
 )
 
 type USDCCCTPTokenDataObserver struct {
-	configs []pluginconfig.USDCCCTPObserverConfig
+	configs           pluginconfig.USDCCCTPObserverConfig
```
makramkd, line 16:
```suggestion
	config           pluginconfig.USDCCCTPObserverConfig
```

Since its just a single config

execute/tokendata/usdc/usdc.go
```diff
@@ -4,43 +4,168 @@ import (
...
+
+func NewUSDCCCTPTokenDataObserver(configs pluginconfig.USDCCCTPObserverConfig) *USDCCCTPTokenDataObserver {
+	return NewUSDCCCTP(
+		configs,
+		nil,
+		nil,
+	)
+}
+
+func NewUSDCCCTP(
```
makramkd, line 30: Both this and the above are public facing constructors, lets have either some docs on which one to use for which scenario or just have one `NewTokenDataObserver` (since package name is usdc, it reads like `usdc.NewTokenDataObserver`) to reduce likelihood of using wrong one and confusion

execute/tokendata/usdc/usdc.go
```diff
@@ -4,43 +4,168 @@ import (
...
 
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+	"golang.org/x/exp/maps"
+
 	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
+	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
 	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
 )
 
 type USDCCCTPTokenDataObserver struct {
```
makramkd, line 15: This name is a mouthful, two questions:

* any reason this needs to be a public struct?
* can this just be `TokenDataObserver` since we're in the USDC package? The "USDCCCTP" prefix seem redundant

similar comments apply elsewhere where this phenomenon is present. If it can be private then just `tokenDataObserver` is good enough IMO. Its perfectly fine to return private structs, see https://go.dev/doc/effective_go#generality

execute/tokendata/usdc/usdc.go
```diff
@@ -4,43 +4,168 @@ import (
...
+			}
+		}
+	}
+	return usdcMessages
+}
+
+func (u *USDCCCTPTokenDataObserver) fetchUSDCMessageHashes(
+	ctx context.Context,
+	usdcMessages map[cciptypes.ChainSelector]map[cciptypes.SeqNum][]int,
+) (map[cciptypes.ChainSelector]map[cciptypes.SeqNum]map[int][]byte, error) {
```
makramkd, line 105: I wonder if some "key flattening" would help here now that you have 3 nested maps? Maybe instead the type can be:

```suggestion
type seqNumTokIdx struct {
  seqNum cciptypes.SeqNum
  // tokIdx is the index into the token amounts array part of the message
  tokIdx int
}

) (map[cciptypes.ChainSelector]map[seqNumTokIdx][]byte, error) {
```

makramkd: Having just the `int` type in the map keys feels pretty opaque to me. Similar thoughts on the other 3x nested map types

pluginconfig/execute.go
```diff
@@ -53,8 +53,9 @@ type ExecuteOffchainConfig struct {
 	// BatchingStrategyID is the strategy to use for batching messages.
 	BatchingStrategyID uint32 `json:"batchingStrategyID"`
 
-	// TokenDataObserver registers different strategies for processing token data.
-	TokenDataObserver []TokenDataObserverConfig `json:"tokenDataObservers"`
+	// TokenDataObservers registers different strategies for processing token data.
+	// TokenDataObservers registers different strategies for processing token data.
```
makramkd, line 57: Duplicated comment, can delete

execute/tokendata/observer.go
mateusz-sekara: At this stage we assume config is deduped, I will add the test to the config parser to cover that edge case

execute/tokendata/observer.go
mateusz-sekara: There are plenty of tests that covers that path in the observer_test.go, you want more granular ones?

execute/tokendata/usdc/usdc.go
mateusz-sekara: yeah, it can be private as long as we expose actors, will do

execute/tokendata/usdc/usdc.go
mateusz-sekara: These nested maps are terrible indeed. I want to reiterate that once I start implementing that interface

execute/tokendata/usdc/usdc.go
mateusz-sekara: > similar comments apply elsewhere where this phenomenon is present. If it can be private then just tokenDataObserver is good enough IMO. Its perfectly fine to return private structs, see https://go.dev/doc/effective_go#generality

Actually, our linter settings don't allow me to return private structs from constructors

```go
execute/tokendata/usdc/usdc.go:26:3: unexported-return: exported func NewTokenDataObserver returns unexported type *usdc.tokenDataObserver, which can be annoying to use (revive)
) *tokenDataObserver {
```

execute/tokendata/observer.go
```diff
@@ -2,42 +2,145 @@ package tokendata
...
 type CompositeTokenDataObserver struct {
-	Observers []TokenDataObserver
+	observers []TokenDataObserver
+}
+
+// nolint early-return
+func NewConfigBasedCompositeObservers(config []pluginconfig.TokenDataObserverConfig) (*CompositeTokenDataObserver, error) {
+	observers := make([]TokenDataObserver, len(config))
+	for i, c := range config {
+		if c.USDCCCTPObserverConfig != nil {
```
asoliman92, line 45: Nit: styling, use switch case instead

execute/tokendata/observer.go
```diff
@@ -2,42 +2,145 @@ package tokendata
...
 ) (exectypes.TokenDataObservations, error) {
-	var tokenObservation exectypes.TokenDataObservations
-	for _, ob := range t.Observers {
-		tokenData, err := ob.Observe(ctx, messages)
+	tokenDataObservations := t.initTokenDataObservations(msgObservations)
+
+	for _, ob := range t.observers {
+		tokenData, err := ob.Observe(ctx, msgObservations)
+		if err != nil {
+			return nil, err
```
asoliman92, line 70: Why return if one observer has an error? Others can complete the operation successfully.

asoliman92: Maybe also check IsTokenSupported before the Observe call?

mateusz-sekara: > Maybe also check IsTokenSupported before the Observe call?

It doesn't work this way, every Observer gets all the data, if it doesn't support any tokens it should just bypass and return empty data. Also `IsTokenSupported` works on a token level, whereas we do batching here and pass everything. Therefore underlying implementation can batch requests more efficiently

execute/tokendata/observer.go
```diff
@@ -2,42 +2,146 @@ package tokendata
...
+		}
+	}
+	return false
+}
+
+// initTokenDataObservations initializes the token data observations with empty token data, it asks the child observers
+// whether the token is supported or not. If token is supported and requires additional processing we set its state to
+// isReady=false. If token is noop (doesn't require offchain processing) we initialize it with empty
+// data and set IsReady=true.
+func (t *compositeTokenDataObserver) initTokenDataObservations(
```
asoliman92, line 106: `getSupportedTokenDataObs` better?

execute/tokendata/observer.go
mateusz-sekara: > Why return if one observer has an error? Others can complete the operation successfully.

Yeah, I guess you are right, no need to return an error here. I think it's not that relevant at this stage as we have only a single observer, but let me change that and create a test for that

execute/tokendata/observer.go
mateusz-sekara: `getSupportedTokenDataObs` that's not true, we initialize everything, but set proper values based on whether the token is supported or not

execute/tokendata/observer.go
mateusz-sekara: `compositeTokenDataObserver` works as an orchestrator here, it asks his children what tokens are supported and what not. Because we don't have an explicit list of tokens that don't require processing we implicitly get that by checking which tokens are supported by observers.

Therefore, we need to build some empty state and that's exactly what we do here by inserting `NewNoopTokenData()`. The comment I provided is true for any other `TokenDataObserver` that will be implemented in the future, but the composite has to cheat a bit here to initialize the empty state properly.

execute/tokendata/observer.go
asoliman92: ahaa... one generic comment on the approach is why use slices and check for lengths and have noop/unsupported? Isn't it better to use a map of messageID to tokenData when only it has data and return the tokenData mapped by messageID?

makramkd (approved):

asoliman92 (approved):

PR #131, opened 2024-09-18
commit plugin - support batch sizes greater than one

dimkouv:
So far commit plugin did not support batch sizes greater than 1, meaning that the reported sequence number ranges were always of size 1 e.g. `chain1: [123, 123]`.

---

In this PR I use the ccipReader to observe the onRamp sequence numbers to increase the batch sizes.

Since every onRamp exists on a different chain we cannot use batch calls, goroutines are spawned instead for each source chain using `errgroup`.

If for a specific source chain we cannot get the onRamp sequence number, onRamp sequence numbers observation stops.

Commit plugin should limit the amount of data it can process in a round. In a scenario when we have too many pending messages, e.g. [123, 10000] we limit this ranges up to a specific configured batch size which is set to a default of `256`. 

Limits can be configured on plugin initialization.


commit/merkleroot/observation.go
```diff
@@ -202,6 +205,55 @@ func (o ObserverImpl) ObserveOffRampNextSeqNums(ctx context.Context) []plugintyp
...
+			if err != nil {
+				return fmt.Errorf("failed to get expected next sequence number for source chain %d: %w", sourceChain, err)
+			}
+			if nextOnRampSeqNum == 0 {
+				return fmt.Errorf("expected next sequence number for source chain %d is 0", sourceChain)
+			}
+
+			latestOnRampSeqNums[i] = plugintypes.SeqNumChain{
+				ChainSel: sourceChain,
+				SeqNum:   nextOnRampSeqNum - 1, // Latest is the next one minus one.
```
winder, line 243: `latestOnRampSeqNum` would be set to 0 for a new chain. Is that OK?

internal/libs/cciptypeutil/seqnumrange.go
```diff
@@ -0,0 +1,21 @@
+package cciptypeutil
+
+import "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+// SeqNumRangeLimit limits the range to n elements by truncating the end if necessary.
+func SeqNumRangeLimit(rng ccipocr3.SeqNumRange, n uint64) ccipocr3.SeqNumRange {
```
winder, line 6: nit: could add a `.Limit` function to the SeqNumRange type in chainlink-common.

commit/merkleroot/outcome.go
```diff
@@ -82,11 +85,24 @@ func reportRangesOutcome(
...
+			rng := cciptypes.NewSeqNumRange(offRampNextSeqNum, onRampMaxSeqNum)
+
 			chainRange := plugintypes.ChainRange{
 				ChainSel:    chainSel,
-				SeqNumRange: [2]cciptypes.SeqNum{offRampNextSeqNum, onRampMaxSeqNum},
+				SeqNumRange: cciptypeutil.SeqNumRangeLimit(rng, rngLimit),
 			}
 			rangesToReport = append(rangesToReport, chainRange)
+
+			if wasTruncated := rng.End() != chainRange.SeqNumRange.End(); wasTruncated {
```
winder, line 101:
```suggestion
			// check if the range was truncated.
			if rng.End() != chainRange.SeqNumRange.End() {
```

commit/merkleroot/processor.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/shared"
 )
 
+// DefaultSeqNumsBatchLimit is the default number of max new messages to scan, we use this value when
+// the config is not set for a specific chain.
+const DefaultSeqNumsBatchLimit = 256
```
winder, line 20: [MaxNumberTreeLeaves](https://github.com/smartcontractkit/chainlink-common/blob/e78a0de3f6847845b18da1797e5dccf9521abed2/pkg/merklemulti/merkle_multi.go#L35)

```suggestion
const DefaultSeqNumsBatchLimit = merklemulti.MaxNumberTreeLeaves
```

commit/merkleroot/outcome_test.go
```diff
@@ -37,3 +39,82 @@ func Test_buildReport(t *testing.T) {
...
+			name: "simple scenario with one chain",
+			consensusObservation: ConsensusObservation{
+				OnRampMaxSeqNums: map[cciptypes.ChainSelector]cciptypes.SeqNum{
+					1: 20,
+				},
+				OffRampNextSeqNums: map[cciptypes.ChainSelector]cciptypes.SeqNum{
+					1: 18, // off ramp next is 18, on ramp max is 20 so new msgs are: [18, 19, 20]
+				},
+			},
+			rangeLimitsPerSourceChain: map[cciptypes.ChainSelector]uint64{},
```
winder, line 70:
```suggestion
			rangeLimitsPerSourceChain: map[cciptypes.ChainSelector]uint64{}, // default limits
```

commit/merkleroot/observation.go
dimkouv: yes, should be good.

commit/merkleroot/processor.go
winder: Isn't this the default unless there's an override? Or you're saying there should be no default and the system needs to crash if the config is missing?

makramkd: Ideally the latter (IMO defaults are often more trouble than they're worth in this scenario, because people would just use the default even though it might be wrong) but to start with we could fall back to the default and enforce that its provided later if we decide to do so

commit/merkleroot/observation.go
dimkouv: yeah, in that case there won't be any range reported, since (onRampMax < offRampNext).

internal/libs/cciptypeutil/seqnumrange.go
dimkouv: Added

commit/merkleroot/processor.go
dimkouv: Added EVM prefix and moved to pluginconfig package

internal/libs/cciptypeutil/seqnumrange.go
```diff
@@ -0,0 +1,21 @@
...
+	if numElems <= 0 {
+		return rng
+	}
+
+	if uint64(numElems) > n {
+		newEnd := rng.Start() + ccipocr3.SeqNum(n) - 1
+		if newEnd > rng.End() { // overflow - do nothing
+			return rng
+		}
+		rng.SetEnd(newEnd)
```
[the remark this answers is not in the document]
dimkouv: Pushed to chainlink-common: https://github.com/smartcontractkit/chainlink-common/pull/781/commits
works like this

internal/libs/cciptypeutil/seqnumrange.go
```diff
@@ -0,0 +1,21 @@
+package cciptypeutil
+
+import "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+// SeqNumRangeLimit limits the range to n elements by truncating the end if necessary.
+func SeqNumRangeLimit(rng ccipocr3.SeqNumRange, n uint64) ccipocr3.SeqNumRange {
+	numElems := rng.End() - rng.Start() + 1
+	if numElems <= 0 {
```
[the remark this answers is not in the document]
dimkouv: `< 0` is impossible but checking `<= 0` instead of `== 0` should be harmless. There's an overflow check below.

pluginconfig/commit.go
```diff
@@ -39,6 +40,11 @@ type CommitPluginConfig struct {
 	// RMNSignaturesTimeout is the timeout for RMN signature verification.
 	// Typically set to `MaxQueryDuration - e`, where e some small duration.
 	RMNSignaturesTimeout time.Duration `json:"rmnSignaturesTimeout"`
+
+	// BatchLimits is the maximum number of messages to include in a single report for a target chain.
+	// If for example in the next round we have 1000 pending messages and a batch limit of 256, only 256 seq nums
+	// will be in the report. If a value is not set we fallback to merkleroot.DefaultSeqNumsBatchLimit.
+	BatchLimits map[cciptypes.ChainSelector]uint64 `json:"batchLimits"`
```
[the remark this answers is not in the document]
dimkouv: Made it like this, one value per dest / plugin instance

makramkd (approved):

PR #135, opened 2024-09-18
Introduce contract reader facade

winder:
Avoid using `ContractReader` interface directly. It can lead to incompatibilities as a result of some chain reader details described [here](https://github.com/smartcontractkit/chainlink-common/pull/680).


asoliman92 (approved):
Thanks!

PR #137, opened 2024-09-19
RMN home and remote readers

0xnogo:
Introduction of an rmn home reader (will be modified with the new RMNHome.sol) and an rmn remote reader.

Also modelled what the data would look like in the offchain code (see types/config.go). This will replace the the /config.go in a later PR.


commit/merkleroot/rmn/types/config.go
```diff
@@ -0,0 +1,78 @@
...
+	ConfigDigest     cciptypes.Bytes32
+	Signers          []RMNRemoteSignerInfo
+	MinSigners       uint64
+	ConfigVersion    uint32
+	RmnReportVersion string // e.g., "RMN_V1_6_ANY2EVM_REPORT"
+}
+
+// RMNRemoteSignerInfo contains information about a signer from the RMNRemote contract.
+type RMNRemoteSignerInfo struct {
+	SignReportsAddress    cciptypes.Bytes // for signing reports
```
dimkouv, line 41: Let's rename this (I believe I had it wrong?), maybe just `SignerOnchainAddress` ?

commit/merkleroot/rmn/types/config.go
```diff
@@ -0,0 +1,78 @@
...
+	ConfigDigest   cciptypes.Bytes32
+	OffchainConfig cciptypes.Bytes // Raw offchain configuration bytes
+}
+
+// RMNHomeNodeInfo contains information about a node from the RMNHome contract.
+type RMNHomeNodeInfo struct {
+	ID                        NodeID // ID is the index of this node in the RMN config
+	PeerID                    cciptypes.Bytes32
+	SupportedSourceChains     mapset.Set[cciptypes.ChainSelector]
+	SignObservationsPublicKey *ed25519.PublicKey // offchainPublicKey
```
dimkouv, line 26: Let's rename this to maybe just `OffchainPublicKey` and add a comment e.g `The private key of this pub key.. was used to signed observations...`

commit/merkleroot/rmn/types/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Config mirrors RMNHome.sol's Config struct
+type Config struct {
+	Nodes          []Node          `json:"nodes"`
+	SourceChains   []SourceChain   `json:"sourceChains"`
+	OffchainConfig cciptypes.Bytes `json:"offchainConfig"`
+}
+
+// Node mirrors RMNHome.sol's Node struct
+type Node struct {
```
dimkouv, line 68: Do we need both? Seems like duplicate

dimkouv: or maybe this should not be exported?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+	services.Service
+}
+
+type rmnHomeState struct {
+	primaryConfigDigest   cciptypes.Bytes32
+	secondaryConfigDigest cciptypes.Bytes32
+	rmnHomeConfig         map[cciptypes.Bytes32]rmntypes.RMNHomeConfig
+}
+
+type RmnHomePoller struct {
```
dimkouv, line 48: maybe a comment?

e.g.
```
RmnHomePoller runs in the background and...
```

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+	wg                   sync.WaitGroup
+	stopCh               services.StopChan
+	sync                 services.StateMachine
+	contractReader       contractreader.ContractReaderFacade
+	rmnHomeBoundContract types.BoundContract
+	lggr                 logger.Logger
+	mutex                *sync.RWMutex
```
dimkouv, line 55: I recommend grouping things like `wg,stopCh, sync, mutex`, very minor

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+		mutex:                &sync.RWMutex{},
+		failedPolls:          0,
+		lggr:                 lggr,
+		pollingDuration:      pollingInterval,
+	}
+}
+
+func (r *RmnHomePoller) Start(ctx context.Context) error {
+	r.lggr.Infow("Start Polling RMNHome")
+	return r.sync.StartOnce(r.Name(), func() error {
```
dimkouv, line 81: I am not familiar with `services.StateMachine`, what happens if we call `Start` twice? returns an error?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+		r.rmnHomeBoundContract.ReadIdentifier(consts.MethodNameGetAllConfigs),
+		primitives.Unconfirmed,
+		map[string]interface{}{},
+		&versionedConfigWithDigests,
+	)
+	if err != nil {
+		return fmt.Errorf("error fetching RMNHomeConfig: %w", err)
+	}
+
+	if len(versionedConfigWithDigests) != 2 {
```
dimkouv, line 130: maybe
```
// expectedConfigs are two, the primary and the secondary...
const expectedConfigs = 2
```

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+		r.lggr.Warnw("expected 2 RMNHomeConfigs, got", "count", len(versionedConfigWithDigests))
+		return fmt.Errorf("expected 2 RMNHomeConfigs, got %d", len(versionedConfigWithDigests))
+	}
+
+	var primaryConfigDigest, secondaryConfigDigest cciptypes.Bytes32
+
+	// check if the versionesconfigwithdigests are set (can be empty)
+	if versionedConfigWithDigests[0].ConfigDigest == (cciptypes.Bytes32{}) {
```
dimkouv, line 138: I believe we have a `cciptypes.Bytes32().IsEmpty()`

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+		r.lggr.Warnw("primary config digest is empty")
+	} else {
+		primaryConfigDigest = versionedConfigWithDigests[0].ConfigDigest
+	}
+
```
dimkouv, line 143: I think we can simplify this with
```
primaryConfigDigest := versionedConfigWithDigests[0].ConfigDigest
if primaryConfigDigest.IsEmpty() {
    log.warn
}
```

dimkouv: because it's empty in any case either if we set it or not

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+	} else {
+		secondaryConfigDigest = versionedConfigWithDigests[1].ConfigDigest
+	}
+
+	r.setRMNHomeState(
+		primaryConfigDigest,
+		secondaryConfigDigest,
+		convertOnChainConfigToRMNHomeChainConfig(r.lggr, versionedConfigWithDigests))
+
+	if len(versionedConfigWithDigests) == 0 {
```
dimkouv, line 155: I think this is always false since few lines above we do
```
if len(versionedConfigWithDigests) != 2 {
   return err
}
```

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+
+	s.primaryConfigDigest = primaryConfigDigest
+	s.secondaryConfigDigest = secondaryConfigDigest
+	s.rmnHomeConfig = rmnHomeConfig
+}
+
+func (r *RmnHomePoller) GetRMNNodesInfo(configDigest cciptypes.Bytes32) ([]rmntypes.RMNHomeNodeInfo, error) {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+	return r.rmnHomeState.rmnHomeConfig[configDigest].Nodes, nil
```
dimkouv, line 180: check if `configDigest` exists in rmnHomeConfig and return err if not?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+	_, ok := r.rmnHomeState.rmnHomeConfig[configDigest]
+	return ok, nil
+}
+
+func (r *RmnHomePoller) GetMinObservers(configDigest cciptypes.Bytes32) (map[cciptypes.ChainSelector]uint64, error) {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+	return r.rmnHomeState.rmnHomeConfig[configDigest].MinObservers, nil
```
dimkouv, line 193: return err if not exists?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+func (r *RmnHomePoller) GetOffChainConfig(configDigest cciptypes.Bytes32) (cciptypes.Bytes, error) {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+	return r.rmnHomeState.rmnHomeConfig[configDigest].OffchainConfig, nil
```
dimkouv, line 199: same

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+) map[cciptypes.Bytes32]rmntypes.RMNHomeConfig {
+	if len(versionedConfigWithDigests) == 0 {
+		lggr.Warnw("no on chain RMNHomeConfigs found")
+		return map[cciptypes.Bytes32]rmntypes.RMNHomeConfig{}
+	}
+
+	rmnHomeConfigs := make(map[cciptypes.Bytes32]rmntypes.RMNHomeConfig)
+	for _, versionedConfigWithDigest := range versionedConfigWithDigests {
+		// check if the versionesconfigwithdigests are set (can be empty)
+		if versionedConfigWithDigest.ConfigDigest == (cciptypes.Bytes32{}) {
```
dimkouv, line 241: Maybe create a separate function for doing validations and keep this to only translate one struct to the other?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,311 @@
...
+
+func (r *RmnHomePoller) Close() error {
+	return r.sync.StopOnce(r.Name(), func() error {
+		defer r.wg.Wait()
+		close(r.stopCh)
+		return nil
+	})
+}
+
+func (r *RmnHomePoller) Ready() error {
```
dimkouv, line 210: Would be nice to unit test this func

commit/merkleroot/rmn/types/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"crypto/ed25519"
+
+	mapset "github.com/deckarep/golang-set/v2"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
+
+type NodeID uint32
```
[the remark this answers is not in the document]
0xnogo: Yes as mentioned in the comment, we will have to clean it up. I decided to leave in order to reduce the scope of the PR.

commit/merkleroot/rmn/types/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+// RMNHomeConfig contains the configuration fetched from the RMNHome contract.
+type RMNHomeConfig struct {
+	Nodes          []RMNHomeNodeInfo
+	MinObservers   map[cciptypes.ChainSelector]uint64
```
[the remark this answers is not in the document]
0xnogo: The rename makes sense. Let's have the variable self-explained.

commit/merkleroot/rmn/types/config.go
```diff
@@ -0,0 +1,78 @@
...
+	SupportedSourceChains     mapset.Set[cciptypes.ChainSelector]
+	SignObservationsPublicKey *ed25519.PublicKey // offchainPublicKey
+}
+
+// RMNRemoteConfig contains the configuration fetched from the RMNRemote contract.
+type RMNRemoteConfig struct {
+	ContractAddress  cciptypes.Bytes
+	ConfigDigest     cciptypes.Bytes32
+	Signers          []RMNRemoteSignerInfo
+	MinSigners       uint64
```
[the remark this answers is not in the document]
0xnogo: I am not sure yet how this is going to work. That would be the ideal case as we would not have to deal with figuring out what is the correct remote config to apply. Let's sync on that when I start the work on RMNRemote

commit/merkleroot/rmn/types/config.go
0xnogo: This one is used as an interface between the contract and the reader. It's actually handy to have it exporter as it is used in the tests.

I wonder if we going to have these struct in the gethwrappers? This would allow us to remote them.

internal/reader/rmn_home.go
0xnogo: Good reading improvement.

internal/reader/rmn_home.go
0xnogo: Exactly we have a check one the state to verify that has never been started

internal/reader/rmn_home.go
0xnogo: Valid point

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+type rmnHomeState struct {
+	primaryConfigDigest   cciptypes.Bytes32
+	secondaryConfigDigest cciptypes.Bytes32
+	rmnHomeConfig         map[cciptypes.Bytes32]rmntypes.RMNHomeConfig
+}
+
+// RmnHomePoller polls the RMNHome contract for the latest RMNHomeConfigs
+// It is running in the backdoung with a polling interval of pollingDuration
```
dimkouv, line 50: typo: `backdoung`

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	primaryConfigDigest := versionedConfigWithDigests[0].ConfigDigest
+	if primaryConfigDigest.IsEmpty() {
+		r.lggr.Warnw("primary config digest is empty")
+	}
+
+	secondaryConfigDigest := versionedConfigWithDigests[1].ConfigDigest
+	if secondaryConfigDigest.IsEmpty() {
+		r.lggr.Warnw("secondary config digest is empty")
```
dimkouv, line 145: I don't know exactly how this works, but would be nice to not have `warn` logs when everything seems ok. Are we always expecting to have a secondary config digest?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,331 @@
...
+	s.secondaryConfigDigest = secondaryConfigDigest
+	s.rmnHomeConfig = rmnHomeConfig
+}
+
+func (r *RmnHomePoller) GetRMNNodesInfo(configDigest cciptypes.Bytes32) ([]rmntypes.RMNHomeNodeInfo, error) {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+	_, ok := r.rmnHomeState.rmnHomeConfig[configDigest]
+	if !ok {
+		if !ok {
```
dimkouv, line 174: if !ok { if !ok { .. } }

internal/reader/rmn_home.go
0xnogo: Just a disclaimer, contracts are not ready. But from the what I can tell, primary is acting as blue and secondary as green. So in most of the cases, we are going to have only 1 config set. But during migration 2 are going to set until we check that everything is working properly and then we are going to remove the previous config.

internal/reader/rmn_home_test.go
```diff
@@ -0,0 +1,372 @@
...
+)
+
+var (
+	rmnHomeBoundContract = types.BoundContract{
+		Address: "0xRMNHomeFakeAddress",
+		Name:    consts.ContractNameRMNHome,
+	}
+)
+
+// test the ready method
```
dimkouv, line 30: comment not necessary

internal/reader/rmn_home.go
0xnogo: Yes we can have a debug log in this case.

dimkouv (approved):
Overall looks good 🚀, please resolve the minor comments before merging.

And maybe ask another look for `sync services.StateMachine`, cause I am not familiar with this library and maybe I didn't spot any issues

commit/merkleroot/rmn/types/config.go
makramkd: The structs will be exported in the gethwrappers but we don't have access to gethwrappers here so we have to do this duplication.

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,329 @@
...
+		rmnHomeState:         rmnHomeState{},
+		mutex:                &sync.RWMutex{},
+		failedPolls:          0,
+		lggr:                 lggr,
+		pollingDuration:      pollingInterval,
+	}
+}
+
+func (r *RmnHomePoller) Start(ctx context.Context) error {
+	r.lggr.Infow("Start Polling RMNHome")
```
makramkd, line 83: Maybe move this log inside the `StartOnce`?

pkg/consts/consts.go
```diff
@@ -73,6 +75,10 @@ const (
 	// Used by the home chain reader.
 	MethodNameGetAllChainConfigs = "GetAllChainConfigs"
 	MethodNameGetOCRConfig       = "GetOCRConfig"
+
+	// RMNHome.sol methods
+	// Used by the rmn home reader.
+	MethodNameGetAllConfigs = "getAllConfigs"
```
makramkd, line 81:
```suggestion
	MethodNameGetAllConfigs = "GetAllConfigs"
```

In theory this doesn't have to match the chain-specific name, so lets make it "generic"

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,329 @@
...
+
+type rmnHomeState struct {
+	primaryConfigDigest   cciptypes.Bytes32
+	secondaryConfigDigest cciptypes.Bytes32
+	rmnHomeConfig         map[cciptypes.Bytes32]rmntypes.RMNHomeConfig
+}
+
+// RmnHomePoller polls the RMNHome contract for the latest RMNHomeConfigs
+// It is running in the background with a polling interval of pollingDuration
+type RmnHomePoller struct {
```
makramkd, line 51: Can be private

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,329 @@
...
+		rmnHomeBoundContract: rmnHomeBoundContract,
+		rmnHomeState:         rmnHomeState{},
+		mutex:                &sync.RWMutex{},
+		failedPolls:          0,
+		lggr:                 lggr,
+		pollingDuration:      pollingInterval,
+	}
+}
+
+func (r *RmnHomePoller) Start(ctx context.Context) error {
```
makramkd, line 82: Where do we expect to call this?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,329 @@
...
+		select {
+		case <-ctx.Done():
+			r.mutex.Lock()
+			r.failedPolls = 0
+			r.mutex.Unlock()
+			return
+		case <-ticker.C:
+			if err := r.fetchAndSetRmnHomeConfigs(ctx); err != nil {
+				r.mutex.Lock()
+				r.failedPolls++
```
makramkd, line 113: Should this be reset if `err == nil`?

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,329 @@
...
+		r.rmnHomeBoundContract.ReadIdentifier(consts.MethodNameGetAllConfigs),
+		primitives.Unconfirmed,
+		map[string]interface{}{},
+		&versionedConfigWithDigests,
+	)
+	if err != nil {
+		return fmt.Errorf("error fetching RMNHomeConfig: %w", err)
+	}
+
+	if len(versionedConfigWithDigests) != expectedRMNHomeConfigs {
```
makramkd, line 133: I don't think this check is right, we won't always have 2 configs necessarily.

commit/merkleroot/rmn/types/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+// RMNRemoteSignerInfo contains information about a signer from the RMNRemote contract.
+type RMNRemoteSignerInfo struct {
+	SignerOnchainAddress  cciptypes.Bytes // The signer's onchain address, used to verify report signature
+	NodeIndex             uint64          // The index of the node in the RMN config
+	SignObservationPrefix string          // The prefix of the observation to sign
+}
+
+// VersionedConfigWithDigest mirrors RMNHome.sol's VersionedConfigWithDigest struct
+type VersionedConfigWithDigest struct {
```
makramkd, line 47: Can we move the mirror types to a separate file, or perhaps in the same file as the rmn_home.go poller

commit/merkleroot/rmn/types/config.go
asoliman92: What Makram said and we should have it generic enough as well. What if we change home chain to be non EVM in the future.

internal/reader/rmn_home.go
asoliman92: This should be similar to HomeChain. Expose RMNHome and New function in [pkg](https://github.com/smartcontractkit/chainlink-ccip/blob/631f15febe7ccea6b0648653b7fe1b5c6d4f0167/pkg/reader/home_chain.go#L22-L22). Then use them from [delegate](https://github.com/smartcontractkit/chainlink/blob/e6c11f2353e0bda76c4c60d700c808942f88b5de/core/capabilities/ccip/delegate.go#L158-L158).

makramkd: No, this won't be like HomeChain. We will have a separate RMNHomeReader for each plugin instance.

asoliman92 (approved):

internal/reader/rmn_home.go
0xnogo: This would be expcted to be called in the commit `NewPluginFactory`. We would also need to call `Start` after it.

It's going to be in my next PR, I just need to feature flag it: https://github.com/smartcontractkit/chainlink-ccip/pull/140

makramkd: And `Stop` will be called by the plugin, in the `Close` method?

internal/reader/rmn_home.go
0xnogo:
```
function getAllConfigs()
    external
    view
    returns (VersionedConfigWithDigest memory primaryConfig, VersionedConfigWithDigest memory secondaryConfig)
```    
It looks to me that it will always return 2 objects, hence the empty checks later on.

makramkd: Oh I see, interesting.

internal/reader/rmn_home.go
0xnogo: Correct

internal/reader/rmn_home.go
0xnogo: We are reporting the failed report in a raw, so you are correct. Good catch.

internal/reader/rmn_home.go
```diff
@@ -0,0 +1,371 @@
...
+			SourceChainMinObservers: minObservers,
+			ConfigDigest:            versionedConfigWithDigest.ConfigDigest,
+			OffchainConfig:          config.OffchainConfig,
+		}
+	}
+	return rmnHomeConfigs
+}
+
+// IsNodeObserver checks if a node is an observer for the given source chain
+func IsNodeObserver(sourceChain SourceChain, nodeIndex int, totalNodes int) (bool, error) {
```
makramkd, line 311: This method is no doubt useful, but does it make more sense as a method on `SourceChain` maybe?

makramkd (approved):

PR #138, opened 2024-09-19
Add Extended versions of QueryKey and GetLatestValue.

winder:
Reduce repetitive and error prone API code.

Specifically, this prevents an error which has happened once already where different chain selectors were used due to copy/paste of boilerplate code:
image


pkg/contractreader/extended.go
```diff
@@ -7,15 +7,38 @@ import (
...
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
 
 	"github.com/smartcontractkit/chainlink-ccip/internal/libs/slicelib"
 	"github.com/smartcontractkit/chainlink-ccip/internal/reader/contractreader"
 )
 
 // Extended version of a ContractReader.
 type Extended interface {
-	contractreader.ContractReaderFacade
```
winder, line 17: We don't use the entire interface.

makramkd (approved):

PR #140, opened 2024-09-19
Accept home chain selector id in commit plugin factory

0xnogo:
Accept the hom chain selector id in the under the `rmnEnabled` flag (otherwise integration test will fail as the RMNHome contract is not ready + RMNHomeAddress is not yet defined in OCR3Config.

The Commit factory will instantiate a `RMNHome reader` per plugin instance as the `rmnHomeAddress` is part of the OCR3Config, hence not uniquely defined ([1 per don id](https://github.com/smartcontractkit/ccip/blob/ee6a5901eb99a369564fcb46250f3a560059db6a/contracts/src/v0.8/ccip/capability/CCIPConfig.sol#L82))

The build of `chainlink` is failing: here is the PR to mirror the interface changes made in this PR: https://github.com/smartcontractkit/chainlink/pull/14500.


commit/factory.go
```diff
@@ -104,6 +110,32 @@ func (p *PluginFactory) NewReportingPlugin(config ocr3types.ReportingPluginConfi
...
+	if rmnEnabled {
+		rmnHomeAddress := p.ocrConfig.Config.RmnHomeAddress
+		rmnCr, ok := p.contractReaders[p.homeChainSelector]
+		if !ok {
+			return nil,
+				ocr3types.ReportingPluginInfo{},
+				fmt.Errorf("failed to find contract reader for home chain %d", p.homeChainSelector)
+		}
+		rmnHomeBoundContract := types.BoundContract{
+			Address: "0x" + hex.EncodeToString(rmnHomeAddress),
```
makramkd, line 124: Lets use `typeconv.AddressBytesToString`

makramkd: It will need to be standardized later, better to use it now so it can correctly update all occurrences.

internal/reader/home_chain.go
```diff
@@ -358,6 +358,7 @@ type OCR3Config struct {
 	F                     uint8                   `json:"F"`
 	OffchainConfigVersion uint64                  `json:"offchainConfigVersion"`
 	OfframpAddress        []byte                  `json:"offrampAddress"`
+	RmnHomeAddress        []byte                  `json:"rmnHomeAddress"`
```
makramkd, line 361: This will cause all tests to fail because this is not on-chain yet. I think we should remove this for now and maybe add later when its finally in there?

0xnogo: Would it be just empty? We are not using it (under the rmn feature flag).

It seems like ccip integ test are passing here: https://github.com/smartcontractkit/chainlink/pull/14500

makramkd (approved):

PR #141, opened 2024-09-19
 Add DiscoveryProcessor to commit plugin. 

winder:
Add contract discovery to the commit plugin.

### DIFF
--- a/commit/plugin.go
+++ b/commit/plugin.go
@@ -209,6 +209,7 @@ func (p *Plugin) Observation(
 		MerkleRootObs: merkleRootObs,
 		TokenPriceObs: tokenPriceObs,
 		ChainFeeObs:   chainFeeObs,
+		DiscoveryObs:  discoveryObs,
 		FChain:        fChain,
 	}
 	return obs.Encode()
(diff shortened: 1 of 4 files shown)


makramkd (approved):

commit/plugin.go
```diff
@@ -164,6 +180,18 @@ func (p *Plugin) Observation(
 		return nil, fmt.Errorf("decode query: %w", err)
 	}
 
+	var discoveryObs dt.Observation
+	if p.discoveryProcessor != nil {
+		discoveryObs, err = p.discoveryProcessor.Observation(ctx, dt.Outcome{}, dt.Query{})
+		if err != nil {
+			p.lggr.Errorw("failed to discover contracts", "err", err)
+		}
+		if !p.contractsInitialized {
```
0xnogo, line 189: If contract addresses change, is there a risk of having stale addresses?

makramkd: Contract addresses won't change but we could have new chains added while the plugin is running.

However, the discovery processor continuously discovers new addresses at the moment, so these new chains get picked up. There's also a test for this in chainlink.

winder: A new address would be a problem because we don't unbind the old one and currently ensure that there is only one bound address. This is something we could support pretty easily if we need to in the future.

PR #145, opened 2024-09-20
only transmit if blue config

makramkd:
Ticket: https://smartcontract-it.atlassian.net/browse/CCIP-3373

Chainlink PR: https://github.com/smartcontractkit/chainlink/pull/14506

### DIFF
--- a/commit/factory.go
+++ b/commit/factory.go
@@ -56,6 +56,7 @@ func (p PluginFactoryConstructor) NewValidationService(ctx context.Context) (cor
 // PluginFactory implements common ReportingPluginFactory and is used for (re-)initializing commit plugin instances.
 type PluginFactory struct {
 	lggr            logger.Logger
+	donID           uint32
 	ocrConfig       reader.OCR3ConfigWithMeta
 	commitCodec     cciptypes.CommitPluginCodec
 	msgHasher       cciptypes.MessageHasher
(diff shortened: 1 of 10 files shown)


commit/factory.go
```diff
@@ -66,6 +67,7 @@ type PluginFactory struct {
 
 func NewPluginFactory(
 	lggr logger.Logger,
+	donID uint32,
```
dimkouv, line 70: do we need a type alias for `uint32`? e.g. `type DonID uint32`

commit/report.go
```diff
@@ -104,3 +112,13 @@ func (p *Plugin) ShouldTransmitAcceptedReport(
 	)
 	return true, nil
 }
+
+func (p *Plugin) isGreenInstance(ctx context.Context) bool {
+	ocrConfigs, err := p.homeChain.GetOCRConfigs(ctx, p.donID, consts.PluginTypeCommit)
+	if err != nil {
+		p.lggr.Errorw("failed to get ocr configs from home chain", "err", err)
+		return false
```
dimkouv, line 120: i think we need an error here, to distinguish between `isGreenInstance` and actual `error`. Because we might assume it's blue while it's green.

execute/plugin.go
```diff
@@ -590,6 +601,16 @@ func (p *Plugin) ShouldTransmitAcceptedReport(
 	return true, nil
 }
 
+func (p *Plugin) isGreenInstance(ctx context.Context) bool {
+	ocrConfigs, err := p.homeChain.GetOCRConfigs(ctx, p.donID, consts.PluginTypeExecute)
+	if err != nil {
+		p.lggr.Errorw("failed to get ocr configs from home chain", "err", err)
+		return false
```
dimkouv, line 608: same here, return an err

commit/factory.go
makramkd: Sure, can do a type alias.

dimkouv (approved):

PR #147, opened 2024-09-20
CCIP-2881 Implementing USDCReader 

mateusz-sekara:
USDCReader fetches matching `MessageSent(bytes)` events from the Circle's Message Transmiters. We don't rely on events ordering as we do in v1 implementation but rather fetch events by nonce similar to the PoC created here https://github.com/smartcontractkit/ccip/pull/1025/files

Integration tests in Chainlink repo https://github.com/smartcontractkit/chainlink/pull/14516


internal/reader/contractreader/contractreader_facade.go
```diff
@@ -18,4 +18,9 @@ type ContractReaderFacade interface {
 	Unbind(ctx context.Context, bindings []types.BoundContract) error
 	QueryKey(ctx context.Context, contract types.BoundContract, filter query.KeyFilter, limitAndSort query.LimitAndSort, sequenceDataType any) ([]types.Sequence, error)
 	//mustEmbedUnimplementedContractReaderServer()
+	Name() string
```
mateusz-sekara, line 21: This make contractreader_facade to implement ContractReader interface

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,308 @@
...
+		source, dest cciptypes.ChainSelector,
+		tokens map[exectypes.MessageTokenID]cciptypes.RampTokenAmount,
+	) (map[exectypes.MessageTokenID]MessageHash, error)
+}
+
+const (
+	CCTPMessageVersion = uint32(0)
+)
+
+// TODO, this should be fetched from USDC Token Pool and cached
```
mateusz-sekara, line 32: Separate PR, this values are fixed, but we should fetch them from chain instead of hardcoding to rely on a single source of the truth

execute/tokendata/usdc/attestation.go
```diff
@@ -32,8 +35,8 @@ type AttestationStatus struct {
 type AttestationClient interface {
 	Attestations(
 		ctx context.Context,
-		msgs map[cciptypes.ChainSelector]map[cciptypes.SeqNum]map[int][]byte,
-	) (map[cciptypes.ChainSelector]map[cciptypes.SeqNum]map[int]AttestationStatus, error)
+		msgs map[cciptypes.ChainSelector]map[exectypes.MessageTokenID]reader.MessageHash,
+	) (map[cciptypes.ChainSelector]map[exectypes.MessageTokenID]AttestationStatus, error)
```
winder, line 39: Moving away from `map[int][]byte` means you'll need to update the `ValidateObservation` function to check that you don't have `MessageTokenID`'s.

execute/exectypes/observation.go
```diff
@@ -126,6 +127,22 @@ func (td TokenData) IsReady() bool {
...
+}
+
+func NewMessageTokenID(seqNr cciptypes.SeqNum, index int) MessageTokenID {
+	return MessageTokenID{SeqNr: seqNr, Index: index}
+}
+
+func (mti MessageTokenID) String() string {
+	return fmt.Sprintf("%d_%d", mti.SeqNr, mti.Index)
+}
+
```
winder, line 145: There are over 100 lines of token data types, maybe worth putting them in their own file.

At this point, it may also be helpful to have some sort of builder to manage working with this complex object. That should allow you to stick with the unwieldy triple-map type without hurting the business logic code so much.

internal/reader/contractreader/contractreader_facade.go
winder: Why do we want that? You could also accomplish that by embedding `service.Service`, [which is how the ContractReader does it](https://github.com/smartcontractkit/chainlink-common/blob/d4cf7aff85cb3f2a0630bc2a11b42cfcbb80c40c/pkg/types/contract_reader.go#L22).

I thought we left out `Unbind` to narrow the interface even.

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,308 @@
...
+	sel.AVALANCHE_MAINNET.Selector:           1,
+	sel.ETHEREUM_MAINNET_OPTIMISM_1.Selector: 2,
+	sel.ETHEREUM_MAINNET_ARBITRUM_1.Selector: 3,
+	sel.ETHEREUM_MAINNET_BASE_1.Selector:     6,
+	sel.POLYGON_MAINNET.Selector:             7,
+	sel.AVALANCHE_TESTNET_FUJI.Selector:      1,
+}
+
+type usdcMessageReader struct {
+	contractReaders map[cciptypes.ChainSelector]types.ContractReader
```
winder, line 44:
```suggestion
	contractReaders map[cciptypes.ChainSelector]contractreader.ContractReaderFacade
```

winder: I'm curious what @asoliman92  things, but at this stage I would prefer we avoid `types.ContractReader` everywhere except for `CCIPReader`. Especially with the new `DiscoveryProcessor` I think it will make things a lot simpler.

execute/exectypes/observation.go
mateusz-sekara: Actually, I wanted to move away from this triple map monster. I guess it's easier to use SeqNr/Index identifier as a single item identifying the token, instead of using these maps everywhere. That saves one nested for loop during the processing ;)

pkg/reader/usdc_reader.go
mateusz-sekara: There is one major issue with `ContractReaderFacade`, it's in the internal package, I can't use that in the integration tests in the Chainlink repository.  Also, we don't need that facade to be used there, because we can rely on the pure interfaces. DiscoveryProcessor is not going to be a thing for USDCReader, because contracts' addresses are passed using the config.

I've added missing methods to the `ContractReaderFacade` and it can be used interchangeably with the `ContractReader`. Mocking works without any additional magic

mateusz-sekara: > I would prefer we avoid types.ContractReader everywhere except for CCIPReader.

What's the reasoning behind this?

internal/reader/contractreader/contractreader_facade.go
mateusz-sekara: Oh, nice it's gonna be cleaner to just embed `services.Service`

I've added it to do something like this in tests
```go
	emptyReader := reader.NewMockContractReaderFacade(t)
	faultyReader := reader.NewMockContractReaderFacade(t)
	validReader := reader.NewMockContractReaderFacade(t)


	contactReaders := map[cciptypes.ChainSelector]types.ContractReader{
		faultyChain: faultyReader,
		emptyChain:  emptyReader,
		validChain:  validReader,
	}
```

execute/tokendata/usdc/attestation.go
mateusz-sekara: This type is used only as an intermediate step in the usdc processing. There are no changes to how observations look like, no changes to the business logic required as well. Am I missing sth?

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,308 @@
...
+				append(
+					append(
+						append(
+							msgVersionBytes[:],
+							sourceDomainBytes[:]...),
+						destDomainBytes[:]...),
+					nonceBytes[:]...),
+				senderBytes[:]...),
+		)
+		messageTransmitterEvents[id] = messageTransmittedEvent
```
winder, line 205: The nested appends were a bit confusing for me, suggest byte buffer: 
```
		var buf bytes.Buffer
		buf.Write(msgVersionBytes[:])
		buf.Write(sourceDomainBytes[:])
		buf.Write(destDomainBytes[:])
		buf.Write(nonceBytes[:])
		buf.Write(senderBytes[:])
		messageTransmitterEvents[id] = buf.Bytes()[0:32]
```

Or if you want to be precise about the packing:
```
		messageTransmittedEvent := [32]byte{}
		copy(messageTransmittedEvent2[:], msgVersionBytes[:])
		copy(messageTransmittedEvent2[4:], sourceDomainBytes[:])
		copy(messageTransmittedEvent2[8:], destDomainBytes[:])
		copy(messageTransmittedEvent2[12:], nonceBytes[:])
		copy(messageTransmittedEvent2[20:], senderBytes[:])
		messageTransmitterEvents[id] = messageTransmittedEvent
```
https://play.golang.com/p/F3t5uBny_VJ

execute/tokendata/usdc/attestation.go
winder: I must be the one missing something. Just want to make sure we don't have illegal observations.

We have a problem in some types like this:
```
some_observation_data: []{
   { chain:1, data:...},
   { chain:2, data:...},
   { chain:1, data:...}, // illegal but not enforced by the type system so we have to validate manually
}
```

execute/exectypes/observation.go
winder: Anything to simplify the usage of this thing is great, just want to make sure we don't lose type safety. may still be worth moving these to a separate file.

mateusz-sekara: Agree, already moved

execute/tokendata/usdc/attestation.go
mateusz-sekara: I don't think we introduced any regression here, but additional tests and wiring up together with ocr is going to be another pr ;)

internal/reader/rmn_home.go
```diff
@@ -10,16 +10,15 @@ import (
...
-
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
 
+	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/contractreader"
+
```
winder, line 21:
```suggestion
```

pkg/contractreader/extended_test.go
```diff
@@ -1,22 +1,22 @@
-package contractreader
+package contractreader_test
```
winder, line 1: :sweat_smile:

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,307 @@
...
+	if err := r.Bind(ctx, []types.BoundContract{contract}); err != nil {
+		return empty,
+			fmt.Errorf(
+				"unable to bind %s for chain %d: %w",
+				consts.ContractNameCCTPMessageTransmitter, chainSel, err,
+			)
+	}
+
+	return contract, nil
+}
```
winder, line 232: This is nearly identical to the generalized version `bindReaderContract` in `ccipChainReader`. Maybe worth unifying?

One difference is that the address in that version is the []byte version rather than the encoded version.

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,307 @@
...
+
+		senderBytes := [12]byte{}
+
+		var buf bytes.Buffer
+		buf.Write(msgVersionBytes[:])
+		buf.Write(sourceDomainBytes[:])
+		buf.Write(destDomainBytes[:])
+		buf.Write(nonceBytes[:])
+		buf.Write(senderBytes[:])
+		messageTransmitterEvents[id] = [32]byte(buf.Bytes()[:32])
```
winder, line 202: Found another option here, you can combine the appending with the BigEndian call
```suggestion
		destDomain, ok := u.cctpDestDomain[uint64(destChainSelector)]
		if !ok {
			return nil, fmt.Errorf("destination domain not found for chain %s", destChainSelector)
		}
		
		var buf []byte
		buf = binary.BigEndian.AppendUint32(buf, CCTPMessageVersion)
		buf = binary.BigEndian.AppendUint32(buf, sourceTokenPayload.SourceDomain)
		buf = binary.BigEndian.AppendUint32(buf, destDomain)
		buf = binary.BigEndian.AppendUint64(buf, sourceTokenPayload.Nonce)
		buf = append(buf2, v_senderBytes[:]...)
		
		messageTransmitterEvents[id] = [32]byte(buf[:32])
```

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,307 @@
...
+			return nil, fmt.Errorf("destination domain not found for chain %s", destChainSelector)
+		}
+
+		destDomainBytes := [4]byte{}
+		binary.BigEndian.PutUint32(destDomainBytes[:], destDomain)
+
+		nonceBytes := [8]byte{}
+		binary.BigEndian.PutUint64(nonceBytes[:], sourceTokenPayload.Nonce)
+
+		senderBytes := [12]byte{}
```
winder, line 194: No sender?

mateusz-sekara: It's always empty for evm chains

pkg/contractreader/extended_test.go
mateusz-sekara: This is because of mocking, I need to have a separate package to avoid cyclic dependency.

pkg/reader/helpers.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	typeconv "github.com/smartcontractkit/chainlink-ccip/internal/libs/typeconv"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/contractreader"
+)
+
+// bindable is a helper interface to represent all different types of contract readers.
+// We expose per-reader type functions, but then cast to this common interface to avoid code duplication.
+type bindable interface {
```
mateusz-sekara, line 16: I couldn't find a better way to do it. It's because we have contractreaderFacade and contracreaderExtended which are different interfaces. I created dedicated function which only convert that to common type and execute the shared logic @winder

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,282 @@
...
+
+// TODO, this should be fetched from USDC Token Pool and cached
+var CCTPDestDomains = map[uint64]uint32{
+	sel.ETHEREUM_MAINNET.Selector:                    0,
+	sel.AVALANCHE_MAINNET.Selector:                   1,
+	sel.ETHEREUM_MAINNET_OPTIMISM_1.Selector:         2,
+	sel.ETHEREUM_MAINNET_ARBITRUM_1.Selector:         3,
+	sel.ETHEREUM_MAINNET_BASE_1.Selector:             6,
+	sel.POLYGON_MAINNET.Selector:                     7,
+	sel.ETHEREUM_TESTNET_SEPOLIA.Selector:            0,
```
asoliman92, line 42: Does the token pool have same values for testnets as mainnets? can't this introduce some mix between mainnet and testnet?

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,282 @@
...
+	// 3. This should be done by ChainReader - picking only events matching eventIDs.
+	// Right now ChainReader doesn't support filtering by specific data words
+	out := make(map[exectypes.MessageTokenID]MessageHash)
+	for tokenID, messageID := range eventIDs {
+		messageHash, ok1 := messageSentEvents[messageID]
+		if !ok1 {
+			// Token not available in the source chain
+			continue
+		}
+		out[tokenID] = messageHash
```
asoliman92, line 162: double checking, the tokens will be always in same order? so out[0] will always mean the same token?

pkg/reader/usdc_reader.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	// 2. Query the MessageTransmitter contract for the MessageSent events based on the 3rd words.
+	// We need entire MessageSent payload to use that with the Attestation API
+	cr, ok := u.boundContracts[source]
+	if !ok {
+		return nil, fmt.Errorf("no contract bound for chain %d", source)
+	}
+
+	iter, err := u.contractReaders[source].QueryKey(
```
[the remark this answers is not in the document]
mateusz-sekara: I'm not sure if it's relevant at this stage. @ilija42 promised to add filtering by words directly and that's all we need here ;)

pkg/reader/usdc_reader.go
mateusz-sekara: Token Pool doesn't, but CCTP does. TokenPool knows only about either mainnet or testnet. However, CCTP uses the same values for the particular chain, regarding on whether it's mainnet and testnet

pkg/reader/usdc_reader.go
mateusz-sekara: Hmm, we do have a map here, so ordering doesn't really matter

mateusz-sekara: `tokenId` is a composite key as you suggested in a previous PR :P
```go
type MessageTokenID struct {
	SeqNr cciptypes.SeqNum
	Index int
}
```

pkg/reader/usdc_reader.go
asoliman92: Can you add a TODO as well to make it family agnostic? We should be checking these TODOs later but you'll find them in multiple places here and chainlink repo

pkg/reader/usdc_reader.go
mateusz-sekara: Meaning that Arbitrum Mainnet and Arbitrum Testnet has always domain set to 3

mateusz-sekara: Btw, these values are replicated from CCTP contracts, we only keep the mapping from our selectors to their domains, so there is not much we can do about conflicting values. This is how they implemented that 🤷

pkg/reader/usdc_reader.go
mateusz-sekara: To be honest entire USDCReader is very evm-specific because it relies on the structure dynamic structure of CCTP we fetched from Circle repository. Also CCTP is present only on evm-chains. If they decide to expand to non-evm chains we would need to solve that problem anyway. 

However, we should be able to apply a similar composite pattern as we have for TokenDataObserver
```go

usdcReader := NewCompositeUSDCReader(
    evmUSDCReader(supportedChains), <- current implementation
    solanaUSDCReader(solanaChainId) <- new implementation
)
```

Not sure if this todo item is important at this stage, but I can add it if you want it

pkg/reader/usdc_reader_test.go
```diff
@@ -0,0 +1,360 @@
...
+			name:           "should return empty dataset when chain doesn't have events",
+			sourceSelector: emptyChain,
+			destSelector:   faultyChain,
+			expectedMsgIDs: []exectypes.MessageTokenID{},
+		},
+		{
+			name:           "should return error when chain reader errors",
+			sourceSelector: faultyChain,
+			destSelector:   emptyChain,
+			errorMessage:   "error querying contract reader for chain 6433500567565415381",
```
asoliman92, line 201: Nit: use format and use the chain selector variable from faultyChain instead of hardcoding.

pkg/reader/usdc_reader_test.go
```diff
@@ -0,0 +1,360 @@
...
+			name:           "should return error when CCTP domain is not supported",
+			sourceSelector: emptyChain,
+			destSelector:   cciptypes.ChainSelector(2),
+			errorMessage:   "destination domain not found for chain ChainSelector(2)",
+		},
+		{
+			name:           "should return error when CCTP domain is not supported",
+			sourceSelector: cciptypes.ChainSelector(sel.POLYGON_MAINNET.Selector),
+			destSelector:   emptyChain,
+			errorMessage:   "no contract bound for chain 4051577828743386545",
```
asoliman92, line 213: same nit

pkg/reader/usdc_reader_test.go
```diff
@@ -0,0 +1,360 @@
...
+		},
+		{
+			name:           "valid chain return events but nothing is matched",
+			sourceSelector: validChain,
+			destSelector:   emptyChain,
+			expectedMsgIDs: []exectypes.MessageTokenID{},
+		},
```
asoliman92, line 220: Need a test from 2 valid chains and see the expected messages?

pkg/reader/usdc_reader_test.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	contactReaders := map[cciptypes.ChainSelector]contractreader.ContractReaderFacade{
+		faultyChain: faultyReader,
+		emptyChain:  emptyReader,
+		validChain:  validReader,
+	}
+
+	tokens := map[exectypes.MessageTokenID]cciptypes.RampTokenAmount{
```
asoliman92, line 175: can't see where's this used

pkg/reader/usdc_reader_test.go
mateusz-sekara: I don't want to go that deep, getting results will require a ton of mocking and byte operations to match events. It's covered in the integration tests in the chainlink repository, I don't want to reinvent the wheel at this stage. It's a simple unit test with basic flows

pkg/reader/helpers.go
makramkd: This is a pretty confusing set of helpers

winder: This is nice, thanks.

winder (approved):

pkg/reader/usdc_reader.go
makramkd: Having different EVM and Solana USDC readers kinda defeats the purpose of using chain reader, which is supposed to be chain agnostic.

PR #149, opened 2024-09-20
pkg/reader: enforce non-copyability of ccip reader

makramkd:
The CCIP reader is a shared object used by all processors from within the plugin.

In order to enforce it always being passed by pointer ensure that it is non-copyable by embedding a mutex.

Note that trying to copy the object will not cause compilation failures but linting failures.

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -30,6 +30,9 @@ import (
 // TODO: unit test the implementation when the actual contract reader and writer interfaces are finalized and mocks
 // can be generated.
 type ccipChainReader struct {
+	// to prevent copying this struct.
+	_ sync.Mutex
+
 	lggr            logger.Logger
 	contractReaders map[cciptypes.ChainSelector]contractreader.Extended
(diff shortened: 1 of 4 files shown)


pkg/reader/ccip.go
```diff
@@ -30,6 +30,9 @@ import (
 // TODO: unit test the implementation when the actual contract reader and writer interfaces are finalized and mocks
 // can be generated.
 type ccipChainReader struct {
+	// to prevent copying this struct.
+	_ sync.Mutex
```
winder, line 34: Are you sure this works?

winder (approved):

pkg/reader/ccip.go
winder: Sounds like there are some linters that prevent the copying, so better than nothing.

winder:
Pretty sure we don't need this because ptr receivers are used everywhere.

pkg/reader/ccip.go
makramkd: Yeah, its linter based, not build based unfortunately.

PR #152, opened 2024-09-20
.github/workflows: run add chain test

makramkd:
Run the `TestAddChainInbound` test from the ccip deployment package so that we ensure that our discovery processor is working as expected.

### DIFF
--- a/.github/workflows/ccip-integration-test.yml
+++ b/.github/workflows/ccip-integration-test.yml
@@ -65,10 +65,23 @@ jobs:
           ./ccip.test local db preparetest
         env:
           CL_DATABASE_URL: ${{ env.DB_URL }}
-      - name: Run ccip ocr3 integration test
+      - name: Run ccip ocr3 initial deploy integration test
         run: |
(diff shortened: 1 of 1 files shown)


asoliman92 (commented):
Can't we just run all ccip tests under deployment instead of adding one by one?

makramkd:
Also, there is a non-trivial subset of the tests that are not really relevant to this codebase.

dimkouv:
If it makes sense we can de-dup this
```
func run_integ_test(test_name)
          cd $GITHUB_WORKSPACE/chainlink/integration-tests
          go test ... $test_name ...
          EXITCODE=${PIPESTATUS[0]}
          if [ $EXITCODE -ne 0 ]; then
            echo "Integration test failed"
          else
            echo "Integration test passed!"
          fi
          exit $EXITCODE
        env:
          CL_DATABASE_URL: ${{ env.DB_URL }}
```

makramkd:
Are there functions in yaml? 😛

dimkouv:
No, but I think the preferred approach is to have a `scripts/` dir and use bash.

PR #154, opened 2024-09-20
Exec v2 scaffolding to observes prices/costs

rstout:
Adds scaffolding to the V2 Exec Plugin to enable observing prices and costs (but does not yet observe these values). Once prices and costs are observed, message fees and costs can be computed, and messages with higher costs than paid fees can be filtered.

The core data type is PriceObservations:

```// PriceObservations contains prices and costs that can be used to compute message fees and execution costs in USD
type PriceObservations struct {
	// The price of Juels (the smallest denomination of LINK) in USD. Each CCIP message has a fee denominated in Juels.
	JuelPriceUSD cciptypes.BigInt `json:"juelPriceUSD"`

	// The price of gas in the destination chain, denoted in the native token of the destination chain
	GasPrice cciptypes.BigInt `json:"gasPrice"`

	// The price of the native token of the destination chain in USD
	DestNativeTokenPriceUSD cciptypes.BigInt `json:"destNativeTokenPriceUSD"`

	// Maps message IDs to the estimated gas cost of executing these messages
	MessageExecutionGasCosts map[string]cciptypes.BigInt `json:"messageExecutionCosts"`
}
```

The message fee in USD will be: `message.jeuls * JuelPriceUSD * feeBoostFactor`
The message execution cost will be `GasPrice * MessageExecutionGasCosts * DestNativeTokenPriceUSD`

### DIFF
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -9,12 +9,13 @@ import (
 	mapset "github.com/deckarep/golang-set/v2"
 	"golang.org/x/exp/maps"
 
-	"github.com/smartcontractkit/chainlink-common/pkg/logger"
-	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 	"github.com/smartcontractkit/libocr/commontypes"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 	libocrtypes "github.com/smartcontractkit/libocr/ragep2p/types"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 
 	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
 	"github.com/smartcontractkit/chainlink-ccip/execute/internal/gas"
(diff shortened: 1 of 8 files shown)


execute/plugin.go
```diff
@@ -285,6 +307,15 @@ func (p *Plugin) Observation(
 			for _, msg := range commitReport.Messages {
 				sender := typeconv.AddressBytesToString(msg.Sender[:], uint64(p.cfg.DestChain))
 				nonceRequestArgs[commitReport.SourceChain][sender] = struct{}{}
+				messages = append(messages, msg)
+			}
+		}
+
+		prices := exectypes.PriceObservations{}
+		if p.messageExecutionEconomicsEnabled {
+			prices, err = p.priceObserver.ObservePrices(messages)
```
makramkd, line 316: Is it really sufficient to just call this here? I believe the price observations are a mix of source and dest information, some nodes may have source info, some may have dest info, wouldn't we need to consensus-ify these values (median, etc.) and only then do the boosting on the medianized values?

makramkd: Or is this stuff going to be done in upcoming PRs?

execute/plugin_functions.go
```diff
@@ -391,6 +454,7 @@ func getConsensusObservation(
 	destChainSelector cciptypes.ChainSelector,
 	F int,
 	fChain map[cciptypes.ChainSelector]int,
+	messageExecutionEconomicsEnabled bool,
```
winder, line 457: I think you could leave this out. The prices would be empty by default on account of no price observations being present. Add a check to VerifiyObservation that there are no price observations when the flag is false.

execute/plugin.go
winder: Medianizing would happen in the outcome phase based on these observations.

winder (approved):
This all looks on track to me.

execute/plugin.go
rstout: > I believe the price observations are a mix of source and dest information

Minor point, but I believe everything should be on the dest chain.

> wouldn't we need to consensus-ify these values (median, etc.) and only then do the boosting on the medianized values?

We do this consensus in `mergePriceObservations` in `execute/plugin_functions.go`

PR #155, opened 2024-09-22
Adding support for RMNRemote discovery

0xnogo:


internal/plugincommon/discovery/processor.go
```diff
@@ -151,6 +154,10 @@ func (cdp *ContractDiscoveryProcessor) Outcome(
 			nonceManagerAddrs,
 			ao.Observation.DestNonceManager,
 		)
+		rmnRemoteAddrs = append(
```
makramkd, line 157: Check for zero length like the above

pkg/reader/ccip.go
```diff
@@ -498,6 +498,9 @@ func (r *ccipChainReader) DiscoverContracts(
 		consts.ContractNameNonceManager: {
 			destChain: staticConfig.NonceManager,
 		},
+		consts.ContractNameRMNRemote: {
+			destChain: staticConfig.RmnProxy,
```
makramkd, line 502: I think this is incorrect, its not called `RmnProxy` anymore but just "rmn"

makramkd: https://github.com/smartcontractkit/chainlink/blob/207b0d502a9cadb600873cce9602bdfacca0e177/contracts/src/v0.8/ccip/offRamp/OffRamp.sol#L91

0xnogo: Good catch

pkg/reader/ccip.go
```diff
@@ -655,7 +658,7 @@ func (r *ccipChainReader) getOfframpStaticConfig(
 //nolint:lll // It's a URL.
 type offrampStaticChainConfig struct {
 	ChainSelector      cciptypes.ChainSelector `json:"chainSelector"`
-	RmnProxy           []byte                  `json:"rmnProxy"`
+	Rmn                []byte                  `json:"rmn"`
```
makramkd, line 661: Hmm tests seem to be failing after this change, maybe i was wrong?

internal/plugincommon/discovery/processor.go
```diff
@@ -147,10 +150,18 @@ func (cdp *ContractDiscoveryProcessor) Outcome(
 			cdp.lggr.Warnf("skipping empty nonce manager address in observation from Oracle %d", ao.OracleID)
 			continue
 		}
+		if len(ao.Observation.RMNRemote) == 0 {
+			cdp.lggr.Warnf("skipping empty RMNRemote address in observation from Oracle %d", ao.OracleID)
+			continue
+		}
```
makramkd, line 156: This will end up skipping a nonce manager that is valid, put this right before the rmnRemoteAddrs append

pkg/reader/ccip.go
0xnogo: The RMNRemote contract is not defined in the dest chain reader defintion. This PR should fix it: https://github.com/smartcontractkit/chainlink/pull/14588

makramkd (approved):

PR #156, opened 2024-09-23
Chain fee processor

asoliman92:


commit/merkleroot/rmn/rmnpb/rmn_offchain.pb.go
```diff
@@ -7,10 +7,11 @@
 package rmnpb
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
+
```
asoliman92, line 12: running linter locally changed this.

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,130 @@
...
+	fChains := plugincommon.GetConsensusMap(p.lggr, "fChain", aggObs.FChain, fMin)
+
+	fDestChain, exists := fChains[p.destChain]
+	if !exists {
+		return ConsensusObservation{},
+			fmt.Errorf("no consensus value for fDestChain, destChain: %d", p.destChain)
+	}
+
+	timestamp := plugincommon.Median(aggObs.Timestamps, plugincommon.TimestampComparator)
+	chainFeeUpdatesConsensus := plugincommon.GetConsensusMapAggregator(
```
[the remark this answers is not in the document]
makramkd: I'm guessing:
* Take the median of the gas price update timestamps observed on the dest
* Take the median ofthe gas price updates themselves

Note that for the gas price updates themselves, they may be two dimensional. I'm not sure if it makes sense to take the median of the 2D number directly vs. taking the median of DA and exec gas prices separately?

commit/chainfee/processor.go
```diff
@@ -3,41 +3,110 @@ package chainfee
...
+		// (dataAvFee * nativeTokenPrice) << 112 | executionFee * nativeTokenPrice
+		// nolint:lll
+		// https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/FeeQuoter.sol#L498
+
+		dataAvailabilityPrice := cciptypes.NewBigIntFromInt64(1).
+			Mul(consensusObs.NativeTokenPrices[chain].Int,
+				feeComp.DataAvailabilityFee)
+		execPrice := cciptypes.NewBigIntFromInt64(1).
+			Mul(consensusObs.NativeTokenPrices[chain].Int,
+				feeComp.ExecutionFee)
```
makramkd, line 93: Units here are pretty important; I think a comment laying out the units + decimals would be helpful for people reading the code to fully understand whats going on

commit/chainfee/observation.go
```diff
@@ -0,0 +1,12 @@
+package chainfee
+
+import cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+func (p *processor) ObserveFChain() map[cciptypes.ChainSelector]int {
+	fChain, err := p.homeChain.GetFChain()
+	if err != nil {
+		p.lggr.Warnw("call to GetFChain failed", "err", err)
```
dimkouv, line 8: Isn't it an actual error? Maybe `Errorw` ?

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugintypes"
+)
+
+func (p *processor) getConsensusObservation(
+	aos []plugincommon.AttributedObservation[Observation],
+) (ConsensusObservation, error) {
+	aggObs := aggregateObservations(aos)
+
+	fMin := mathslib.RepeatedF(func() int { return p.bigF }, maps.Keys(aggObs.FChain))
```
dimkouv, line 21: Without reading `RepeatedF`, I don't understand what's going on here.

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,130 @@
...
+		aggObs.ChainFeePriceUpdates,
+		mathslib.RepeatedF(
+			func() int { return mathslib.TwoFPlus1(fDestChain) },
+			maps.Keys(aggObs.ChainFeePriceUpdates),
+		),
+		plugincommon.TimestampedBigAggregator,
+	)
+
+	// Stop early if earliest updated timestamp is still fresh
+	earliestUpdateTime := plugincommon.EarliestTimestamp(maps.Values(chainFeeUpdatesConsensus), timestamp)
```
dimkouv, line 45: I would expect `EarliestTimestamp` just returns the minimum from the provided slice.
I don't understand what `EarliestTimestamp` is doing with the second argument (`timestamp`).

commit/chainfee/outcome.go
winder: Hopefully this is a little clearer
```suggestion
	fMin := consensus.MakeMultiThreshold(fChains, consensus.TwoFPlus1)
```

commit/chainfee/processor.go
```diff
@@ -3,41 +3,110 @@ package chainfee
...
 )
 
-type Processor struct {
+type processor struct {
+	destChain                        cciptypes.ChainSelector
+	lggr                             logger.Logger
+	homeChain                        reader.HomeChain
+	ccipReader                       readerpkg.CCIPReader
+	ChainFeePriceBatchWriteFrequency commonconfig.Duration
+	bigF                             int
```
winder, line 26: This can be `F`, the other one is standardizing on `fRoleDON`.

commit/chainfee/outcome.go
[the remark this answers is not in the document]
asoliman92: Good Point about the 2D fees. We don't really use them so far so I can just make it return timestamps only as well and worry about the 2D if we need to use them later.

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,119 @@
...
+	donThresh := consensus.MakeConstantThreshold[cciptypes.ChainSelector](consensus.TwoFPlus1(p.fRoleDON))
+	fChains := consensus.GetConsensusMap(p.lggr, "fChain", aggObs.FChain, donThresh)
+
+	fDestChain, exists := fChains[p.destChain]
+	if !exists {
+		return ConsensusObservation{},
+			fmt.Errorf("no consensus value for fDestChain, destChain: %d", p.destChain)
+	}
+
+	timestamp := consensus.Median(aggObs.Timestamps, consensus.TimestampComparator)
```
winder, line 30: Validate the number of observations?

internal/plugincommon/consensus/consensus.go
```diff
@@ -85,10 +87,45 @@ func TimestampComparator(a, b time.Time) bool {
...
 func BigIntComparator(a, b cciptypes.BigInt) bool {
 	return a.Cmp(b.Int) == -1
 }
 
 func TokenPriceComparator(a, b cciptypes.TokenPrice) bool {
 	return a.Price.Int.Cmp(b.Price.Int) == -1
 }
+
+// TimestampedBigAggregator aggregates the fee quoter updates by taking the median of the prices and timestamps
+var TimestampedBigAggregator = func(updates []plugintypes.TimestampedBig) plugintypes.TimestampedBig {
```
winder, line 103:
```suggestion
func TimestampedBigAggregator(updates []plugintypes.TimestampedBig) plugintypes.TimestampedBig {
```

commit/chainfee/types.go
```diff
@@ -1,16 +1,44 @@
...
 }
+
+// ConsensusObservation holds the consensus values for all observations in a round
+type ConsensusObservation struct {
+	FeeComponents         map[cciptypes.ChainSelector]types.ChainFeeComponents `json:"feeComponents"`
+	NativeTokenPrices     map[cciptypes.ChainSelector]cciptypes.BigInt         `json:"nativeTokenPrice"`
+	FChain                map[cciptypes.ChainSelector]int                      `json:"fChain"`
+	ChainFeeLatestUpdates map[cciptypes.ChainSelector]time.Time                `json:"chainFeePriceUpdates"`
+	Timestamp             time.Time                                            `json:"timestamp"`
+}
```
winder, line 42: This appears to be a duplicate of `Observation`

commit/chainfee/types.go
```diff
@@ -1,16 +1,44 @@
...
+
+var EmptyConsensusObservation = ConsensusObservation{}
```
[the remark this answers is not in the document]
asoliman92: I actually ditched using it and forgot to remove it. Thanks!

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,124 @@
...
+		func(vals []types.ChainFeeComponents) types.ChainFeeComponents {
+			executionFees := make([]cciptypes.BigInt, len(vals))
+			dataAvailabilityFees := make([]cciptypes.BigInt, len(vals))
+			for i, feeComp := range vals {
+				executionFees[i] = cciptypes.NewBigInt(feeComp.ExecutionFee)
+				dataAvailabilityFees[i] = cciptypes.NewBigInt(feeComp.DataAvailabilityFee)
+			}
+			return types.ChainFeeComponents{
+				ExecutionFee:        consensus.Median(executionFees, consensus.BigIntComparator).Int,
+				DataAvailabilityFee: consensus.Median(dataAvailabilityFees, consensus.BigIntComparator).Int,
```
winder, line 61: This doesn't handle overflows

commit/chainfee/outcome_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}{
+		{
+			name:          "Outcome gas prices when earliest update is before batch write frequency duration",
+			aos:           SameObs(5, obsNeedUpdate),
+			expectedError: false,
+			expectedOutcome: func() Outcome {
+				gas1 := new(big.Int)
+				// {ExecutionFee: big.NewInt(100), DataAvailabilityFee: big.NewInt(200)},
+				// (200 * 10) << 112 | (100 * 10)
+				gas1, ok := gas1.SetString("10384593717069655257060992658440193000", 10) // base 10
```
winder, line 162: What is `(200 * 10) << 112 | (100 * 10)`?

asoliman92: explained in details here:
https://github.com/smartcontractkit/chainlink-ccip/pull/156/files#diff-8737ab5aae07ab1c494adf5c23edbd6847798e4b8293fb46b8c57aa89f7d4cc7R102-R106

commit/chainfee/processor.go
```diff
@@ -3,41 +3,134 @@ package chainfee
...
-	return &Processor{}
+// nolint: revive
+func NewProcessor(
+	lggr logger.Logger,
+	destChain cciptypes.ChainSelector,
+	homeChain reader.HomeChain,
+	ccipReader readerpkg.CCIPReader,
+	chainFeePriceBatchWriteFrequency commonconfig.Duration,
+	fRoleDON int,
+) *processor {
```
winder, line 41: Either: `processor` should be public, or this should return `PluginProcessor[blah, blah, blah]`

pkg/reader/ccip.go
```diff
@@ -439,29 +441,100 @@ func (r *ccipChainReader) Nonces(
...
+			map[string]any{
+				"token": nativeTokenAddress,
+			},
+			&price,
+		)
+		if err != nil {
+			r.lggr.Errorw("failed to get native token price", "chain", chain, "err", err)
+			continue
+		}
+		prices[chain] = cciptypes.NewBigInt(price)
```
winder, line 503: We have that concurrent call pattern used in other methods, but I'm not sure it's worth the extra complexity. Maybe we keep it simple like this and optimize with the Batch calls later.

commit/chainfee/processor.go
```diff
@@ -3,41 +3,134 @@ package chainfee
...
+
+		// Price is the combination ofr both dataAvailabilityPriceUSD and execPriceUSD
+		price := dataAvailabilityPriceUSD.Lsh(dataAvailabilityPriceUSD, 112)
+		combinedPricesUSD := new(big.Int).Or(price, execPriceUSD)
+
+		gasPrice := cciptypes.GasPriceChain{
+			ChainSel: chain,
+			GasPrice: cciptypes.NewBigInt(combinedPricesUSD),
+		}
+		gasPrices = append(gasPrices, gasPrice)
```
winder, line 120: This seems super EVM specific... how can it be generalized?

asoliman92: Extracted into new structs that can later be changed

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,185 @@
...
+	// because all nodes can observe the home chain.
+	donThresh := consensus.MakeConstantThreshold[cciptypes.ChainSelector](consensus.TwoFPlus1(p.fRoleDON))
+	fChains := consensus.GetConsensusMap(p.lggr, "fChain", aggObs.FChain, donThresh)
+
+	fDestChain, exists := fChains[p.destChain]
+	if !exists {
+		return Observation{},
+			fmt.Errorf("no consensus value for fDestChain, destChain: %d", p.destChain)
+	}
+	if len(aggObs.Timestamps) < fDestChain {
```
makramkd, line 32: I don't think this is the right way to check for consensus one the number of observations - is `Timestamps` simply the `time.Now().UTC()` stuff that each node observes? I think it should be safe to assert that `len(aggObs.Timestamps) > consensus.TwoFPlus1(fDestChain)`

commit/chainfee/outcome.go
makramkd: I believe `Median` does a `sort` then picks out the middle element, there's no addition going on.

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return aggObs
+}
+
+// getGasPricesToUpdate checks which chain fees need to be updated based on the observed chain fee prices and
+// the fee quoter updates.
+// A chain fee is selected for update if it meets one of 2 conditions:
+// 1. If time passed since the last update is greater than the stale threshold.
+// 2. If deviation between the fee quoter and feed exceeds the chain's configured threshold.
```
makramkd, line 133:
```suggestion
// 2. If deviation between the fee quoter and latest observed chain fee exceeds the chain's configured threshold.
```

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,185 @@
...
+
+		// Checks if executionFee or dataAvFee deviates from the last update
+		if mathslib.Deviates(
+			currentChainFee.ExecutionFeePriceUSD,
+			lastUpdate.ChainFee.ExecutionFeePriceUSD,
+			ci.ExecDeviationPPB.Int64()) ||
+			mathslib.Deviates(
+				currentChainFee.DataAvFeePriceUSD,
+				lastUpdate.ChainFee.DataAvFeePriceUSD,
+				ci.DataAvDeviationPPB.Int64()) {
```
makramkd, line 172: nit: If the if condition is more than one line its better to break out the subconditions into vars IMO

commit/chainfee/processor.go
```diff
@@ -3,41 +3,126 @@ package chainfee
...
+func (p *processor) Outcome(
 	prevOutcome Outcome,
 	query Query,
 	aos []plugincommon.AttributedObservation[Observation],
 ) (Outcome, error) {
-	return Outcome{}, nil
+
+	consensusObs, err := p.getConsensusObservation(aos)
+	if err != nil {
+		return Outcome{}, err
```
makramkd, line 83: Do we need to wrap this error to provide more context?

commit/chainfee/processor.go
```diff
@@ -3,41 +3,126 @@ package chainfee
...
+	}
+	// No need to update yet
+	if len(consensusObs.FeeComponents) == 0 {
+		return Outcome{}, nil
```
makramkd, line 87: Worth adding a log here so we can track behavior of this component and aid in debugging?

```suggestion
		p.lggr.Debug("no consensus on fee components, nothing to update", "consensusObs", consensusObs)
		return Outcome{}, nil
```

commit/chainfee/observation.go
```diff
@@ -0,0 +1,12 @@
+package chainfee
+
+import cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+func (p *processor) ObserveFChain() map[cciptypes.ChainSelector]int {
```
makramkd, line 5: Does this really need a separate file?

commit/chainfee/processor.go
```diff
@@ -3,41 +3,126 @@ package chainfee
...
+	//	return Outcome{}, nil
+	//}
+
+	chainFeeUSDPrices := make(map[cciptypes.ChainSelector]plugincommon.ChainFeeUSDPrices)
+	for chain, feeComp := range consensusObs.FeeComponents {
+		// We need to report a packed GasPrice
+		// The packed GasPrice is a 224-bit integer with the following format:
+		// (dataAvFeePriceUSD) << 112 | (executionFeePriceUSD)
+		// nolint:lll
+		// https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/FeeQuoter.sol#L498
```
makramkd, line 103: Where does the packing happen in this loop? Or is this comment better placed elsewhere?

commit/chainfee/types.go
```diff
@@ -1,16 +1,35 @@
...
 	GasPrices []cciptypes.GasPriceChain `json:"gasPrices"`
 }
 
 type Observation struct {
-	GasPrices []cciptypes.GasPriceChain `json:"gasPrices"`
+	FeeComponents     map[cciptypes.ChainSelector]types.ChainFeeComponents    `json:"feeComponents"`
+	NativeTokenPrices map[cciptypes.ChainSelector]cciptypes.BigInt            `json:"nativeTokenPrice"`
+	FChain            map[cciptypes.ChainSelector]int                         `json:"fChain"`
+	ChainFeeUpdates   map[cciptypes.ChainSelector]plugincommon.ChainFeeUpdate `json:"chainFeeUpdates"`
+	Timestamp         time.Time                                               `json:"timestamp"`
```
makramkd, line 25:
```suggestion
	TimestampNow         time.Time                                               `json:"timestamp"`
```

This naming feels more descriptive, `timestamp` can mean literally anything related to time.

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,185 @@
...
+
+	chainFeeUpdatesConsensus := consensus.GetConsensusMapAggregator(
+		p.lggr,
+		"ChainFeeUpdates",
+		aggObs.ChainFeeUpdates,
+		consensus.MakeConstantThreshold[cciptypes.ChainSelector](consensus.TwoFPlus1(fDestChain)),
+		plugincommon.ChainFeeUpdateAggregator,
+	)
+
+	twoFPlus1 := consensus.MakeMultiThreshold(fChains, consensus.TwoFPlus1)
```
makramkd, line 47: nit: naming for clarity

```suggestion
	twoFChainPlus1 := consensus.MakeMultiThreshold(fChains, consensus.TwoFPlus1)
```

commit/chainfee/types.go
```diff
@@ -1,16 +1,35 @@
...
+	// Each Gas Price is the combination of Execution and DataAvailability Fees using bitwise operations
 	GasPrices []cciptypes.GasPriceChain `json:"gasPrices"`
 }
 
 type Observation struct {
-	GasPrices []cciptypes.GasPriceChain `json:"gasPrices"`
+	FeeComponents     map[cciptypes.ChainSelector]types.ChainFeeComponents    `json:"feeComponents"`
+	NativeTokenPrices map[cciptypes.ChainSelector]cciptypes.BigInt            `json:"nativeTokenPrice"`
+	FChain            map[cciptypes.ChainSelector]int                         `json:"fChain"`
+	ChainFeeUpdates   map[cciptypes.ChainSelector]plugincommon.ChainFeeUpdate `json:"chainFeeUpdates"`
```
makramkd, line 24: I think its helpful for readers to know where these parts of the observation come from:

* FeeComponents: from the source chains, via chain writer
* NativeTokenPrices: from the source chains, via fee quoter
* ChainFeeUpdates: from the dest chain, via fee quoter

internal/plugincommon/chainfee.go
```diff
@@ -0,0 +1,71 @@
+package plugincommon
+
+import (
+	"math/big"
+	"time"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/consensus"
+)
+
+type ChainFeeUSDPrices struct {
```
makramkd, line 10: q: is there a reason these types are in plugincommon and not in the chainfee package? Circular deps or something?

pkg/reader/ccip.go
```diff
@@ -439,29 +443,101 @@ func (r *ccipChainReader) Nonces(
...
-			gasPrices[i] = cciptypes.NewBigInt(gasPrice.ExecutionFee)
-			return nil
-		})
+func (r *ccipChainReader) GetWrappedNativeTokenPriceUSD(
+	ctx context.Context,
+	selectors []cciptypes.ChainSelector,
+) map[cciptypes.ChainSelector]cciptypes.BigInt {
+	// 1. Call chain's router to get native token address https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/Router.sol#L189:L191
+	// nolint:lll
+	// 2. Call FeeQuoter to get native tokens price  https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/FeeQuoter.sol#L229-L229
```
makramkd, line 467:
```suggestion
	// 2. Call chain's FeeQuoter to get native tokens price  https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/FeeQuoter.sol#L229-L229
```

pkg/reader/ccip.go
```diff
@@ -12,6 +12,10 @@ import (
 	"golang.org/x/exp/maps"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugintypes"
+
 	types2 "github.com/smartcontractkit/libocr/offchainreporting2plus/types"
```
makramkd, line 19: I only see this kind of type aliasing in intellij 😛 can we name this `ocr3types` instead?

pkg/reader/ccip.go
```diff
@@ -439,29 +443,101 @@ func (r *ccipChainReader) Nonces(
...
+	// 2. Call FeeQuoter to get native tokens price  https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/FeeQuoter.sol#L229-L229
+	prices := make(map[cciptypes.ChainSelector]cciptypes.BigInt, len(selectors))
+	for _, chain := range selectors {
+		reader, ok := r.contractReaders[chain]
+		if !ok {
+			r.lggr.Warnw("contract reader not found", "chain", chain)
+			continue
+		}
+
+		var nativeTokenAddress types2.Account
```
makramkd, line 476: Can we add a TODO here to use batching in the future

pkg/reader/ccip.go
```diff
@@ -439,29 +443,101 @@ func (r *ccipChainReader) Nonces(
...
+	return prices
+}
 
-	if err := eg.Wait(); err != nil {
-		return nil, err
+// GetChainFeePriceUpdate Read from Destination chain FeeQuoter latest fee updates for the provided chains.
+// It unpacks the packed fee into the ChainFeeUSDPrices struct.
+// nolint:lll
+// https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/FeeQuoter.sol#L263-L263
+func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors []cciptypes.ChainSelector) map[cciptypes.ChainSelector]plugincommon.ChainFeeUpdate {
```
makramkd, line 514: Should we have an explicit param for the dest chain here? Or are we still not sure we want to follow that pattern? cc @winder

pluginconfig/commit.go
```diff
@@ -62,6 +62,10 @@ func (c CommitPluginConfig) Validate() error {
 	return c.OffchainConfig.Validate()
 }
 
+type FeeInfo struct {
+	ExecDeviationPPB   cciptypes.BigInt `json:"execDeviationPPB"`
+	DataAvDeviationPPB cciptypes.BigInt `json:"dataAvailabilityDeviationPPB"`
+}
```
makramkd, line 68: nit: newline after this one

pluginconfig/commit.go
```diff
@@ -62,6 +62,10 @@ func (c CommitPluginConfig) Validate() error {
 	return c.OffchainConfig.Validate()
 }
 
+type FeeInfo struct {
+	ExecDeviationPPB   cciptypes.BigInt `json:"execDeviationPPB"`
+	DataAvDeviationPPB cciptypes.BigInt `json:"dataAvailabilityDeviationPPB"`
```
makramkd, line 67: nit: Struct field name matches json tag name ideally

```suggestion
	DataAvailabilityDeviationPPB cciptypes.BigInt `json:"dataAvailabilityDeviationPPB"`
```

makramkd (commented):
Looking good!

pkg/reader/ccip.go
winder: Isn't that what selectors is?

makramkd: I believe `selectors` here are sources. Would be nice to be more explicit

commit/chainfee/observation.go
asoliman92: Just for consistency between all commit processors. I was even going to make proxies for ccipReader calls that happens in actual observation but decided against it for simplicity.

commit/chainfee/outcome.go
asoliman92: Yes, good call 👍

commit/chainfee/outcome.go
asoliman92: It's because I'm using `cciptypes.BigInt`, to get the `*big.Int` I call `.Int`. TBH this cciptypes.BigInt is annoying and I'd rather use only `*big.Int` directly where possible.

commit/chainfee/processor.go
asoliman92: It's happening on the wrapper type level when calling `getGasPricesToUpdate` later. I have similar comment in multiple places to make it obvious to anyone who's reading the code. I think it's good to keep it here or at least the Outcome method to make sure whoever reading the code from top level understands what's happening.

internal/plugincommon/chainfee.go
asoliman92: cyclic dependency because I was returning it in ccip reader. On a second thought, I'm returning TimestampedBig and doing the conversion inside the processor for better encapsulation.

pkg/reader/ccip.go
asoliman92: Added TODO 👍

pkg/reader/ccip.go
asoliman92: selectors are sources (which are dest chain selectors on the dest chain). You want to make the API clearer by providing the destChain @makramkd ?

commit/chainfee/outcome.go
makramkd: Ah shoot I meant `len(aggObs.Timestamps) >= consensus.TwoFPlus1(fDestChain)` (notice >= not >)

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,185 @@
...
+	fDestChain, exists := fChains[p.destChain]
+	if !exists {
+		return Observation{},
+			fmt.Errorf("no consensus value for fDestChain, destChain: %d", p.destChain)
+	}
+
+	if len(aggObs.Timestamps) < int(consensus.TwoFPlus1(fDestChain)) {
+		return Observation{},
+			fmt.Errorf("not enough observations for timestamps to reach consensus, have %d, need %d",
+				len(aggObs.Timestamps), fDestChain)
```
makramkd, line 35:
```suggestion
				len(aggObs.Timestamps), consensus.TwoFPlus1(fDestChain))
```

commit/chainfee/processor.go
```diff
@@ -3,41 +3,137 @@ package chainfee
...
+	// https://github.com/smartcontractkit/chainlink/blob/60e8b1181dd74b66903cf5b9a8427557b85357ec/contracts/src/v0.8/ccip/FeeQuoter.sol#L498
+	// In next loop we calculate the price in USD for the data availability and execution fees.
+	// And getGasPricesToUpdate will select and calculate the **packed** gas price to update based.
+	for chain, feeComp := range consensusObs.FeeComponents {
+		nativeTokenPriceUSD := consensusObs.NativeTokenPrices[chain].Int
+
+		// Calculate the price in USD for the data availability and execution fees.
+		// Raw fee components are in native token units
+		chainFeeUsd := ComponentsUSDPrices{
+			ExecutionFeePriceUSD: new(big.Int).Mul(nativeTokenPriceUSD, feeComp.ExecutionFee),
```
makramkd, line 113: So this is native token price in USD to 36 decimals, ExecutionFee is the fee in the native token its the smallest denomination (e.g if its ethereum, its wei)?

So this means `USD/NATIVE * NATIVE == USD`, but the question I have is, is `nativeTokenPriceUSD` per smallest denomination, or is it per 1e18 of the smallest denomination? The denominations have to be the same in order for this multiplication to make sense.

makramkd (approved):

commit/chainfee/outcome.go
```diff
@@ -0,0 +1,185 @@
...
+
+	return aggObs
+}
+
+// getGasPricesToUpdate checks which chain fees need to be updated based on the observed chain fee prices and
+// the fee quoter updates.
+// A chain fee is selected for update if it meets one of 2 conditions:
+// 1. If time passed since the last update is greater than the stale threshold.
+// 2. If deviation between the fee quoter and latest observed chain fee exceeds the chain's configured threshold.
+func (p *processor) getGasPricesToUpdate(
```
dimkouv, line 134: Maybe we want to rename `Gas Price` to `Chain Fee` in several places?

commit/chainfee/observation.go
dimkouv: I would expect `observation.go` to have processor observation logic, but it's in processor.go, which is confussing

commit/chainfee/processor.go
```diff
@@ -3,46 +3,172 @@ package chainfee
...
 	prevOutcome Outcome,
 	query Query,
 ) (Observation, error) {
-	return Observation{}, nil
+	// Get the fee components for all available chains that we can read from
+	feeComponents := p.ccipReader.GetAvailableChainsFeeComponents(ctx)
+	// Get the native token prices for all available chains that we can read from
+	nativeTokenPrices := p.ccipReader.GetWrappedNativeTokenPriceUSD(ctx, maps.Keys(feeComponents))
+	// Get the latest chain fee price updates for the source chains
+	timestampedPriceUpdates := p.ccipReader.GetChainFeePriceUpdate(ctx, maps.Keys(feeComponents))
```
dimkouv, line 64: Minor performance: dup call to `maps.Keys(feeComponents)`

commit/chainfee/processor.go
```diff
@@ -3,46 +3,172 @@ package chainfee
...
+func (p *processor) Outcome(
 	prevOutcome Outcome,
 	query Query,
 	aos []plugincommon.AttributedObservation[Observation],
 ) (Outcome, error) {
-	return Outcome{}, nil
+
+	consensusObs, err := p.getConsensusObservation(aos)
+	if err != nil {
+		return Outcome{}, fmt.Errorf("failed to get consensus observation: %w", err)
```
dimkouv, line 94: general tip, no need to have `failed to` in errors. `get consensus observation: %w` is ok.
because if we do `failed to` in each wrap, then we get:

```
failed to do this: failed to do that: failed to ...
```

instead of

```
do this: do that: ...
```

dimkouv (approved):
Approving, maybe we can resolve comments with `ValidateObservation` implementation since there's still work left.

commit/chainfee/outcome.go
asoliman92: The current final report is GasPrice, that's why I use it only here

commit/chainfee/outcome.go
asoliman92: I'm have the condition reversed and using `<` so should be okay

PR #160, opened 2024-09-24
Remove temporary call to DiscoverContracts.

winder:
Removing temporary code which is no longer needed. The Discovery Processor handles this, and it is now fully integrated with plugins.


winder:
Resolved by #163

PR #163, opened 2024-09-25
internal/plugincommon/discovery: logic cleanup, unit tests

makramkd:
This PR cleans up some discovery processor logic and adds unit tests for the `Observation` and `Outcome` methods.


internal/plugincommon/discovery/processor.go
```diff
@@ -53,19 +53,28 @@ func (cdp *ContractDiscoveryProcessor) Query(_ context.Context, _ dt.Outcome) (d
 func (cdp *ContractDiscoveryProcessor) Observation(
 	ctx context.Context, _ dt.Outcome, _ dt.Query,
 ) (dt.Observation, error) {
+	// all oracles should be able to read from the home chain, so we
+	// can fetch f chain reliably.
+	// TODO: should we error out here or just try to "observe everything we can"?
+	// i.e, similar to the commit plugin.
```
makramkd, line 59: @winder lmk your thoughts on this

internal/plugincommon/discovery/processor.go
```diff
@@ -53,19 +53,28 @@ func (cdp *ContractDiscoveryProcessor) Query(_ context.Context, _ dt.Outcome) (d
...
 	contracts, err := (*cdp.reader).DiscoverContracts(ctx, cdp.dest)
 	if err != nil {
 		if errors.Is(err, reader.ErrContractReaderNotFound) {
-			// Not a dest reader, no observations will be made.
+			// Not a dest reader, only fChain observation will be made.
 			// Processor is not disabled because the outcome phase will bind observed contracts.
-			return dt.Observation{}, nil
+			return dt.Observation{
+				FChain: fChain,
+			}, nil
```
makramkd, line 72: @winder and this.

pkg/reader/ccip.go
```diff
@@ -504,18 +504,19 @@ func (r *ccipChainReader) DiscoverContracts(
...
-			return fmt.Errorf("sync: %w", err)
-		}
-	}
-
 	var errs []error
 	for contractName, chainSelToAddress := range contracts {
 		for chainSel, address := range chainSelToAddress {
+			// defense in depth: don't bind if the address is empty.
+			// callers should ensure this but we double check here.
+			if len(address) == 0 {
```
makramkd, line 512: @AnieeG your change is here

pkg/reader/ccip.go
```diff
@@ -582,6 +585,23 @@ func (r *ccipChainReader) getSourceChainsConfig(
...
+			// chain data -> go struct.
+			if len(resp.OnRamp) == 0 {
+				return fmt.Errorf(
+					"onRamp misconfigured/didn't unmarshal for chain %d: %x",
+					chainSel,
+					resp.OnRamp,
+				)
+			}
+
+			if !resp.IsEnabled {
```
makramkd, line 599: This is actually a regression, we had it in `Sync` but seems like it was deleted.

pkg/reader/ccip.go
```diff
@@ -582,6 +585,23 @@ func (r *ccipChainReader) getSourceChainsConfig(
 			if err != nil {
 				return fmt.Errorf("failed to get source chain config: %w", err)
 			}
+
+			// This may happen due to some sort of regression in the codec that unmarshals
+			// chain data -> go struct.
+			if len(resp.OnRamp) == 0 {
```
makramkd, line 591: I think it makes more sense to return an error here than invalid data. Lmk your thoughts @winder

internal/plugincommon/discovery/processor.go
```diff
@@ -100,32 +110,55 @@ func (cdp *ContractDiscoveryProcessor) Outcome(
...
+	)
+	if len(onrampConsensus) == 0 {
+		cdp.lggr.Warnw("No consensus on onramps, onrampConsensus map is empty")
+	}
+	contracts[consts.ContractNameOnRamp] = onrampConsensus
 
-	contracts[consts.ContractNameNonceManager] = consensus.GetConsensusMap(
+	// TODO: doesn't seem correct to use the fChain of the source chain to determine consensus on onramps,
+	// which are read from the dest chain.
+	nonceManagerConsensus := consensus.GetConsensusMap(
```
makramkd, line 150: cc @winder

makramkd: I've fixed this

execute/plugin.go
```diff
@@ -406,6 +407,7 @@ func (p *Plugin) Outcome(
 				Observation: decodedAos[i].Observation.Contracts,
 			}
 		}
+		p.lggr.Infow("Processing discovery observations", "discoveryObservations", discoveryAos)
```
winder, line 410: Let's generalize this. Move it to the top of `Outcome(`
```
p.lggr.Debugw("execute Outcome", "outctx": outctx, "query": query", "aos": aos)
```

winder: Before decoding would be more general, but require a tool for interpreting data. After would be better for ad-hoc log inspection.

internal/plugincommon/consensus/consensus.go
```diff
@@ -15,30 +15,30 @@ import (
 func GetConsensusMap[K comparable, T any](
 	lggr logger.Logger,
 	objectName string,
-	itemsByChain map[K][]T,
+	itemsByKey map[K][]T,
```
winder, line 18: Thank you, I had the same change pending. But I'm starting to question whether key should be generic since the thresholds become more indirect

internal/plugincommon/discovery/processor.go
winder: Let's error early and often, we can relax them later as we harden the system.

For fChain specifically, it's a problem. The current design requires observing it multiple times per round of observations.

The simplest way to address this would be adding fChain as a parameter to the `PluginProcesssor` type.

internal/plugincommon/discovery/processor.go
winder: Aha, yes this is required even since we need to use fRoleDON for the threshold.

internal/plugincommon/discovery/processor.go
```diff
@@ -53,19 +53,28 @@ func (cdp *ContractDiscoveryProcessor) Query(_ context.Context, _ dt.Outcome) (d
...
 		}
-		return dt.Observation{}, fmt.Errorf("unable to discover contracts: %w", err)
-	}
 
-	fChain, err := cdp.homechain.GetFChain()
-	if err != nil {
-		return dt.Observation{}, fmt.Errorf("unable to get fchain: %w", err)
+		// otherwise a legitimate error occurred when discovering.
+		// TODO: should we still return the f chain observation w/ a nil error?
+		return dt.Observation{}, fmt.Errorf("unable to discover contracts: %w", err)
```
winder, line 77: No. Let's not create a situation where we return an error but still need to handle the observation.

internal/plugincommon/discovery/processor.go
```diff
@@ -98,34 +108,63 @@ func (cdp *ContractDiscoveryProcessor) Outcome(
...
 
-	// collect onramp addresses
+	// collect onramp and nonce manager addresses
 	onrampAddrs := make(map[cciptypes.ChainSelector][][]byte)
+	var nonceManagerAddrs [][]byte
 	for _, ao := range aos {
 		for chain, addr := range ao.Observation.OnRamp {
+			// we don't want invalid observations to "poison" the consensus.
+			if len(addr) == 0 {
+				continue
```
winder, line 119:
```suggestion
				cdp.lggr.Warnf("skipping empty onramp address in observation from Oracle %d", ao.Observation.OracleID)
				continue
```

internal/plugincommon/discovery/processor.go
```diff
@@ -98,34 +108,63 @@ func (cdp *ContractDiscoveryProcessor) Outcome(
...
+			}
 			onrampAddrs[chain] = append(onrampAddrs[chain], addr)
 		}
-	}
 
-	// collect nonce manager addresses
-	var nonceManagerAddrs [][]byte
-	for _, ao := range aos {
+		if len(ao.Observation.DestNonceManager) == 0 {
+			continue
```
winder, line 125:
```suggestion
				cdp.lggr.Warnf("skipping empty nonce manager address in observation from Oracle %d", ao.Observation.OracleID)
			continue
```

pkg/reader/ccip.go
winder: Yes, let's detect problems and generate the errors early. Full address validation could happen in `DiscoverContracts`

internal/plugincommon/discovery/processor_test.go
```diff
@@ -0,0 +1,436 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery/discoverytypes"
+	mock_home_chain "github.com/smartcontractkit/chainlink-ccip/mocks/internal_/reader"
+	mock_reader "github.com/smartcontractkit/chainlink-ccip/mocks/pkg/reader"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/reader"
+)
+
+func TestContractDiscoveryProcessor_Observation(t *testing.T) {
+	t.Run("support the dest, happy path", func(t *testing.T) {
```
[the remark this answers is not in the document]
makramkd: Got it. VS Code lets me run these separately, but not sure about vim/emacs/etc.

execute/plugin.go
makramkd: Before decoding I believe it would log the base64-encoded bytes, which should be decodable into JSON after getting the raw bytes.

internal/plugincommon/consensus/consensus.go
makramkd: Good point, lets consider in a followup.

internal/plugincommon/discovery/processor.go
makramkd: Yeah actually you're right, fChain is pretty much a dependency for all upcoming consensus operations, so it should error.

pkg/reader/ccip.go
makramkd: Good note about address validation, I think here you can put the address validation, i.e if the source chain is EVM and the # of bytes is not what is expected, error.

internal/plugincommon/discovery/processor.go
winder: [Created a ticket for this.](https://smartcontract-it.atlassian.net/jira/software/c/projects/CCIP/boards/423?selectedIssue=CCIP-3526)

PR #166, opened 2024-09-25
chainlink-common bump

0xnogo:
Bumping `chainlink-common` to fill the gap between `chainlink` and `chainlink-ccip`.

Currently working on passing the home chain selector id from the delegate to the commit plugin factory and I am getting some regression error since we are using an older version in chainlink-ccip.


makramkd:
Lets include reasoning for bumping, especially chainlink-common which is a notorious dep

makramkd (approved):

PR #167, opened 2024-09-25
Exec v2 scaffolding to determine costly messages

rstout:
Adds scaffolding to the V2 Exec Plugin to enable observing which messages are too costly to execute.

### DIFF
--- a/execute/exectypes/commit_data.go
+++ b/execute/exectypes/commit_data.go
@@ -25,6 +25,11 @@ type CommitData struct {
 	// ExecutedMessages are the messages in this report that have already been executed.
 	ExecutedMessages []cciptypes.SeqNum `json:"executedMessages"`
 
+	// CostlyMessages are the message IDs of messages that cost more to execute than was paid to execute them (i.e.
+	// source fee < execution cost). These messages will not be executed in the current round, but may be executed in
+	// future rounds (e.g. if gas prices decrease or if these messages' fees are boosted high enough).
+	CostlyMessages []cciptypes.Bytes32 `json:"costlyMessages"`
+
 	// TokenData for each message.
 	MessageTokenData []MessageTokenData `json:"messageTokenData"`
 }
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -9,12 +9,13 @@ import (
 	mapset "github.com/deckarep/golang-set/v2"
 	"golang.org/x/exp/maps"
 
-	"github.com/smartcontractkit/chainlink-common/pkg/logger"
-	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 	"github.com/smartcontractkit/libocr/commontypes"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 	libocrtypes "github.com/smartcontractkit/libocr/ragep2p/types"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 
 	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
 	"github.com/smartcontractkit/chainlink-ccip/execute/internal/gas"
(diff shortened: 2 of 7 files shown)


execute/exectypes/commit_data.go
```diff
@@ -25,6 +25,11 @@ type CommitData struct {
 	// ExecutedMessages are the messages in this report that have already been executed.
 	ExecutedMessages []cciptypes.SeqNum `json:"executedMessages"`
 
+	// CostlyMessages are the message IDs of messages that cost more to execute than was paid to execute them (i.e.
+	// source fee < execution cost). These messages will not be executed in the current round, but may be executed in
+	// future rounds (e.g. if gas prices decrease or if these messages' fees are boosted high enough).
```
makramkd, line 30: I have a question about the logic surrounding this - re: executed in future rounds. How long can these messages potentially accumulate before we kick them out? Can they stick around for the whole permissionless execution window?

execute/plugin.go
```diff
@@ -272,7 +276,13 @@ func (p *Plugin) Observation(
 			return types.Observation{}, fmt.Errorf("unable to process token data %w", err1)
 		}
 
-		return exectypes.NewObservation(groupedCommits, messages, tkData, nil, discoveryObs).Encode()
+		costlyMessages, err := p.costlyMessageObserver.Observe(ctx, messages)
```
makramkd, line 279: Trying to understand this: at this point, we have messages from potentially many commit reports, and we filter out those that are "too costly" to execute, even after boosting. Next round comes around, say a commit report has some subset of its messages executed, the remaining not executed due to being too costly. We pick up from there again, and we determine if they're still too costly, etc?

execute/plugin_functions.go
```diff
@@ -382,6 +382,30 @@ func mergeNonceObservations(
...
+func mergeCostlyMessages(
+	aos []plugincommon.AttributedObservation[exectypes.Observation],
+	fChainDest int,
+) []cciptypes.Bytes32 {
+	costlyMessages := mapset.NewSet[cciptypes.Bytes32]()
+	counts := make(map[cciptypes.Bytes32]int)
+	for _, ao := range aos {
+		for _, costlyMessage := range ao.Observation.CostlyMessages {
+			counts[costlyMessage]++
+			if counts[costlyMessage] > fChainDest {
```
makramkd, line 396: small nit: can use `if counts[costlyMessage] >= int(consensus.FPlus1(fChainDest))` here to make this condition clearer.

The goal behind this helper is to be able to easily audit where we require at least one honest observation vs. where we require an honest majority.

execute/plugin_functions.go
```diff
@@ -382,6 +382,30 @@ func mergeNonceObservations(
...
+				costlyMessages.Add(costlyMessage)
+			}
+		}
+	}
+
+	if costlyMessages.Cardinality() == 0 {
+		return nil
+	}
```
makramkd, line 404: Is this check necessary? If cardinality is zero it should just return an empty slice right?

execute/report/report.go
```diff
@@ -196,6 +197,29 @@ func (b *execReportBuilder) checkMessage(
 		"data", messageTokenData.ToByteSlice())
 
 	// 3. Check if the message has a valid nonce.
+	status := b.checkMessageNonce(msg, execReport)
+	if status != "" {
+		return execReport, status, nil
+	}
+
+	// 4. Check if the message is too costly to execute.
+	if slices.Contains(execReport.CostlyMessages, msg.Header.MessageID) {
```
makramkd, line 206: Consider: if `execReport.CostlyMessages` is a map, this could be a constant time lookup. I'm not sure how big these slices could be, I imagine it could be bad if there's a gas spike and there are 1000s of ccip messages across different chains.

execute/exectypes/commit_data.go
makramkd: Another question re: observation size. Does this get capped, i.e if there is a gas spike on the dest and many messages are underpaid, we could have many messages backed up and too costly.

rstout: I think we can proceed, discuss offline and address any issues in future PRs

execute/plugin.go
rstout: Yes, that's correct

execute/plugin_functions.go
rstout: Done!

execute/plugin_functions.go
rstout: This helps with testing. Without this check, a lot of tests would need to expect 
```
exectypes.Observation{
    CostlyMessages: []cciptypes.Bytes32{}
}
```
Instead of simply:
```
exectypes.Observation{}
```

rstout: I've added a comment

execute/report/report.go
rstout: Good point, I'll use a map

rstout: These are the costly messages per report, they won't be more than the message limit per commit report. I did try making CostlyMessages a map/set, but I ran into JSON encoding issues. We can revisit this if it becomes an issue.

makramkd (approved):

PR #168, opened 2024-09-26
pkg/reader: DiscoverContracts unit tests

makramkd:
Write some unit tests for the discover contracts logic covering different scenarios.

This is in an effort to catch any bugs as early as possible in unit tests, especially when it comes to the "not all nodes support all chains" scenario.

### DIFF
--- a/pkg/reader/ccip_test.go
+++ b/pkg/reader/ccip_test.go
@@ -390,7 +390,13 @@ func TestCCIPChainReader_DiscoverContracts_HappyPath(t *testing.T) {
 		primitives.Unconfirmed,
 		map[string]any{"sourceChainSelector": sourceChain1},
 		mock.Anything,
-	).Return(nil).Run(func(ctx context.Context, contractName, methodName string, confidenceLevel primitives.ConfidenceLevel, params, returnVal interface{}) {
+	).Return(nil).Run(func(
+		ctx context.Context,
+		contractName,
+		methodName string,
+		confidenceLevel primitives.ConfidenceLevel,
+		params,
+		returnVal interface{}) {
 		v := returnVal.(*sourceChainConfig)
 		v.OnRamp = s1Onramp
 		v.IsEnabled = true
@@ -403,7 +409,13 @@ func TestCCIPChainReader_DiscoverContracts_HappyPath(t *testing.T) {
 		primitives.Unconfirmed,
 		map[string]any{"sourceChainSelector": sourceChain2},
 		mock.Anything,
-	).Return(nil).Run(func(ctx context.Context, contractName, methodName string, confidenceLevel primitives.ConfidenceLevel, params, returnVal interface{}) {
+	).Return(nil).Run(func(
+		ctx context.Context,
+		contractName,
+		methodName string,
+		confidenceLevel primitives.ConfidenceLevel,
+		params,
+		returnVal interface{}) {
 		v := returnVal.(*sourceChainConfig)
 		v.OnRamp = s2Onramp
 		v.IsEnabled = true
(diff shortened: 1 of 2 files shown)


pkg/reader/ccip_test.go
```diff
@@ -371,3 +371,288 @@ func TestCCIPChainReader_Sync_BindError(t *testing.T) {
...
+		ctx context.Context,
+		contractName,
+		methodName string,
+		confidenceLevel primitives.ConfidenceLevel,
+		params,
+		returnVal interface{}) {
+		v := returnVal.(*offrampStaticChainConfig)
+		v.NonceManager = destNonceMgr
+	})
+	defer destExtended.AssertExpectations(t)
```
mateusz-sekara, line 441: Is this call really needed? I think tooling calls it at the end of the test anyway, doesn't it?

pkg/reader/ccip_test.go
```diff
@@ -371,3 +371,288 @@ func TestCCIPChainReader_Sync_BindError(t *testing.T) {
...
+
+	// create the reader
+	ccipChainReader := &ccipChainReader{
+		destChain: destChain,
+		contractReaders: map[cciptypes.ChainSelector]contractreader.Extended{
+			destChain: destExtended,
+			// these won't be used in this test, but are needed because
+			// we determine the source chain selectors to query from the chains
+			// that we have readers for.
```
mateusz-sekara, line 450: 👍

pkg/reader/ccip_test.go
```diff
@@ -371,3 +371,288 @@ func TestCCIPChainReader_Sync_BindError(t *testing.T) {
...
+			sourceChain1: reader_mocks.NewMockExtended(t),
+			sourceChain2: reader_mocks.NewMockExtended(t),
+		},
+		lggr: logger.Test(t),
+	}
+
+	contractAddresses, err := ccipChainReader.DiscoverContracts(ctx, destChain)
+	require.NoError(t, err)
+	expectedContractAddresses := ContractAddresses{
```
mateusz-sekara, line 459: nit, but I think "expectations" are usually built at the beginning of the test to easier read the input and output of the test ;)

pkg/reader/ccip_test.go
```diff
@@ -371,3 +371,288 @@ func TestCCIPChainReader_Sync_BindError(t *testing.T) {
...
+	})
+	// mock the call for sourceChain2
+	destExtended.EXPECT().ExtendedGetLatestValue(
+		mock.Anything,
+		consts.ContractNameOffRamp,
+		consts.MethodNameGetSourceChainConfig,
+		primitives.Unconfirmed,
+		map[string]any{"sourceChainSelector": sourceChain2},
+		mock.Anything,
+	).Return(nil).Run(func(
```
mateusz-sekara, line 412: I wonder if we can reduce this boilercode a bit, for instance by passing lambda

```go
func withReturnValueOverridden(mapper func(returnVal interface{})) func(ctx context.Context,
	contractName,
	methodName string,
	confidenceLevel primitives.ConfidenceLevel,
	params,
	returnVal interface{}) {
	return func(ctx context.Context,
		contractName,
		methodName string,
		confidenceLevel primitives.ConfidenceLevel,
		params,
		returnVal interface{}) {
		mapper(returnVal)
	}
}


	destExtended.EXPECT().ExtendedGetLatestValue(
		mock.Anything,
		consts.ContractNameOffRamp,
		consts.MethodNameGetSourceChainConfig,
		primitives.Unconfirmed,
		map[string]any{"sourceChainSelector": sourceChain1},
		mock.Anything,
	).Return(nil).Run(withReturnValueOverridden(func(returnVal interface{}) {
		v := returnVal.(*sourceChainConfig)
		v.OnRamp = s1Onramp
		v.IsEnabled = true
	}))

pkg/reader/ccip_test.go
makramkd: Seems like you're right, this is called in the `Cleanup` function of each mock, will remove.

pkg/reader/ccip_test.go
makramkd: Added

pkg/reader/ccip_test.go
makramkd: Fixed

mateusz-sekara (approved):

pkg/reader/ccip_test.go
makramkd: Unfortunately the signature is slightly different than ExtendedGetLatestValue. I think should be OK since this only occurs once.

PR #170, opened 2024-09-26
CCIP-3538 Consensus layer for TokenData observations

mateusz-sekara:


execute/exectypes/observation_test.go
```diff
@@ -12,13 +12,11 @@ func Test_MessageTokensData(t *testing.T) {
 		name         string
 		msgTokenData MessageTokenData
 		ready        bool
-		error        error
```
mateusz-sekara, line 15: No longer used

execute/plugin_functions.go
```diff
@@ -305,31 +305,75 @@ func mergeCommitObservations(
...
 		}
-		if cnt > maxCnt {
-			maxCnt = cnt
-			maxV = ao.Observation.TokenData
+	}
+
+	for selector, seqNrs := range validators {
+		for tokenID, validator := range seqNrs {
+			mtd := results[selector][tokenID.SeqNr]
+			if values := validator.GetValid(); len(values) > 0 {
```
makramkd, line 339: In the other consensus functions we've been checking that `len(values) == 1` exactly, idea being if there's more than one value its probably also a consensus error - i.e, there isn't a single value that is "agreed upon". Not sure if this makes sense in the F+1 context which seems to be the current one though.

Assuming we have 3 * F + 1 nodes, you could have up to two different values that have been observed F+1 times, which in my mind is a consensus error. If you do 2*F+1 this isn't possible, but I think for exec we're more lenient than commit.

makramkd: That said, you could just go with this cuz at the end of the day the attestation is verified on-chain, and if an invalid one is chosen the execution will just fail. So its not a big damage.

makramkd (approved):
Small comment about `len(values) > 0` condition, lmk what you think. Otherwise LGTM.

execute/plugin_functions.go
mateusz-sekara: Makes sense, fixing

PR #171, opened 2024-09-26
Discover Router and FeeQuoters

winder:
Extend the DiscoverContract function:
* Accepts a slice of chains. In the future this should be the total list of all supported chains.
* Best-effort discovery of addresses known to the onramps. These will not be discovered until the second round.
* Lookup additional offramp contracts: FeeQuoter, Router.

### DIFF
--- a/internal/mocks/inmem/ccipreader_inmem.go
+++ b/internal/mocks/inmem/ccipreader_inmem.go
@@ -138,7 +138,7 @@ func (r InMemoryCCIPReader) GetChainFeePriceUpdate(
 }
 
 func (r InMemoryCCIPReader) DiscoverContracts(
-	ctx context.Context, destChain cciptypes.ChainSelector,
+	ctx context.Context, allChains []cciptypes.ChainSelector,
 ) (reader.ContractAddresses, error) {
 	return nil, nil
 }
--- a/internal/plugincommon/discovery/discoverytypes/types.go
+++ b/internal/plugincommon/discovery/discoverytypes/types.go
@@ -1,6 +1,8 @@
 package discoverytypes
 
-import "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+import (
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+)
 
 // Outcome isn't needed for this processor.
 type Outcome struct {
@@ -13,10 +15,13 @@ type Query []byte
 
 // Observation of contract addresses.
 type Observation struct {
-	FChain           map[ccipocr3.ChainSelector]int
-	OnRamp           map[ccipocr3.ChainSelector][]byte
-	DestNonceManager []byte
-	RMNRemote        []byte
+	//SourceChains     []ccipocr3.ChainSelector
+	FChain map[cciptypes.ChainSelector]int
+	// See reader.ContractAddresses for more info on this data structure.
+	Addresses map[string]map[cciptypes.ChainSelector][]byte
+
+	// TODO: fix circular dependency /w token reader
+	//Addresses  reader.ContractAddresses
 
 	// TODO: some sort of request flag to avoid including this every time.
 	// Request bool
--- a/internal/plugincommon/discovery/processor.go
+++ b/internal/plugincommon/discovery/processor.go
@@ -200,7 +200,7 @@ func (cdp *ContractDiscoveryProcessor) Outcome(
 	donThresh := consensus.MakeConstantThreshold[cciptypes.ChainSelector](consensus.TwoFPlus1(cdp.fRoleDON))
 	fChain := consensus.GetConsensusMap(cdp.lggr, "fChain", fChainObs, donThresh)
 	fChainThresh := consensus.MakeMultiThreshold(fChain, consensus.TwoFPlus1)
-	destThresh := consensus.MakeConstantThreshold[cciptypes.ChainSelector](consensus.TwoFPlus1(cdp.fRoleDON))
+	destThresh := consensus.MakeConstantThreshold[cciptypes.ChainSelector](consensus.TwoFPlus1(fChain[cdp.dest]))
 
 	agg := aggregateObservations(cdp.lggr, cdp.dest, aos)
 
--- a/internal/plugincommon/discovery/processor.go
+++ b/internal/plugincommon/discovery/processor.go
@@ -66,7 +66,8 @@ func (cdp *ContractDiscoveryProcessor) Observation(
 		return dt.Observation{}, fmt.Errorf("unable to get fchain: %w", err)
 	}
 
-	contracts, err := (*cdp.reader).DiscoverContracts(ctx, cdp.dest)
+	// TODO: discover the full list of source chain selectors and pass it into DiscoverContracts.
+	contracts, err := (*cdp.reader).DiscoverContracts(ctx, nil)
 	if err != nil {
 		if errors.Is(err, reader.ErrContractReaderNotFound) {
 			// Not a dest reader, only fChain observation will be made.
@@ -81,10 +82,8 @@ func (cdp *ContractDiscoveryProcessor) Observation(
 	}
 
 	return dt.Observation{
-		FChain:           fChain,
-		OnRamp:           contracts[consts.ContractNameOnRamp],
-		DestNonceManager: contracts[consts.ContractNameNonceManager][cdp.dest],
-		RMNRemote:        contracts[consts.ContractNameRMNRemote][cdp.dest],
+		FChain:    fChain,
+		Addresses: contracts,
 	}, nil
 }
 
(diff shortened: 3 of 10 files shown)


internal/plugincommon/discovery/discoverytypes/types.go
```diff
@@ -13,10 +15,12 @@ type Query []byte
...
 type Observation struct {
-	FChain           map[ccipocr3.ChainSelector]int
-	OnRamp           map[ccipocr3.ChainSelector][]byte
-	DestNonceManager []byte
-	RMNRemote        []byte
+	//SourceChains     []ccipocr3.ChainSelector
+	FChain    map[cciptypes.ChainSelector]int
+	Addresses map[string]map[cciptypes.ChainSelector][]byte
+
+	// TODO: fix circular dependency /w token reader
```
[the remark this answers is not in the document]
winder: I think TokenReader should probably not be depending on something in the plugin packages

pkg/reader/ccip.go
```diff
@@ -550,45 +550,113 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
 
-func (r *ccipChainReader) DiscoverContracts(
+func (r *ccipChainReader) discoverDestinationContracts(
 	ctx context.Context,
-	destChain cciptypes.ChainSelector,
+	allChains []cciptypes.ChainSelector,
 ) (ContractAddresses, error) {
-	if err := validateExtendedReaderExistence(r.contractReaders, destChain); err != nil {
-		return nil, err
+	// Exit without an error if we cannot read the destination.
```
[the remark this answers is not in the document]
winder: Moved this to `DiscoverContracts` along with a comment. The handling is a little different than elsewhere because it's important to continue after this error in order to handle cases where a node is configured such that they can read from a source chain but not the dest chain.

internal/plugincommon/discovery/discoverytypes/types.go
```diff
@@ -13,10 +15,12 @@ type Query []byte
 
 // Observation of contract addresses.
 type Observation struct {
-	FChain           map[ccipocr3.ChainSelector]int
-	OnRamp           map[ccipocr3.ChainSelector][]byte
-	DestNonceManager []byte
-	RMNRemote        []byte
+	//SourceChains     []ccipocr3.ChainSelector
+	FChain    map[cciptypes.ChainSelector]int
+	Addresses map[string]map[cciptypes.ChainSelector][]byte
```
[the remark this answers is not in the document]
winder: We're not consistent with addresses. They use `[]byte`, `types.Account` and `cciptypse.Bytes`. The typeconv utility uses []byte so thats what I would opt for until we're able to add a proper address type.

internal/plugincommon/discovery/processor.go
```diff
@@ -66,7 +66,8 @@ func (cdp *ContractDiscoveryProcessor) Observation(
 		return dt.Observation{}, fmt.Errorf("unable to get fchain: %w", err)
 	}
 
-	contracts, err := (*cdp.reader).DiscoverContracts(ctx, cdp.dest)
+	// TODO: discover the full list of source chain selectors and pass it into DiscoverContracts.
```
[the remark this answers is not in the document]
winder: Should not be a blocker. I tried to avoid breaking the current behavior where things work if everyone is a dest reader.

internal/mocks/inmem/ccipreader_inmem.go
```diff
@@ -138,7 +138,7 @@ func (r InMemoryCCIPReader) GetChainFeePriceUpdate(
 }
 
 func (r InMemoryCCIPReader) DiscoverContracts(
-	ctx context.Context, destChain cciptypes.ChainSelector,
+	ctx context.Context, allChains []cciptypes.ChainSelector,
```
asoliman92, line 141: `allChains` might be a bit misleading. What is `all`?

internal/plugincommon/discovery/discoverytypes/types.go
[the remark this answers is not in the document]
asoliman92: Maybe add a comment to explain the first string key here.

internal/plugincommon/discovery/processor.go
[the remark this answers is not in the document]
asoliman92: Can't we rely on the Keys of contractReaders to get the full list? (maybe needs filtering with homeChain as well?)

internal/plugincommon/discovery/processor.go
```diff
@@ -204,16 +239,46 @@ func (cdp *ContractDiscoveryProcessor) Outcome(
...
+		"fee quoter",
+		agg.feeQuoterAddrs,
+		fDestChainThresh,
+	)
+	cdp.lggr.Infow("Determined consensus fee quoter",
+		"feeQuoterConsensus", feeQuoterConsensus,
+		"feeQuoterAddrs", agg.feeQuoterAddrs,
+		"fDestChainThresh", fDestChainThresh,
+	)
+	if len(feeQuoterConsensus) == 0 {
```
asoliman92, line 277: Why only this condition for feeQuoter and not the others?

internal/plugincommon/discovery/processor.go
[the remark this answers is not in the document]
winder: ccipReader only has readers for supported chains. In todays integration test we have an all-to-all configuration, but production does not need to be all-to-all.

internal/plugincommon/discovery/processor_test.go
```diff
@@ -41,20 +43,31 @@ func TestContractDiscoveryProcessor_Observation_SupportsDest_HappyPath(t *testin
 	expectedOnRamp := map[cciptypes.ChainSelector][]byte{
 		source: []byte("onRamp"),
 	}
+	expectedFeeQuoter := map[cciptypes.ChainSelector][]byte{
+		source: []byte("onRamp"),
```
asoliman92, line 47: Why are we adding source as onRamp here?

internal/plugincommon/discovery/processor.go
```diff
@@ -114,6 +113,72 @@ func (cdp *ContractDiscoveryProcessor) ValidateObservation(
 	return nil
 }
 
+type aggObs struct {
+	onrampAddrs       map[cciptypes.ChainSelector][][]byte
+	feeQuoterAddrs    map[cciptypes.ChainSelector][][]byte
+	nonceManagerAddrs [][]byte
```
asoliman92, line 119: won't nonceManager and routers be easier to deal with based on ChainSelectors as well?

internal/mocks/inmem/ccipreader_inmem.go
winder: The OffRamp has a function to get source chain configurations by chain selector, it has all configurations regardless of whether or not the current node supports them. So currently the full list needs to come from outside of the ccip reader.

The onchain team is hopefully going to address this by providing a `getSourceChainSelectors()` function which would replace this parameter.

pkg/reader/ccip.go
```diff
@@ -550,45 +550,116 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
 	}
 
-	chains := maps.Keys(r.contractReaders)
+	// build up resp as we go.
+	var resp ContractAddresses
 
-	// OnRamps are in the offramp SourceChainConfig.
-	sourceConfigs, err := r.getSourceChainsConfig(ctx, chains)
+	// OnRamps are in the offRamp SourceChainConfig.
+	sourceConfigs, err := r.getOffRampSourceChainsConfig(ctx, allChains)
```
asoliman92, line 566: took me some time to get why we need the `allChains` here. Maybe add a doc to the function.

internal/plugincommon/discovery/processor.go
winder: The others should always be present. But maybe thats even more of a reason to add add the warning.

internal/plugincommon/discovery/discoverytypes/types.go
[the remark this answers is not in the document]
winder: Added a comment that directs you to the type I wanted to use here

internal/plugincommon/discovery/discoverytypes/types.go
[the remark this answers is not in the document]
winder: The token reader types are defined in the `exectypes` package, they should probably be moved into the reader package.

internal/plugincommon/discovery/processor.go
winder: added.

internal/plugincommon/discovery/processor.go
winder: updated.

pkg/reader/ccip.go
winder: Updated comment on DiscoverContract. But also, this should hopefully go away soon.

internal/plugincommon/discovery/processor.go
winder: added others.

internal/plugincommon/discovery/processor_test.go
winder: This is the fee quoter from the onramp, the other entry here is a fee quoter from the destination. I updated the names for clarity.

PR #172, opened 2024-09-27
CCIP-3431 Plugin e2e tests using USDC token transfers 

mateusz-sekara:
Basic e2e test for USDC transfer and token data observers.
* sends both regular and token messages
* only one USDC message is ready in the first round, so it's added to the report
* in the 2nd round, another attestation is ready so all messages are ready for the execution

There is a lot of code copy-pasted from `plugin_e2e_test`, but please don't focus on that. I want to have the test working first, merge that, and then as a follow-up extract e2e tools to the common place and de-duplicate the code between this test and plugin_e2e_test.

If I did extract the tooling in this PR it would be very hard to understand the test scenario, because PR would be very populated by the refactoring.

### DIFF
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -56,7 +56,7 @@ type Plugin struct {
 	contractsInitialized bool
 }
 
-func (p *Plugin) RreportingCfg() ocr3types.ReportingPluginConfig {
+func (p *Plugin) ReportingCfg() ocr3types.ReportingPluginConfig {
 	return p.reportingCfg
 }
 
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -56,6 +56,10 @@ type Plugin struct {
 	contractsInitialized bool
 }
 
+func (p *Plugin) RreportingCfg() ocr3types.ReportingPluginConfig {
+	return p.reportingCfg
+}
+
 func NewPlugin(
 	donID plugintypes.DonID,
 	reportingCfg ocr3types.ReportingPluginConfig,
--- a/execute/plugin_usdc_test.go
+++ b/execute/plugin_usdc_test.go
@@ -104,7 +104,20 @@ func Test_USDC_Transfer(t *testing.T) {
 	})
 	require.ElementsMatch(t, sequenceNumbers, []cciptypes.SeqNum{102, 103, 104, 105})
 	// Attestation data added to the USDC
-	require.NotEmpty(t, outcome.Report.ChainReports[0].OffchainTokenData[3])
+	//require.NotEmpty(t, outcome.Report.ChainReports[0].OffchainTokenData[3])
+
+	res, err = runner.RunRound(ctx)
+	require.NoError(t, err)
+	res, err = runner.RunRound(ctx)
+	require.NoError(t, err)
+	res, err = runner.RunRound(ctx)
+	require.NoError(t, err)
+	outcome, err = exectypes.DecodeOutcome(res.Outcome)
+	require.NoError(t, err)
+	sequenceNumbers = slicelib.Map(outcome.Report.ChainReports[0].Messages, func(m cciptypes.Message) cciptypes.SeqNum {
+		return m.Header.SequenceNumber
+	})
+	require.ElementsMatch(t, sequenceNumbers, []cciptypes.SeqNum{105})
 }
 
 type nodeSetup struct {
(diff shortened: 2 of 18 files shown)


execute/plugin.go
```diff
@@ -56,6 +56,10 @@ type Plugin struct {
 	contractsInitialized bool
 }
 
+func (p *Plugin) RreportingCfg() ocr3types.ReportingPluginConfig {
```
mateusz-sekara, line 59: Temporary, gonna remove that as followup once I extract the shared parts of tests

execute/plugin.go
```diff
@@ -56,6 +56,10 @@ type Plugin struct {
 	contractsInitialized bool
 }
 
+func (p *Plugin) ReportingCfg() ocr3types.ReportingPluginConfig {
```
mateusz-sekara, line 59: Temporary, gonna remove that as followup once I extract the shared parts of tests

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+	// Contract Discovery round.
+	res, err := runner.RunRound(ctx)
+	require.NoError(t, err)
+	outcome, err := exectypes.DecodeOutcome(res.Outcome)
+	require.NoError(t, err)
+	require.Equal(t, exectypes.Initialized, outcome.State)
+
+	// Round 1 - Get Commit Reports
+	// One pending commit report only.
+	// Two of the messages are executed which should be indicated in the Outcome.
```
makramkd, line 83: This naming is kinda confusing to me - messages are executed, but they're not, right, since the report only gets generated in the `Filter` state and transmitted on-chain thereafter?

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+	readermock "github.com/smartcontractkit/chainlink-ccip/mocks/pkg/contractreader"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/contractreader"
+	readerpkg "github.com/smartcontractkit/chainlink-ccip/pkg/reader"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+	plugintypes2 "github.com/smartcontractkit/chainlink-ccip/plugintypes"
+)
+
+const (
+	address = "0x888888cfaebbed5554c3f36bfbd233f822e9455f"
```
dimkouv, line 49: add some comment explaining what's this address, e.g. was it randomly picked or it means sth.

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+)
+
+func Test_USDC_Transfer(t *testing.T) {
+	ctx := context.Background()
```
dimkouv, line 53: You can use tests.Context(t) instead of bg

execute/plugin_usdc_test.go
makramkd: Or are we saying they're pre-executed as part of the test setup?

makramkd: Seems like they're pre-executed, but wondering whats the significance of that for this test. Or is it just how the other tests are set up as well?

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+						"status": "complete",
+						"attestation": "0x720502893578a89a8a87982982ef781c18b194"
+					}`)
+
+	// Run 3 more rounds to get all attestations
+	res, err = runner.RunRound(ctx)
+	require.NoError(t, err)
+	res, err = runner.RunRound(ctx)
+	require.NoError(t, err)
+	res, err = runner.RunRound(ctx)
```
dimkouv, line 126: you can use a loop maybe

dimkouv (approved):

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+	require.Len(t, outcome.Report.ChainReports, 0)
+	require.Len(t, outcome.PendingCommitReports, 1)
+
+	// Round 3 - Filter
+	// An execute report with the following messages executed: 102, 103, 104, 105.
+	res, err = runner.RunRound(ctx)
+	require.NoError(t, err)
+	outcome, err = exectypes.DecodeOutcome(res.Outcome)
+	require.NoError(t, err)
+	sequenceNumbers := slicelib.Map(outcome.Report.ChainReports[0].Messages, func(m cciptypes.Message) cciptypes.SeqNum {
```
makramkd, line 107: I really wish such basic utilities are part of https://pkg.go.dev/golang.org/x/exp/slices :-(

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+	require.ElementsMatch(t, sequenceNumbers, []cciptypes.SeqNum{102, 103, 104})
+	//Attestation data added to the USDC
+	require.NotEmpty(t, outcome.Report.ChainReports[0].OffchainTokenData[2])
+
+	server.AddResponse(
+		"0x70ef528624085241badbff913575c0ab50241e7cb6db183a5614922ab0bcba5d",
+		`{
+						"status": "complete",
+						"attestation": "0x720502893578a89a8a87982982ef781c18b194"
+					}`)
```
makramkd, line 119: Small nit: could use better formatting

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+				readIdentifier string,
+				confidenceLevel primitives.ConfidenceLevel,
+				params,
+				returnVal interface{},
+			) {
+				*returnVal.(*[]reader.OCR3ConfigWithMeta) = []reader.OCR3ConfigWithMeta{{}}
+			}).
+		Return(nil)
+
+	homeChain := reader.NewHomeChainConfigPoller(
```
makramkd, line 198: I know this isn't part of your PR but do we need to run the real home chain poller instead of just a mock?

execute/plugin_usdc_test.go
```diff
@@ -0,0 +1,515 @@
...
+				SourceChainSelector: src,
+				SequenceNumber:      seqNum,
+			},
+		},
+		Destination: dest,
+		Executed:    executed,
+	}
+}
+
+func setupSimpleTest(
```
makramkd, line 226: Small description of test scenario would be helpful, e.g "This test sets up 6 messages, 2 of which are already executed, and 3 of which have USDC attestions, 2 of which are not ready immediately", etc.

makramkd (approved):

execute/plugin_usdc_test.go
mateusz-sekara: This is how it's setup in other places

PR #173, opened 2024-09-27
Fix flaky RMN test

dimkouv:
Verified that the issue was in the test logic and not the RMN logic itself.

Test never failed locally after running with `-count=1000`.
So issue is on lower-tier hardware, like the gh workflow one.

Checked the failed test logs, test logic was waiting for invalid request ids.
Adjusted test code in `waitForReportSignatureRequestsToBeSent` to not wait for `RMN responses` but wait for `RMN signature report responses`. Because in the lower tier hardware it might receive `Observation Responses` from previous step which took long to get processed.

### DIFF
--- a/commit/merkleroot/rmn/controller.go
+++ b/commit/merkleroot/rmn/controller.go
@@ -633,6 +633,8 @@ func (c *controller) listenForRmnReportSignatures(
 						ReportSignatureRequest: reportSigReq,
 					},
 				}
+
+				c.lggr.Infow("sending report signature request", "node", node.ID, "requestID", req.RequestId)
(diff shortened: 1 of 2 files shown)


0xnogo (approved):

commit/merkleroot/rmn/controller.go
0xnogo, line 1: In case of network delay or any issue on a production setup, can the controller mixed observation with report signature responses?

dimkouv: Good point. There are validations that catch all this scenarios.
If we receive a response that we are not waiting for it's simply dropped.

PR #176, opened 2024-09-27
Deterministic RMN observation requests

dimkouv:
Part of RMN Validity Criteria

> Fixed dest lane update requests (in ObservationRequest), fixed dest lane updates (in Observation/Report) must be in strictly (no duplicates) ascending order of source chain selector.


commit/merkleroot/rmn/controller_test.go
```diff
@@ -312,6 +316,17 @@ func (ts *testSetup) waitForReportSignatureRequestsToBeSent(
...
+				for _, ao := range req.GetReportSignatureRequest().AttributedSignedObservations {
+					assert.Greater(t, ao.SignerNodeIndex, prevRmnNodeIDX)
+					prevRmnNodeIDX = ao.SignerNodeIndex
+
+					prevSrcChain := uint64(0)
+					for _, lu := range ao.SignedObservation.Observation.FixedDestLaneUpdates {
+						assert.Greater(t, lu.LaneSource.SourceChainSelector, prevSrcChain)
+						prevSrcChain = lu.LaneSource.SourceChainSelector
+					}
+				}
```
makramkd, line 329: This asserts that the signed observations are sorted right? I think you can use https://pkg.go.dev/sort#SliceIsSorted and do `assert.True(t, sort.SliceIsSorted(...))` to make that clear?

PR #180, opened 2024-09-27
Dk/debug 2

dimkouv:


execute/tokendata/usdc/usdc_int_test.go
makramkd, line 1: Should not be deleted

internal/plugincommon/discovery/discoverytypes/types.go
makramkd, line 1: should be OK

dimkouv:
closing, figured out what's wrong

PR #183, opened 2024-09-29
RMN Controller Improvements

dimkouv:
- Improved logging.
- In the Outcome phase exclude source chains that are not signed to prevent on-chain revert.
- Add comments explaining critical components.
- Trigger additional RMN requests instantly on the first error instead of waiting for timer expiration.
- Fix bug where duplicate lane updates were added in the same RMN report prior to verifying sigs.
- Cleaner less complicated logic.
- Handle edge case where some RMN node mix valid and invalid roots for different chains in their observation.
- Stop earlier after all RMN responses are received instead of waiting for timer expiration (on signature requests).

### DIFF
--- a/commit/merkleroot/outcome.go
+++ b/commit/merkleroot/outcome.go
@@ -164,6 +164,8 @@ func buildReport(
 		for _, root := range roots {
 			if signedRoots.Contains(root) {
 				rootsToReport = append(rootsToReport, root)
+			} else {
+				lggr.Warnw("skipping merkle root not signed by RMN", "root", root)
 			}
 		}
 		roots = rootsToReport
--- a/commit/merkleroot/outcome.go
+++ b/commit/merkleroot/outcome.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"time"
 
+	mapset "github.com/deckarep/golang-set/v2"
 	"golang.org/x/exp/maps"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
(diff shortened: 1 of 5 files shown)


commit/merkleroot/outcome.go
```diff
@@ -138,15 +139,34 @@ func buildReport(
...
-		lggr.Warn("RMNSignatures are nil")
-	} else {
+	if q.RMNSignatures != nil { // TODO: should never be nil, error after e2e RMN integration.
 		parsedSigs, err := rmn.NewECDSASigsFromPB(q.RMNSignatures.Signatures)
 		if err != nil {
-			lggr.Errorw("Failed to parse RMN signatures", "error", err)
-		} else {
-			sigs = parsedSigs
+			lggr.Errorw("Failed to parse RMN signatures returning an empty outcome", "error", err)
+			return Outcome{}
```
makramkd, line 146: Should this situation return an error to the top-level plugin?

commit/merkleroot/outcome.go
```diff
@@ -138,15 +139,34 @@ func buildReport(
...
 		}
+		sigs = parsedSigs
+
+		signedRoots := mapset.NewSet[cciptypes.MerkleRootChain]()
+		for _, laneUpdate := range q.RMNSignatures.LaneUpdates {
```
makramkd, line 151: I'm not sure if we do these validations elsewhere, but one thing I can think of is do we validate that RMN nodes don't send overlapping intervals / different roots for same interval / etc. ?

commit/merkleroot/rmn/controller.go
```diff
@@ -323,14 +347,53 @@ func (c *controller) listenForRmnObservationResponses(
...
+		sort.Slice(values, func(i, j int) bool { return values[i] < values[j] })
+		if values[len(values)-1] < minObservers {
+			return false
+		}
+	}
+
+	return true
+}
+
 // nolint:gocyclo // todo
```
makramkd, line 384: Didn't notice this `nolint` before, is there any way we can reduce cyclomatic complexity of this func?

commit/merkleroot/rmn/controller.go
```diff
@@ -463,70 +625,22 @@ func (c *controller) getRmnReportSignatures(
...
-	return &ReportSignatures{
-		Signatures:  ecdsaSignatures,
-		LaneUpdates: fixedDestLaneUpdates,
-	}, nil
+	return requestIDs, signersRequested, nil
 }
 
-// reportSigWithNodeID is a helper struct to store the report signature and the node ID that provided it.
-type reportSigWithNodeID struct {
+// reportSigWithSignerAddress is a helper struct to store the report signature and the address of the SK that signed it.
```
makramkd, line 642:
```suggestion
// reportSigWithSignerAddress is a helper struct to store the report signature and the address of the private key that signed it.
```

commit/merkleroot/outcome.go
```diff
@@ -138,15 +139,34 @@ func buildReport(
...
+		}
+
+		// Only report roots that are present in RMN signatures.
+		rootsToReport := make([]cciptypes.MerkleRootChain, 0)
+		for _, root := range roots {
+			if signedRoots.Contains(root) {
+				rootsToReport = append(rootsToReport, root)
+			}
+		}
+		roots = rootsToReport
```
makramkd, line 169: Lets log the excluded roots, if any

commit/merkleroot/outcome.go
dimkouv: Yes we handle all this

commit/merkleroot/rmn/controller.go
dimkouv: It's a bit hard actually, i'll try in a follow up.
I enable gocyclo linter recently.

commit/merkleroot/outcome.go
dimkouv: The way plugin is designed, it's the same thing, just that plugin would log the error.

makramkd (approved):

PR #185, opened 2024-09-30
Allow zero value for data availability

asoliman92:


makramkd (approved):

commit/chainfee/validate_observation.go
```diff
@@ -39,35 +37,39 @@ func (p *processor) ValidateObservation(
...
 	}
-	if b.Cmp(zero) < 0 || b.Cmp(zero) == 0 {
+	if b.Cmp(zero) < 0 {
 		return fmt.Errorf("%s must be positive", name)
 	}
+	if !allowZero && b.Cmp(zero) == 0 {
+		return fmt.Errorf("%s must be non zero", name)
+	}
+
 	return nil
```
dimkouv, line 73: ValidateBigInt - returns an error if number is negative or if number is zero and we don't allow zero


I think instead of
```
if err = validate(v, "some name", true); err != nil { return err}
```

i recommend
```
if v.Cmp(zero) <= 0 { return fmt.Errorf("some name...") } // not positive
if v.Cmp(zero) < 0 { return fmt.Errorf("some name...") } // is negative
```

IMO the extra function makes things more complex to understand.

PR #187, opened 2024-09-30
init wlog package

dimkouv:
Usage:

```
w := NewWrappedLogger(baseLogger)
w.WithField("foo", "bar")

w.Info("something") // something foo=bar	
```

WrappedLogger is compliant with cl-common Logger and can be provided to functions like that, meaning there won't be any breaking changes.

Comparing output with set fields against existing logger:

```
    wrappedlogger_test.go:18: ------------------
    logger.go:146: 17:15:42.822006000	DEBUG	some messagesome other message
    logger.go:146: 17:15:42.822274000	DEBUG	some messagesome other message str=map[a:1 b:123] foo=bar baz=123
    wrappedlogger_test.go:23: ------------------
    logger.go:146: 17:15:42.822284000	DEBUG	hello 123 world
    logger.go:146: 17:15:42.822291000	DEBUG	hello 123 world foo=bar baz=123 str=map[a:1 b:123]
    wrappedlogger_test.go:28: ------------------
    logger.go:146: 17:15:42.822299000	DEBUG	hello friend	{"foo": 123, "bar": "baz"}
    logger.go:146: 17:15:42.822338000	DEBUG	hello friend foo=bar baz=123 str=map[a:1 b:123]	{"foo": 123, "bar": "baz"}
```

Looking for some early feedback.
Fields will be ordered, ignore them being unordered in the output above.

### DIFF
--- a/internal/libs/wlog/wrappedlogger.go
+++ b/internal/libs/wlog/wrappedlogger.go
@@ -1,0 +1,54 @@
+package wlog
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+)
+
+type WrappedLogger struct {
+	baseLogger logger.Logger
+	data       map[string]any
+}
+
+func NewWrappedLogger(baseLogger logger.Logger) *WrappedLogger {
+	return &WrappedLogger{
+		baseLogger: baseLogger,
+		data:       make(map[string]any),
+	}
+}
+
+func (w *WrappedLogger) WithField(key string, value any) {
+	w.data[key] = value
+}
+
+func (w *WrappedLogger) dataToString(prefix, suffix string) string {
+	extraData := make([]string, 0, len(w.data))
+	for k, v := range w.data {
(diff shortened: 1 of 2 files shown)


internal/libs/wlog/wrappedlogger.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func NewWrappedLogger(baseLogger logger.Logger) *WrappedLogger {
+	return &WrappedLogger{
+		baseLogger: baseLogger,
+		data:       make(map[string]any),
+	}
+}
+
+func (w *WrappedLogger) WithField(key string, value any) {
```
makramkd, line 22: I would love to be able to restrict this for only "simple" types like integers and strings. It would be great if, at a glance, I know the oracle ID and e.g, current state of the plugin.

```
[oracle 1] [state Filter] building report with 10 msgs
```

And in the log fields I have more details.

Then again, I think we could write some tools that do this kind of log analysis / transformation after pulling down the logs into a text file. The only issue with this is that it may not be possible due to grafana limitations etc.

PR #190, opened 2024-10-01
Fix Protobuf Generation by Specifying Full Path to protoc

0xnogo:
This PR updates the Makefile to specify the full path to the protoc binary, ensuring that the correct version (libprotoc 28.0) is used during protobuf generation. This change avoids conflicts when multiple protoc versions are present in the system PATH.


Makefile
```diff
@@ -14,6 +14,8 @@ else
 endif
 PROTOC_URL := https://github.com/protocolbuffers/protobuf/releases/download/v28.0/$(PROTOC_ZIP)
 
+PROTOC_BIN := /usr/local/bin/protoc
```
makramkd, line 17: Isn't this macOS specific?

0xnogo: this is the path where we put the bin (see line 50). Btw i'll also replace the path there with this variable

0xnogo: True. The other option is to provide it through environment variable or default to PROTOC_BIN.

For linux it could be something like:
```
make generate-protobuf PROTOC_BIN=/usr/bin/protoc
```

dimkouv (approved):

PR #191, opened 2024-10-01
CCIP-3593 Adding more logs to non-trivial places in USDC token processing

mateusz-sekara:


execute/tokendata/usdc/http.go
```diff
@@ -108,14 +116,19 @@ func (r httpResponse) attestationToBytes() (cciptypes.Bytes, error) {
 
 func (h *httpClient) Get(ctx context.Context, messageHash cciptypes.Bytes32) (cciptypes.Bytes, HTTPStatus, error) {
 	// Terminate immediately when rate limited
-	if h.inCoolDownPeriod() {
+	if coolDown, duration := h.inCoolDownPeriod(); coolDown {
+		h.lggr.Errorw(
```
asoliman92, line 120: I don't think this should be an error, maybe a warning?

asoliman92 (approved):

execute/tokendata/usdc/http.go
mateusz-sekara: We have a self-rate limitting process in place which should always prevent from reaching coolDown. Cooldown is very destructive because the penalty of not respecting rate limits is high (e.g. backoff for 5 minutes).

mateusz-sekara: I don't have strong opinion on the logging level, but I'm okay with warn I guess

asoliman92: No strong opinion either. Feel free to use what you think is better.

PR #193, opened 2024-10-01
RMNRemoteConfig integration in Merkleroot processor

0xnogo:
Adding `RMNRemote` config in the the Merkleroot processor.

The process should go as:
Round 0
1. Query: PrevOutcome is empty, then the Query is empty
2. Observation: each node observe the RMNRemote data by calling the contract. The address should have been discovered as part of the Discovery processor
3. Outcome: reach consensus and built the Outcome with the RMNRemote config. The fetching is done as part of the `ReportIntervalsSelected` state.

Round 1
1. Query: next state is BuildingReport, so we should be able to call rmn to compute the signature with the RMNRemote config data
2. Observation: same for the report signing
3. Outcome: no change
4. Report: report submission

The RMN controller is not yet using the `RMNRemoteCfg`.


commit/merkleroot/rmn/config.go
```diff
@@ -33,7 +34,8 @@ type RMNNodeInfo struct {
 	IsSigner                  bool
 	SignReportsAddress        cciptypes.Bytes
 	SignObservationsPublicKey *ed25519.PublicKey // offChainPublicKey
-	SignObservationPrefix     string             // e.g. "chainlink ccip 1.6 rmn observation"
+	// TODO: clarify this field
+	SignObservationPrefix string // e.g. "chainlink ccip 1.6 rmn observation"
```
0xnogo, line 38: @dimkouv this field is not fetched by the `GetRMNRemoteCfg`. Is it something offline?

pkg/reader/ccip.go
```diff
@@ -550,6 +551,63 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
+	err = r.contractReaders[destChainSelector].ExtendedGetLatestValue(
+		ctx,
+		consts.ContractNameRMNRemote,
+		consts.MethodNameGetVersionedConfig,
+		primitives.Unconfirmed,
+		map[string]any{},
+		&vc,
+	)
+	if err != nil {
+		return rmntypes.RMNRemoteConfig{}, fmt.Errorf("failed to get RMNRemote config: %w", err)
```
dimkouv, line 577: General tip. Don't say `failed to` in the errors.
Otherwise with wrapping it becomes:
```
failed to do this: failed to do that: failed to foo
```

we can just do
```
do this: do that: foo
```

commit/merkleroot/observation.go
```diff
@@ -58,15 +59,9 @@ func (w *Processor) verifyQuery(ctx context.Context, prevOutcome Outcome, q Quer
...
-	// If we are in the BuildingReport state, and we are not retrying RMN signatures in the next round, we expect RMN
-	// signatures to be provided by the leader.
-	if nextState == BuildingReport && !q.RetryRMNSignatures && q.RMNSignatures == nil {
-		return fmt.Errorf("RMN signatures are required in the BuildingReport state but not provided by leader")
-	}
-
-	// If we are not in the BuildingReport state, we do not expect RMN signatures to be provided.
-	if nextState != BuildingReport && q.RMNSignatures != nil {
-		return fmt.Errorf("RMN signatures are provided but not expected in the %d state", nextState)
+	err := checkRMNRequirements(nextState, q, prevOutcome)
```
dimkouv, line 62: nice, clean

commit/merkleroot/observation.go
```diff
@@ -338,6 +358,28 @@ func (o ObserverImpl) computeMerkleRoot(ctx context.Context, msgs []cciptypes.Me
...
+		return rmntypes.RMNRemoteConfig{}
+	}
+
+	if !supportsDestChain {
+		return rmntypes.RMNRemoteConfig{}
+	}
+
+	rmnRemoteCfg, err := o.ccipReader.GetRMNRemoteConfig(ctx, dstChain)
+	if err != nil {
+		o.lggr.Warnw("call to GetRMNRemoteConfig failed", "err", err)
```
dimkouv, line 377: I believe this two should be `Errorw` not `Warnw`

commit/plugin_e2e_test.go
```diff
@@ -75,6 +77,21 @@ func TestPlugin_E2E_AllNodesAgree(t *testing.T) {
 		sourceChain2: 19, // no new msg, still on 19
 	}
 
+	rmnReportCfg := rmntypes.RMNRemoteConfig{
```
dimkouv, line 80: This exact same `rmnReportCfg` is defined in 3 places. Maybe just declare it once and use random values if possible.

commit/merkleroot/outcome.go
```diff
@@ -112,10 +117,17 @@ func reportRangesOutcome(
 		return offRampNextSeqNums[i].ChainSel < offRampNextSeqNums[j].ChainSel
 	})
 
+	if observedRMNRemoteConfig[dstChain].IsEmpty() {
+		lggr.Warn("RMNRemoteConfig is nil")
```
dimkouv, line 121: Isn't it supposed to be empty in some specific states?
If yes, let's only `Warn` in the state that we actually expect to have it.

commit/merkleroot/outcome.go
```diff
@@ -245,12 +257,16 @@ func getConsensusObservation(
 			fmt.Errorf("no consensus value for fDestChain, destChain: %d", destChain)
 	}
 
+	// convert aggObs.RMNRemoteConfigs to a map of RMNRemoteConfigs
+	rmnRemoteConfigs := map[cciptypes.ChainSelector][]rmntypes.RMNRemoteConfig{destChain: aggObs.RMNRemoteConfigs}
```
dimkouv, line 261: I think the plugin only cares about the destination chain remote config.
Maybe no need to keep a map.

commit/merkleroot/rmn/types/config.go
```diff
@@ -30,17 +30,29 @@ type RMNHomeNodeInfo struct {
...
+	ContractAddress  cciptypes.Bytes       `json:"contractAddress"`
+	ConfigDigest     cciptypes.Bytes32     `json:"configDigest"`
+	Signers          []RMNRemoteSignerInfo `json:"signers"`
+	MinSigners       uint64                `json:"minSigners"`
+	ConfigVersion    uint32                `json:"configVersion"`
+	RmnReportVersion string                `json:"rmnReportVersion"` // e.g., "RMN_V1_6_ANY2EVM_REPORT"
+}
+
+func (r RMNRemoteConfig) IsEmpty() bool {
+	return len(r.ContractAddress) == 0 ||
```
dimkouv, line 42: can we use something like `r == RMNRemoteConfig{}` to check if it's empty?
to prevent adding another field and forgetting to update this function.

commit/merkleroot/rmn/types/config.go
```diff
@@ -30,17 +30,29 @@ type RMNHomeNodeInfo struct {
...
+		r.RmnReportVersion == ""
 }
 
 // RMNRemoteSignerInfo contains information about a signer from the RMNRemote contract.
 type RMNRemoteSignerInfo struct {
-	OnchainPublicKey      cciptypes.Bytes // The signer's onchain address, used to verify report signature
-	NodeIndex             uint64          // The index of the node in the RMN config
-	SignObservationPrefix string          // The prefix of the observation to sign
+	// The signer's onchain address, used to verify report signature
+	OnchainPublicKey cciptypes.Bytes `json:"onchainPublicKey"`
```
dimkouv, line 53: Should be renamed I think `OnchainPublicKey` -> `SignerAddress` or something. Let's use a name similar to what the contract uses.

commit/merkleroot/outcome.go
dimkouv: I see, because you wanted to reuse `GetConsensusMap`

dimkouv: I recommend adding some tests to `GetConsensusMap` with this specific map we provide here.

dimkouv (commented):
Overall logic looks good, added comments about code improvements.

commit/merkleroot/query.go
```diff
@@ -48,6 +48,10 @@ func (w *Processor) Query(ctx context.Context, prevOutcome Outcome) (Query, erro
 		})
 	}
 
+	if prevOutcome.RMNRemoteCfg.IsEmpty() {
```
dimkouv, line 51: We can probably do this validation before everything else.

pkg/reader/ccip.go
0xnogo: Makes sense, thanks for the tips

commit/merkleroot/outcome.go
0xnogo: Valid comment. This method is executed in the context of the `SelectingRangesForReport` state, which mean that we should always have an rmn remote config.

0xnogo: Which means that we should always warn if we don't have it.

commit/merkleroot/rmn/types/config.go
0xnogo: Excellent point

commit/merkleroot/observation.go
```diff
@@ -338,6 +358,28 @@ func (o ObserverImpl) computeMerkleRoot(ctx context.Context, msgs []cciptypes.Me
...
+		return rmntypes.RMNRemoteConfig{}
+	}
+	return rmnRemoteCfg
```
makramkd, line 380: This is a nit, but I think all of the "do I support this chain" logic is already done in the CCIP reader. I think this code can just be simplified to call the CCIP reader and checking `ErrContractReaderNotFound` or whatever the error is.

```golang
rmnRemoteCfg, err := o.ccipReader.GetRMNRemoteConfig(ctx, dstChain)
if err != nil {
  if errors.Is(err, reader.ErrContractReaderNotFound) {
    // don't support dest chain
    return rmntypes.RMNRemoteConfig{}
  }
  // legitimate error, we should log it.
  o.lggr.Warnw("call to GetRMNRemoteConfig failed", "err", err)
  return rmntypes.RMNRemoteConfig{}
}

return rmnRemoteCfg
```

commit/merkleroot/types.go
```diff
@@ -73,10 +80,14 @@ func aggregateObservations(aos []plugincommon.AttributedObservation[Observation]
 				append(aggObs.OffRampNextSeqNums[seqNumChain.ChainSel], seqNumChain.SeqNum)
 		}
 
+		// RMNRemoteConfig
+		aggObs.RMNRemoteConfigs = append(aggObs.RMNRemoteConfigs, obs.RMNRemoteConfig)
```
makramkd, line 84: The nodes that can't read the destination chain will return empty `rmntypes.RMNRemoteConfig{}` objects, should we filter those out before aggregating?

commit/merkleroot/rmn/types/config.go
```diff
@@ -30,17 +30,29 @@ type RMNHomeNodeInfo struct {
 
 // RMNRemoteConfig contains the configuration fetched from the RMNRemote contract.
 type RMNRemoteConfig struct {
```
makramkd, line 32: We should drop all of the `RMN` prefixes from these types since the package is already scoped to RMN.

commit/merkleroot/rmn/types/config.go
makramkd: Its tricky because `r == RMNRemoteConfig{}` will probably return true if they're all exactly equal, but if there is a slice of 0 length (declared via `var slice cciptypes.Bytes` and another slice that's just nil `var slice cciptypes.Bytes = nil` they wouldn't be equal, which is funky.

commit/merkleroot/rmn/types/config.go
makramkd: Contract uses `onchainPublicKey` which is why I wanted to rename to this https://github.com/smartcontractkit/ccip/blob/7d4f8e56e27faba2007edff7b6623ac7c29ec94a/contracts/src/v0.8/ccip/rmn/RMNRemote.sol#L43

commit/merkleroot/types.go
makramkd: We saw this issue with discovery where nil slices were not filtered out.

commit/merkleroot/outcome.go
```diff
@@ -245,12 +257,16 @@ func getConsensusObservation(
...
+
 	// Get consensus using strict 2f+1 threshold.
 	twoFPlus1 := consensus.MakeMultiThreshold(fChains, consensus.TwoFPlus1)
```
makramkd, line 264: Confusing var names and comments lead to confusing code - my suggestion:

```suggestion
	// Get consensus using strict 2fChain+1 threshold.
	twoFChainPlus1 := consensus.MakeMultiThreshold(fChains, consensus.TwoFPlus1)
```

commit/merkleroot/types.go
makramkd: I wonder if this issue applies to all the aggregateObservation functions...

commit/merkleroot/observation.go
0xnogo: A node not supporting the dest chain is a valid case right? Hence only warning. But for the second one, i think you are right, it is a legitimate error.

commit/merkleroot/observation.go
0xnogo: Nice one

pkg/reader/ccip.go
makramkd: @dimkouv I wonder if we can create a linter just for that.

makramkd: I can look into it as a side project 😛

commit/merkleroot/outcome.go
0xnogo: Done

commit/merkleroot/types.go
0xnogo: I am not sure about the other element of the observation but I added an if condition to check whether the RMNRemoteConfig is empty for now. It makes sense to filter out the nil values.

commit/merkleroot/rmn/types/config.go
0xnogo: I like that this naming is closely following the contract. But I don't have any straight opinion, so I don't mind removing the RMN prefix

commit/merkleroot/outcome.go
```diff
@@ -71,12 +72,16 @@ func reportRangesOutcome(
...
+	dstChain cciptypes.ChainSelector,
 ) Outcome {
 	rangesToReport := make([]plugintypes.ChainRange, 0)
 
 	observedOnRampMaxSeqNumsMap := consensusObservation.OnRampMaxSeqNums
 	observedOffRampNextSeqNumsMap := consensusObservation.OffRampNextSeqNums
+	observedRMNRemoteConfig := consensusObservation.RMNRemoteConfig
+
 	offRampNextSeqNums := make([]plugintypes.SeqNumChain, 0)
+	rmnRemoteConfig := rmntypes.RemoteConfig{}
```
makramkd, line 84: This looks pretty misplaced over here, lets move it closer to where its assigned.

makramkd: Prefer `var rmnRemoteConfig rmntypes.RemoteConfig` for declarations.

commit/merkleroot/rmn/config.go
makramkd: This feels like something that might need to be in either the Commit offchain config or the shared config b/w RMN and CCIP.

commit/merkleroot/rmn/types/config.go
```diff
@@ -12,35 +12,47 @@ import (
...
+	RmnReportVersion string             `json:"rmnReportVersion"` // e.g., "RMN_V1_6_ANY2EVM_REPORT"
 }
 
-// RMNRemoteSignerInfo contains information about a signer from the RMNRemote contract.
-type RMNRemoteSignerInfo struct {
-	OnchainPublicKey      cciptypes.Bytes // The signer's onchain address, used to verify report signature
-	NodeIndex             uint64          // The index of the node in the RMN config
-	SignObservationPrefix string          // The prefix of the observation to sign
+func (r RemoteConfig) IsEmpty() bool {
+	return (len(r.ContractAddress) == 0 || r.ContractAddress == nil) &&
```
makramkd, line 42: Checking `len(r.ContractAddress) == 0` is equivalent to checking for nilness so the nil check is redundant.

commit/merkleroot/types.go
```diff
@@ -73,10 +80,16 @@ func aggregateObservations(aos []plugincommon.AttributedObservation[Observation]
 				append(aggObs.OffRampNextSeqNums[seqNumChain.ChainSel], seqNumChain.SeqNum)
 		}
 
+		// RMNRemoteConfig
+		if !obs.RMNRemoteConfig.IsEmpty() {
+			aggObs.RMNRemoteConfigs = append(aggObs.RMNRemoteConfigs, obs.RMNRemoteConfig)
```
makramkd, line 85: Not sure if we do this elsewhere but where do we validate that RMNRemoteConfig is only fetched by dest readers?

commit/merkleroot/validate_observation.go
```diff
@@ -50,6 +51,10 @@ func (w *Processor) ValidateObservation(
 		return fmt.Errorf("failed to validate OffRampNextSeqNums: %w", err)
 	}
 
+	if err := validateRMNRemoteConfig(ao.OracleID, supportsDestChain); err != nil {
```
makramkd, line 54: Noice, see it here.

commit/merkleroot/validate_observation.go
```diff
@@ -50,6 +51,10 @@ func (w *Processor) ValidateObservation(
...
+		return fmt.Errorf("failed to validate RMNRemoteConfig: %w", err)
```
makramkd, line 55:
```suggestion
		return fmt.Errorf("validate RMNRemoteConfig: %w", err)
```

commit/merkleroot/validate_observation.go
```diff
@@ -129,6 +134,17 @@ func validateObservedOffRampMaxSeqNums(
 	return nil
 }
 
+func validateRMNRemoteConfig(
+	observer commontypes.OracleID,
+	supportsDestChain bool,
+) error {
+	if !supportsDestChain {
+		return fmt.Errorf("observer %d does not support dest chain, but has observed a RMNRemoteConfig", observer)
```
makramkd, line 142:
```suggestion
		return fmt.Errorf("oracle %d does not support dest chain, but has observed a RMNRemoteConfig", observer)
```

commit/merkleroot/validate_observation.go
makramkd: Do we have tests for this new addition?

pkg/reader/ccip.go
```diff
@@ -898,5 +956,33 @@ func (r *ccipChainReader) getOnRampDynamicConfigs(
...
+	RMNHomeContractConfigDigest []byte   `json:"rmnHomeContractConfigDigest"`
+	Signers                     []signer `json:"signers"`
+	MinSigners                  uint64   `json:"minSigners"`
+}
+
+// versionnedConfig is used to parse the response from the RMNRemote contract's getVersionedConfig method.
+// See: https://github.com/smartcontractkit/ccip/blob/ccip-develop/contracts/src/v0.8/ccip/rmn/RMNRemote.sol#L167-L169
+//
+//nolint:lll // It's a URL.
+type versionnedConfig struct {
```
makramkd, line 982:
```suggestion
type versionedConfig struct {
```

pkg/reader/ccip.go
```diff
@@ -550,6 +551,63 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
+		consts.MethodNameGetVersionedConfig,
+		primitives.Unconfirmed,
+		map[string]any{},
+		&vc,
+	)
+	if err != nil {
+		return rmntypes.RemoteConfig{}, fmt.Errorf("get RMNRemote config: %w", err)
+	}
+
+	var dh string
```
makramkd, line 580: Hm I'm not sure this'll work, the contract returns `bytes32`, which means `cciptypes.Bytes32` or `[32]byte` depending on your needs (i.e JSON serialization)

pkg/reader/ccip.go
```diff
@@ -550,6 +551,63 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
+		})
+	}
+
+	return rmntypes.RemoteConfig{
+		ContractAddress:  contractAddress,
+		ConfigDigest:     cciptypes.Bytes32(vc.Config.RMNHomeContractConfigDigest),
+		Signers:          signers,
+		MinSigners:       vc.Config.MinSigners,
+		ConfigVersion:    vc.Version,
+		RmnReportVersion: dh,
```
makramkd, line 607: I think there's a mismatch b/w what you're expecting and what the contract returns here.

```solidity
bytes32 private constant RMN_V1_6_ANY2EVM_REPORT = keccak256("RMN_V1_6_ANY2EVM_REPORT");

function getReportDigestHeader() external pure returns (bytes32 digestHeader) {
  return RMN_V1_6_ANY2EVM_REPORT;
}
```

`RMN_V1_6_ANY2EVM_REPORT` is a hash - so this field should be `cciptypes.Bytes32` ideally.

0xnogo: Good catch

commit/merkleroot/outcome.go
0xnogo: The idea was to have it declare where all the returned values where. But I got your point, moving it closer to where it's being used

commit/merkleroot/rmn/config.go
0xnogo: I see. Makes sense I was confused if we considered it to be in the contracts. I could maybe have it as part of this [PR](https://github.com/smartcontractkit/chainlink-ccip/pull/195) where I am removing the CommonPluginConfig in favor of the Offchain config.

commit/merkleroot/rmn/types/config.go
dimkouv: Isn't that wrong though?

PR #195, opened 2024-10-02
Remove CommitPluginConfig and use OffchainConfig instead

0xnogo:
This PR will need some modification in the test setup in chainlink.


commit/factory.go
```diff
@@ -28,9 +27,9 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
 )
 
-const maxReportTransmissionCheckAttempts = 5
 const maxQueryLength = 1024 * 1024 // 1MB
-const rmnEnabled = false
+// const maxReportTransmissionCheckAttempts = 5
```
0xnogo, line 31: We need to have some default value setup. I'll figure out where to put those

0xnogo: As discussed offline with @makramkd, added a method `ApplyDefaults`.

pluginconfig/commit.go
```diff
@@ -153,6 +154,14 @@ func (c CommitOffchainConfig) Validate() error {
 		}
 	}
```
makramkd, line 156: I was thinking we'd call `ApplyDefaults` here in `Validate` since `Validate` is more likely to be called. And defaults are only applied on fields that need default values - wdyt?

makramkd:
Integration test is failing with this log:

```
failed to process diff: failed to create CCIP commit oracle: failed to create readers and writers: failed to validate commit offchain config: newMsgScanBatchSize not set
```

I think if you apply the above suggestion (calling `applyDefaults` inside `Validate`) it should fix things by setting the right values.

pluginconfig/commit.go
```diff
@@ -9,58 +9,23 @@ import (
...
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 
+	"github.com/smartcontractkit/chainlink-common/pkg/merklemulti"
+
 	commonconfig "github.com/smartcontractkit/chainlink-common/pkg/config"
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 )
 
 // EvmDefaultMaxMerkleTreeSize is the default number of max new messages to put in a merkle tree.
 // We use this default value when the config is not set for a specific chain.
```
makramkd, line 21: Lets move this comment closer to the var declaration

pluginconfig/commit.go
```diff
@@ -9,58 +9,23 @@ import (
...
-	if c.RMNEnabled && c.RMNSignaturesTimeout == 0 {
-		return fmt.Errorf("rmnSignaturesTimeout not set")
-	}
-
-	return c.OffchainConfig.Validate()
-}
+const (
+	defaultRMNSignaturesTimeout               = 5 * time.Second
+	defaultNewMsgScanBatchSize                = 256
+	EvmDefaultMaxMerkleTreeSize               = merklemulti.MaxNumberTreeLeaves
```
makramkd, line 25: To match the naming of other variables, perhaps:

```suggestion
	DefaultMaxMerkleTreeSize               = merklemulti.MaxNumberTreeLeaves
```

makramkd: This is exported because it seems to be checked again in plugin.go but I'm not sure thats necessary if we always call `Validate` and that calls `applyDefaults`

pluginconfig/commit.go
0xnogo: If we want to apply both default application and validation, we probably also need to rename the function to something like `ApplyDefaultsAndValidate`. This will avoid any surprise in the outcome.

makramkd: Yeah, that sounds good to me.

pluginconfig/commit.go
0xnogo: I don't think this comment is adding much meaning to the var name. I'll just remove it.

pluginconfig/commit.go
0xnogo: Making it private and renamed using the same pattern

pluginconfig/commit.go
```diff
@@ -130,9 +95,49 @@ type CommitOffchainConfig struct {
 	// This will typically be an arbitrum testnet/mainnet chain depending on
 	// the deployment.
 	PriceFeedChainSelector cciptypes.ChainSelector `json:"tokenPriceChainSelector"`
+
+	// NewMsgScanBatchSize is the number of max new messages to scan, typically set to 256.
+	NewMsgScanBatchSize int `json:"newMsgScanBatchSize"`
+
+	// The maximum number of times to check if the previous report has been transmitted
+	MaxReportTransmissionCheckAttempts uint
```
makramkd, line 103: Lets add a `json` tag for this to be consistent

pluginconfig/commit.go
```diff
@@ -9,58 +9,23 @@ import (
...
-
-	if c.RMNEnabled && c.RMNSignaturesTimeout == 0 {
-		return fmt.Errorf("rmnSignaturesTimeout not set")
-	}
-
-	return c.OffchainConfig.Validate()
-}
+const (
+	defaultRMNSignaturesTimeout               = 5 * time.Second
+	defaultNewMsgScanBatchSize                = 256
```
makramkd, line 23: I think 256 here is just `merklemulti.MaxNumberTreeLeaves`, lets use that constant instead since its more informative.

pluginconfig/commit.go
```diff
@@ -130,9 +95,49 @@ type CommitOffchainConfig struct {
...
+	RMNEnabled bool `json:"rmnEnabled"`
+
+	// MaxMerkleTreeSize is the maximum size of a merkle tree to create prior to calculating the merkle root.
+	// If for example in the next round we have 1000 pending messages and a max tree size of 256, only 256 seq nums
+	// will be in the report. If a value is not set we fallback to EvmDefaultMaxMerkleTreeSize.
+	MaxMerkleTreeSize uint64 `json:"maxTreeSize"`
 }
 
-func (c CommitOffchainConfig) Validate() error {
+func (c *CommitOffchainConfig) ApplyDefaults() {
```
makramkd, line 118: Nit: does this need to be a public method?

0xnogo: Making it private as not used elsewhere

makramkd (approved):

PR #198, opened 2024-10-02
Fix circular dependency with token reader type.

winder:


internal/plugincommon/discovery/discoverytypes/types.go
```diff
@@ -18,10 +20,7 @@ type Observation struct {
 	//SourceChains     []ccipocr3.ChainSelector
 	FChain map[cciptypes.ChainSelector]int
 	// See reader.ContractAddresses for more info on this data structure.
-	Addresses map[string]map[cciptypes.ChainSelector][]byte
-
-	// TODO: fix circular dependency /w token reader
-	//Addresses  reader.ContractAddresses
+	Addresses reader.ContractAddresses
```
winder, line 23: This is what triggered the circular dependency

makramkd: So the cycle is discoverytypes imports reader imports exectypes imports reader?

mateusz-sekara (approved):

PR #200, opened 2024-10-02
Discover router from offRamp contracts.

winder:


pkg/consts/consts.go
```diff
@@ -36,6 +36,8 @@ const (
 	// OnRamp methods
 	MethodNameOnRampGetDynamicConfig        = "OnRampGetDynamicConfig"
 	MethodNameOnRampGetStaticConfig         = "OnRampGetStaticConfig"
+	MethodNameOnRampGetChainConfig          = "OnRampGetDynamicConfig"
```
makramkd, line 39: Doesn't seem to be used

pkg/reader/ccip.go
```diff
@@ -957,6 +953,60 @@ func (r *ccipChainReader) getOnRampDynamicConfigs(
 	return result, nil
 }
 
+type onRampDestChainConfig struct {
+	SequenceNumber   uint64 `json:"sequenceNumber"`
+	AllowListEnabled bool   `json:"allowListEnabled"`
+	Router           []byte `json:"router"`
+}
+
+//nolint:dupl // It's not quite duplicate code...
```
makramkd, line 962: I wonder if there's some generics that can be extracted. e.g:

```golang
func getOnRampConfigs[ConfigType any](
  ctx context.Context,
  srcChains []cciptypes.ChainSelector,
  methodName string,
) (map[cciptypes.ChainSelector]ConfigType, error) {
  // initial validation code ...

  result := make(map[cciptypes.ChainSelector]ConfigType)
  mu := new(sync.Mutex)
  eg := new(errgroup.Group)
  for _, chainSel := range srcChains {
    // dest chain check

    chainSel := chainSel
    eg.Go(func() error {
      var resp ConfigType
      err := r.contractReaders[chainSel].ExtendedGetLatestValue(
        ctx, consts.ContractNameOnRamp, methodName, primitives.Unconfirmed, map[string]any{}, &resp)
// .. rest of code
}
```

You might be able to make this more general, e.g also pass in the args for the call, but for now this might be enough, could reduce some duplication.

winder: The main problem is that there are more than just these 2 -- I just did a better job of copy/pasting here and the linter caught it :sweat_smile: 

There's already a followup ticket for switching these to using the Batch API. I updated the description to mention the refactoring as well: https://smartcontract-it.atlassian.net/browse/CCIP-3129

internal/plugincommon/discovery/processor.go
```diff
@@ -182,28 +180,55 @@ func aggregateObservations(
 	return obs
 }
 
-// Outcome comes to consensus on the contract addresses and updates the chainreader. It doesn't actually
-// return an Outcome.
+func getRouterConsensus(
```
winder, line 183: This function could use some specific attention.

TODO: add a comment describing the approach used here.

makramkd: Can we add a unit test suite for this func?

internal/plugincommon/discovery/processor.go
```diff
@@ -182,28 +180,55 @@ func aggregateObservations(
...
+		lggr,
+		"router",
+		routerAddrs,
+		thresh,
+	)
+	// All the routers in routerAddrs are "valid", but we can only choose one.
+	routerCounts := make(map[string]int)
+	for _, addr := range routerConsensus {
+		routerCounts[string(addr)]++
+	}
```
makramkd, line 198: Would be useful to add this as a test case, but I'm trying to figure out if this is needed if you have 2F + 1 thresholds. If you have N = 3F + 1 and you do get consensus on a router address then you can just use that right away, because 2F+1 means that there's only F that observed otherwise, not enough to form another majority.

winder: Imagine srcChain1 has 2F+1 agreement on `0xRouter1` and srcChain2 has 2F+1 on `0xRouter2`. Which should we use?

makramkd: So each source chain will have a different router. If srcChain1 has 2FChain+1 honest majority on 0xRouter1 and srcChain2 has 2FChain+1 honest majority on 0xRouter2, we should just use each of these routers respectively for each chain. Isn't that what we're trying to get to? I don't think there's an ask to get the most common address across the sources, the router addresses will be different for different chains (in order to avoid confusion mostly)

winder: This was the approach @asoliman92 suggested. Your suggestion involves configuring multiple router addresses on the destination, how would you choose which one to use later? At the moment we're not configuring multiple bindings for any of the other contracts.

makramkd: Ah wait I think I'm confusing something - the routers in the "per source" config here refer to local router addresses, i.e those on the dest chain itself, so my above comment is incorrect.

makramkd: I think thats whats causing some confusion right now, maybe some renames would clarify it a bit more.

internal/plugincommon/discovery/processor.go
```diff
@@ -182,28 +180,65 @@ func aggregateObservations(
...
 
-// Outcome comes to consensus on the contract addresses and updates the chainreader. It doesn't actually
-// return an Outcome.
+// getRouterConsensus figures out a single router address from the multiple observed addresses from multiple source
+// chains. This happens in two phases, first we get the consensus on the router address for each source chain
+// according the threshold (fChain). Second, once we have consensus across the source chains, we select whichever
+// address was observed the most times. If there is a tie, we return an error.
+//
+// For a typical configuration, we expect all source chains to report the same destination router address.
+func getRouterConsensus(
```
makramkd, line 189:
```suggestion
func getDestRouterConsensus(
```

internal/plugincommon/discovery/processor.go
```diff
@@ -182,28 +180,65 @@ func aggregateObservations(
...
 }
 
-// Outcome comes to consensus on the contract addresses and updates the chainreader. It doesn't actually
-// return an Outcome.
+// getRouterConsensus figures out a single router address from the multiple observed addresses from multiple source
+// chains. This happens in two phases, first we get the consensus on the router address for each source chain
+// according the threshold (fChain). Second, once we have consensus across the source chains, we select whichever
+// address was observed the most times. If there is a tie, we return an error.
+//
+// For a typical configuration, we expect all source chains to report the same destination router address.
```
makramkd, line 188: I think this comment is confusingly worded because we are _not_ reading source chain routers - we are reading the dest chain's local router reference for messages coming from each source chain (see the link I pasted below).

I've attempted to clarify this below but let me know if there's any further misunderstandings.

```suggestion
// getRouterConsensus figures out a single dest router address from the multiple observed addresses from multiple offramp source
// chain configurations. This happens in two phases, first we get the consensus on the router address for 
// each offramp source chain config according the dest 2*fChain + 1 threshold. Second, once we have consensus across the source chains, 
// we select whichever address was observed the most times. If there is a tie, we return an error.
//
// For a typical configuration, we expect all offramp source chain configurations to report the same router address.
// See https://github.com/smartcontractkit/chainlink/blob/05c8b45da51340aa9575ac909486decc9c164792/contracts/src/v0.8/ccip/offRamp/OffRamp.sol#L105
```

internal/plugincommon/discovery/processor.go
```diff
@@ -182,28 +180,65 @@ func aggregateObservations(
...
+func getRouterConsensus(
+	lggr logger.Logger,
+	routerAddrs map[cciptypes.ChainSelector][][]byte,
+	thresh consensus.MultiThreshold[cciptypes.ChainSelector],
```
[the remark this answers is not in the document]
winder: That wont work, suppose the dest chain has F=5 and for one of the sources has F=1 because there are only 3 readers. That source chain would never reach consensus. You could take it one step further and imagine all source readers have F=1, in this case there would never be consensus from any source readers if F=5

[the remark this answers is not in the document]
makramkd: Not sure I understand, why would there never be consensus? This is a purely dest chain read and no sources are touched.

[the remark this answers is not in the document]
winder: The dest router address is being discovered by reading from the source chain onramp's

pkg/reader/ccip_test.go
```diff
@@ -624,7 +612,7 @@ func TestCCIPChainReader_DiscoverContracts_HappyPath_OnlySupportDest(t *testing.
 	destExtended.EXPECT().ExtendedGetLatestValue(
```
0xnogo, line 612: Do you think we could create a method to factorize the ExtendedGetLatestValue? It's repeated quite a lot in this file.

Something like:

```
func setupMockExpectations(mock *MockExtended, destChain cciptypes.ChainSelector, destFeeQuoter, destRouter []byte) {
    mock.EXPECT().ExtendedGetLatestValue(
        mock.Anything,
        consts.ContractNameOffRamp,
        consts.MethodNameOffRampGetDynamicConfig,
        primitives.Unconfirmed,
        map[string]any{},
        mock.Anything,
    ).Return(nil).Run(withReturnValueOverridden(func(returnVal interface{}) {
        v := returnVal.(*offRampDynamicChainConfig)
        v.FeeQuoter = destFeeQuoter
        v.Router = destRouter
    }))
}
```

pkg/reader/ccip.go
```diff
@@ -628,6 +628,10 @@ func (r *ccipChainReader) discoverDestinationContracts(
 	}
 	for chain, cfg := range sourceConfigs {
 		resp = resp.Append(consts.ContractNameOnRamp, chain, cfg.OnRamp)
+		// The local router is located in each source chain config. Add it once.
+		if len(resp[consts.ContractNameOnRamp][r.destChain]) == 0 {
+			resp = resp.Append(consts.ContractNameRouter, r.destChain, cfg.Router)
+		}
```
winder, line 634: Previous versions of this PR had some complex logic dealing with the fact that there are multiple local router configurations. This has been simplified, we now take the first one and call it a day.

pkg/reader/ccip.go
```diff
@@ -628,6 +628,10 @@ func (r *ccipChainReader) discoverDestinationContracts(
 	}
 	for chain, cfg := range sourceConfigs {
```
makramkd, line 629: Discussed offline: maybe sort these to ensure deterministic ordering across oracles, so that the first router grab is the same for all nodes.

pkg/reader/ccip.go
```diff
@@ -628,6 +628,10 @@ func (r *ccipChainReader) discoverDestinationContracts(
...
 		resp = resp.Append(consts.ContractNameOnRamp, chain, cfg.OnRamp)
+		// The local router is located in each source chain config. Add it once.
+		if len(resp[consts.ContractNameOnRamp][r.destChain]) == 0 {
+			resp = resp.Append(consts.ContractNameRouter, r.destChain, cfg.Router)
```
makramkd, line 633: Discussed offline: explain that we want to discover the wrapped native address and that both router and testrouter should point to the same wrapped native, and that this is a system-level assumption that is not explicitly checked offchain.

pkg/reader/ccip.go
```diff
@@ -708,17 +695,36 @@ func (r *ccipChainReader) DiscoverContracts(
...
+		if err != nil {
+			r.lggr.Infow("unable to lookup source fee quoters, this is expected during initialization", "err", err)
 
-		// ErrNoBindings is an allowable error.
-		if !errors.Is(err, contractreader.ErrNoBindings) {
-			return nil, fmt.Errorf("unable to lookup source fee quoters (onRamp dynamic config): %w", err)
+			// ErrNoBindings is an allowable error.
+			if !errors.Is(err, contractreader.ErrNoBindings) {
+				return nil, fmt.Errorf("unable to lookup source fee quoters (onRamp dynamic config): %w", err)
+			}
```
makramkd, line 706: I'm confused by this comment, its an allowable error but we still return an error wrapping `ErrNoBindings`, which is not checked in the discovery processor. Is this expected to only be temporary?

pkg/reader/ccip.go
```diff
@@ -708,17 +695,36 @@ func (r *ccipChainReader) DiscoverContracts(
...
+	// Read onRamps for Router in DestChainConfig.
+	{
+		destChainConfig, err := r.getOnRampDestChainConfig(ctx, myChains)
+		if err != nil {
+			r.lggr.Infow("unable to lookup source routers, this is expected during initialization", "err", err)
+
+			// ErrNoBindings is an allowable error.
+			if !errors.Is(err, contractreader.ErrNoBindings) {
+				return nil, fmt.Errorf("unable to lookup source routers (onRamp dest chain config): %w", err)
+			}
```
makramkd, line 723: Ditto here, kind of confused what the control flow is supposed to be.

pkg/reader/ccip.go
winder: Updated control flow to avoid the nested condition:
```
		if errors.Is(err, contractreader.ErrNoBindings) {
			// ErrNoBindings is an allowable error.
			r.lggr.Infow("unable to lookup source routers, this is expected during initialization", "err", err)
		} else if err != nil {
			return nil, fmt.Errorf("unable to lookup source routers (onRamp dest chain config): %w", err)
		} else {
			for chain, cfg := range destChainConfig {
				resp = resp.Append(consts.ContractNameRouter, chain, cfg.Router)
			}
		}
```

makramkd: Ah shoot I totally didn't read the `!` haha

makramkd: I like the new control flow! LGTM

makramkd (approved):

PR #201, opened 2024-10-02
Update chainlink-common in order to pull in Message.FeeValueJuels

rstout:


execute/report/report_test.go
```diff
@@ -480,7 +480,7 @@ func Test_Builder_Build(t *testing.T) {
 		{
 			name: "half report",
 			args: args{
-				maxReportSize: 2529,
+				maxReportSize: 2654,
```
asoliman92, line 483: where do we get these numbers from?

asoliman92 (approved):

execute/report/report_test.go
rstout: The tests failed because the new FeeValueJuels field increased the size of the reports. These tests failed with the new size of the reports, so I updated maxReportSize with these values.

PR #203, opened 2024-10-03
Deprecating the ExecutePluginConfig

0xnogo:
* ExecutePluginConfig in favor of the ExecuteOffchainConfig object
* remove the apply default logic in Validate


pluginconfig/execute_test.go
```diff
@@ -8,110 +8,6 @@ import (
 	commonconfig "github.com/smartcontractkit/chainlink-common/pkg/config"
 )
 
-func TestExecuteOffchainConfig_Validate(t *testing.T) {
```
makramkd, line 11: Why delete?

winder (approved):

pluginconfig/execute_test.go
0xnogo: Mistake, restored it.

makramkd (approved):

pluginconfig/commit_test.go
```diff
@@ -111,124 +111,124 @@ func TestArbitrumPriceSource_Validate(t *testing.T) {
...
+		TokenPriceBatchWriteFrequency      commonconfig.Duration
+		TokenInfo                          map[types.Account]TokenInfo
+		TokenPriceChainSelector            cciptypes.ChainSelector
+		TokenDecimals                      map[types.Account]uint8
+		NewMsgScanBatchSize                uint32
+		MaxReportTransmissionCheckAttempts uint32
+		MaxMerkleTreeSize                  uint32
+	}
+	//nolint:gosec
+	const remoteTokenAddress = "0x260fAB5e97758BaB75C1216873Ec4F88C11E57E3"
```
mateusz-sekara, line 126: How about using randomized values? (or at least making it a regular variable and get rid of this linter ignore) 

report_test.go contains
```go
func randomAddress() string {
	b := make([]byte, 20)
	_, _ = crand.Read(b) // Assignment for errcheck. Only used in tests so we can ignore.
	return cciptypes.Bytes(b).String()
}
```

Maybe we could just move this function to some place like testhelpers/common.go and reuse whenever we need Eth like addresses in tests and have

```go
remoteTokenAddress := testhelpers.RandomAddress()
aggregatorAddress := testhelpers.RandomAddress()
```

mateusz-sekara (approved):

PR #206, opened 2024-10-03
Move exec Outcome into a separate file.

winder:
Attempt to make the function easier to read by splitting out state specific observation code into helper functions.

This addresses the disabled lint rule // nolint:gocyclo // todo, but is otherwise a no-op change.

### DIFF
--- a/execute/outcome.go
+++ b/execute/outcome.go
@@ -1,0 +1,187 @@
+package execute
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	mapset "github.com/deckarep/golang-set/v2"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+
+	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
+	"github.com/smartcontractkit/chainlink-ccip/execute/report"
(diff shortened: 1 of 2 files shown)


execute/outcome.go
```diff
@@ -0,0 +1,187 @@
...
+	return outcome.Encode()
+}
+
+func (p *Plugin) getCommitReportsOutcome(observation exectypes.Observation) exectypes.Outcome {
+	// flatten commit reports and sort by timestamp.
+	var commitReports []exectypes.CommitData
+	for _, report := range observation.CommitReports {
+		commitReports = append(commitReports, report...)
+	}
+	sort.Slice(commitReports, func(i, j int) bool {
```
rstout, line 109: Is this sort needed if exectypes.NewOutcome also sorts?

rstout (approved):

mateusz-sekara (commented):
Hmm, do you think it's the right direction for the Exec plugin? Seems to me we are diverging from what we have in the Commit plugin, whereas it would be great to follow similar patterns for both implementations.

In Commit we use this Processor pattern in which code is structured horizontally. Logic is moved to specific processors and therefore Commit plugin struct works mainly as an orchestrator
```go
type commit struct {
     processor1
     processor2
     processor3
}
```

By keeping exec as large struct but splitting into multiple files by the ocr phase we organize the code in a different way making it even more diverged from what we have in commit. 

```go
//outcome.go
func (e *exec) Outcome()

//observation.go
func (e *exec) Observation()
```

Don't get me wrong, I like the refactoring you did there, but I wonder if that's the right way for us if we want both implementations to follow similar principles 

cc @asoliman92

PR #207, opened 2024-10-03
Implement ccipChainReader.LinkPriceUSD()

rstout:
Implement a function in ccipChainReader to get the price of the LINK token. We first call the dest chain FeeQuoter's `GetStaticConfig` method to get the address of the LINK token, then we call the `GetTokenPrices` method to get the price of LINK.

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -771,6 +771,83 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
 	return addressBytes, nil
 }
 
+// LinkPriceUSD gets the LINK price in 1e-18 USDs from the FeeQuoter contract on the destination chain.
+// For example, if the price is 1 LINK = 10 USD, this function will return 10e18 (10 * 1e18). You can think of this
+// function returning the price of LINK not in USD, but in a small denomination of USD, similar to returning
+// the price of ETH not in ETH but in wei (1e-18 ETH).
+func (r *ccipChainReader) LinkPriceUSD(ctx context.Context) (cciptypes.BigInt, error) {
+	// Ensure we can read from the destination chain.
+	if err := validateExtendedReaderExistence(r.contractReaders, r.destChain); err != nil {
+		return cciptypes.BigInt{}, fmt.Errorf("failed to validate dest chain reader existence: %w", err)
+	}
+
+	// TODO: consider caching this value.
+	feeQuoterCfg, err := r.getDestFeeQuoterStaticConfig(ctx)
+	if err != nil {
+		return cciptypes.BigInt{}, fmt.Errorf("get destination fee quoter static config: %w", err)
+	}
+
+	linkPriceUSD, err := r.getFeeQuoterTokenPriceUSD(ctx, feeQuoterCfg.LinkToken)
+	if err != nil {
+		return cciptypes.BigInt{}, fmt.Errorf("get LINK price in USD: %w", err)
+	}
+
+	return linkPriceUSD, nil
+}
+
+// getDestFeeQuoterStaticConfig returns the destination chain's Fee Quoter's StaticConfig
+func (r *ccipChainReader) getDestFeeQuoterStaticConfig(ctx context.Context) (feeQuoterStaticConfig, error) {
+	var staticConfig feeQuoterStaticConfig
+	err := r.getDestinationData(
+		ctx,
+		r.destChain,
+		consts.ContractNameFeeQuoter,
+		consts.MethodNameFeeQuoterGetStaticConfig,
+		&staticConfig,
+	)
+
+	if err != nil {
+		return feeQuoterStaticConfig{}, fmt.Errorf("unable to lookup fee quoter (offramp static config): %w", err)
+	}
+
+	return staticConfig, nil
+}
+
+// getFeeQuoterTokenPriceUSD gets the token price in USD of the given token address from the FeeQuoter contract on the
+// destination chain.
+func (r *ccipChainReader) getFeeQuoterTokenPriceUSD(ctx context.Context, tokenAddr []byte) (cciptypes.BigInt, error) {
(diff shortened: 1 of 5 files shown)


pkg/reader/ccip.go
```diff
@@ -775,6 +775,81 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
 	return addressBytes, nil
 }
 
+// GetLinkPriceUSD gets the LINK price in USD from the FeeQuoter contract on the destination chain.
+func (r *ccipChainReader) GetLinkPriceUSD(ctx context.Context) (cciptypes.BigInt, error) {
+	// Exit without an error if we cannot read the destination.
```
makramkd, line 780: This comment contradicts the code below - do we wanna exit w/ and error or without?

pkg/reader/ccip.go
```diff
@@ -775,6 +775,81 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
 	return addressBytes, nil
 }
 
+// GetLinkPriceUSD gets the LINK price in USD from the FeeQuoter contract on the destination chain.
+func (r *ccipChainReader) GetLinkPriceUSD(ctx context.Context) (cciptypes.BigInt, error) {
```
makramkd, line 779: nit: call this just `LinkPriceUSD`?

pkg/reader/ccip.go
```diff
@@ -775,6 +775,81 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
...
+	// Exit without an error if we cannot read the destination.
+	if err := validateExtendedReaderExistence(r.contractReaders, r.destChain); err != nil {
+		return cciptypes.BigInt{}, fmt.Errorf("validate extended reader existence: %w", err)
+	}
+
+	feeQuoterCfg, err := r.getDestFeeQuoterStaticConfig(ctx)
```
makramkd, line 785: Is this cache-able? Don't need to cache the static configs now, but can put a TODO for later perhaps.

pkg/reader/ccip.go
```diff
@@ -775,6 +775,81 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
...
+	if err != nil {
+		return cciptypes.BigInt{}, fmt.Errorf("get destination fee quoter static config: %w", err)
+	}
+
+	linkTokenAddr := feeQuoterCfg.LinkToken
+
+	getLinkPriceUSD, err := r.getLinkPriceUSD(ctx, linkTokenAddr)
+
+	if err != nil {
```
makramkd, line 794: Inline and reduce newlines

```suggestion
	getLinkPriceUSD, err := r.getLinkPriceUSD(ctx, feeQuoterCfg.LinkToken)
	if err != nil {
```

pkg/reader/ccip.go
```diff
@@ -775,6 +775,81 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
...
+	err := r.getDestinationData(
+		ctx,
+		r.destChain,
+		consts.ContractNameFeeQuoter,
+		consts.MethodNameFeeQuoterGetStaticConfig,
+		&staticConfig,
+	)
+
+	if err != nil {
+		return feeQuoterStaticConfig{}, fmt.Errorf("unable to lookup nonce manager (offramp static config): %w", err)
```
makramkd, line 813:
```suggestion
		return feeQuoterStaticConfig{}, fmt.Errorf("unable to lookup fee quoter static config: %w", err)
```

pkg/reader/ccip.go
```diff
@@ -775,6 +775,81 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
...
+// TODO: doc
+func (r *ccipChainReader) getLinkPriceUSD(ctx context.Context, linkTokenAddr []byte) (cciptypes.BigInt, error) {
+	if len(linkTokenAddr) == 0 {
+		return cciptypes.BigInt{}, fmt.Errorf("linkTokenAddr is empty")
+	}
+
+	reader, ok := r.contractReaders[r.destChain]
+	if !ok {
+		return cciptypes.BigInt{}, fmt.Errorf("contract reader not found for chain %d", r.destChain)
+	}
```
makramkd, line 828: This is the second time this check is done in this call stack, dedup?

pkg/reader/ccip.go
```diff
@@ -775,6 +775,81 @@ func (r *ccipChainReader) GetContractAddress(contractName string, chain cciptype
 	return addressBytes, nil
 }
 
+// GetLinkPriceUSD gets the LINK price in USD from the FeeQuoter contract on the destination chain.
```
makramkd, line 778: Maybe state the denomination here (i.e how many decimals)

pkg/reader/ccip.go
rstout: The alternative is to get rid of the `validateExtendedReaderExistence` call, but that's a stronger check. I'm fine with leaving both in.

pkg/reader/ccip.go
rstout: Aside: these implicit decimals/denominations are very confusing imo

pkg/reader/ccip.go
asoliman92: This function is a duplicate from `OnchainTokenPricesReader.GetFeeQuoterTokenUpdates`. Maybe pass the price reader to CCIP and use it instead.

makramkd (approved):

pkg/reader/ccip.go
rstout: `GetWrappedNativeTokenPriceUSD` also makes the same FeeQuoter call, and there's other FeeQuoter calls in `ccipChainReader`. I think it makes sense to consolidate FeeQuoter functionality in a future PR, "after the dust settles".

PR #209, opened 2024-10-04
CCIP-3664 Single rate limiter per attestation http client

mateusz-sekara:


execute/tokendata/usdc/http.go
```diff
@@ -60,7 +60,37 @@ type httpClient struct {
...
-func NewHTTPClient(
+var (
+	clientInstances = make(map[string]*httpClient)
+	mutex           sync.Mutex
+)
+
+// GetHTTPClient returns a singleton instance of the httpClient for the given API URL.
+// It's critical to reuse existing clients because of the self-rate limiting mechanism. Being rate limited by
+// Circle comes with a long cool down period, so we should always self-rate limit before hitting the API rate limit.
+func GetHTTPClient(
```
makramkd, line 71: Trying to understand the benefits of this - I see that this is the following call stack:

* factory.go: calls tokendata.NewConfigBasedCompositeObservers
* observer.go: calls createUSDCTokenObserver
* observer.go: calls NewSequentialAttestationClient
* attestation.go: calls NewHTTPClient

To me it seems like only one client is created for a single plugin instance. In the LOOPP world, I _believe_ we will spawn a different LOOPP chainlink-ccip process for each chain (right now, its not a separate process, so this singleton logic would work as expected).

However in a LOOPP world, with different processes, this won't have the expected behavior.

makramkd (commented):
Wh

execute/tokendata/usdc/http.go
mateusz-sekara: Yeah, loops will make this kind of synchronization would be painful, maybe we will need to make httpclient loop or something? I don't know the answer. Correct me if I'm wrong but we are far from a loop world and we need this rate limiting to work properly anyway.

I can add a comment about loopification

makramkd: Fair enough, yeah in LOOPP settings lots of things will probably break :P

makramkd (approved):

PR #210, opened 2024-10-04
CCIP-3677 Observe different token data observers in parallel

mateusz-sekara:


execute/tokendata/observer.go
```diff
@@ -116,18 +116,41 @@ func (c *compositeTokenDataObserver) Observe(
...
+	for i := 0; i < len(c.observers); i++ {
+		select {
+		case tokenData := <-resultsCh:
+			var err error
+			tokenDataObservations, err = merge(tokenDataObservations, tokenData)
+			if err != nil {
+				return nil, fmt.Errorf("merging token data together failed: %w", err)
+			}
+		case <-ctx.Done():
+			// Return whatever was processed so far when context is canceled
```
asoliman92, line 148: To be sure. failed ones will be picked up in the next round?

mateusz-sekara: Yep, if message is ready to be executed, it should be included in the outcome and transmitted as an exec report to the chain. All failed ones would be retried in the next round if consensus is not reached

execute/tokendata/observer.go
```diff
@@ -116,18 +116,41 @@ func (c *compositeTokenDataObserver) Observe(
 	msgObservations exectypes.MessageObservations,
 ) (exectypes.TokenDataObservations, error) {
 	tokenDataObservations := c.initTokenDataObservations(msgObservations)
+	// Terminate immediately if there are no observers
+	if len(c.observers) == 0 {
+		return tokenDataObservations, nil
+	}
+
+	resultsCh := make(chan exectypes.TokenDataObservations, len(c.observers))
```
winder, line 124: Some thoughts about the general approach.

1. I'd like to see this be asynchronous to the Observe function. (i.e. persistent result channel + go routine collecting results).
2. During Observe, send new/failed attestation requests out for observation and grab ones that are ready.
3. Is there a timeout on the context?
4. Channel size should be one or none: [one or none](https://github.com/uber-go/guide/blob/master/style.md#channel-size-is-one-or-none).

execute/tokendata/observer.go
```diff
@@ -116,18 +116,41 @@ func (c *compositeTokenDataObserver) Observe(
...
+			if err != nil {
+				c.lggr.Errorw("Error while observing token data", "error", err)
+				tokenData = exectypes.TokenDataObservations{}
+			}
+			resultsCh <- tokenData
+		}(&ob)
+	}
+
+	for i := 0; i < len(c.observers); i++ {
+		select {
```
winder, line 140: The granularity of a result is an entire set of observations which causes problems when the context is done:
1. The observers are notified to start returning accumulated attestations.
2. The `ctx.Done()` case is hit when immediately returns accumulated attestations.

This means we're very likely to ignore all attestations when the timeout is hit.

PR #216, opened 2024-10-08
Commit / RMN - Error on multiple valid roots

dimkouv:
When we have multiple roots with enough observers we should error instead of moving forward with the most voted one, since this is a violation.

### DIFF
--- a/commit/merkleroot/rmn/controller.go
+++ b/commit/merkleroot/rmn/controller.go
@@ -669,6 +669,9 @@ func selectRoots(
 			selectedRoot = root
 		}
 
+		if selectedRoot.IsEmpty() {
+			return nil, fmt.Errorf("no valid root for chain %d", chain)
+		}
 		selectedRoots[chain] = selectedRoot
 	}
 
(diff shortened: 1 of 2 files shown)


commit/merkleroot/rmn/controller.go
```diff
@@ -637,20 +648,31 @@ func getMostVotedRootsFromObservations(
...
+
+			if !selectedRoot.IsEmpty() {
+				return nil, fmt.Errorf("more than one valid root for chain %d", chain)
+			}
+
+			selectedRoot = root
 		}
-		mostVotedRoots[chain] = mostVotedRoot
+
+		selectedRoots[chain] = selectedRoot
```
0xnogo, line 672: Is it adding an empty root is no root is valid?

If yes, we might want to add an non-empty check before adding it to the selectedRoots

```
if selectedRoot.IsEmpty() {
    return nil, fmt.Errorf("no valid root found for chain %d", chain)
}
```

0xnogo (approved):

PR #218, opened 2024-10-08
Split CostlyMessageObserver into Fee and ExecCost components

rstout:
Implements CostlyMessageObserver in terms of a feeCalculator and a execCostCalculator. When checking if a message is too costly, we use these 2 components to compute the message fee and exec cost. If the cost is greater than the fee, the message is considered too costly. Also adds messageTimestamps as a param to CostlyMessageObserver.Observe() so that fee boosting can use the message timestamps.

### DIFF
--- a/execute/exectypes/costly_messages.go
+++ b/execute/exectypes/costly_messages.go
@@ -1,0 +1,187 @@
+package exectypes
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugintypes"
+)
+
+// CostlyMessageObserver observes messages that are too costly to execute.
+type CostlyMessageObserver interface {
+	// Observe takes a set of messages and returns a slice of message IDs that are too costly to execute.
+	// Also takes in a map from message ID to the message's timestamp, needed to calculate fee boosting.
+	Observe(
+		ctx context.Context,
+		messages []cciptypes.Message,
+		messageTimestamps map[cciptypes.Bytes32]time.Time,
+	) ([]cciptypes.Bytes32, error)
+}
+
+func NewCostlyMessageObserver() CostlyMessageObserver {
+	return &CCIPCostlyMessageObserver{
+		// TODO: Implement fee and exec cost calculators
+		feeCalculator:      &ZeroMessageFeeUSD18Calculator{},
+		execCostCalculator: &ZeroMessageExecCostUSD18Calculator{},
+	}
+}
+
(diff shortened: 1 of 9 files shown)


execute/exectypes/costly_messages.go
```diff
@@ -0,0 +1,135 @@
...
+type MessageFeeE18USDCalculator interface {
+	// MessageFeeE18USD Returns a map from message ID to the message's fee in e-18 USDs. For example, if the message's
+	// fee is 12USD, this function return this message's fee as 12 * 1e18. You can think of this function returning the
+	// fee not in USD, but in a small denomination of USD, analogous to returning the cost in wei instead of ETH
+	// (1 wei = 1e-18 ETH).
+	MessageFeeE18USD(
+		ctx context.Context,
+		messages []cciptypes.Message,
+		messageTimestamps map[cciptypes.Bytes32]time.Time,
+	) (map[cciptypes.Bytes32]cciptypes.BigInt, error)
```
makramkd, line 90: Taking a stab at designing the types for this to ensure the denominations are always honored:

```golang
type USD18 *big.Int

func NewUSD18FromUSD(usd uint64) USD18 {
  return new(big.Int).Mul(new(big.Int).SetUint64(usd), big.NewInt(1e18))
}

// ... operations ...
```

Maybe this'll become clearer when we see what kind of operations we do on these values, including whats in gas prices and token prices. cc @asoliman92

execute/exectypes/costly_messages.go
```diff
@@ -0,0 +1,135 @@
...
+			costlyMessages = append(costlyMessages, msg.Header.MessageID)
+		}
+	}
+
+	return costlyMessages, nil
+}
+
+var _ CostlyMessageObserver = &CcipCostlyMessageObserver{}
+
+// MessageFeeE18USDCalculator Calculates the fees (paid at source) of a set of messages in e-18 USDs.
```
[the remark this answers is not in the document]
rstout: I created a new type

internal/plugintypes/common.go
```diff
@@ -22,3 +24,15 @@ type ChainRange struct {
...
 
 type DonID = uint32
+
+// USD18 is a small unit of USD, where 1 USD18 is 1e-18 USD, meaning it is 18 decimal places smaller than 1 USD.
+//
+// 1 USD18 = 1e-18 USD   = 0.000000000000000001 USD
+// 1 USD   = 1e18  USD18 = 1,000,000,000,000,000,000 USD18
+//
+// Token prices stored in many contracts (e.g. FeeQuoter) are denominated in USD18
+type USD18 = *big.Int
```
makramkd, line 34: Its a good start 🎉

makramkd (approved):

PR #221, opened 2024-10-09
CCIP-3713 Filtering by the 3rd word in USDC Reader

mateusz-sekara:


pkg/reader/usdc_reader.go
```diff
@@ -120,14 +129,34 @@ func (u usdcMessageReader) MessageHashes(
...
+					Operator: primitives.Eq,
+				}),
+		)
+	}
+
+	filter := []query.Expression{
+		query.Or(eventFilter...),
+		query.Confidence(primitives.Finalized),
+	}
+
```
ilija42, line 148: You can format the key filter with query.Where
```suggestion
		keyFilter, err := query.Where(consts.EventNameCCTPMessageSent,
		query.Or(eventFilter...),
		query.Confidence(primitives.Finalized))
	if err != nil {
		return nil, err
	}

	iter, err := u.contractReaders[source].QueryKey(
		ctx,
		cr,
		keyFilter,
```

pkg/reader/usdc_reader.go
```diff
@@ -120,14 +129,34 @@ func (u usdcMessageReader) MessageHashes(
 		return nil, fmt.Errorf("no contract bound for chain %d", source)
 	}
 
+	eventFilter := make([]query.Expression, 0, len(eventIDs))
+	for _, id := range eventIDs {
+		eventFilter = append(
+			eventFilter,
+			query.Comparator(consts.EventFilterCCIPMessageSent,
```
ilija42, line 136: Name for value filter shouldn't be mixed up with the key name. I would prefer if the comparatror was called something like MessageValue

mateusz-sekara: I don't follow. Is it about the name of the const (`consts.EventFilterCCIPMessageSent`) or its value? What do you suggest instead?

> Name for value filter primitive shouldn't be mixed up with the key name.
I think it's different, isn't it?
Query key is `consts.EventNameCCTPMessageSent`, but comparator is `consts.EventFilterCCIPMessageSent`, am I missing sth?

ilija42: Sry I misread that multiple times while looking at the code. Still, wouldn't constants like this: `consts.CCIPMessageSentValue `and for the event itself `consts.CCIPMessageSent` make more sense?

mateusz-sekara: Trying to follow existing patters in naming, so I've renamed to
```go
consts.CCTPMessageSentValue # for the comparator
consts.EventNameCCTPMessageSent # kept existing for the event
```

ilija42: sgtm

ilija42 (approved):

winder (approved):
neat

PR #222, opened 2024-10-09
Fix flaky test in RMN home reader

0xnogo:


internal/reader/rmn_home_test.go
```diff
@@ -54,79 +54,62 @@ func TestRMNHomeChainConfigPoller_Ready(t *testing.T) {
...
-	require.Equal(t, map[string]error{configPoller.Name(): error(nil)}, healthy)
-
-	// give some time for polling to happen
-	time.Sleep(totalSleepTime)
-	errors := configPoller.HealthReport()
-	require.Equal(t, 1, len(errors))
-	require.Errorf(t, errors[configPoller.Name()], "polling failed %d times in a row", MaxFailedPolls)
-
-	// give some time for successful polling to happen
-	time.Sleep(totalSleepTime * 1)
```
mateusz-sekara, line 118: General observation, we should not use sleep directly unless necessary. IMO `require.Eventually` is the right way of verifying the condition to be met in some amount of time

internal/reader/rmn_home_test.go
```diff
@@ -54,79 +54,62 @@ func TestRMNHomeChainConfigPoller_Ready(t *testing.T) {
...
-	).Return(fmt.Errorf("error")).Times(50)
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 
-	// Set up the mock for a successful call after 10 errors
-	primaryConfig, secondaryConfig := createTestRMNHomeConfigs(false, false)
-	rmnHomeOnChainConfigs := []VersionedConfigWithDigest{primaryConfig, secondaryConfig}
+			// Create a mock ContractReaderFacade
```
mateusz-sekara, line 81: Are these comments required? Seems very straightforward what these are doing ;)

mateusz-sekara:
Oh shoot, it's in draft, ignore my comments if you are still working on that @0xnogo

internal/reader/rmn_home_test.go
```diff
@@ -53,80 +54,97 @@ func TestRMNHomeChainConfigPoller_Ready(t *testing.T) {
...
+							Config: Config{
+								Nodes:        []Node{},
+								SourceChains: []SourceChain{},
+							},
+						},
+					},
+				}
+				mu.Lock()
+				getLatestValueCalled = true
+				mu.Unlock()
```
makramkd, line 109: Why is this needed? The mock can tell you if its called or not - if its called, the test passes, if not, it fails. Does this do anything different?

internal/reader/rmn_home_test.go
```diff
@@ -53,80 +54,97 @@ func TestRMNHomeChainConfigPoller_Ready(t *testing.T) {
...
-	// give some time for polling to fail again
-	time.Sleep(totalSleepTime * 1)
-	errors = configPoller.HealthReport()
-	require.Equal(t, 1, len(errors))
-	require.Errorf(t, errors[configPoller.Name()], "polling failed %d times in a row", MaxFailedPolls)
+			// Wait for the initial fetch to complete
+			require.Eventually(t, func() bool {
+				mu.Lock()
+				defer mu.Unlock()
+				return getLatestValueCalled
```
makramkd, line 125: Ah I see, ok this is interesting.

makramkd: I think you can check if something is `Called` using the mock? Lemme check the API.

internal/reader/rmn_home_test.go
```diff
@@ -53,80 +54,97 @@ func TestRMNHomeChainConfigPoller_Ready(t *testing.T) {
...
+
+			if tt.wantErr {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "polling failed")
+			} else {
+				require.NoError(t, err)
+			}
+
+			require.NoError(t, poller.Close())
+			homeChainReader.AssertExpectations(t)
```
makramkd, line 145: Not needed, this is called within the mock setup code in a `t.Cleanup`

internal/reader/rmn_home_test.go
makramkd: This is working for me, inside the `func() bool` for the `Eventually`:

```golang
return homeChainReader.AssertCalled(t, "GetLatestValue", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything)
```

0xnogo: Nice one! thanks

makramkd (approved):

PR #224, opened 2024-10-11
RMNhome reader lifecycle

0xnogo:
Starting and Cancelling the RMNHome Reader in the Commit Plugin

### DIFF
--- a/commit/factory.go
+++ b/commit/factory.go
@@ -134,9 +134,9 @@ func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types
 		if err := rmnHomeReader.Ready(); err != nil {
 			return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to initialize RMNHome reader: %w", err)
 		}
-		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		childCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 		defer cancel()
-		if err := rmnHomeReader.Start(ctx); err != nil {
+		if err := rmnHomeReader.Start(childCtx); err != nil {
 			return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to start RMNHome reader: %w", err)
 		}
 	}
--- a/commit/factory.go
+++ b/commit/factory.go
@@ -130,6 +130,15 @@ func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types
 			p.lggr,
 			100*time.Millisecond,
 		)
+
+		if err := rmnHomeReader.Ready(); err != nil {
+			return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to initialize RMNHome reader: %w", err)
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := rmnHomeReader.Start(ctx); err != nil {
+			return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to start RMNHome reader: %w", err)
+		}
 	}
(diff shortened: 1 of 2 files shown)


commit/plugin.go
```diff
@@ -79,6 +80,18 @@ func NewPlugin(
...
 
+	if offchainCfg.RMNEnabled {
+		if err := rmnHomeReader.Ready(); err == nil {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			if err := rmnHomeReader.Start(ctx); err != nil {
+				lggr.Errorw("Failed to start RMNHome reader", "err", err)
+			}
+		} else {
+			lggr.Errorw("Failed to initialize RMNHome reader", "err", err)
```
makramkd, line 91: If RMNHome reader isn't working and RMN is enabled, should we even start the plugin?

commit/plugin.go
```diff
@@ -79,6 +80,18 @@ func NewPlugin(
 		offchainCfg.MaxMerkleTreeSize = merklemulti.MaxNumberTreeLeaves
 	}
 
+	if offchainCfg.RMNEnabled {
+		if err := rmnHomeReader.Ready(); err == nil {
```
makramkd, line 84: What does `Ready` do exactly?

0xnogo: Checks that the service has not been started yet, which in this case should not be happening.

commit/plugin.go
0xnogo: As discussed offline, I moved the init logic in the factory to let the libocr retry when `NewReportingPlugin` errored

commit/factory.go
```diff
@@ -130,6 +130,15 @@ func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types
 			p.lggr,
 			100*time.Millisecond,
 		)
+
+		if err := rmnHomeReader.Ready(); err != nil {
+			return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to initialize RMNHome reader: %w", err)
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
makramkd, line 137: We get a ctx in the method already, can we just use that?

0xnogo: My goal is to create a context with a timeout, but I guess we could create a child context

commit/factory.go
```diff
@@ -130,6 +130,15 @@ func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types
 			p.lggr,
 			100*time.Millisecond,
 		)
+
+		if err := rmnHomeReader.Ready(); err != nil {
+			return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to initialize RMNHome reader: %w", err)
+		}
+		childCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+		defer cancel()
+		if err := rmnHomeReader.Start(childCtx); err != nil {
```
makramkd, line 139: Why not just rmnHomeReader.Start(ctx)? I believe that if the parent context cancels then this child context will cancel even if you have a longer timeout than the parent context.

0xnogo: done

commit/plugin.go
```diff
@@ -346,6 +346,22 @@ func (p *Plugin) Outcome(
...
+			}
+		} else {
+			p.lggr.Warn("RMNHome reader was nil during Close")
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("errors during Close: %v", errs)
+	}
 	return nil
```
makramkd, line 365: Use [errors.Join](https://pkg.go.dev/errors#Join) instead:

```suggestion
	return errors.Join(errs)
```

makramkd (approved):

PR #225, opened 2024-10-11
Add new const for GetAllSourceChainConfigs.

winder:
New const for chainlink repo to use.


winder:
Closed in favor of #227

PR #226, opened 2024-10-11
Implement RMN peer client and use it in the Commit Plugin.

dimkouv:
# Commit/RMN Communication using rage p2p

How it works:

1. We come to consensus on RMN remote config digest while computing intervals.
2. In the next round the leader's Query Initializes the peer client connection and everyone else on their Observation phase using the agreed config digest from the previous round.
3. Connection is now initialized and plugin can talk to RMN nodes by creating streams (one for each RMN node) on demand.

### Resolves
- https://github.com/smartcontractkit/chainlink/pull/14750

### DIFF
--- a/.github/workflows/ccip-integration-test.yml
+++ b/.github/workflows/ccip-integration-test.yml
@@ -30,7 +30,7 @@ jobs:
         uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
         with:
           repository: smartcontractkit/chainlink
-          ref: dk/ccip-rmn-plumbing
+          ref: develop
           path: chainlink
       - name: Get the correct commit SHA via GitHub API
         id: get_sha


commit/merkleroot/rmn/controller.go
```diff
@@ -48,6 +48,18 @@ var (
...
+	// further Controller interaction. If called twice it overwrites the previous connection.
+	InitConnection(
+		ctx context.Context,
+		commitConfigDigest cciptypes.Bytes32,
+		rmnHomeConfigDigest cciptypes.Bytes32,
+		peerIDs []string, // union of oraclePeerIDs and rmnNodePeerIDs
+	) error
+
+	// Close closes the connection to the generic peer group endpoint and all the underlying streams.
+	Close() error
```
makramkd, line 61: Should we call this in the commit plugin's `Close`?

commit/merkleroot/rmn/peerclient.go
```diff
@@ -3,21 +3,42 @@
...
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 
 	rmntypes "github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn/types"
 )
 
-var (
-	ErrRMNNodeNotFound = errors.New("rmn node not found")
-)
+var ErrNoConn = fmt.Errorf("no connection, please call InitConnection before further interraction")
```
makramkd, line 23:
```suggestion
var ErrNoConn = fmt.Errorf("no connection, please call InitConnection before further interaction")
```

commit/merkleroot/observation.go
```diff
@@ -55,6 +59,52 @@ func (w *Processor) Observation(
...
+		return fmt.Errorf("failed to get RMN nodes info: %w", err)
+	}
+
+	peerIDs := make([]string, 0, len(rmnNodesInfo))
+	for _, node := range rmnNodesInfo {
+		peerIDs = append(peerIDs, node.PeerID.String())
+	}
+	for _, p2pID := range w.oracleIDToP2pID {
+		peerIDs = append(peerIDs, p2pID.String())
+	}
```
makramkd, line 93: Why do we need the P2P ID's of the CCIP oracles? AFAIK we don't communicate with them using these RMN-specific streams.

commit/merkleroot/rmn/peerclient.go
```diff
@@ -3,21 +3,42 @@
...
 // PeerClient performs low-level communication with RMN peers.
 type PeerClient interface {
-	Send(rmnNodeID rmntypes.NodeID, request []byte) error
+	// InitConnection initializes the connection to the peer group endpoint and must be called before
+	// further PeerClient interaction.
+	InitConnection(
+		ctx context.Context,
+		commitConfigDigest cciptypes.Bytes32,
+		rmnHomeConfigDigest cciptypes.Bytes32,
+		peerIDs []string, // union of oraclePeerIDs and rmnNodePeerIDs
```
makramkd, line 33: Explanation needed for why the union is required and not just the RMN node peer IDs. Is it something to do with peer discovery or something else?

commit/merkleroot/observation.go
```diff
@@ -55,6 +59,52 @@ func (w *Processor) Observation(
...
+
+	if err := w.rmnController.InitConnection(
+		ctx,
+		cciptypes.Bytes32(w.reportingCfg.ConfigDigest),
+		prevOutcome.RMNRemoteCfg.ConfigDigest,
+		peerIDs,
+	); err != nil {
+		return fmt.Errorf("failed to init connection to RMN: %w", err)
+	}
+	w.rmnControllerInitialized = true
```
makramkd, line 103: I think we'd want to re-initialize the connection if the RMN node set changes or there's a new config digest (maybe just the latter, since I believe that the config digest will change if the node set changes though worth confirming).

commit/merkleroot/rmn/peerclient.go
```diff
@@ -31,27 +52,150 @@ type PeerResponse struct {
...
+
+	return nil
+}
+
+func (r *peerClient) Send(rmnNode rmntypes.HomeNodeInfo, request []byte) error {
+	if r.peerGroup == nil {
+		return ErrNoConn
+	}
+
+	stream, err := r.getOrCreateRageP2PStream(rmnNode)
```
makramkd, line 128: I think this code is a bit racey, it seems cleaner to me to grab the mutex on line 123 and then hold it for the entirety of fetching the stream + sending the message. The reasoning here is that in between this call returning and line 129, the stream could technically get closed?

commit/merkleroot/rmn/peerclient.go
```diff
@@ -31,27 +52,150 @@ type PeerResponse struct {
...
+	if ok {
+		return stream, nil
 	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	// todo: versioning for stream names e.g. for 'v1_7'
+	streamName := fmt.Sprintf("ccip-rmn/v1_6/%x", r.genericEndpointConfigDigest)
+	r.lggr.Info("Creating new stream", "streamName", streamName)
```
makramkd, line 153:
```suggestion
	r.lggr.Infow("Creating new stream", "streamName", streamName)
```

Otherwise log will not look correct

commit/merkleroot/rmn/peerclient.go
```diff
@@ -31,27 +52,150 @@ type PeerResponse struct {
...
+			IncomingBufferSize: 1,
+			MaxMessageLength:   2_097_152, // 2MB
+			MessagesLimit: ragep2p.TokenBucketParams{
+				Rate:     20,
+				Capacity: 100,
+			},
+			BytesLimit: ragep2p.TokenBucketParams{
+				Rate:     20_971_520,  // 20MB
+				Capacity: 104_857_600, // 100MB
+			},
```
makramkd, line 170: Should these be defined via configuration or are these good enough to be constants?

commit/merkleroot/rmn/peerclient.go
```diff
@@ -31,27 +52,150 @@ type PeerResponse struct {
...
+		}
 	}
-	panic("implement actual send")
 }
 
 func (r *peerClient) Recv() <-chan PeerResponse {
 	return r.respChan
 }
+
+// writePrefix writes the prefix to the rightmost 2 bytes of the hash.
```
makramkd, line 195: Wait is this a prefix or a suffix? A prefix would mean (in my mind) the first 2 bytes of a 32 byte array, this is how OCR config digests work. Is this the same or different?

go.mod
```diff
@@ -23,10 +23,12 @@ require (
 	github.com/buger/jsonparser v1.1.1 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
+	github.com/ethereum/go-ethereum v1.13.8 // indirect
```
makramkd, line 26: 😱

commit/merkleroot/rmn/controller.go
makramkd: Sounds good, please create a ticket.

commit/merkleroot/observation.go
makramkd: OK, I think I'm starting to remember this, lets put this comment somewhere in the interface.

commit/merkleroot/rmn/peerclient.go
makramkd: Well right now its writing the prefix to the leftmost 2 bytes, so this comment is incorrect

commit/merkleroot/rmn/peerclient.go
```diff
@@ -31,27 +52,173 @@ type PeerResponse struct {
...
+				Capacity: 104_857_600, // 100MB
+			},
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("new stream %s: %w", streamName, err)
 	}
+
+	r.rageP2PStreams[rmnNode.ID] = stream
+	go r.listenToStream(rmnNode.ID, stream)
```
0xnogo, line 183: how streams are expected to be closed?

makramkd (approved):

PR #227, opened 2024-10-11
discovery: Use getAllSourceChainConfigs

winder:
Allows destination readers to discover all source chain configurations. Previously the algorithm would only discover sources when there was a corresponding contract reader. Now the "getAllSourceChainConfigs"


pkg/reader/ccip_test.go
```diff
@@ -618,22 +637,24 @@ func TestCCIPChainReader_DiscoverContracts_HappyPath_Round2(t *testing.T) {
...
+	destRouter := []byte{0x6}
 
 	var expectedContractAddresses ContractAddresses
+	for i := range onramps {
+		expectedContractAddresses = expectedContractAddresses.Append(
+			consts.ContractNameOnRamp, sourceChain[i], onramps[i])
+	}
 	// All dest chain contracts should be discovered, they do not require source chain support.
-	//expectedContractAddresses = expectedContractAddresses.Append(consts.ContractNameRouter, destChain, destRouter)
+	expectedContractAddresses = expectedContractAddresses.Append(consts.ContractNameRouter, destChain, destRouter)
```
winder, line 653: Previously dest-only readers were unable to discover their own router :joy:

pkg/reader/ccip.go
```diff
@@ -625,89 +625,88 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
...
 	}
-	resp = resp.Append(consts.ContractNameFeeQuoter, r.destChain, dynamicConfig.FeeQuoter)
 
 	return resp, nil
 }
 
 func (r *ccipChainReader) DiscoverContracts(
 	ctx context.Context,
-	allChains []cciptypes.ChainSelector,
+	_ []cciptypes.ChainSelector, //Deprecated
```
asoliman92, line 700: Why not remove it in this PR? No external calls for this AFAIK.

pkg/reader/ccip.go
```diff
@@ -972,6 +978,64 @@ func (r *ccipChainReader) getOffRampSourceChainsConfig(
 	return res, nil
 }
 
+// selectorsAndConfigs wraps the return values from getAllSourceChainConfigs.
+type selectorsAndConfigs struct {
+	Selectors          []uint64            `mapstructure:"F0"`
```
asoliman92, line 983: I believe we don't need to use `mapstructure` here. check this [example](https://github.com/smartcontractkit/chainlink-ccip/blob/ad2992ddf2377fe55c65e1aca9cb03c1c8cd8b2a/internal/reader/home_chain.go#L404-L407) where it's just using normal tag

winder: I believe it's actually the json tags in that example which aren't needed. When a tag isn't provided the field name is used.

pkg/reader/ccip.go
winder: done

pkg/reader/ccip.go
```diff
@@ -625,89 +625,88 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
...
+		var staticConfig offRampStaticChainConfig
+		err := r.getDestinationData(
+			ctx,
+			chain,
+			consts.ContractNameOffRamp,
+			consts.MethodNameOffRampGetStaticConfig,
+			&staticConfig,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to lookup nonce manager (offramp static config): %w", err)
```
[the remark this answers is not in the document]
winder: done

asoliman92 (approved):

PR #217, opened 2024-10-08
Compute message fee in USD

rstout:
Adds a real implementation for the MessageFeeUSD18Calculator iface that gets the LINK price using CCIPReader and also does fee boosting.

PR #229, opened 2024-10-14
Misc. fixes

asoliman92:
* Fix bug: Nodes that didn't support feed chain were not participating in the observation at all. They can observe dest chain's FeeQuoter updates though.
* Enable fee quoter in token price processor
* Add integration tests for token processor
* Refactor e2e_plugin tests and prepare it for adding chain fee tests next
* Fix `TimestampedBig` decoding when reading from on-chain. The struct on-chain is uint32. Used a new struct internally without touching the APIs to make the change less invasive.
* Fixed onRamp discovery. GetLatestValue parameters were not correct.
* Converting any string addresses to bytes.


pkg/reader/price_reader.go
```diff
@@ -66,23 +72,28 @@ type LatestRoundData struct {
...
+	chain ccipocr3.ChainSelector,
 ) (map[ocr2types.Account]plugintypes.TimestampedBig, error) {
-	if !pr.feeQuoterEnabled {
-		return make(map[ocr2types.Account]plugintypes.TimestampedBig), nil
+	updates := make([]plugintypes.TimestampedBig, len(tokens))
+	updateMap := make(map[ocr2types.Account]plugintypes.TimestampedBig)
+
+	feeQuoterAddress, err := pr.ccipReader.GetContractAddress(consts.ContractNameFeeQuoter, chain)
+	if err != nil {
+		pr.lggr.Debugf("failed to get fee quoter address for chain %d: %v", chain, err)
```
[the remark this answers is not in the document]
asoliman92: Don't get the question, it's already `Debug`, or you mean remove the `f`?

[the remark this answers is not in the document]
asoliman92: Yes. FeeQuoter will not be available if it's still to be discovered.

rstout (approved):

commit/plugin_e2e_test.go
```diff
@@ -40,53 +46,53 @@ import (
...
+var (
+	oracleIDs = []commontypes.OracleID{1, 2, 3}
+	peerIDs   = []libocrtypes.PeerID{{1}, {2}, {3}}
 
-	donID := uint32(1)
-	oracleIDs := []commontypes.OracleID{1, 2, 3}
-	peerIDs := []libocrtypes.PeerID{{1}, {2}, {3}}
-	require.Equal(t, len(oracleIDs), len(peerIDs))
+	ArbPrice = big.NewInt(1).Mul(big.NewInt(5), big.NewInt(1e18))
+	EthPrice = big.NewInt(1).Mul(big.NewInt(7), big.NewInt(1e18))
```
winder, line 64: Just left a comment about this [in another PR](https://github.com/smartcontractkit/chainlink-ccip/pull/217#discussion_r1801148135). I like that you used 1 instead of 0, but I think this is a little better
```suggestion
	ArbPrice = new(big.Int).Mul(big.NewInt(5), big.NewInt(1e18))
	EthPrice = new(big.Int).Mul(big.NewInt(7), big.NewInt(1e18))
```

pkg/reader/price_reader.go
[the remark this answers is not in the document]
winder: nit: we use structured logging pretty much everywhere
```suggestion
		pr.lggr.Debugw("failed to get fee quoter address.", "chain", chain, "err", err)
```

commit/plugin_e2e_test.go
asoliman92: agreed.

pkg/reader/price_reader.go
```diff
@@ -66,23 +72,28 @@ type LatestRoundData struct {
...
+		return updateMap, nil
 	}
-	var updates []plugintypes.TimestampedBig
 
 	boundContract := commontypes.BoundContract{
-		Address: string(pr.FeeQuoterAddress),
+		Address: string(feeQuoterAddress),
```
makramkd, line 90: How does this work in the tests? This seems wrong

commit/tokenprice/processor.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
 	"github.com/smartcontractkit/chainlink-ccip/internal/reader"
+	reader2 "github.com/smartcontractkit/chainlink-ccip/pkg/reader"
```
makramkd, line 18: Nit: bad import alias name

pkg/reader/price_reader.go
```diff
@@ -104,24 +114,28 @@ func (pr *OnchainTokenPricesReader) GetFeeQuoterTokenUpdates(
 	return updateMap, nil
 }
 
-func (pr *OnchainTokenPricesReader) GetTokenFeedPricesUSD(
+func (pr *priceReader) GetFeedPricesUSD(
 	ctx context.Context, tokens []ocr2types.Account,
 ) ([]*big.Int, error) {
 	prices := make([]*big.Int, len(tokens))
+	if pr.feedChainReader == nil {
+		pr.lggr.Debugf("node does not support feed chain")
```
makramkd, line 122: Nit: Log feed chain selector here

pkg/reader/price_reader.go
asoliman92: Why does it seem wrong? Did you see the code just above it that sets the feeQuoter?

pkg/reader/price_reader.go
asoliman92: We already log this on a higher level. We shouldn't even reach this point. Just extra safety.

pkg/reader/price_reader.go
makramkd: `string([]byte)` doesn't give you a hex address, but base64 I think. I'm pretty sure the chain reader needs a hex address. So how does this work 😄

asoliman92: Good catch man. Seems it's failing gracefully and that's why the test is passing

internal/plugintypes/timestamp.go
```diff
@@ -12,13 +12,29 @@ type TimestampedBig struct {
 	Value     cciptypes.BigInt `json:"value"`
 }
 
+// TimestampedUnixBig Maps to on-chain struct
+// https://github.com/smartcontractkit/chainlink/blob/37f3132362ec90b0b1c12fb1b69b9c16c46b399d/contracts/src/v0.8/ccip/libraries/Internal.sol#L43-L47
+//
+//nolint:lll //url
+type TimestampedUnixBig struct {
+	Timestamp uint32           `json:"timestamp"`
```
makramkd, line 20: What kind of timestamp is this, is this millis, seconds, ?

internal/plugintypes/timestamp.go
```diff
@@ -12,13 +12,29 @@ type TimestampedBig struct {
 	Value     cciptypes.BigInt `json:"value"`
 }
 
+// TimestampedUnixBig Maps to on-chain struct
+// https://github.com/smartcontractkit/chainlink/blob/37f3132362ec90b0b1c12fb1b69b9c16c46b399d/contracts/src/v0.8/ccip/libraries/Internal.sol#L43-L47
+//
+//nolint:lll //url
+type TimestampedUnixBig struct {
```
makramkd, line 19: Can we just use this in the observations instead of having two different types?

internal/plugintypes/timestamp.go
```diff
@@ -12,13 +12,29 @@ type TimestampedBig struct {
...
+	Timestamp uint32           `json:"timestamp"`
+	Value     cciptypes.BigInt `json:"value"`
```
0xnogo, line 21: you might want to use `*big.Int`. I had `deserialization` issue using the chain reader.

internal/plugintypes/timestamp.go
asoliman92: In a next PR. For now let's make this PR more about fixes. Changing this will change a lot of things about observations with timestamps

internal/plugintypes/timestamp.go
asoliman92: What's used on chain? 

https://github.com/smartcontractkit/chainlink/blob/37f3132362ec90b0b1c12fb1b69b9c16c46b399d/contracts/src/v0.8/ccip/libraries/Internal.sol#L43-L47

internal/plugintypes/timestamp.go
asoliman92: I think I've been getting some Nil values as well, let me check this out 👍

internal/plugintypes/timestamp.go
makramkd: Check the usage, if its set to block.timestamp then its seconds since the epoch

PR #231, opened 2024-10-15
RMN home reader plumbing

0xnogo:
- Create a reader pkg
- Bug fixing
- go version bump to match chainlink's


winder:
Could you move this to `pkg` instead of creating an alias?

makramkd (approved):

PR #239, opened 2024-10-16
Fix: Token Price uses correct reader

asoliman92:


pkg/reader/price_reader.go
```diff
@@ -184,7 +190,7 @@ func (pr *priceReader) getFeedDecimals(
 ) (uint8, error) {
 	var decimals uint8
 	if err :=
-		pr.feedChainReader.GetLatestValue(
+		pr.feedChainReader().GetLatestValue(
```
makramkd, line 193: Can be nil, is this covered by the nil check above?

pkg/reader/price_reader.go
```diff
@@ -106,7 +112,7 @@ func (pr *priceReader) GetFeeQuoterTokenUpdates(
 	// MethodNameFeeQuoterGetTokenPrices returns an empty update with
 	// a timestamp and price of 0 if the token is not found
 	if err :=
-		pr.feedChainReader.GetLatestValue(
+		pr.chainReaders[chain].GetLatestValue(
```
makramkd, line 115: This can be nil, need to do a nil check somewhere

pkg/reader/price_reader.go
asoliman92: Yes, it's covered by the nil check above, doing extra won't hurt though.

commit/factory.go
```diff
@@ -159,7 +159,7 @@ func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types
 	)
 
 	// The node supports the chain that the token prices are on.
-	tokenPricesCr, ok := p.contractReaders[offchainConfig.PriceFeedChainSelector]
+	_, ok := readers[offchainConfig.PriceFeedChainSelector]
```
winder, line 162: I'm surprised that you need to switch from `p.contractReaders` to `readers`. Is that really required?

We can move the facade mapping to `NewPluginFactory` and avoid using the actual ContractReader type outside of the constructor.

asoliman92: I don't "need" to, it felt more consistent to use the same `facade` everywhere though.

winder (approved):

mateusz-sekara (commented):
Can we have a test covering this fix? I mean failing before applying your fixes

PR #248, opened 2024-10-17
Introduce unknown address type.

winder:
This PR replaces the many account types (string, []byte, types.Account) with `UnknownAddress` and `UnknownEncodedAddress`. These types are now used across the CCIP code and can extended in the future to add chain agnostic encoding and validation functions.

Related chainlink changes:
* https://github.com/smartcontractkit/chainlink/pull/14836


.github/workflows/ccip-integration-test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         with:
           repository: smartcontractkit/chainlink
 #          ref: develop
-          ref: fix-fee-quoter
```
mateusz-sekara, line 34: How about making it work as a manual trigger as well? I mean overriding this value with every PR is very error prone. Maybe ability to trigger the job from GHA with custom branch from chainlink would be sufficient just to run these tests before merging? cc @b-gopalswami

mateusz-sekara (approved):

.github/workflows/ccip-integration-test.yml
makramkd: Yeah, I pitched this idea before and received some pushback. I'm not against it, it should be straightforward as a flow:

* comment on the pull request "cmd branch <chainlink-branch>" to on demand run the test using the appropriate chainlink branch
* some sort of trigger on comment with github actions. Parse the comment and ensure it comes from PR author, etc. then trigger the workflow with the branch name

makramkd (approved):

makramkd:
Please make sure to revert back to `develop` in the GH action prior to merging.

PR #249, opened 2024-10-18
Fix: OnRamp discovery

asoliman92:
* Use another chainSelector other than self to get the destChainConfig
* Fix `onRampDynamicConfig` type to match on-chain

Note: It was silently failing to get the `onRampDynamicConfig` so we were seeing only the symptoms when we were not able to read NativeToken as the flow is:
1. Get Router address
2. Get OnRamp Address
3. Get wrapped token address from Router (step 1)
4. Get FeeQuoter address from OnRamp (step 2)
5. Get native token using fee quoter and the wrapped token address.

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -627,7 +627,7 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
 
 func (r *ccipChainReader) isValidAddress(address []byte, contractName string, chain cciptypes.ChainSelector) bool {
 	if len(address) == 0 {
-		r.lggr.Errorw("address for contract is empty",
+		r.lggr.Warnw("address for contract is empty",
 			"contract", contractName,
 			"chain", chain,
 		)
@@ -1108,6 +1108,9 @@ type onRampDynamicConfig struct {
 	AllowListAdmin         []byte `json:"allowListAdmin"`
 }
 
+// We're wrapping the onRampDynamicConfig this way to map to on-chain return type which is a named struct
+// nolint:lll
+// https://github.com/smartcontractkit/chainlink/blob/12af1de88238e0e918177d6b5622070417f48adf/contracts/src/v0.8/ccip/onRamp/OnRamp.sol#L328
 type getOnRampDynamicConfigResponse struct {
 	DynamicConfig onRampDynamicConfig `json:"dynamicConfig"`
 }
@@ -1187,7 +1190,9 @@ func (r *ccipChainReader) getOnRampDestChainConfig(
 			continue
 		}
 
-		// For chain X, all DestChainConfigs will have same OnRamp address which is the chainX OnRamp
+		// For chain X, all DestChainConfigs will have one of 2 values for the Router address
+		// 1. Chain X Test Router in case we're testing a new lane
+		// 2. Chain X Router
 		chainSel := chainSel
 		eg.Go(func() error {
 			resp := onRampDestChainConfig{}
@@ -1197,15 +1202,12 @@ func (r *ccipChainReader) getOnRampDestChainConfig(
 				consts.MethodNameOnRampGetDestChainConfig,
 				primitives.Unconfirmed,
 				map[string]any{
-					// Any other chain other than self should work as a chain don't have self as a destination
-					// Here we use destChain as the chainSel is looping over all source chains, so all of them should have
-					// a config for the destChain
 					"destChainSelector": r.destChain,
 				},
 				&resp,
 			)
 			if err != nil {
-				return fmt.Errorf("failed to get onramp dynamic config: %w", err)
+				return fmt.Errorf("failed to get onramp dest chain config: %w", err)
 			}
 			mu.Lock()
 			result[chainSel] = resp
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -727,6 +727,10 @@ func (r *ccipChainReader) DiscoverContracts(
 			return nil, fmt.Errorf("unable to lookup source fee quoters (onRamp dynamic config): %w", err)
 		} else {
 			for chain, cfg := range dynamicConfigs {
+				r.lggr.Infow("appending source fee quoter contract address",
+					"chain", chain,
+					"address", cfg.FeeQuoter,
+				)
 				resp = resp.Append(consts.ContractNameFeeQuoter, chain, cfg.FeeQuoter)
 			}
 		}
@@ -742,6 +746,10 @@ func (r *ccipChainReader) DiscoverContracts(
 			return nil, fmt.Errorf("unable to lookup source routers (onRamp dest chain config): %w", err)
 		} else {
 			for chain, cfg := range destChainConfig {
+				r.lggr.Infow("appending Router contract address",
+					"chain", chain,
+					"address", cfg.Router,
+				)
 				resp = resp.Append(consts.ContractNameRouter, chain, cfg.Router)
 			}
 		}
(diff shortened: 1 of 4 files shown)


pkg/reader/ccip.go
```diff
@@ -1087,22 +1100,27 @@ func (r *ccipChainReader) getDestinationData(
...
+type onRampDynamicConfig struct {
+	FeeQuoter              []byte `json:"feeQuoter"`
+	ReentrancyGuardEntered bool   `json:"reentrancyGuardEntered"`
+	MessageInterceptor     []byte `json:"messageInterceptor"`
+	FeeAggregator          []byte `json:"feeAggregator"`
+	AllowListAdmin         []byte `json:"allowListAdmin"`
+}
+
+type getOnRampDynamicConfigResponse struct {
+	DynamicConfig onRampDynamicConfig `json:"dynamicConfig"`
```
makramkd, line 1112: Some comments on why this is structured this way would be useful

pkg/reader/ccip.go
```diff
@@ -1166,17 +1187,20 @@ func (r *ccipChainReader) getOnRampDestChainConfig(
...
 				primitives.Unconfirmed,
 				map[string]any{
-					"destChainSelector": chainSel,
+					// Any other chain other than self should work as a chain don't have self as a destination
+					// Here we use destChain as the chainSel is looping over all source chains, so all of them should have
+					// a config for the destChain
+					"destChainSelector": r.destChain,
 				},
 				&resp,
 			)
```
makramkd, line 1206: Please fix the error message below which is incorrect

```golang
return fmt.Errorf("failed to get onramp dest chain config: %w", err)
```

pkg/reader/ccip.go
```diff
@@ -1166,17 +1187,20 @@ func (r *ccipChainReader) getOnRampDestChainConfig(
...
 			err := r.contractReaders[chainSel].ExtendedGetLatestValue(
 				ctx,
 				consts.ContractNameOnRamp,
 				consts.MethodNameOnRampGetDestChainConfig,
 				primitives.Unconfirmed,
 				map[string]any{
-					"destChainSelector": chainSel,
+					// Any other chain other than self should work as a chain don't have self as a destination
+					// Here we use destChain as the chainSel is looping over all source chains, so all of them should have
+					// a config for the destChain
```
makramkd, line 1202: Wait, this comment doesn't make sense to me. You're getting the dest chain config from each onramp, correct? Then we should always use `r.destChain`. I believe this was the fix I made here but elected not to merge: https://github.com/smartcontractkit/chainlink-ccip/pull/236/files#diff-a16bcafc2684d6be248d96e8fdf5fa77ec115f311c0c73d13f3f942ad0edb0c4R1123

mateusz-sekara (approved):

pkg/reader/ccip.go
```diff
@@ -625,6 +625,17 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
 	}, nil
 }
 
+func (r *ccipChainReader) isValidAddress(address []byte, contractName string, chain cciptypes.ChainSelector) bool {
+	if len(address) == 0 {
+		r.lggr.Errorw("address for contract is empty",
```
makramkd, line 630: Don't use error, it will spam the logs with a big stack trace. Warn here makes more sense to me

makramkd (changes requested):

pkg/reader/ccip.go
asoliman92: I wrote it and deleted it locally 😁. re-adding them 👍

pkg/reader/ccip.go
mateusz-sekara: Hmm, I would say it depends. If this is something that should never happen or happens because of the wrong configuration then error/critical is easier to spot in the logs. If this is a regular check that can fail from time to time then definitely should not be `error`

pkg/reader/ccip.go
asoliman92: > You're getting the dest chain config from each onramp

Yes

> Then we should always use r.destChain

All dest configs have same router address (or maybe test address). Why should it be always `r.destChain`?  In this specific line of code it needs to be `r.destChain` but in general any address other than self that has a destChainConfig should work. I'll rephrase it to reduce confusion.

pkg/reader/ccip.go
asoliman92: I agree with @mateusz-sekara here. If this error was present I could've figured out somethings earlier.

pkg/reader/ccip.go
```diff
@@ -625,6 +625,17 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
 	}, nil
 }
 
+func (r *ccipChainReader) isValidAddress(address []byte, contractName string, chain cciptypes.ChainSelector) bool {
```
winder, line 628: nit: if you return rather than logging this might be helpful for other functions as well.
```suggestion
func isValidAddress(address []byte, contractName string, chain cciptypes.ChainSelector) (bool, error) {
```

pkg/reader/ccip.go
```diff
@@ -625,6 +625,17 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
...
+	if len(address) == 0 {
```
winder, line 629: Check for zero addresses as well? The discovery processor has this helper:
```
// isZero returns true if data is nil or all zeros, otherwise returns false.
func isZero(data []byte) bool {
	for _, v := range data {
		if v != 0 {
			return false
		}
	}
	return true
}
```

pkg/reader/ccip.go
```diff
@@ -1087,22 +1100,31 @@ func (r *ccipChainReader) getDestinationData(
...
+	FeeAggregator          []byte `json:"feeAggregator"`
+	AllowListAdmin         []byte `json:"allowListAdmin"`
+}
+
+// We're wrapping the onRampDynamicConfig this way to map to on-chain return type which is a named struct
+// https://github.com/smartcontractkit/chainlink/blob/12af1de88238e0e918177d6b5622070417f48adf/contracts/src/v0.8/ccip/onRamp/OnRamp.sol#L328
+//
+//nolint:lll
+type getOnRampDynamicConfigResponse struct {
+	DynamicConfig onRampDynamicConfig `json:"dynamicConfig"`
```
winder, line 1116: Which one of these is the fix? If you were to remove the unused fields are the others still returned?

asoliman92: Main fix is the wrapper. I didn't try removing the fields again after adding the wrapper though.

pkg/reader/ccip.go
asoliman92: I don't remember seeing the processor warnings about the empty router though while debugging 🤔. hmmmmm..... maybe I just missed it in the huge log file. Let me remove it so we don't have duplicates for now and keep my eye on it for next couple of PRs.

pkg/reader/ccip.go
asoliman92: I removed it completely for now.

makramkd (approved):

pkg/reader/ccip.go
winder: I meant these fields:
```
type onRampDynamicConfig struct {
	FeeQuoter              []byte `json:"feeQuoter"`
-	ReentrancyGuardEntered bool   `json:"reentrancyGuardEntered"`
-	MessageInterceptor     []byte `json:"messageInterceptor"`
-	FeeAggregator          []byte `json:"feeAggregator"`
-	AllowListAdmin         []byte `json:"allowListAdmin"`
}
```

Hopefully removing those would be OK otherwise we have to worry about things breaking every time the onchain types change.

It probably works for ignoring the implicit structure for named return values also.

The default json encoding ignores "extra" data: https://play.golang.com/p/RN_ICGCaVqV

winder (approved):
LGTM

pkg/reader/ccip.go
makramkd: Yeah, I was referring to those fields as well, I think we tried without them and it didn't work.

PR #255, opened 2024-10-22
RMN related improvements and fixes

dimkouv:
- Improved logging
- Address type related bugfixes
- Update rmn stream cfg

*I will create a ticket to properly address as a follow-up the non chain-agnostic address part*

### Unblocks
https://github.com/smartcontractkit/chainlink/pull/14796

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -1,6 +1,6 @@
 # Plugin Components
 
-* shared types and interfaces - [chainlink-common, ccipocr3](https://github.com/smartcontractkit/chainlink-common/tree/main/pkg/types/ccipocr3)
+* shared types and interfaces - [chainlink-common, ccipocr3](https://github.com/smartcontractkit/chainlink-ccip/tree/main/pkg/types/ccipocr3)
 * [OCR Plugins](https://github.com/smartcontractkit/chainlink-ccip)
   * [commit](https://github.com/smartcontractkit/chainlink-ccip/tree/main/commit)
   * [execute](https://github.com/smartcontractkit/chainlink-ccip/tree/main/execute)


commit/merkleroot/rmn/controller.go
```diff
@@ -310,12 +312,26 @@ func (c *controller) sendObservationRequests(
...
 
+		fixedDestLaneUpdateRequests := make([]*rmnpb.FixedDestLaneUpdateRequest, 0)
+
+		// convert OnRamp address from 32bytes (abi encoded) to 20bytes (evm address) before sending the request
+		// todo: make this part chain-agnostic
+		for _, request := range requests {
+			fixedDestLaneUpdateRequests = append(fixedDestLaneUpdateRequests, &rmnpb.FixedDestLaneUpdateRequest{
+				LaneSource: &rmnpb.LaneSource{
+					SourceChainSelector: request.LaneSource.SourceChainSelector,
+					OnrampAddress:       typconv.KeepNRightBytes(request.LaneSource.OnrampAddress, 20),
```
makramkd, line 323: If we want this to be more general (at least on an interface level) maybe this function can be something like 

```golang
func BytesToAddress(bytesAddress []byte, chainSel uint64) ([]byte, error) {
  family, err := chainsel.GetSelectorFamily(chainSel)
  if err != nil { return nil, fmt.Errorf("get selector family: %w", err) }
  switch family {
    case chainsel.FamilyEVM:
      return bytesAddress[uint(len(b))-n:]
    case chainsel.FamilySol:
      // ...
    default:
      return nil, fmt.Errorf("unknown family: %s", family)
  }
}
```

This API is actually available right now: https://github.com/smartcontractkit/chain-selectors/blob/5b55b2f9e17e46e14481baa94eef8c33a6998b1f/selectors.go#L90. I think we should start cutting over to it in the chainlink-ccip codebase.

makramkd (approved):

PR #257, opened 2024-10-22
Update go version.

winder:


.github/workflows/ccip-integration-test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: ['1.22.5']
+        go-version: ['1.22.0', '1.23.0']
```
makramkd, line 19: Do we need to keep 1.22 in the matrix?

makramkd: Ah 1.22 is required in the workflows, got it.

makramkd:
Still needed?

PR #260, opened 2024-10-23
Billing doc 01

asoliman92:


docs/billing.md
```diff
@@ -1,5 +1,51 @@
 # Billing
 
+Before diving into billing let's do a recap of some parts of important parts of the system.
+
+Note: During the rest of the documentation we'll consider ChainA as source chain and chainB as destinationChain.
```
winder, line 5: If you rename `chainA` to `sourceChain` and `chainB` to `destinationChain` you can skip the note

matYang: +1, source + dest is typically easier to understand

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+
+
+## On Chain
+
+We have multiple contracts interactions that happens to send message from ChainA to ChainB. These contracts are deployed on each chain that need to support CCIP
```
matYang, line 10: `We have multiple contract interactions that happen`

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+1. Router: Non changing initial contract that users/dApps call to send a message
```
matYang, line 11: these contracts should be well-defined enough in other places, should be unnecessary to redefine them in billing doc

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+2. OnRamp: Forwards Router's send request
+3. FeeQuoter: OnRamp calls to estimate the fees for sending a message <<< Important
+4. OffRamp: Receives the messages (Offchain part plays the part here to send the message to the other chain's OffRamp)
+
+So if ChainA is sending a message to ChainB the flow looks like
+
+1. ChainA: Router -> OnRamp -> FeeQuoter
+2. OffChain using OffChain Reporting 3 (OCR3) processes send events emitted from OnRamp in Commit Plugin send a transaction commit on ChainB's OffRamp
+3. ChanB: OffRamp -> FeeQuoter
```
matYang, line 20: this would happen technically not at Commit but at msg execution right

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+
+## OffChain
+
+We have 2 plugins (Commit and Execute).
+
+Commit is the one responsible for initial billing.
```
matYang, line 26: `billing` is too broad, specifically, Commit is responsible for reporting gas prices, and token prices that don't come from keystone

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+
+Execute is responsible for fee boosting [Add Link] during the actual execution of messages.
+
+## Fee Structure
+
+To send a message from ChainA to ChainB we need to account for multiple fees
+
+1. Network/Premium fees.
+2. ChainB Transaction fees (execution fees on the destination chain)
+3. Data Availability fees.
```
matYang, line 36: DA fee is part of transaction fee, you can refer to the public doc: https://docs.chain.link/ccip/billing

Network Fee + Blockchain Transaction Fee, where blockchain transaction fee accounts for:
```
commit + bless overhead
exec overhead
message gas
```
in the form of
EVM Execution cost + DA cost

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+2. OffChain using OffChain Reporting 3 (OCR3) processes send events emitted from OnRamp in Commit Plugin send a transaction commit on ChainB's OffRamp
+3. ChanB: OffRamp -> FeeQuoter
+
+## OffChain
+
+We have 2 plugins (Commit and Execute).
+
+Commit is the one responsible for initial billing.
+
+Execute is responsible for fee boosting [Add Link] during the actual execution of messages.
```
matYang, line 28: Exec is also responsible for batch building, to ensure we fit correct amount of messages into a single offramp transmission;

you might also wanna define what `fee boosting` is, it is basically assigning inflight messages that were previously skipped due to being underpaid an increasing weight for execution as time passes

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+To send a message from ChainA to ChainB we need to account for multiple fees
+
+1. Network/Premium fees.
+2. ChainB Transaction fees (execution fees on the destination chain)
+3. Data Availability fees.
+
+Of-course the details of the calculation will depend on the message being sent and whether it has tokens to send or not, data availability..etc.
+For in details look on how the fees is calculated you can check FeeQuoter's `getValidatedFee` (TODO: Put link)
+
+To be able to pay in one of the available fee tokens on ChainA we need to convert whatever the estimated fees the user will pay on ChainB
```
matYang, line 41: we need to arrive at a quote in USD and convert the USD amount into corresponding fee token amount.

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+2. ChainB Transaction fees (execution fees on the destination chain)
+3. Data Availability fees.
+
+Of-course the details of the calculation will depend on the message being sent and whether it has tokens to send or not, data availability..etc.
+For in details look on how the fees is calculated you can check FeeQuoter's `getValidatedFee` (TODO: Put link)
+
+To be able to pay in one of the available fee tokens on ChainA we need to convert whatever the estimated fees the user will pay on ChainB 
+
+So the components we need to calculate the final price are:
+1. ChainA fee token price, usually LINK and the native token of the chain. This is what the user pays in the end. TokenPriceProcessor will update them. [TODO: link]
```
matYang, line 44: my understanding is long-term we want to ingest from KS, is that still correct?

docs/billing.md
```diff
@@ -1,5 +1,51 @@
...
+2. ChainB Fee/Gas price. FeeChainProcessor will update them. [TODO: link]
+3. ChainB native token price (to be able to calculate the full fees on execution as the Gas from step 2 is usually in the native token price)
```
matYang, line 46: `to be able to calculate the fees denominated in USD`

docs/billing.md
asoliman92: yes for the reverse directions

docs/billing.md
asoliman92: That's still correct yes.

docs/billing.md
asoliman92: I purposefully didn't use source and dest so that when I reach the part where each source (which is dest) adds other sources' gas prices. I'll finish the doc and if you still think it's better to rename I'll do so.

docs/billing.md
asoliman92: Added commit plugin on the arrow, does that make it clearer?

docs/billing.md
asoliman92: done

docs/billing.md
```diff
@@ -1,11 +1,84 @@
 # Billing
 
-## Token Prices
+Before diving into billing let's do a recap of some parts of important parts of the system.
```
matYang, line 3: `some important parts`

docs/billing.md
matYang: Router: Immutable entry point contract that users/dapps call to get quotes and send messages

matYang (approved):
other than the small nits mentioned above, LGTM

docs/billing.md
```diff
@@ -1,11 +1,84 @@
...
+## Token Prices Processor
+
+During commit round. Token prices are updated by TokenPriceProcessor. 
+
+For each round these are the steps:
+1. **Observation:**  
+   a. Fetches the token prices from USD feed for tokens we [configure](https://github.com/smartcontractkit/chainlink-ccip/blob/f03ff5183eb8323ba8e0a13dc58d1da13b755307/pluginconfig/commit.go#L89-L91) during the plugin initiation.   
+   b. Fetches current token prices stored in **destination chain FeeQuoter** (the chain the current node is supposed to commit to).
+2. **Outcome:**  
+Cross-check values from 1a and 1b. and posts the tokens that needs updating in the Outcome. The prices from the feed (1a) will be used when:
```
0xAustinWang, line 54: Does this mean the tokenPrice from the FeeQuoter is used in other situations? Or what happens if FeeQuoter and Feed both have price?

Is there a situation when the Feed is unavailable?

docs/billing.md
```diff
@@ -1,11 +1,84 @@
...
+We have multiple contracts interactions that happen to send message from sourceChain to destinationChain. These contracts are deployed on each chain that need to support CCIP
+1. Router: Non changing initial contract that users/dApps call to send a message
+2. OnRamp: Forwards Router's send request
+3. FeeQuoter: OnRamp calls to estimate the fees for sending a message <<< Important
+4. OffRamp: Receives the messages (Offchain part plays the part here to send the message to the other chain's OffRamp)
+
+So if sourceChain is sending a message to destinationChain the flow looks like
+
+1. sourceChain: Router -> OnRamp -> FeeQuoter
+2. OffChain using OffChain Reporting 3 (OCR3) processes send events emitted from OnRamp in Commit Plugin send a transaction commit on destinationChain's OffRamp
```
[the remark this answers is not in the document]
asoliman92: token and gas prices are commit plugin's responsibility only. What do you think we need the exec plugin here for?

docs/billing.md
asoliman92: If FeeQuoter has price that don't lie in one of the 3 categories mentioned will not be updated.


> Is there a situation when the Feed is unavailable?

We don't account for this. This is an offchain config that's injected during plugin initialization and need to be present before the plugin starts.

makramkd:
Diagram would be good

winder (approved):

asoliman92:
billing documentation

PR #262, opened 2024-10-23
Add gasLimit and tokenTransferGas to CalculateMessageMaxGas

rstout:
Parse gasLimit and tokenTransferGas from ccipocr3.Message in CalculateMessageMaxGas and add these values to the gas estimate total.


execute/internal/gas/evm/gas_helpers.go
```diff
@@ -54,7 +92,28 @@ func (gp EstimateProvider) CalculateMessageMaxGas(msg ccipocr3.Message) uint64 {
 	var data []byte = msg.Data
 	dataLength := len(data)
 
-	// TODO: parse max gas from ExtraArgs.
+	// If msg is an EVM message, its ExtraArgs field is an encoded EVMExtraArgsV2 struct, defined in Client.sol here:
```
matYang, line 95: should reference the contracts in chainlink repo, afaik v2 code has been ported and actively developed there

execute/internal/gas/evm/gas_helpers.go
```diff
@@ -54,7 +92,28 @@ func (gp EstimateProvider) CalculateMessageMaxGas(msg ccipocr3.Message) uint64 {
...
+	//nolint:lll
+	// https://github.com/smartcontractkit/ccip/blob/dee09c782c17de1e37e3a1cf625d430330532c6d/contracts/src/v0.8/ccip/libraries/Client.sol#L47
+	gasLimit, err := decodeEVMExtraArgsV2GasLimit(msg.ExtraArgs)
```
matYang, line 98: can we switch based on the first 4 bytes of selector and use the corresponding extra arg decoder version, as opposed to hardcoded v2; the v2 extra args selector is defined in `clients.sol`: 
```
bytes4 public constant EVM_EXTRA_ARGS_V2_TAG = 0x181dcf10;
```

execute/internal/gas/evm/gas_helpers.go
asoliman92: I think it's okay if it's a permanent link. contracts shouldn't be changing anyways.

execute/internal/gas/evm/gas_helpers.go
rstout: This logic will only handle v2 encoded args though, right? I guess it makes sense to confirm the first 4 bytes are the V2 tag.

> btw shouldn't this be implemented somewhere else to also parse the OOO flag?

I copied from this function in `chainlink`, which also just picks out the gasLimit:
https://github.com/smartcontractkit/chainlink/blob/c1a39ebbf0226d1efafac07034b16625d45b3f82/core/capabilities/ccip/ccipevm/helpers.go#L11

execute/internal/gas/evm/gas_helpers.go
rstout: I think it's better to have a link that may one day break than not have any link

execute/internal/gas/evm/gas_helpers.go
```diff
@@ -54,7 +92,28 @@ func (gp EstimateProvider) CalculateMessageMaxGas(msg ccipocr3.Message) uint64 {
...
+	if err != nil {
+		// TODO: metrics
+		gp.lggr.Error("decoding EVMExtraArgsV2 gas limit", err)
```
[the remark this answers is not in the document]
rstout: Agreed

execute/internal/gas/evm/gas_helpers.go
```diff
@@ -30,7 +37,38 @@ const (
 		100 //# SLOAD_GAS = WARM_STORAGE_READ_COST for rewriting from non-zero (in-progress) to non-zero (success/failure)
 )
 
+var (
+	abiUint32  = ABITypeOrPanic("uint32")
+	abiUint256 = ABITypeOrPanic("uint256")
+	abiBool    = ABITypeOrPanic("bool")
+
+	// ExtraArgsV2ABI ABI types for EVMExtraArgsV2 (see Client.sol)
+	ExtraArgsV2ABI = abi.Arguments{
```
mateusz-sekara, line 46: I guess referencing ABI/go-ethereum structs in the plugin code is forbidden as we want to stay chain agnostic. Some time ago I faced similar issues with abi encode/decode for the USDC specific attestation payloads. I was told to solve that for defining the interface in the `chailink-ccip` but providing implementation in the chainlink [here](https://github.com/smartcontractkit/chainlink/blob/66cd65e8e84848749853f907c700af154619cb2c/core/capabilities/ccip/ccipevm/tokendata.go)

PR #272, opened 2024-10-28
Da price logic

0xnogo:


pkg/consts/consts.go
```diff
@@ -44,6 +44,7 @@ const (
 	MethodNameFeeQuoterGetTokenPrice        = "GetTokenPrice"
 	MethodNameGetFeePriceUpdate             = "GetDestinationChainGasPrice"
 	MethodNameFeeQuoterGetStaticConfig      = "GetStaticConfig"
+	MethodNameFeeQuoterGetDynamicConfig     = "GetDynamicConfig"
```
rstout, line 47: afaict `GetDynamicConfig` is not a method of FeeQuoter, I do see it in OffRamp and OnRamp though.

pkg/reader/ccip.go
```diff
@@ -916,6 +929,34 @@ func (r *ccipChainReader) getFeeQuoterTokenPriceUSD(ctx context.Context, tokenAd
...
+type destChainConfig struct {
+	DestDataAvailabilityOverheadGas   uint32 `json:"destDataAvailabilityOverheadGas"`
+	DestGasPerDataAvailabilityByte    uint16 `json:"destGasPerDataAvailabilityByte"`
+	DestDataAvailabilityMultiplierBps uint16 `json:"destDataAvailabilityMultiplierBps"`
+}
+
+// getDestChainConfig returns the destination chain's Fee Quoter's DestChainConfig
+func (r *ccipChainReader) getDestChainConfig(ctx context.Context) (destChainConfig, error) {
+	var dynamicConfig destChainConfig
+	err := r.getDestinationData(
```
rstout, line 945: I think there's a few issues here. First, I don't think `GetDynamicConfig` is a FeeQuoter method. I think the correct method is `getDestChainConfig`, but this function takes `destChainSelector` as a param, so that would need to be supplied. Also, we can't get the dest chain's `FeeQuoter.DestChainConfig` from the FeeQuoter instance on the dest chain, [see this thread](https://chainlink-core.slack.com/archives/C02PBKHCP28/p1729635837527769). So we'll need to get `DestChainConfig` from the FeeQuoter on a different chain, but maybe there's a better way. I think we'd also need to ensure that the encoded `DestChainConfig` we get from FeeQuoter can be parsed into the elided/simplified `destChainConfig` struct we've defined above.

execute/exectypes/observation.go
```diff
@@ -8,6 +8,13 @@ import (
 	dt "github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery/discoverytypes"
 )
 
+// MessageWithSrcChain is a message wrapper that includes the source chain selector.
+// It is used to facilitate the fetching of DestChainConfig from the src chain FeeQuoter.
+type MessageWithSrcChain struct {
+	Message     cciptypes.Message
+	SourceChain cciptypes.ChainSelector
```
makramkd, line 15: Source chain selector is already included in the message header, this is not needed

pkg/reader/ccip.go
```diff
@@ -851,6 +851,23 @@ func (r *ccipChainReader) LinkPriceUSD(ctx context.Context) (cciptypes.BigInt, e
 	return linkPriceUSD, nil
 }
 
+func (r *ccipChainReader) GetDataAvailabilityConfig(ctx context.Context, srcChain cciptypes.ChainSelector) (
+	destDAOverheadGas, destGasPerDAByte, destDAMultiplierBps int64, err error) {
```
makramkd, line 855: This is super EVM specific, whats the plan for non-EVM?

pkg/reader/ccip.go
0xnogo: As discussed offline, we will fetch all the possible configs for a dest chain and take the median

pkg/reader/ccip.go
rstout: Either non-EVM chains will need to expose cost params or there needs to be a cost-estimation service we can call. We should have a plan for this, but we should move forward with this in the meantime.

PR #275, opened 2024-10-29
Add DA cost estimation to CCIPMessageExecCostUSD18Calculator

rstout:


execute/exectypes/costly_messages.go
```diff
@@ -364,4 +376,55 @@ func (c *CCIPMessageExecCostUSD18Calculator) computeExecutionCostUSD18(
...
+	if dataAvailabilityFee == nil || dataAvailabilityFee.Cmp(big.NewInt(0)) == 0 {
+		return big.NewInt(0)
+	}
+
+	messageGas := CalculateMessageMaxDAGas(message, daConfig)
+	return big.NewInt(0).Mul(messageGas, dataAvailabilityFee)
+}
+
+// CalculateMessageMaxDAGas calculates the total DA gas needed for a CCIP message
+func CalculateMessageMaxDAGas(
```
mateusz-sekara, line 394: Why public? Seems we don't use it anywhere outside of this pkg

pkg/reader/ccip.go
```diff
@@ -1292,5 +1294,71 @@ func (r *ccipChainReader) getRMNRemoteAddress(
...
+
+	if err != nil {
+		return cciptypes.FeeQuoterDestChainConfig{}, fmt.Errorf("failed to get dest chain config: %w", err)
+	}
+
+	return destChainConfig, nil
+}
+
+// GetMedianDataAvailabilityGasConfig returns the median of the DataAvailabilityGasConfig values from all FeeQuoters
+func (r *ccipChainReader) GetMedianDataAvailabilityGasConfig(
```
mateusz-sekara, line 1327: Some doc clarifying the decision behind it would be great. Why do we get all the configs? Why do compute a median from them etc

pkg/reader/ccip.go
```diff
@@ -1292,5 +1294,71 @@ func (r *ccipChainReader) getRMNRemoteAddress(
...
+	ctx context.Context,
+) (cciptypes.DataAvailabilityGasConfig, error) {
+	overheadGasValues := make([]uint32, 0)
+	gasPerByteValues := make([]uint16, 0)
+	multiplierBpsValues := make([]uint16, 0)
+	for chain := range r.contractReaders {
+		config, err := r.getFeeQuoterDestChainConfig(ctx, chain)
```
mateusz-sekara, line 1334: We have a ticket for reviewing all the bottlenecks like this so we will need to rework that sooner or later. Looping over all contract readers and fetching data from them synchronously and sequentially will be a major bottleneck. Can you please leave a `TODO/FIXME` comment in the code to take a closer look at the performance? It's gonna be easier to identify this place when working on perf issues globally

execute/exectypes/costly_messages.go
```diff
@@ -364,4 +376,55 @@ func (c *CCIPMessageExecCostUSD18Calculator) computeExecutionCostUSD18(
...
+	msg cciptypes.Message,
+	daConfig cciptypes.DataAvailabilityGasConfig,
+) *big.Int {
+	// Calculate token data length
+	var totalTokenDataLen int
+	for _, tokenAmount := range msg.TokenAmounts {
```
mateusz-sekara, line 400: I guess this is inspired by what we have in the solidity right? Can you link in the comment source of the truth?

mateusz-sekara (changes requested):

pkg/reader/ccip.go
0xnogo: Good point

mateusz-sekara (approved):

execute/exectypes/costly_messages.go
```diff
@@ -364,4 +376,55 @@ func (c *CCIPMessageExecCostUSD18Calculator) computeExecutionCostUSD18(
...
+	if dataAvailabilityFee == nil || dataAvailabilityFee.Cmp(big.NewInt(0)) == 0 {
+		return big.NewInt(0)
+	}
+
+	messageGas := calculateMessageMaxDAGas(message, daConfig)
+	return big.NewInt(0).Mul(messageGas, dataAvailabilityFee)
+}
+
+// calculateMessageMaxDAGas calculates the total DA gas needed for a CCIP message
+func calculateMessageMaxDAGas(
```
matYang, line 394: DA is quite specific to EVM, I'm not expert of v2 code base, wondering if you want to put it under an EVM folder somewhere?

execute/exectypes/costly_messages.go
```diff
@@ -364,4 +376,55 @@ func (c *CCIPMessageExecCostUSD18Calculator) computeExecutionCostUSD18(
...
+	msg cciptypes.Message,
+	daConfig cciptypes.DataAvailabilityGasConfig,
+) *big.Int {
+	// Calculate token data length
+	var totalTokenDataLen int
+	for _, tokenAmount := range msg.TokenAmounts {
+		totalTokenDataLen += len(tokenAmount.SourcePoolAddress) +
```
matYang, line 401: given we are already calcing DA gas in EVM specific manner, we can follow the fee quoter logic:
```
total_len = fixed_overhead + len(tokenAmount.ExtraData) + len(tokenAmount.DestExecData)
where
fixed_overhead = ANY_2_EVM_MESSAGE_FIXED_BYTES_PER_TOKEN = 32 * ((2 * 3) + 3);
see Internal.sol for breakdown
```

execute/exectypes/costly_messages.go
```diff
@@ -15,6 +15,12 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
+const (
+	EvmWordBytes             = 32
+	ConstantMessagePartBytes = 10 * 32 // A message consists of 10 abi encoded fields 32B each (after encoding)
```
matYang, line 20: Internal.sol marked the overhead as 14 lots, see
```
ANY_2_EVM_MESSAGE_FIXED_BYTES = 32 * 14;
```
it's 14 as opposed to 10 because EVM ABI encoding adds 2 slots (offset + length) for every dynamic struct field

execute/exectypes/costly_messages.go
```diff
@@ -364,4 +376,55 @@ func (c *CCIPMessageExecCostUSD18Calculator) computeExecutionCostUSD18(
...
+	}
+
+	// Calculate total message data length
+	dataLen := ConstantMessagePartBytes +
+		len(msg.Data) +
+		len(msg.Sender) +
+		len(msg.Receiver) +
+		len(msg.ExtraArgs) +
+		len(msg.FeeToken) +
+		EvmWordBytes*2 + // FeeTokenAmount and FeeValueJuels
```
matYang, line 415: ExtraArgs /FeeToken/FeeTokenAmount/FeeValueJuels are not posted to dest chain, so we don't need to account for them here;

I think `Receiver` is already accounted for in the overhead too, so you just need
```
overhead + len(data) + len(sender) + len_token
```

execute/exectypes/costly_messages.go
```diff
@@ -364,4 +376,55 @@ func (c *CCIPMessageExecCostUSD18Calculator) computeExecutionCostUSD18(
...
+	dataAvailabilityFee *big.Int,
+	daConfig cciptypes.DataAvailabilityGasConfig,
+	message cciptypes.Message,
+) plugintypes.USD18 {
+	if dataAvailabilityFee == nil || dataAvailabilityFee.Cmp(big.NewInt(0)) == 0 {
+		return big.NewInt(0)
+	}
+
+	messageGas := calculateMessageMaxDAGas(message, daConfig)
+	return big.NewInt(0).Mul(messageGas, dataAvailabilityFee)
```
matYang, line 390: is this actually returning fee in USD or fee in the native currency? 

there is a pre-existing comment that says `executionFee (USD18/gas)` so I presume dataAvailabilityFee is denoted in USD/gas as opposed to native/gas, but after reading the code, it seems `feeComponents` is read straight from the ChainWriter, does chain writer return native/gas (which I thought should be the case and will make this impl incorrect). or actually usd/gas, please double check

execute/exectypes/costly_messages.go
0xnogo: You right. We will work on making this logic chain abstracted in a second step

execute/exectypes/costly_messages.go
rstout: Looks like USD/gas: https://github.com/smartcontractkit/chainlink/blob/a1e4f8e960d4b5bf05e353dec3254d13c9a3b4c8/contracts/src/v0.8/ccip/FeeQuoter.sol#L303

```  
function getDestinationChainGasPrice(
    uint64 destChainSelector
  ) external view returns (Internal.TimestampedPackedUint224 memory) {
    return s_usdPerUnitGasByDestChainSelector[destChainSelector];
  }
```

execute/exectypes/costly_messages.go
rstout: The priority is to unblock the itests right now, so our plan is to merge this, then move this logic to chainlink with the other EVM specific logic

jmank88 (approved):

execute/exectypes/costly_messages.go
```diff
@@ -15,6 +15,13 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
+const (
+	EvmWordBytes              = 32
```
jmank88, line 19: nit/
```suggestion
	EVMWordBytes              = 32
```

execute/exectypes/costly_messages.go
matYang: I'm still confused, isn't the value fetched from ChainWriter, maybe I missed it, where is it read from FeeQuoter?

matYang (approved):
approving this PR, with the expectation that a conversion from native in units of wei to usd will be added later

PR #277, opened 2024-10-30
Improve RMN stream configuration

dimkouv:
Improve RMN stream configuration by computing the parameters.

### DIFF
--- a/commit/merkleroot/rmn/streamconfig.go
+++ b/commit/merkleroot/rmn/streamconfig.go
@@ -26,6 +26,10 @@ const (
 	rateScale     = 1.2
 	capacityScale = 3
 
+	// bufferSize should be set to 1 as advised by the RMN team.
+	outgoingBufferSize = 1
+	incomingBufferSize = 1
+
 	estimatedRoundInterval = time.Second
 )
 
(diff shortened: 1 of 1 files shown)


commit/merkleroot/rmn/streamconfig.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+func newStreamConfig(
+	lggr logger.Logger,
+	streamName string,
+) networking.NewStreamArgs1 {
+	cfg := networking.NewStreamArgs1{
+		StreamName:         streamName,
+		OutgoingBufferSize: 1, // should always be set to 1.
+		IncomingBufferSize: 1, // should always be set to 1.
```
0xnogo, line 44: Could we add these magic numbers as a const? An explanation to why they should always be set to 1 would be helpful.

commit/merkleroot/rmn/streamconfig.go
```diff
@@ -0,0 +1,160 @@
...
+				ClosedInterval: &rmnpb.ClosedInterval{
+					MinMsgNr: math.MaxUint64,
+					MaxMsgNr: math.MaxUint64,
+				},
+			},
+		)
+	}
+	reqBytes, err := proto.Marshal(req)
+	if err != nil {
+		panic(err)
```
0xnogo, line 112: what if it panics here? Should we propagate an error instead of panicking?

0xnogo (approved):

PR #288, opened 2024-11-01
pkg/peergroup: use libocr interfaces directly

makramkd:
Chainlink fix: https://github.com/smartcontractkit/chainlink/pull/15066


.mockery.yaml
```diff
@@ -26,14 +26,18 @@ packages:
...
             Extended:
             ContractReaderFacade:
     github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3:
         interfaces:
             ExecutePluginCodec:
+    github.com/smartcontractkit/libocr/networking:
+        interfaces:
+            PeerGroupFactory:
+                config:
+                    dir: mocks/libocr_networking/
```
makramkd, line 40: Needed otherwise the dir would be `github.com/smartcontrackit/libocr/networking`

PR #291, opened 2024-11-01
Remove abusive check on destchainConfig

0xnogo:
`TestInitialDeploy` and `TestAddLane` are successful.

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -1397,10 +1397,6 @@ func (r *ccipChainReader) GetMedianDataAvailabilityGasConfig(
 		}
 	}
 
-	if len(overheadGasValues) == 0 {
-		return cciptypes.DataAvailabilityGasConfig{}, fmt.Errorf("no valid fee quoter destChainConfigs found")
-	}
-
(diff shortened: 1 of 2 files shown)


pkg/reader/ccip.go
```diff
@@ -1397,10 +1397,6 @@ func (r *ccipChainReader) GetMedianDataAvailabilityGasConfig(
 		}
 	}
 
-	if len(overheadGasValues) == 0 {
-		return cciptypes.DataAvailabilityGasConfig{}, fmt.Errorf("no valid fee quoter destChainConfigs found")
-	}
-
```
makramkd, line 1403: If this slice is empty, what are the returned values? All zeros? Will that break stuff downstream or nah?

makramkd (approved):

pkg/reader/ccip.go
0xnogo: the rest of the flow only involve multiplication and addition. I don't think it will have an impact on it.

Basically:
```
totalCost = exec cost + da cost
da cost will be equal to 0
```

PR #295, opened 2024-11-04
Commit plugin factory improvements

dimkouv:
- Choose commit plugin query size based on computations
- Validations and tests
- Choose better panic messages

### DIFF
--- a/commit/factory.go
+++ b/commit/factory.go
@@ -29,7 +29,7 @@ import (
 
 // maxQueryLength is set to twice the maximum size of a theoretical merkle root processor query
 // that assumes 1,000 source chains and 256 (theoretical max) RMN nodes.
-const maxQueryLength = 603_872
+const maxQueryLength = 615_520
 
 // PluginFactoryConstructor implements common OCR3ReportingPluginClient and is used for initializing a plugin factory
 // and a validation service.


commit/factory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	// This test will verify that the maxQueryLength constant is set to a proper value.
+
+	// Estimate the maximum number of source chains we are going to ever have.
+	// This value should be tweaked after we are close to supporting that many chains.
+	const estimateMaxNumberOfSourceChains = 1000
+
+	// Estimate the maximum number of RMN report signers we are going to ever have.
+	// This value is defined in RMNRemote contract as `f`.
+	// This value should be tweaked if necessary in order to define new limits.
+	const estimatedMaxRmnReportSigners = 200
```
makramkd, line 27: Lets use the theoretical max which is 256 here

commit/factory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn/rmnpb"
+	"github.com/smartcontractkit/chainlink-ccip/commit/tokenprice"
+)
+
+func Test_maxQueryLength(t *testing.T) {
+	// This test will verify that the maxQueryLength constant is set to a proper value.
+
+	// Estimate the maximum number of source chains we are going to ever have.
+	// This value should be tweaked after we are close to supporting that many chains.
+	const estimateMaxNumberOfSourceChains = 1000
```
makramkd, line 22:
```suggestion
	const estimatedMaxNumberOfSourceChains = 1000
```

commit/factory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	q := Query{
+		MerkleRootQuery: merkleroot.Query{
+			RetryRMNSignatures: true,
+			RMNSignatures: &rmn.ReportSignatures{
+				Signatures:  sigs,
+				LaneUpdates: laneUpdates,
+			},
+		},
+		TokenPriceQuery: tokenprice.Query{},
+		ChainFeeQuery:   chainfee.Query{},
```
makramkd, line 58: If these are not specified do we save on some space or are these not `omitempty`?

commit/factory.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
 )
 
-const maxQueryLength = 1024 * 1024 // 1MB
+const maxQueryLength = 603_872
```
makramkd, line 30:
```suggestion
// maxQueryLength is set to twice the maximum size of a theoretical merkle root processor query
// that assumes 1,000 source chains and 256 (theoretical max) RMN nodes.
const maxQueryLength = 603_872
```

commit/factory_test.go
```diff
@@ -0,0 +1,65 @@
...
+	for i := range sigs {
+		sigs[i] = &rmnpb.EcdsaSignature{R: make([]byte, 32), S: make([]byte, 32)}
+	}
+
+	laneUpdates := make([]*rmnpb.FixedDestLaneUpdate, estimateMaxNumberOfSourceChains)
+	for i := range laneUpdates {
+		laneUpdates[i] = &rmnpb.FixedDestLaneUpdate{
+			LaneSource: &rmnpb.LaneSource{
+				SourceChainSelector: math.MaxUint64,
+				OnrampAddress:       make([]byte, 40),
```
makramkd, line 39: Does this actually get encoded to enough base64 zeros?

makramkd (commented):
Awesome idea, just some cleanup + questions

makramkd (approved):

PR #296, opened 2024-11-04
Commit plugin parameter tuning

dimkouv:
Estimate and set more proper OCR parameters

### DIFF
--- a/commit/factory.go
+++ b/commit/factory.go
@@ -46,9 +46,10 @@ const (
 	// maxObservationLength is set to the maximum size of an observation
 	maxObservationLength = 1_047_202
 
-	maxOutcomeLength = 10_000
+	// maxOutcomeLength is set to the maximum size of an outcome
+	maxOutcomeLength = 1_167_765
 
-	maxReportLength = 10_000
+	maxReportLength = 993_982
 
 	// maxReportCount is set to 1 because the commit plugin only generates one report per round.
 	maxReportCount = 1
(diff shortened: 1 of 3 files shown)


commit/factory.go
```diff
@@ -27,9 +27,35 @@ import (
...
+	// Estimated maximum number of RMN nodes the system will support.
+	estimatedMaxRmnNodesCount = 256
+
+	// Estimated maximum number of priced tokens that the Commit DON supports.
+	// This value does not indicate a system limitation but just an estimation to properly tune the OCR parameters.
+	// The value can be adjusted as needed.
+	estimatedMaxNumberOfPricedTokens = 10_000
+
+	// maxQueryLength is set to twice the maximum size of a theoretical merkle root processor query
+	// that assumes 1,000 source chains and 256 (theoretical max) RMN nodes.
```
makramkd, line 45:
```suggestion
	// that assumes 900 source chains and 256 (theoretical max) RMN nodes.
```

Should be 900 since you updated the above. Or maybe reference `estimatedMaxNumberOfSourceChains` directly

commit/factory.go
```diff
@@ -27,9 +27,35 @@ import (
...
+	maxQueryLength = 559_320
+
+	// maxObservationLength is set to the maximum size of an observation
```
makramkd, line 48: Maybe refer to the tests for each of these maxes too

makramkd (approved):

PR #308, opened 2024-11-08
add fchain validation for fee token observation

0xAustinWang:
https://smartcontract-it.atlassian.net/browse/CCIP-3980

### DIFF
--- a/commit/tokenprice/validate_observation.go
+++ b/commit/tokenprice/validate_observation.go
@@ -2,7 +2,9 @@ package tokenprice
 
 import (
 	"fmt"
+
 	mapset "github.com/deckarep/golang-set/v2"
+
 	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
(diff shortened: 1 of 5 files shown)


commit/tokenprice/validate_observation.go
```diff
@@ -0,0 +1,64 @@
...
+	if err := validateObservedTokenPrices(obs.FeedTokenPrices); err != nil {
+		return fmt.Errorf("failed to validate observed token prices: %w", err)
+	}
+
+	return nil
+}
+
+func validateFChain(fChain map[cciptypes.ChainSelector]int) error {
+	for chainSelector, f := range fChain {
+		if f < 0 {
```
asoliman92, line 43: I don't think we should allow 0 fChain.
```suggestion
		if f <= 0 {
```

commit/tokenprice/validate_observation_test.go
```diff
@@ -64,3 +64,40 @@ func Test_validateObservedTokenPrices(t *testing.T) {
...
+			},
+			expErr: true,
+		},
+		{
+			name: "FChain valid",
+			fChain: map[cciptypes.ChainSelector]int{
+				12: 6,
+				7:  9,
+			},
+			expErr: false,
```
asoliman92, line 88: Maybe add a test for zero fChain as well

commit/tokenprice/validate_observation.go
asoliman92: same for other validations in other files.

asoliman92 (approved):

PR #312, opened 2024-11-08
background / cached / concurrent token data observations

dimkouv:
Implement `NewBackgroundObserver()` and replace the existing observer with it.

1. It accepts an existing token data observer to make the actual calls.
2. Performs token data fetching asynchronously. This requires at least two `Observe` calls to get the token data.
3. Performs the requests concurrently up to a maximum configured number of spawned goroutines.
4. Keeps the results in the cache for a configurable duration.
5. Cleans up the expired data from the cache on a configurable duration.
6. Handling of re-orgs.
7. Rate limiting for retries.

### DIFF
--- a/execute/factory.go
+++ b/execute/factory.go
@@ -14,7 +14,7 @@ import (
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
 	ragep2ptypes "github.com/smartcontractkit/libocr/ragep2p/types"
 
-	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
+	"github.com/smartcontractkit/chainlink-ccip/execute/costlymessages"
 	"github.com/smartcontractkit/chainlink-ccip/execute/internal/gas"
 	"github.com/smartcontractkit/chainlink-ccip/execute/tokendata"
 	"github.com/smartcontractkit/chainlink-ccip/internal/plugintypes"
@@ -165,7 +165,7 @@ func (p PluginFactory) NewReportingPlugin(
 		return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to create token data observer: %w", err)
 	}
 
-	costlyMessageObserver := exectypes.NewCostlyMessageObserverWithDefaults(
+	costlyMessageObserver := costlymessages.NewObserverWithDefaults(
 		p.lggr,
 		true,
 		ccipReader,
--- a/execute/costlymessages/costly_messages.go
+++ b/execute/costlymessages/costly_messages.go
@@ -1,4 +1,4 @@
-package exectypes
+package costlymessages
 
 import (
 	"context"
@@ -23,8 +23,8 @@ const (
 	daMultiplierBase          = 10_000  // DA multiplier is in multiples of 0.0001, i.e. 1/daMultiplierBase
 )
 
-// CostlyMessageObserver observes messages that are too costly to execute.
-type CostlyMessageObserver interface {
+// Observer observes messages that are too costly to execute.
+type Observer interface {
 	// Observe takes a set of messages and returns a slice of message IDs that are too costly to execute.
 	// Also takes in a map from message ID to the message's timestamp, needed to calculate fee boosting.
 	Observe(
@@ -34,18 +34,18 @@ type CostlyMessageObserver interface {
 	) ([]cciptypes.Bytes32, error)
 }
 
-// NewCostlyMessageObserverWithDefaults creates a new CostlyMessageObserver with default calculators.
+// NewObserverWithDefaults creates a new Observer with default calculators.
 // The default calculators are:
 // - CCIPMessageFeeUSD18Calculator
 // - CCIPMessageExecCostUSD18Calculator
-func NewCostlyMessageObserverWithDefaults(
+func NewObserverWithDefaults(
 	lggr logger.Logger,
 	enabled bool,
 	ccipReader readerpkg.CCIPReader,
 	relativeBoostPerWaitHour float64,
 	estimateProvider gas.EstimateProvider,
-) CostlyMessageObserver {
-	return NewCostlyMessageObserver(
+) Observer {
+	return NewObserver(
 		lggr,
 		enabled,
 		NewCCIPMessageFeeUSD18Calculator(
@@ -62,15 +62,15 @@ func NewCostlyMessageObserverWithDefaults(
 	)
 }
 
-// NewCostlyMessageObserver allows to specific feeCalculator and execCostCalculator.
+// NewObserver allows to specific feeCalculator and execCostCalculator.
 // Therefore, it's very convenient for testing.
-func NewCostlyMessageObserver(
+func NewObserver(
 	lggr logger.Logger,
 	enabled bool,
 	feeCalculator MessageFeeE18USDCalculator,
 	execCostCalculator MessageExecCostUSD18Calculator,
-) CostlyMessageObserver {
-	return &CCIPCostlyMessageObserver{
+) Observer {
+	return &observer{
 		lggr:               lggr,
 		enabled:            enabled,
 		feeCalculator:      feeCalculator,
@@ -78,7 +78,7 @@ func NewCostlyMessageObserver(
 	}
 }
 
-type CCIPCostlyMessageObserver struct {
+type observer struct {
 	lggr               logger.Logger
 	enabled            bool
 	feeCalculator      MessageFeeE18USDCalculator
@@ -88,13 +88,13 @@ type CCIPCostlyMessageObserver struct {
 // Observe returns a slice of message IDs that are too costly to execute.
 // It calculates the fee and execution cost of each message. The messages are considered too costly if the fee is less
 // than the execution cost.
-func (o *CCIPCostlyMessageObserver) Observe(
+func (o *observer) Observe(
 	ctx context.Context,
 	messages []cciptypes.Message,
 	messageTimestamps map[cciptypes.Bytes32]time.Time,
 ) ([]cciptypes.Bytes32, error) {
 	if !o.enabled {
-		o.lggr.Infof("CostlyMessageObserver is disabled")
+		o.lggr.Infof("Observer is disabled")
 		return nil, nil
 	}
 
@@ -128,7 +128,7 @@ func (o *CCIPCostlyMessageObserver) Observe(
 	return costlyMessages, nil
 }
 
-var _ CostlyMessageObserver = &CCIPCostlyMessageObserver{}
+var _ Observer = &observer{}
 
 // MessageFeeE18USDCalculator Calculates the fees (paid at source) of a set of messages in USD18s.
 type MessageFeeE18USDCalculator interface {
--- a/execute/costlymessages/costly_messages_test.go
+++ b/execute/costlymessages/costly_messages_test.go
@@ -1,4 +1,4 @@
-package exectypes
+package costlymessages
 
 import (
 	"context"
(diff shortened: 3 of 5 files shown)


execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,430 @@
...
+	c.runExpirationLoop(cleanupInterval)
+	return c
+}
+
+// get returns the token data for the given message ID if it exists in the cache.
+func (c *inMemTokenDataCache) get(msgID cciptypes.Bytes32) (exectypes.MessageTokenData, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	msgData, ok := c.inMemTokenData[msgID]
```
0xAustinWang, line 362: I think we can just return `c.inMemTokenData[msgID]` here

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// size returns the number of messages in the queue
+func (q *msgQueue) size() int {
+	q.mu.RLock()
+	defer q.mu.RUnlock()
+	return len(q.msgs)
+}
+
+type inMemTokenDataCache struct {
```
0xAustinWang, line 330: would it make sense to manage the lock for the cache and its expiration separately? That way we can handle cleanup independently of accessing the token data from the cache itself.

0xAustinWang (approved):

execute/factory.go
```diff
@@ -135,6 +136,15 @@ func (p PluginFactory) NewReportingPlugin(
 		return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to create token data observer: %w", err)
 	}
 
+	backgroundTokenDataObserver := tokendata.NewBackgroundObserver(
+		logger.Named(p.lggr, "BackgroundObserver"),
+		baseObserver,
```
mateusz-sekara, line 141: I'm not sure about the execution flow here, if `baseObserver` is part of `backgroundTokenObserver` it means that we will process every token in the background automatically right? Although it's probably a good pattern, I'm afraid of using integrations with Batch APIs efficiently. (e.g. all attestations for particular token resolved in one request)

Let's imagine the following case. We have two different integrations registered in `baseObserver`
* USDC requiring message-by-message processing
* MyFancyToken provides Batch API which can serve all observations within a single HTTP call

image


If backgroundTokenDataObserver wraps `baseObserver`, we enforce message-by-message processing for every underlying tokenDataObserver, right? As long as we have only USDC it's not a problem, but the next integration is going to provide Batch API to reduce that pain and I wonder how this implementation suits it. Is there any easy way to adjust it to properly support batching? I was thinking that `backgroundTokenDataObserver` should be on the particular tokenDataObserver instead of wrapping `baseObserver`, what're your thoughts on that? For instance sth like this

image

execute/factory.go
```diff
@@ -135,6 +136,15 @@ func (p PluginFactory) NewReportingPlugin(
...
+		10,             /* numWorkers */
+		10*time.Minute, /* cacheExpirationInterval */
+		15*time.Minute, /* cacheCleanupInterval */
+		5*time.Second,  /* observeTimeout */
```
mateusz-sekara, line 145: Similarly to the previous comment, probably these values are going to be specific to the integration, no?

execute/tokendata/observer_background_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/smartcontractkit/chainlink-common/pkg/utils/tests"
+
+	"github.com/smartcontractkit/chainlink-ccip/execute/exectypes"
+	"github.com/smartcontractkit/chainlink-ccip/internal/libs/testhelpers/rand"
+	"github.com/smartcontractkit/chainlink-ccip/internal/mocks"
+	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+)
+
+//nolint:gocyclo // this is a test
```
mateusz-sekara, line 19: Probably you can avoid that by putting the test into `tokendata_test` package (as long as you don't reach to unexported fields/structs)

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,431 @@
...
+	go func() {
+		ticker := time.NewTicker(cleanupInterval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-c.done:
+				c.lggr.Debug("expiration loop gracefully stopped")
+				return
+			case <-ticker.C:
```
mateusz-sekara, line 399: Why not use `cache.Cache` for that? It's widely adopted in our codebase and probably has all the required features to handle that, no?

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,431 @@
...
+	return cciptypes.Message{}, false
+}
+
+// containsMsg returns true if the message is already in the queue
+func (q *msgQueue) containsMsg(msg cciptypes.Message) bool {
+	q.mu.RLock()
+	defer q.mu.RUnlock()
+
+	for _, qMsg := range q.msgs {
+		if qMsg.msg.Header.MessageID == msg.Header.MessageID {
```
mateusz-sekara, line 315: `containsMsg` has linear complexity which might become a bottleneck during a message spike scenarios. Probably not an issue right now, but may be worth adding TODO in the future. If it's easy then maybe `qMsg` should also have a set indicating message presence `map[MessageID]struct{}`

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,431 @@
...
+
+			if _, seqExists := tokenData[msg.Header.SourceChainSelector][msg.Header.SequenceNumber]; !seqExists {
+				lggr.Errorw("underlying observer did not return token data for the sequence number")
+				o.msgQueue.enqueue(msg, o.reprocessInterval)
+				continue
+			}
+
+			if !tokenData[msg.Header.SourceChainSelector][msg.Header.SequenceNumber].SupportedAreReady() {
+				lggr.Infow("token data not ready by the underlying observer, message pushed again to the queue")
+				o.msgQueue.enqueue(msg, o.reprocessInterval)
```
mateusz-sekara, line 209: How many times are we going to do this? I'm worried that retrying indefinitely some faulty messages might impact processing of "valid" ones

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,431 @@
...
+	for i := 0; i < o.numWorkers; i++ {
+		o.wg.Add(1)
+		workerID := i
+		go o.worker(workerID)
+	}
+}
+
+// worker is a goroutine that processes messages from the queue.
+// It will stop using the o.done channel.
+func (o *backgroundObserver) worker(id int) {
```
mateusz-sekara, line 154: I think we miss QoS/ isolation for processing between different integrations. Let's imagine 99% of traffic is from USDC and USDC is down for 1 hour. However, we keep retrying USDC messages leading into starvation of "Integration B" messages because of keeping them in the same queue. 

I think it could be solved by either:
* partitioning / having a queue per integration 
* having `backgroundWorker` per tokenDataObserver instead of global one

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,431 @@
...
+			select {
+			case <-c.done:
+				c.lggr.Debug("expiration loop gracefully stopped")
+				return
+			case <-ticker.C:
+				func() {
+					c.mu.Lock()
+					defer c.mu.Unlock()
+
+					for msgID, expiresAt := range c.expiresAt {
```
mateusz-sekara, line 404: Let's imagine the message has two tokens:
* USDC
* MyFancyToken

MyFancyToken got successful attestation immediately, but USDC is lagging behind. Does it mean that we will keep purging successful MyFancyToken attestation while waiting for the successful USDC response?

execute/tokendata/observer_background_test.go
```diff
@@ -0,0 +1,146 @@
...
+func Test_backgroundObserver(t *testing.T) {
```
mateusz-sekara, line 20: I think we missed some advanced scenarios here, for instance
* multiple tokenDataObservers in baseObserver
  * some faulty, some succesful
  * all faulty
  * all successful

execute/tokendata/observer_background.go
mateusz-sekara: Shouldn't successful responses expire after reaching `permissionlessExecutionThreshold`? I think we assume that attestation APIs are idempotent after returning success. Therefore, successful attestation returned once is valid forever and should never change. Is that a valid assumption?

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+// IsTokenSupported simply forwards the call to the underlying observer.
+func (o *backgroundObserver) IsTokenSupported(
+	sourceChain cciptypes.ChainSelector, msgToken cciptypes.RampTokenAmount) bool {
+	return o.observer.IsTokenSupported(sourceChain, msgToken)
+}
+
+// Close stops the background goroutines and cleans up resources.
+func (o *backgroundObserver) Close() {
```
mateusz-sekara, line 134: Do we ever call that from the plugin? I guess we should call that during plugin shutdown

mateusz-sekara:
@winder you mind taking a look at the PR?

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,431 @@
...
+	close(o.done)
+	o.wg.Wait()
+}
+
+// startWorkers starts the worker goroutines that process messages from the queue.
+func (o *backgroundObserver) startWorkers() {
+	o.lggr.Info("waiting for existing (if any) workers to stop")
+	o.wg.Wait()
```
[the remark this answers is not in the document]
mateusz-sekara: Hmm, but that will hang the parent thread. If you accidentally try to start workers you will hang forever on `o.wg.Wait()`. Shouldn't that be some kind of atomicBitFlag to indicate whether workers have already started?

execute/factory.go
mateusz-sekara: > I was also thinking about a similar approach but unfortunately it doesn't fit well, since the TokenDataObserver interface, accepts msgs and not individual tokens, so the work cannot be sliced evenly.

That was intentional design. Therefore every tokenDataObserver (every integration) can pick which tokens/messages it should process. Therefore you don't need any "parent" orchestrator to know how to it for them

mateusz-sekara: Although it will work well when there is always single tokenDataObserver, I think it will require rework right after adding new integration. I wonder how much effort is required to have it properly designed from the day one? I like the idea that right now everything is separate and doesn't require changes to existing code, but on the other hand, I do wonder how much rework this requires
> I tried to sketch this approach in the code but I ended up touching many components, so I felt that it's in the category of premature optimization.

execute/tokendata/observer_background.go
```diff
@@ -0,0 +1,436 @@
...
+
+	o.startWorkers()
+	return o
+}
+
+// Observe fetches token data for the given messages that are already present in memory.
+// If token data are not in memory, it enqueues the messages for background processing.
+// Meaning that at least two calls to Observe are needed to get token data for a message.
+// If data are already in the queue for processing, it is a nop.
+func (o *backgroundObserver) Observe(
```
winder, line 69: Observe is probably the wrong name. This background process is fetching attestations as quickly as possible, it shouldn't worry about what they're used for. Maybe something like `GetCachedTokenData` or `GetAvailableTokenData`?

execute/factory.go
winder: I'm also concerned with embedding the base observer into the background task. Token observations could be completely decoupled from the observation logic, which could create a simpler interface with a single responsibility.

execute/tokendata/observer_background.go
winder: Could we let the plugin dictate retries by waiting until the next time it asks for the token data?

execute/tokendata/observer.go
```diff
@@ -217,13 +246,48 @@ func (n *NoopTokenDataObserver) Observe(
 			tokenData := make([]exectypes.TokenData, len(message.TokenAmounts))
 			for i := range message.TokenAmounts {
 				tokenData[i] = exectypes.NewNoopTokenData()
+				if n.isError(selector, seq, i) {
```
mateusz-sekara, line 249: nit, but could be 
```go
// it doesn't rely on the details of the struct, but rather use defined constructors 
if n.isError() {
   tokenData[i] = exectypes.NewErrorTokenData(errors.New("some error"))
} else {
   tokenData[i] = exectypes.NewNoopTokenData()
}
```

mateusz-sekara (approved):

PR #315, opened 2024-11-14
create F utility comparators

0xAustinWang:
https://smartcontract-it.atlassian.net/browse/CCIP-4050


commit/merkleroot/rmn/rmnpb/rmn_offchain.pb.go
```diff
@@ -7,10 +7,11 @@
 package rmnpb
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
+
```
asoliman92, line 12: This is what's causing the error in CI, don't change/fmt the generated file.

internal/plugincommon/consensus/threshold_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			f:           4,
+			val:         4,
+			expectedRes: false,
+		},
+	}
+
+	for _, tc := range testCase {
+		res := GteFPlusOne(tc.f, tc.val)
+		require.Equal(t, tc.expectedRes, res, "they should be equal")
```
asoliman92, line 35: maybe use a clearer message instead of "they"

0xAustinWang: deleted the string message as these validations are pretty straightforward

asoliman92 (approved):

PR #318, opened 2024-11-15
USDC Integration Tests Fixes

mateusz-sekara:
* Adding proper padding when decoding EVM events
* Lower casing addresses to avoid drift with checksumed/not-checksumed addresses

### DIFF
--- a/pkg/reader/usdc_reader.go
+++ b/pkg/reader/usdc_reader.go
@@ -45,10 +45,6 @@ var CCTPDestDomains = map[uint64]uint32{
 	sel.ETHEREUM_TESTNET_SEPOLIA_ARBITRUM_1.Selector: 3,
 	sel.ETHEREUM_TESTNET_SEPOLIA_BASE_1.Selector:     6,
 	sel.POLYGON_TESTNET_AMOY.Selector:                7,
-	// Tests
-	sel.GETH_TESTNET.Selector:  100,
-	sel.GETH_DEVNET_2.Selector: 101,
-	sel.GETH_DEVNET_3.Selector: 103,
 }
 
 type usdcMessageReader struct {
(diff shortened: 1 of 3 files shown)


pkg/reader/usdc_reader.go
```diff
@@ -109,11 +105,34 @@ func NewUSDCMessageReader(
...
 		lggr:            lggr,
 		contractReaders: contractReaders,
-		cctpDestDomain:  CCTPDestDomains,
+		cctpDestDomain:  AllAvailableDomains(),
 		boundContracts:  boundContracts,
 	}, nil
 }
 
+// FIXME It adds test selectors to the domains
+func AllAvailableDomains() map[uint64]uint32 {
```
mateusz-sekara, line 114: This is temporary support for test domains. Ideally, we want to fetch these from the chain

pkg/reader/usdc_reader.go
```diff
@@ -109,11 +105,34 @@ func NewUSDCMessageReader(
...
+	chainIDs := make([]uint64, 3+101)
```
makramkd, line 115: What's the 3 + 101, just some big number?

pkg/reader/usdc_reader.go
```diff
@@ -269,14 +288,14 @@ func NewSourceTokenDataPayload(nonce uint64, sourceDomain uint32) *SourceTokenDa
...
-		return nil, fmt.Errorf("extraData is too short, expected at least 12 bytes")
+	if len(extraData) < 64 {
+		return nil, fmt.Errorf("extraData is too short, expected at least 64 bytes")
 	}
 
-	// Extract the nonce (first 8 bytes)
-	nonce := binary.BigEndian.Uint64(extraData[:8])
-	// Extract the sourceDomain (next 4 bytes)
-	sourceDomain := binary.BigEndian.Uint32(extraData[8:12])
+	// Extract the nonce (first 8 bytes), padded to 32 bytes
```
makramkd, line 295: nit might be worth mentioning that this is due to EVM abi encoding

makramkd: Or maybe the whole func is evm specific

makramkd (approved):

pkg/reader/usdc_reader.go
mateusz-sekara: Just some big number

pkg/reader/usdc_reader.go
mateusz-sekara: I will clarify that

PR #320, opened 2024-11-19
Better commit limits.

winder:
OCR allows providing message size limits at the plugin level. The limits set previously were placeholders. This PR sets some better values based on a new test function that allows for experimenting with different values.

The test function attempts to create maximum sized observations and outcomes. Because the commit plugin can process messages faster than the exec plugin, the number of pending commit reports and message are essentially unbounded when there is high load. From testing it became clear that as few as 350 messages could completely fill the OCR observation limit.

Because of this we need some specific limiting to truncate commit reports and messages if the final observation would be oversized. That will be done in a followup PR.


execute/README.md
```diff
@@ -0,0 +1,74 @@
+# Execute Plugin
```
winder, line 1: Wrote down some docs from my notes while reviewing the data structures

execute/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+The execute plugin observes the destination chain for pending commit reports
+and manages their execution. The data required is spread across source and
+destination chains, so the process takes multiple rounds of consensus before
+any given commit can be executed.
+
+## Commit Plugin Design
```
rstout, line 8:
```suggestion
## Execute Plugin Design
```

rstout (approved):

execute/factory.go
```diff
@@ -26,6 +25,31 @@ import (
...
+	// maxOutcomeLength is set to the maximum size of an outcome
+	// check factory_test for the calculation
+	maxOutcomeLength = ocr3types.MaxMaxOutcomeLength
+
+	// maxReportLength is set to an estimate of a maximum report size
+	// check factory_test for the calculation
+	maxReportLength = ocr3types.MaxMaxReportLength
+
+	// maxReportCount is set to 1 because the commit plugin only generates one report per round.
+	maxReportCount = 1
```
asoliman92, line 50: But exec takes more time to finish a round so we might have multiple commit reports? the comment is a bit confusing for me.

execute/factory_test.go
```diff
@@ -0,0 +1,221 @@
...
+Total size of observation: 1565257
+100 CommitReports: 30335
+350 Messages: 1019335
+TokenData: 196035
+Costly Messages: 24295
+Nonces: 21035
+Contracts: 274962
+*/
+func TestObservationSize(t *testing.T) {
+	t.Skip("This test is for estimating message sizes, not for running in CI")
```
asoliman92, line 74: In the PR description you mention 350 messages fill the observation limit. 
1. Where do we get the observation limit from? Is it dictated by OCR protocl?
2. In all cases I think having a test to assert that 350 messages (or whatever the limit we set) doesn't exceed the observation limit to prevent issues in the future.

execute/factory.go
winder: This is referring to OCR reports. I'll update the comment here with something along the lines of "not to be mistaken by the ExecutePluginReport which is a single OCR report which may contain several per-chain sub-reports."

execute/factory_test.go
winder: 350 messages is a worst case with 1000 bytes of data and 250 bytes of token data. This doesn't even account for the discovery observation or commit reports. For small messages with one token transfer and no token data we can fit over 1100 messages into a single observation.

All of this is totally informational to help us choose a reasonable number. We would need metrics about what the average message looks like in order to figure out the expected throughput, for example:
* token only and how many tokens.
* data only and how much data.
* both tokens and data.
* number of messages per commit report.
* frequency of messages requiring token data.

PR #321, opened 2024-11-20
rmn - cursing checks and reverts prevention

dimkouv:
core ref: b009ab4b34d251297f9db643eaf6822b3980211c

### DIFF
--- a/execute/costlymessages/costly_messages.go
+++ b/execute/costlymessages/costly_messages.go
@@ -1,4 +1,4 @@
-package exectypes
+package costlymessages
 
 import (
 	"context"
@@ -23,8 +23,8 @@ const (
 	daMultiplierBase          = 10_000  // DA multiplier is in multiples of 0.0001, i.e. 1/daMultiplierBase
 )
 
-// CostlyMessageObserver observes messages that are too costly to execute.
-type CostlyMessageObserver interface {
+// Observer observes messages that are too costly to execute.
+type Observer interface {
 	// Observe takes a set of messages and returns a slice of message IDs that are too costly to execute.
 	// Also takes in a map from message ID to the message's timestamp, needed to calculate fee boosting.
 	Observe(
@@ -34,18 +34,18 @@ type CostlyMessageObserver interface {
 	) ([]cciptypes.Bytes32, error)
 }
 
-// NewCostlyMessageObserverWithDefaults creates a new CostlyMessageObserver with default calculators.
+// NewObserverWithDefaults creates a new Observer with default calculators.
 // The default calculators are:
 // - CCIPMessageFeeUSD18Calculator
 // - CCIPMessageExecCostUSD18Calculator
-func NewCostlyMessageObserverWithDefaults(
+func NewObserverWithDefaults(
 	lggr logger.Logger,
 	enabled bool,
 	ccipReader readerpkg.CCIPReader,
 	relativeBoostPerWaitHour float64,
 	estimateProvider gas.EstimateProvider,
-) CostlyMessageObserver {
-	return NewCostlyMessageObserver(
+) Observer {
+	return NewObserver(
 		lggr,
 		enabled,
 		NewCCIPMessageFeeUSD18Calculator(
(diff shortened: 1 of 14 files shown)


commit/report.go
```diff
@@ -121,6 +122,11 @@ func (p *Plugin) ShouldAcceptAttestedReport(
 		return false, nil
 	}
 
+	if err := cursingValidation(ctx, p.ccipReader, decodedReport.MerkleRoots); err != nil {
+		p.lggr.Errorw("report not accepted due to cursing", "err", err)
+		return false, nil
+	}
```
winder, line 128: Are you sure this works? What if the transmitter can't read from the source?

winder: Are we only looking for curses on the destination in all cases?

commit/report.go
```diff
@@ -172,6 +178,40 @@ func (p *Plugin) ShouldTransmitAcceptedReport(
...
+
+	for sourceChain, isCursed := range curseInfo.CursedSourceChains {
+		if isCursed {
+			return fmt.Errorf("source chain %d is cursed", sourceChain)
+		}
+	}
+
+	return nil
+}
```
winder, line 213: Maybe this could be part of ChainSupport, we'd have to give it the ccipReader

pkg/reader/ccip_interface.go
```diff
@@ -148,6 +148,13 @@ type CCIPReader interface {
 		destChainSelector cciptypes.ChainSelector,
 	) (rmntypes.RemoteConfig, error)
 
+	// GetRmnCurseInfo returns rmn curse/pausing information about the provided chains
+	// from the destination chain RMN remote contract. Caller should be able to access destination.
+	GetRmnCurseInfo(
+		ctx context.Context,
+		sourceChainSelectors []cciptypes.ChainSelector,
```
winder, line 155:
```suggestion
		ctx context.Context,
		destChainSelector cciptypes.ChainSelector,
		sourceChainSelectors []cciptypes.ChainSelector,
```

commit/merkleroot/observation.go
```diff
@@ -273,16 +275,40 @@ func (o observerImpl) ObserveOffRampNextSeqNums(ctx context.Context) []plugintyp
...
+	sourceChains := make([]cciptypes.ChainSelector, 0, len(allSourceChains))
+	for _, ch := range allSourceChains {
+		if !curseInfo.CursedSourceChains[ch] {
+			sourceChains = append(sourceChains, ch)
+		}
+	}
 	sort.Slice(sourceChains, func(i, j int) bool { return sourceChains[i] < sourceChains[j] })
+
+	if len(sourceChains) == 0 {
+		o.lggr.Debugw("nothing to observe from the offramp, no active source chains exist")
```
makramkd, line 308: This might be better as a warn/info log?

makramkd: Since it would be exceptional for this to happen (chains have to be cursed)

commit/report.go
makramkd: Yeah, cursing logic is as follows. If we want to curse a chain:

1. We `curse(LEGACY_CURSE_SUBJECT)` on the chain that is being cursed. This causes `Router.routeMessage` and `Router.ccipSend` to revert on that chain.
2. We `curse(bytes16(chainSelector))` on all other chains, causing `ccipSend` calls to revert from those chains to `chainSelector`.

commit/report.go
```diff
@@ -172,6 +178,40 @@ func (p *Plugin) ShouldTransmitAcceptedReport(
...
 
+// cursingValidation will make one contract call to get rmn curse info.
+// If destination is cursed or some source chain is cursed it returns error.
+func cursingValidation(
+	ctx context.Context,
+	ccipReader ccipreader.CCIPReader,
+	merkleRoots []cciptypes.MerkleRootChain,
+) error {
+	if len(merkleRoots) == 0 {
+		return nil
```
makramkd, line 189: Maybe add a comment for this case, i.e we still want to transmit tokprices/gasprices if cursed

pkg/reader/ccip.go
```diff
@@ -667,6 +669,59 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
...
+	if err := validateExtendedReaderExistence(r.contractReaders, r.destChain); err != nil {
+		return nil, fmt.Errorf("validate dest=%d extended reader existence: %w", r.destChain, err)
+	}
+
+	type retTyp struct {
+		CursedSubjects [][16]byte
+	}
+	var cursedSubjects retTyp
+
+	err := r.contractReaders[r.destChain].ExtendedGetLatestValue(
```
makramkd, line 687: I assume you tested this; the named returns have messed up contract reader queries before

pkg/reader/ccip.go
```diff
@@ -667,6 +669,59 @@ func (r *ccipChainReader) GetRMNRemoteConfig(
...
+		consts.MethodNameGetCursedSubjects,
+		primitives.Unconfirmed,
+		map[string]any{},
+		&cursedSubjects,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("get latest value of %s: %w", consts.MethodNameGetCursedSubjects, err)
+	}
+
+	r.lggr.Debugw("got cursed subjects", "cursedSubjects", cursedSubjects.CursedSubjects)
```
makramkd, line 699: This will log arrays like `[1 2 3 4 ...]` maybe can hexify prior for easier reading?

pkg/reader/ccip_interface.go
makramkd: Dest chain selector should be in the OCR config, that can be fetched pretty easily in the plugin itself

PR #322, opened 2024-11-21
#DO NOT MERGE: refactor commit and exec codec to generic codec

huangzhen1997:
Replaing the codec in plugin [here](https://github.com/smartcontractkit/chainlink-ccip/blob/b58de304ea19ad434bdbd302a219513a2abbdde3/commit/plugin.go#L44). Rather than using commit codec interface type, we could use the generic [codec](https://github.com/smartcontractkit/chainlink-common/blob/e0189e5db1ec1fa675cdacccdae5baaa40239c8a/pkg/types/codec.go#L92) instead. And later update the chainlink core repo with the actual impl change.

This will not be merged until compatibility check is done in chainlink core node


winder:
Hi @huangzhen1997, thanks for the pull request! Unfortunately, this isn't compatible with how we extend functionality in the plugins. You should implement the existing `cciptypes.CommitPluginCodec` interface using `Codec` and `RemoteCodec`. They can be injected into the plugins via the existing factorys.

PR #328, opened 2024-11-26
Bump common and fix Chain Writer naming

ilija42:
core ref: fa346afd62ef6f6f97b18ddae4eee2298ebc5001


jmank88 (approved):

.github/workflows/ccip-ocr3-build-lint-test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        go-version: ['1.22']
+        go-version: ['1.23.3']
```
jmank88, line 14: Was `1.22` actually a stale reference to `1.22.0`? Or does that mean latest patch `1.22.x` in this context? In which case we'd want:
```suggestion
        go-version: ['1.23']
```

jmank88:
> Did CI get stuck?

Oh, or are the required CI jobs hard-coded with a Go version `1.22` in the name?

ilija42:
> > Did CI get stuck?
> 
> Oh, or are the required CI jobs hard-coded with a Go version `1.22` in the name?

seems like it

winder (dismissed):
LGTM. I believe CI will pass after #339 is merged.

.golangci.yml
```diff
@@ -31,6 +31,10 @@ linters:
     - unparam
     - unused
 linters-settings:
+  gosec:
+    # TODO - remove this and fix linter complaint
+    excludes:
+      - G115
```
jmank88, line 37: How many were there? It would be a little better to mark them directly as `//nolint:gosec // G115 ....` if it's not too much trouble.

Makefile
```diff
@@ -57,13 +57,13 @@ install-protoc:
...
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.62.2
 
 ensure_go_version:
-	@go version | grep -q 'go1.22' || (echo "Please use go1.22" && exit 1)
+	@go version | grep -q 'go1.23' || (echo "Please use go1.23" && exit 1)
 
-ensure_golangcilint_1_59:
-	@golangci-lint --version | grep -q '1.59' || (echo "Please use golangci-lint 1.59" && exit 1)
+ensure_golangcilint_1_62_2:
```
jmank88, line 65: Does this pattern serve a purpose? Or could we drop it?
```suggestion
ensure_golangcilint_version:
```

.golangci.yml
ilija42: 10+

Makefile
winder: should drop it

winder (approved):

PR #330, opened 2024-11-27
Handle execution observation  size constraints [CCIP-4324]

asoliman92:
This PR addresses size constraints at the full commit report level. A follow-up PR will handle splitting individual commit reports that exceed the allowed size limit.

Changes in the GetMessages State:
After receiving all messages, an observation is prepared with all messages.
If the encoded observation exceeds the size limit, it is truncated using the following approach:
Truncation Algorithm:
Sequentially remove commit reports from each chain until the encoded size falls within the limit.
Continue until only one chain remains with a single commit report.
Key Notes:

Entire messages (including token data and executed messages) are removed during truncation.
Nonces are retained as their size is minimal, and their removal adds unnecessary complexity.


execute/plugin_functions.go
```diff
@@ -167,17 +167,11 @@ func filterOutExecutedMessages(
 		for i := reportIdx; i < len(reports); i++ {
 			reportRange := reports[i].SequenceNumberRange
 			if executed.End() < reportRange.Start() {
-				// need to go to the next set of executed messages.
+				// No messages in current executed range are in reports[i]
+				// need to go to the next set of executed range.
 				break
 			}
 
-			if executed.End() < reportRange.Start() {
```
asoliman92, line 174: Un needed check. Identical to the first one and wasn't correct.

execute/test_utils.go
```diff
@@ -89,33 +89,48 @@ func SetupSimpleTest(t *testing.T, srcSelector, dstSelector cciptypes.ChainSelec
 	}
 }
 
-func (it *IntTest) WithMessages(messages []inmem.MessagesWithMetadata, crBlockNumber uint64, crTimestamp time.Time) {
```
asoliman92, line 92: Adding numReports to create multi-report in tests

internal/libs/testhelpers/ocr3runner.go
```diff
@@ -100,7 +100,16 @@ func (r *OCR3Runner[RI]) RunRound(ctx context.Context) (result RoundResult[RI],
 
 	// check that all the outcomes are the same.
 	if countUniqueOutcomes(outcomes) > 1 {
-		return RoundResult[RI]{}, fmt.Errorf("outcomes are not equal")
+		decodedOs := make([]exectypes.Outcome, len(outcomes))
```
asoliman92, line 103: Was failing with non-equal outcomes when we were not sorting by seqNumbers, bug showed when added multiple reports for same chain. This is just logging for future references if something like this happen again.

execute/exectypes/outcome.go
```diff
@@ -90,6 +90,11 @@ func newSortedOutcome(
 		func(i, j int) bool {
 			return pendingCommitsCP[i].SourceChain < pendingCommitsCP[j].SourceChain
 		})
+	sort.Slice(
```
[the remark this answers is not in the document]
mateusz-sekara: You don't need to sort twice, do you? I think you can do a single `sort.slice` that combines both SourceChain and SequenceNumber comparisons

execute/observation.go
```diff
@@ -241,6 +241,25 @@ func (p *Plugin) getMessagesObservation(
 	observation.CostlyMessages = costlyMessages
 	observation.TokenData = tkData
 
+	encodedObs, err := observation.Encode()
+	if err != nil {
+		return exectypes.Observation{}, fmt.Errorf("unable to encode observation: %w", err)
+	}
+
+	p.lggr.Infow("encoded observation",
```
mateusz-sekara, line 249: I would also log `maxObservationLength`

execute/plugin_functions.go
```diff
@@ -208,6 +202,117 @@ func filterOutExecutedMessages(
 	return filtered, nil
 }
 
+// truncateObservation truncates the observation to fit within the given maxSize after encoding.
+// It removes data from the observation in the following order:
```
mateusz-sekara, line 206: I don't think this logic is correct. You should remove messages instead of entire commits, right? Offramp allows partial execution in which only some messages from the CommitRoot are executed, right?

Let's imagine the case in which you have a single, extremely large CommitReport (256 messages) that hits the limit. If you cut off a single CommitReport you will end up producing empty Observation forever, right?

mateusz-sekara: https://github.com/smartcontractkit/ccip/blob/80dd5d2b4867ee67c348fc1b05669c015bafa14a/core/services/ocr2/plugins/ccip/ccipexec/ocr2.go#L425

execute/plugin_functions.go
```diff
@@ -208,6 +202,117 @@ func filterOutExecutedMessages(
...
+			if len(observation.CommitReports[chain]) > 1 {
+				truncateLastCommit(chain, observation)
+			} else {
+				truncateChain(chain, observation)
+			}
+			chains = maps.Keys(observation.CommitReports)
+			break
+		}
+		// If there's only one chain with one report left, stop.
+		if len(observation.CommitReports) == 1 && len(observation.CommitReports[chains[0]]) == 1 {
```
mateusz-sekara, line 238: Hmm, so that still could overflow and be stuck forever in case of a very large single CommitReport?

execute/plugin_functions.go
asoliman92: The PR description addresses your main comment. For this specific PR we're dealing with large observation with multiple commits. This is a good first line of defense. The task I'm working on currently is if one commit is larger than the size limit. We can't just remove messages easily because of how merkel roots are working so it's a bit tricky and I'm still figuring out how to use hashes instead of messages and pseudo delete a message for example.

simply removing single messages will fail the rest of the flow for now.

execute/plugin_functions.go
asoliman92: That's the subsequent PR I'm working on currently.

execute/plugin_functions.go
```diff
@@ -208,6 +202,117 @@ func filterOutExecutedMessages(
...
+
+	return true
+}
+
+// truncateChain removes all data related to the given chain from the observation.
+// returns true if the chain was found and truncated, false otherwise.
+func truncateChain(
+	chain cciptypes.ChainSelector,
+	observation *exectypes.Observation,
+) bool {
```
makramkd, line 286: Ditto here re: param ordering

execute/plugin_functions.go
```diff
@@ -208,6 +202,117 @@ func filterOutExecutedMessages(
...
+			delete(observation.Messages[chain], seqNum)
+			// Remove the token data from the observation.
+			delete(observation.TokenData[chain], seqNum)
+			// Remove costly messages
+			for i, costlyMessage := range observation.CostlyMessages {
+				if costlyMessage == msg.Header.MessageID {
+					observation.CostlyMessages = append(observation.CostlyMessages[:i], observation.CostlyMessages[i+1:]...)
+				}
+			}
+			// Leaving Nonces untouched
```
makramkd, line 274: I think this would be confusing for debugging, why not remove it?

execute/plugin_functions.go
```diff
@@ -208,6 +202,117 @@ func filterOutExecutedMessages(
...
+		}
+	}
+
+	deleteCostlyMessages := func() {
+		for i, costlyMessage := range observation.CostlyMessages {
+			if _, ok := messageIDs[costlyMessage]; ok {
+				observation.CostlyMessages = append(observation.CostlyMessages[:i], observation.CostlyMessages[i+1:]...)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
asoliman92: Not really, thought it's more concise like this.

execute/plugin_functions.go
```diff
@@ -208,6 +202,117 @@ func filterOutExecutedMessages(
...
+) {
+	encodedObs, err := observation.Encode()
+	if err != nil {
+		return
+	}
+
+	chains := maps.Keys(observation.CommitReports)
+	sort.Slice(chains, func(i, j int) bool {
+		return chains[i] < chains[j]
+	})
```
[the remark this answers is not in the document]
asoliman92: Seems like comparable interface we get from `Keys` is not the same as `Ordered` that's needed for `Sorted`

execute/plugin_functions.go
asoliman92: Size is negligible compared to the rest of the objects and removing is complex to do. It shouldn't be really confusing IMO.

execute/plugin_functions.go
asoliman92: Changed and erroring now when it reaches zero.

mateusz-sekara (dismissed):

execute/plugin_functions.go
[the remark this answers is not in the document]
makramkd: Dangit

execute/exectypes/outcome.go
```diff
@@ -88,7 +88,10 @@ func newSortedOutcome(
 	sort.Slice(
 		pendingCommitsCP,
 		func(i, j int) bool {
-			return pendingCommitsCP[i].SourceChain < pendingCommitsCP[j].SourceChain
+			if pendingCommitsCP[i].SourceChain != pendingCommitsCP[j].SourceChain {
+				return pendingCommitsCP[i].SourceChain < pendingCommitsCP[j].SourceChain
+			}
+			return pendingCommitsCP[i].SequenceNumberRange.Start() < pendingCommitsCP[j].SequenceNumberRange.Start()
```
makramkd, line 94: qq: can this be just

```golang
return pendingCommitsCP[i].SourceChain < pendingCommitsCP[j].SourceChain && pendingCommitsCP[i].SequenceNumberRange.Start() < pendingCommitsCP[j].SequenceNumberRange.Start()
```

or nah?

makramkd (approved):

PR #331, opened 2024-11-27
Skip stale report

asoliman92:


pkg/reader/ccip.go
```diff
@@ -1425,5 +1425,24 @@ func (r *ccipChainReader) GetMedianDataAvailabilityGasConfig(
 	return daConfig, nil
 }
 
+func (r *ccipChainReader) GetLatestPriceSeqNr(ctx context.Context) (uint64, error) {
+
+	var latestSeqNr uint64
+	err := r.contractReaders[r.destChain].ExtendedGetLatestValue(
```
makramkd, line 1431: Lets check that this contract reader is available first, like the other methods.

commit/report.go
```diff
@@ -109,8 +109,18 @@ func (p *Plugin) Reports(
...
 func (p *Plugin) ShouldAcceptAttestedReport(
-	ctx context.Context, u uint64, r ocr3types.ReportWithInfo[[]byte],
+	ctx context.Context, seqNr uint64, r ocr3types.ReportWithInfo[[]byte],
 ) (bool, error) {
+	// first check seqNr is stale or not
+	latestSeqNr, err := p.ccipReader.GetLatestPriceSeqNr(ctx)
+	if err != nil {
+		return false, fmt.Errorf("get latest price seq nr: %w", err)
+	}
+	if seqNr < latestSeqNr {
```
makramkd, line 119: This check only makes sense for prices, lets just duplicate the exact check onchain

commit/report.go
```diff
@@ -109,8 +109,18 @@ func (p *Plugin) Reports(
...
+		p.lggr.Infow("skipping stale report", "seqNr", seqNr, "latestSeqNr", latestSeqNr)
+		return false, nil
+	}
```
makramkd, line 122: Logic should also be in ShouldTransmit

pkg/reader/ccip.go
```diff
@@ -1439,5 +1439,26 @@ func (r *ccipChainReader) GetMedianDataAvailabilityGasConfig(
...
+func (r *ccipChainReader) GetLatestPriceSeqNr(ctx context.Context) (uint64, error) {
+	if err := validateExtendedReaderExistence(r.contractReaders, r.destChain); err != nil {
+		return 0, fmt.Errorf("validate dest=%d extended reader existence: %w", r.destChain, err)
+	}
+	var latestSeqNr uint64
+	err := r.contractReaders[r.destChain].ExtendedGetLatestValue(
+		ctx,
+		consts.ContractNameOffRamp,
+		consts.MethodNameGetLatestPriceSequenceNumber,
+		primitives.Finalized,
```
mateusz-sekara, line 1451: Isn't that wrong? If you get only for finalized state here, you don't get information about increasing sequence numbers, therefore you will keep spamming until finalization is reached (around 15 minutes on Ethereum)

mateusz-sekara (changes requested):

commit/report.go
```diff
@@ -110,47 +110,87 @@ func (p *Plugin) Reports(
...
 func (p *Plugin) ShouldAcceptAttestedReport(
-	ctx context.Context, u uint64, r ocr3types.ReportWithInfo[[]byte],
+	ctx context.Context, seqNr uint64, r ocr3types.ReportWithInfo[[]byte],
 ) (bool, error) {
-	decodedReport, err := p.reportCodec.Decode(ctx, r.Report)
+	if isActive, err := p.checkActiveInstance(ctx); err != nil || !isActive {
+		return false, err
+	}
+
+	latestSeqNr, err := p.ccipReader.GetLatestPriceSeqNr(ctx)
```
mateusz-sekara, line 119: Probably good enough for now, but I wonder if there is a better way of getting these sequence numbers? Fetching this from chain might lead to high throughput of RPC reads, which isn't ideal.

Does OCR3 emit some logs with Sequence Number by any chance or does ChainWriter keeps some sort of information about the latest sequence number transmitted?

asoliman92: I thought about this initially and didn't find another way to my knowledge. ChainWriter doesn't have this AFAIK. I think it's okay to leave it for now and see what the metrics will tell.

mateusz-sekara (approved):

commit/report.go
makramkd: Hm it seems like we don't emit the sequence number at all in `CommitReportAccepted` - @mateusz-sekara if this is critical for performance in your view we need to make onchain folks aware.

makramkd: Ah actually this is the price sequence number, might be a bit trickier getting it into CommitReportAccepted :(

mateusz-sekara: I was hoping for OCR3 events. For instance, OCR2 emits `Transmit` event with every report created (including epoch and round), not sure how it works for OCR3, might be worth investigating once this become a bottleneck

PR #335, opened 2024-12-03
RMN - prevent DoS from a single malicious RMN node.

dimkouv:
If a compromised RMN node responds to the signed observation requests with duplicate valid lane updates then all of them are included to the report signature requests, which legit RMN nodes reject, leading to DoS.

This PR provides a fix that prevents this scenario.

### DIFF
--- a/commit/merkleroot/rmn/controller_test.go
+++ b/commit/merkleroot/rmn/controller_test.go
@@ -599,6 +599,140 @@ func TestClient_ComputeReportSignatures(t *testing.T) {
 	})
 }
 
+func Test_controller_validateSignedObservationResponse(t *testing.T) {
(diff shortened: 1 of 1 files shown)


commit/merkleroot/rmn/controller.go
```diff
@@ -554,12 +554,19 @@ func (c *controller) validateSignedObservationResponse(
 			signedObs.Observation.RmnHomeContractConfigDigest)
 	}
 
+	seenSourceChainSelectors := mapset.NewSet[uint64]()
```
mateusz-sekara, line 557: Can we have a test exploiting that scenario?

mateusz-sekara (approved):

PR #343, opened 2024-12-04
ccipreader optimization - sort and limit

0xnogo:
### Sorting
This change ensures that queries on the `evm.logs` table sort by (`block_number`, `log_index`) instead of `block_timestamp`. Sorting by (block_number, log_index) is aligned with the structure of the existing `idx_logs_chain_block_logindex` index, which is optimized for this ordering.

Sorting by block_timestamp was less efficient as it required a full table scan and explicit sorting, given no dedicated index exists for block_timestamp alone.

### Limit
* Less Data to Process: Postgre can stop scanning as soon as the required number of rows (n) is retrieved.
* Less Data to Transfer: the database sends fewer rows to the client, so it's less network I/O overhead.

### Explain Analyze

Before
```
Sort  (cost=1506.07..1514.40 rows=3333 width=826) (actual time=35.534..38.011 rows=10000 loops=1)
  Sort Key: logs.block_timestamp
  Sort Method: external merge  Disk: 8328kB
  InitPlan 1 (returns $0)
    ->  Limit  (cost=0.15..5.50 rows=1 width=16) (actual time=0.020..0.020 rows=1 loops=1)
          ->  Index Scan using idx_log_poller_blocks_chain_block on log_poller_blocks  (cost=0.15..16.20 rows=3 width=16) (actual time=0.019..0.019 rows=1 loops=1)
                Index Cond: (evm_chain_id = '4'::numeric)
  ->  Bitmap Heap Scan on logs  (cost=106.79..1305.55 rows=3333 width=826) (actual time=2.480..8.719 rows=10000 loops=1)
        Recheck Cond: ((evm_chain_id = '4'::numeric) AND (block_number <= $0))
        Filter: ((block_timestamp >= '2024-12-04 15:32:17+00'::timestamp with time zone) AND (address = '\x1c7ca710baa009147a42e2b56ff53f44ecfe541d'::bytea) AND (event_sig = '\x35c02761bcd3ef995c6a601a1981f4ed3934dcbe5041e24e286c89f5531d17e4'::bytea))
        Rows Removed by Filter: 100
        Heap Blocks: exact=1123
        ->  Bitmap Index Scan on idx_logs_chain_block_logindex  (cost=0.00..105.95 rows=3367 width=0) (actual time=2.145..2.145 rows=10100 loops=1)
              Index Cond: ((evm_chain_id = '4'::numeric) AND (block_number <= $0))
Planning Time: 0.404 ms
Execution Time: 38.860 ms
```

After
```
Limit  (cost=496.55..496.75 rows=83 width=1317) (actual time=0.627..0.664 rows=256 loops=1)
  InitPlan 1 (returns $0)
    ->  Limit  (cost=0.15..5.50 rows=1 width=16) (actual time=0.006..0.007 rows=1 loops=1)
          ->  Index Scan using idx_log_poller_blocks_chain_block on log_poller_blocks  (cost=0.15..16.20 rows=3 width=16) (actual time=0.006..0.006 rows=1 loops=1)
                Index Cond: (evm_chain_id = '1'::numeric)
  ->  Sort  (cost=491.05..491.25 rows=83 width=1317) (actual time=0.627..0.645 rows=256 loops=1)
"        Sort Key: logs.block_number, logs.log_index, logs.tx_hash"
        Sort Method: quicksort  Memory: 541kB
        ->  Index Scan using evm_logs_idx_data_word_five on logs  (cost=0.42..488.40 rows=83 width=1317) (actual time=0.036..0.354 rows=256 loops=1)
              Index Cond: ((address = '\x03a38591daec576a98cc11617a541e249f519b95'::bytea) AND (event_sig = '\x192442a2b2adb6a7948f097023cb6b57d29d3a7a5dd33e6666d33c39cc456f32'::bytea) AND (evm_chain_id = '1'::numeric) AND (SUBSTRING(data FROM 129 FOR 32) >= '\x00000000000000000000000000000000000000000000000000000000000185a0'::bytea) AND (SUBSTRING(data FROM 129 FOR 32) <= '\x00000000000000000000000000000000000000000000000000000000000186a0'::bytea))
              Filter: ((block_number <= $0) AND (SUBSTRING(data FROM 65 FOR 32) = '\x0000000000000000000000000000000000000000000000000000000000000001'::bytea) AND (SUBSTRING(data FROM 97 FOR 32) = '\x0000000000000000000000000000000000000000000000000000000000000004'::bytea))
Planning Time: 0.328 ms
Execution Time: 0.713 ms
```


pkg/reader/ccip.go
```diff
@@ -136,7 +136,10 @@ func (r *ccipChainReader) CommitReportsGTETimestamp(
 			},
 		},
 		query.LimitAndSort{
-			SortBy: []query.SortBy{query.NewSortByTimestamp(query.Asc)},
+			SortBy: []query.SortBy{query.NewSortBySequence(query.Asc)},
+			Limit: query.Limit{
+				Count: uint64(limit * 2),
```
mateusz-sekara, line 141: Hmm, why limit*2? :D

pkg/reader/ccip.go
```diff
@@ -260,7 +263,10 @@ func (r *ccipChainReader) ExecutedMessageRanges(
 			},
 		},
 		query.LimitAndSort{
-			SortBy: []query.SortBy{query.NewSortByTimestamp(query.Asc)},
+			SortBy: []query.SortBy{query.NewSortBySequence(query.Asc)},
+			Limit: query.Limit{
```
mateusz-sekara, line 267: It would be great to have a test for that, but probably in chainlink repo ;)

mateusz-sekara (approved):

pkg/reader/ccip.go
0xnogo: The limit is applied at the end of the method

```
	if len(reports) < limit {
		return reports, nil
	}
	return reports[:limit], nil
```
I thought maybe have more so that we are sure the count will be higher than the limit.

PR #348, opened 2024-12-09
Reduce commit report size in observation phase in exec plugin

asoliman92:
* Hash messages earlier in the observation and use hashes directly in ConstructMerkleTree function
* Truncate observations' reports on messages basis


execute/observation.go
```diff
@@ -206,6 +213,10 @@ func (p *Plugin) getMessagesObservation(
 	previousOutcome exectypes.Outcome,
 	observation exectypes.Observation,
 ) (exectypes.Observation, error) {
+	// Phase 2: Get messages and determine which messages are too costly to execute.
+	//          This phase also determines which messages are too costly to execute.
```
winder, line 217:
```suggestion
```

pkg/types/ccipocr3/generic_types.go
```diff
@@ -129,6 +129,8 @@ type Message struct {
 	FeeValueJuels BigInt `json:"feeValueJuels"`
 	// TokenAmounts is the array of tokens and amounts to transfer.
 	TokenAmounts []RampTokenAmount `json:"tokenAmounts"`
+
+	PseudoDeleted bool `json:"pseudoDeleted"`
```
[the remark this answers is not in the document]
winder: @ilija42 does the codec consider `json:"-"` or `mapstructure:"-"` tags?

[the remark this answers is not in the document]
asoliman92: But this structure already have different fields like the hash inside the `RampMessageHeader`, I can add an extra map outside but I think this might work fine.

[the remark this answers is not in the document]
winder: Hmm. Does our hasher and/or encoder specifically skip over these? This might be a bug.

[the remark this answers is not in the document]
ilija42: I don't think so, you can use drop modifiers if you want to ignore some fields

execute/outcome.go
```diff
@@ -121,31 +121,59 @@ func (p *Plugin) getMessagesOutcome(
...
+				// Always add the message to the report, even if it is pseudo-deleted.
 				report.Messages = append(report.Messages, msg)
+
+				report.Hashes = append(report.Hashes, observation.Hashes[report.SourceChain][j])
+
+				// Add the pseudo-deleted status to the report.
+				if observation.PseudoDeleted[report.SourceChain] != nil {
+					report.MessagePseudoDeleted = append(report.MessagePseudoDeleted,
+						observation.PseudoDeleted[report.SourceChain][j],
+					)
```
winder, line 156: What do you think about using empty messages to represent the deleted state?
`report.Messages = append(report.Messages, cciptypes.Message{})`

execute/outcome.go
```diff
@@ -121,31 +121,59 @@ func (p *Plugin) getMessagesOutcome(
...
+				} else {
+					report.MessagePseudoDeleted = append(report.MessagePseudoDeleted, false)
+				}
 				if costlyMessagesSet.Contains(msg.Header.MessageID) {
 					report.CostlyMessages = append(report.CostlyMessages, msg.Header.MessageID)
 				}
```
winder, line 162: Can you remove these now?

execute/plugin_functions.go
```diff
@@ -245,9 +319,36 @@ func truncateObservation(
...
 }
 
+func observationFitsSize(obs exectypes.Observation, maxSize int) bool {
+	encodedObs, err := obs.Encode()
+	if err != nil {
+		return false
+	}
+	return len(encodedObs) <= maxSize
+}
+func PseudoDeleteMessage(msg cciptypes.Message) cciptypes.Message {
```
winder, line 333: Why not delete the whole message?

execute/plugin_functions.go
```diff
@@ -227,15 +269,47 @@ func truncateObservation(
...
-			} else {
+			commits := observation.CommitReports[chain]
+			if len(commits) == 0 {
+				continue
+			}
+			lastCommit := &commits[len(commits)-1]
+			seqNum := lastCommit.SequenceNumberRange.Start()
+
+			observation.PseudoDeleted[chain] = make(map[cciptypes.SeqNum]bool)
+			for seqNum <= lastCommit.SequenceNumberRange.End() {
```
winder, line 280: You're deleting messages in the commit from first to last, did you mean to delete them last to first?

```suggestion
			seqNum := lastCommit.SequenceNumberRange.End()

			observation.PseudoDeleted[chain] = make(map[cciptypes.SeqNum]bool)
			for seqNum >= lastCommit.SequenceNumberRange.Start() {
```

execute/plugin_functions.go
```diff
@@ -227,15 +269,47 @@ func truncateObservation(
 	// If the encoded observation is too large, start filtering data.
```
winder, line 269: Should costly messages be removed first?

execute/plugin_functions.go
```diff
@@ -227,15 +269,47 @@ func truncateObservation(
...
 	for len(encodedObs) > maxSize {
```
winder, line 270: You're not using this for loop anymore due to the early exit when `observationFitsSize(observation, maxSize)` becomes true. Is that OK?

execute/plugin_functions.go
```diff
@@ -227,15 +269,47 @@ func truncateObservation(
...
+			if observationFitsSize(observation, maxSize) {
+				return observation, nil
+			}
 			chains = maps.Keys(observation.CommitReports)
-			break
 		}
-
 		// Truncated all chains.
 		if len(observation.CommitReports) == 0 {
 			return exectypes.Observation{}, fmt.Errorf("no more data to truncate")
```
winder, line 315: Isn't there other stuff? Like the discovery data?

execute/outcome.go
asoliman92: Yes. I added verification code [earlier](https://github.com/smartcontractkit/chainlink-ccip/pull/348/files#diff-da84b98cf1e979eb462000a946f90ba56747f7af3453826ddc0280bdd2186df5R131-R138)

execute/outcome.go
asoliman92: Yes. Much better. Thanks!

execute/plugin_functions.go
asoliman92: It won't be executed on chain if seqNr is later?

execute/plugin_functions.go
asoliman92: That's not true. We iterate over all the chains one by one and remove only one commit from each. To iterate over the chains again we need to have the outer loop.

execute/exectypes/observation.go
```diff
@@ -14,6 +15,10 @@ type CommitObservations map[cciptypes.ChainSelector][]CommitData
 // and sequence number.
 type MessageObservations map[cciptypes.ChainSelector]map[cciptypes.SeqNum]cciptypes.Message
 
+type MessageHashes map[cciptypes.ChainSelector]map[cciptypes.SeqNum]cciptypes.Bytes32
+
+type PseudoDeletedMessages map[cciptypes.ChainSelector]map[cciptypes.SeqNum]bool
+
```
winder, line 21: No longer used
```suggestion
```

execute/outcome.go
```diff
@@ -121,31 +121,45 @@ func (p *Plugin) getMessagesOutcome(
...
+
+	reports := observation.CommitReports.Flatten()
 	// add messages to their commitReports.
-	for i, report := range commitReports {
+	for i, report := range reports {
 		report.Messages = nil
 		report.CostlyMessages = nil
 		for j := report.SequenceNumberRange.Start(); j <= report.SequenceNumberRange.End(); j++ {
 			if msg, ok := observation.Messages[report.SourceChain][j]; ok {
+				// Always add the message to the report, even if it is pseudo-deleted.
```
winder, line 147:
```suggestion
				// Always add the message to the report, even if it wont be executed.
				// This slice must have an entry for each message in the commit range.
```

execute/outcome.go
```diff
@@ -121,31 +121,45 @@ func (p *Plugin) getMessagesOutcome(
...
+
 				if costlyMessagesSet.Contains(msg.Header.MessageID) {
 					report.CostlyMessages = append(report.CostlyMessages, msg.Header.MessageID)
 				}
 			}
-
-			if tokenData, ok := observation.TokenData[report.SourceChain][j]; ok {
-				report.MessageTokenData = append(report.MessageTokenData, tokenData)
-			}
+			report.MessageTokenData = append(report.MessageTokenData, observation.TokenData[report.SourceChain][j])
```
winder, line 156: This is the same deal as message and hash, need an entry for each message. Could it be moved up?

execute/observation.go
```diff
@@ -236,8 +241,14 @@ func (p *Plugin) getMessagesObservation(
...
 	}
 
+	hashes, err := exectypes.GetHashes(ctx, messageObs, p.msgHasher)
+	if err != nil {
+		return exectypes.Observation{}, fmt.Errorf("unable to get message hashes: %w", err)
+	}
+
 	observation.CommitReports = commitReportCache
 	observation.Messages = messageObs
+	observation.Hashes = hashes
```
winder, line 251: The hash is technically only required if the message data is removed, so you could have a final pass at the end of `truncateObservation` which removes redundant hashes to save even more space.

execute/plugin_functions.go
```diff
@@ -204,11 +243,14 @@ func filterOutExecutedMessages(
 
 // truncateObservation truncates the observation to fit within the given maxSize after encoding.
 // It removes data from the observation in the following order:
-// For each chain, remove commit reports one by one, if the encoded observation is still too large,
-// remove the entire chain.
-// Keep repeating this process until the encoded observation fits within the maxSize or there's only
-// one chain with one report left
-// Note: This function doesn't split one report into multiple parts.
+// For each chain, pick last report and start *pseudoDelete message by message.
```
winder, line 246: nit: now that there's no bool map, I don't think there needs to be a term for this. Just say messages are removed one by one until the report is small enough.
```suggestion
// For each chain, pick last report and start removing messages one at a time.
```

execute/plugin_functions.go
```diff
@@ -204,11 +243,14 @@ func filterOutExecutedMessages(
...
+// If removed all messages from the report, remove the report.
+// If removed last report in the chain, remove the chain.
+// After removing full report from a chain, move to the next chain and repeat. This ensures that we don't
+// exclude messages from one chain only.
+// Keep repeating this process until the encoded observation fits within the maxSize
+// *pseudo delete means putting any bytes as nil.
```
winder, line 252: You're not using pointers yet. I'd save that for a followup so that we can merge in the main algorithm.

execute/plugin_functions.go
```diff
@@ -416,6 +500,55 @@ func mergeCommitObservations(
 	return results, nil
 }
 
+func mergeMessageHashes(
```
winder, line 503: I didn't review these merge functions :(

Somehow the commit versions are significantly simpler, I haven't been able to wrap my head around why.

pkg/types/ccipocr3/generic_types.go
```diff
@@ -131,11 +131,16 @@ type Message struct {
...
-func (c Message) String() string {
-	js, _ := json.Marshal(c)
+func (m Message) String() string {
+	js, _ := json.Marshal(m)
 	return string(js)
 }
 
+// IsEmpty returns true if the message is empty. Can't use == Message{} without using reflect.DeepEqual.
+func (m Message) IsEmpty() bool {
+	return m.Header.MessageID == Bytes32{} && m.Header.SourceChainSelector == 0 && m.Header.DestChainSelector == 0 && m.Header.SequenceNumber == 0 && m.Header.Nonce == 0
```
winder, line 141: I was gonna nitpick the missing fields... but if all this is missing from the header it really might as well be empty.

execute/plugin_functions.go
asoliman92: It took me some time to wrap my head around it as well 😅

execute/plugin_functions.go
asoliman92: going to do it in next PR

execute/observation.go
asoliman92: good optimization going forward for sure 👍

execute/outcome.go
asoliman92: Thought I moved it up already. Thanks!

winder (approved):

PR #351, opened 2024-12-09
CCIP-4541 ChainReader metrics

mateusz-sekara:
Wrapping ContractReaderFacade with metrics. I wrapped only RPC calls, because DB / LogPoller interactions are already covered by the LogPoller promwrapper, we miss only RPC interactions.

Goal of these metrics is to better understand how many requests and how often (and what are the durations) are targeting a particular chain.

Intentionally, implementing it for CCIP only, because we want to start something simple first and iterate on these metrics as we go. Hopefully evolving into something that can be reused by other products as well.


pkg/contractreader/observed.go
```diff
@@ -0,0 +1,159 @@
...
+		prometheus.HistogramOpts{
+			Name:    "contract_reader_batch_request_duration",
+			Help:    "The amount of time elapsed during the ChainReader's BatchGetLatestValues execution",
+			Buckets: buckets,
+		},
+		[]string{"chainID"},
+	)
+	crBatchSizes = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "contract_reader_batch_request_szie",
```
emate, line 50: s/szie/size/

pkg/contractreader/observed.go
```diff
@@ -0,0 +1,159 @@
...
+		float64(5 * time.Second),
+	}
+
+	crDirectRequestsDurations = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "contract_reader_direct_request_duration",
+			Help:    "The amount of time elapsed during the ChainReader's GetLatestValue execution",
+			Buckets: buckets,
+		},
+		[]string{"chainID", "function", "contract"},
```
emate, line 38: Assuming 40 chains, 4 functions, and 3 contracts, this gives us 6240 new metrics. Do we really need this number of buckets?

mateusz-sekara: Yep, it's gonna be per "RPC" function called. I think this metrics is gonna be super helpful

commit/factory.go
```diff
@@ -148,11 +152,21 @@ func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types
 		oracleIDToP2PID[commontypes.OracleID(oracleID)] = node.P2pID
 	}
 
+	// map types to the facade.
+	readers := make(map[cciptypes.ChainSelector]contractreader.ContractReaderFacade, len(p.contractReaders))
+	for chain, cr := range p.contractReaders {
+		chainID, err1 := sel.ChainIdFromSelector(uint64(chain))
```
winder, line 158: This appears to be deprecated:
> // Deprecated, this only supports EVM chains, use the chain agnostic `GetChainIDFromSelector` instead

pkg/contractreader/observed.go
```diff
@@ -0,0 +1,159 @@
...
+	chainID string
+
+	// Prometheus components for tracking metrics
+	directRequestsDurations *prometheus.HistogramVec
+	batchRequestsDurations  *prometheus.HistogramVec
+	batchSizes              *prometheus.CounterVec
+	errors                  *prometheus.CounterVec
+}
+
+func NewObserverReader(
```
winder, line 76: We already have `ExtendedContractReader`, could you put this functionality there?

mateusz-sekara: `ExtendedContractReader` exposes different methods, it's an adapter on top of ContractReaderFacade. Also, it's not used everywhere so it would require a lot of refactoring now to begin with switching everything to `Extended`.

I want to start with building a simple decorator first that wraps only `ContractReader` interface. At the end of the day, every execution flows through these basic methods. Additionally, if we decide to make it generic it should be as simple as moving this `NewObserverReader` to the chainlink-common without any changes to the code.

pkg/contractreader/observed.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		[]string{"chainID"},
+	)
+	CrBatchSizes = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "contract_reader_batch_request_size",
+			Help: "The size of the batch request",
+		},
+		[]string{"chainID"},
+	)
```
winder, line 54: Shouldn't size be a histogram as well?

pkg/contractreader/observed.go
```diff
@@ -0,0 +1,159 @@
...
+func (o *Observed) maybeTrackErrors(err error, function string, contract string) {
+	if err == nil {
+		return
+	}
+	o.errors.
+		WithLabelValues(o.chainID, function, contract).
+		Inc()
+}
+
+func unpackReadIdentifier(readIdentifier string) (contract string, function string) {
```
0xnogo, line 153: I guess this is safe as readIdentifier are never going to change?

pkg/contractreader/observed.go
mateusz-sekara: Histogram works well for bounded values (when the lower and upper limit is known). I think for case like this counter is more accurate and more lightweight, you can see:
* how often we batch and what is the average batch size
* what is the rate of batching

pkg/contractreader/observed.go
winder: Thanks, good rational :+1:

pkg/contractreader/observed.go
mateusz-sekara: Probably it deserved a comment. Read identifier has the following format `address-contract-method`. It's unacceptable for the prom, because addresses change on every deployment and cardinality will explode. Therefore I needed to cut of the address part from the identifier

commit/factory.go
mateusz-sekara: You are right, let me bump chain selectors there

0xnogo (approved):

0xnogo (approved):

PR #352, opened 2024-12-09
More consistent logging.

winder:
Added a couple utilities to add `With` fields to the plugins. Moved the previous usage to the top of our factory functions. Add a new "context" field to use in place of `Named`. This all helps ensure we have the all important logging context available in a consistent way.

Example log:
```json
{
  "level": "info",
  "ts": "2024-12-09T20:59:54.826Z",
  "caller": "discovery/processor.go:254",
  "msg": "Determined consensus nonce manager",
  "version": "2.18.0@732cc15",
  // Other than removing a few instances of Named adding onto this
  // the logger field is unchanged. We might consider simplifying it to
  // just "CCIP" in the future and leveraging the new fields for filtering.
  "logger": "CCIPExecPlugin.evm.2337.0xe0b39353f69b54e945364ffcddd7901697ca0166",

  // These ones added by the log helpers
  "plugin": "Execute",
  "oracleID": 2,
  "donID": 2,
  "processor": "Discovery",
  "configDigest": "000aa1248edb0c67746b5f72ab7c9f8bcca341ea301fabb3c9e22f850db1a483",

  // remaining fields for this log message omitted
  ...
}
```


pkg/logger/logger.go
```diff
@@ -0,0 +1,66 @@
...
+var _ cc.Logger = Logger{}
+
+func appendArgs(lggr cc.Logger, newArgs ...interface{}) Logger {
+	var args []interface{}
+
+	switch t := lggr.(type) {
+	case Logger:
+		args = t.args
+		lggr = t.Logger
+	}
```
winder, line 19: There's some funny business in chainlink-common. It uses reflection to find `With` and `Named` in the `chainlink` implementation. In order to re-wrap the logger, we have to avoid nesting.

I think thats what happens anyway.

winder:
Moving back to draft, I noticed log lines with duplicate entries for `donID` and `oracleID`.

mateusz-sekara (approved):

PR #354, opened 2024-12-10
merkle root processor - unit tests and various fixes

dimkouv:
## Pull Request Summary

This PR introduces the following improvements and fixes:

### Features and Enhancements
- Added various unit tests for critical Merkle root processor logic, including observation and outcome validation.
- Performance optimizations:
  - Parallelization of several algorithms to enhance processing speed.
- Improved logging for better traceability and debugging.
- General code cleanup for improved maintainability.

### Bug Fixes
- **Contract Reader Response Validation**: Ensures all fetched messages match the sequence number range of the query to prevent invalid Merkle roots from being posted on-chain in edge cases where the database is missing logs.
- If onRamp sequence number is not observed for some source chain we were not observing the offRamp sequence numbers for that chain.
- **Edge Case Handling**:
  - Addressed a state issue where `onRamp seqNum-1` could be less than `offRamp next seqNum`, a condition that should be impossible under normal scenarios since it indicates a message not part of the onRamp was executed.
  - Fixed a state issue where a previous outcome sequence number is greater than the current outcome sequence number, this was mistakenly treated as valid logic, indicating a report had been transmitted on-chain (since seq nums are different).

### Miscellaneous
- Resolved a TODO: to return an error when RMN signatures are not provided while RMN is enabled.


commit/merkleroot/observation.go
```diff
@@ -440,6 +447,33 @@ func (o observerImpl) ObserveMerkleRoots(
 					return
 				}
 
+				if uint64(len(msgs)) != uint64(chainRange.SeqNumRange.End()-chainRange.SeqNumRange.Start()+1) {
```
winder, line 450: How about a `chainRange.Size()` function. It's in this repo now, I think there are a couple places in `SeqNumRange` that would use it also.

commit/merkleroot/outcome.go
```diff
@@ -43,57 +49,70 @@ func (p *Processor) getOutcome(
...
 		if !exists {
 			continue
 		}
 
-		if offRampNextSeqNum <= onRampMaxSeqNum {
+		if onRampMaxSeqNum < offRampNextSeqNum-1 {
+			lggr.Errorw("sequence numbers between offRamp and onRamp reached an impossible state, "+
+				"offRamp latest executed sequence number is greater than onRamp latest executed sequence number",
+				"chain", chainSel, "onRampMaxSeqNum", onRampMaxSeqNum, "offRampNextSeqNum", offRampNextSeqNum)
+		}
```
winder, line 112: I guess we'd see this any time a node's RPC is out of date, maybe even each time the node is restarted. Do you think it makes sense to include a hint in the comment?

commit/merkleroot/outcome.go
```diff
@@ -138,10 +152,10 @@ func reportRangesOutcome(
 	return outcome
 }
 
-// Given a set of observed merkle roots, gas prices and token prices, and roots from RMN, construct a report
-// to transmit on-chain
+// Given a set of agreed observed merkle roots and RMN signatures, construct an outcome.
```
winder, line 155: godoc formatting
```suggestion
// buildReport given a set of agreed observed merkle roots and RMN signatures, construct an outcome.
```

commit/merkleroot/outcome.go
```diff
@@ -153,10 +167,15 @@ func buildReport(
 		outcomeType = ReportEmpty
 	}
 
+	if len(roots) > 0 && rmnEnabled && q.RMNSignatures == nil {
+		lggr.Errorw("RMN signatures are nil while RMN is enabled. Returning an empty outcome")
+		return Outcome{}
+	}
```
winder, line 173: Should this also return an error? Maybe include it in the verify function so that we fail prior to building a report.

commit/merkleroot/outcome_test.go
```diff
@@ -3,18 +3,302 @@ package merkleroot
...
+					return Observation{
+						OnRampMaxSeqNums: []plugintypes.SeqNumChain{
+							{ChainSel: chainA, SeqNum: 10},
+							{ChainSel: chainB, SeqNum: 20},
+							{ChainSel: chainC, SeqNum: 30},
+						},
+						OffRampNextSeqNums: []plugintypes.SeqNumChain{
+							{ChainSel: chainA, SeqNum: 10}, // we have to execute 10
+							{ChainSel: chainB, SeqNum: 21}, // this is still at 20, nothing to execute
+							{ChainSel: chainC, SeqNum: 35}, // this is an unexpected state but should not lead to issues
```
winder, line 63: The inline comments you write are extremely helpful :+1:

winder (dismissed):
Nice improvements, only minor feedback

winder (approved):

commit/merkleroot/observation.go
```diff
@@ -440,6 +447,33 @@ func (o observerImpl) ObserveMerkleRoots(
...
+				// there is nothing to observe for this chain since messages are missing.
+				msgIdx := 0
+				for seqNum := chainRange.SeqNumRange.Start(); seqNum <= chainRange.SeqNumRange.End(); seqNum++ {
+					msgSeqNum := msgs[msgIdx].Header.SequenceNumber
+					if msgSeqNum != seqNum {
+						o.lggr.Warnw("message sequence number does not match seqNum range chain skipped",
+							"chain", chainRange.ChainSel,
+							"seqNum", seqNum,
+							"msgSeqNum", msgSeqNum,
+							"range", chainRange.SeqNumRange,
```
makramkd, line 470: Would it be useful to log the `msgs[*].Header.SequenceNumber` array as well? Under which scenario would this happen also? Like a query/sorting bug in the chain reader?

commit/merkleroot/observation.go
```diff
@@ -474,28 +508,40 @@ func (o observerImpl) ObserveMerkleRoots(
...
-// computeMerkleRoot computes the merkle root of a list of messages
+// computeMerkleRoot computes the merkle root of a list of messages.
+// Messages should be sorted by sequence number and not have any gaps.
 func (o observerImpl) computeMerkleRoot(ctx context.Context, msgs []cciptypes.Message) (cciptypes.Bytes32, error) {
-	var hashes [][32]byte
-	sort.Slice(msgs, func(i, j int) bool { return msgs[i].Header.SequenceNumber < msgs[j].Header.SequenceNumber })
+	hashes := make([][32]byte, len(msgs))
+	hashesStr := make([]string, len(hashes)) // also keep hashes as strings for logging purposes
 
+	eg := &errgroup.Group{}
```
makramkd, line 517: Does this really need parallelization?

makramkd (approved):

commit/merkleroot/observation.go
makramkd: Eh I think the benchmarks would be misleading anyway, the benchmark has the whole PC to run, in prod there's way more CPU overhead / thread spawning / goroutine ctx switching. I'm fine with it though, doesn't seem to hurt that much.

PR #357, opened 2024-12-11
ChainFee Processor - observation phase tests and improvements

dimkouv:
- Unit test the observation phase
- Parallelization of external calls to speedup observation phase
- Minor code cleanup (move functions out of types package / un-export items / ..)


asoliman92 (approved):
LGTM

PR #362, opened 2024-12-12
CCIP-4448 Basic Execution plugin prom metrics

mateusz-sekara:


execute/metrics/reporter.go
```diff
@@ -0,0 +1,179 @@
...
+		Add(float64(messagesCount))
+}
+
+func (p *PromReporter) trackTokenData(tokens exectypes.TokenDataObservations) {
+	tokensCount := 0
+	for _, chainTokens := range tokens {
+		tokensCount += len(chainTokens)
+	}
+
+	p.observationDetailsCounter.
```
asoliman92, line 152: Would be nice to see how many are Ready vs Not Ready for token data

execute/observation.go
```diff
@@ -78,6 +78,7 @@ func (p *Plugin) Observation(
 		return nil, err
 	}
 
+	p.observer.TrackObservation(observation)
```
asoliman92, line 81: Adding the state to the observation metrics will be helpful. For example we expect all `getCommitReportsObservation` to have no message observations.

execute/outcome.go
```diff
@@ -97,6 +97,7 @@ func (p *Plugin) Outcome(
 		return nil, nil
 	}
 
+	p.observer.TrackOutcome(outcome)
```
asoliman92, line 100: same about the state

execute/observation.go
mateusz-sekara: State is going to increase the cardinality of the metric, I'm not sure if it's required here. We can assume that proper structs are populated according to the state and there is no bug in the code regarding that, right? If length of reports/observations is zero, then counter won't be increased maintaining the same state. I guess it's fine, isn't it?

mateusz-sekara: On the other hand, it seems that we should report sizes only depending on the state right? Otherwise last state would report observations/reports from the previous states, right? If so then, maybe state doesn't have to be part of the metric, but we should pick what to track based on the state, no?

asoliman92: > We can assume that proper structs are populated according to the state and there is no bug in the code regarding that

Part of having the metrics is debugging and making sure it's working as expected. If we have any bug it will make it much harder to find.

> if length of reports/observations is zero, then counter won't be increased maintaining the same state. I guess it's fine, isn't it?

Theoretically if we truncating removed all reports we'll have zero reports as well but in GetMessages state.

> State is going to increase the cardinality of the metric

It's going to be 3X current cardinality. I don't think that's a lot given the visibility it will give us TBH

makramkd (approved):
LGTM, a good start

PR #365, opened 2024-12-12
Include execute report info.

winder:
Add commit report information to the OCR `ReportInfo` field. This data will be needed by the contract writer in the future.

The raw commit report data was not readily available in the `Reports` function, so the `GetMessages` state's Outcome phase was changed slightly to repurpose the unused `PendingCommitReports` object. It is now used to store CommitReports which correspond to the the report.


execute/plugin.go
```diff
@@ -292,17 +319,17 @@ func (p *Plugin) Reports(
 	p.lggr.Debugw("transmission schedule override",
 		"transmissionSchedule", transmissionSchedule, "oracleIDToP2PID", p.oracleIDToP2pID)
 
-	report := []ocr3types.ReportPlus[[]byte]{
+	r := []ocr3types.ReportPlus[[]byte]{
 		{
 			ReportWithInfo: ocr3types.ReportWithInfo[[]byte]{
-				Report: encoded,
-				Info:   nil,
```
asoliman92, line 299: How did we miss this earlier? :D

pkg/types/ccipocr3/plugin_execute_types.go
```diff
@@ -11,3 +17,35 @@ type ExecutePluginReportSingleChain struct {
...
+
+// ExecuteReportInfo contains metadata needed by transmitter and contract
+// writer.
+type ExecuteReportInfo []MerkleRootChain
+
+// Encode v1 execute report info. Very basic versioning in the first byte to
+// allow for future encoding optimizations.
+func (eri ExecuteReportInfo) Encode() ([]byte, error) {
+	data, err := json.Marshal(eri)
+	data = append([]byte{1}, data...)
```
asoliman92, line 34:
```suggestion
	data = append([]byte{01}, data...)
```
2 digits just in case

pkg/types/ccipocr3/plugin_execute_types.go
```diff
@@ -11,3 +17,35 @@ type ExecutePluginReportSingleChain struct {
 	Proofs              []Bytes32     `json:"proofs"`
 	ProofFlagBits       BigInt        `json:"proofFlagBits"`
 }
+
+type CommitReportWithMessages struct {
```
asoliman92, line 21: Where's this used?

execute/plugin.go
winder: Until now it wasn't needed. A followup PR will decode the info and pass some data into the contract writer.

pkg/types/ccipocr3/plugin_execute_types.go
winder: Oops, didn't end up using this. It might still be needed, but I'll remove it for this PR.

asoliman92 (approved):

PR #370, opened 2024-12-13
CCIP-4448 Prometheus Commit plugin metrics impl

mateusz-sekara:


execute/metrics/prom.go
```diff
@@ -0,0 +1,189 @@
+package metrics
```
mateusz-sekara, line 1: This is not changed, it was moved to a separate file to reflect the commit's structure

commit/metrics/prom.go
```diff
@@ -0,0 +1,224 @@
...
+		merkleProcessorObservationCounter:   promMerkleProcessorObservationDetails,
+		merkleProcessorOutcomeCounter:       promMerkleOutcomeDetails,
+		tokenProcessorObservationCounter:    promTokenPriceObservationDetails,
+		tokenProcessorOutcomeCounter:        promTokenPriceOutcomeDetails,
+		chainFeeProcessorObservationCounter: promChainFeeObservationDetails,
+		chainFeeProcessorOutcomeCounter:     promChainFeeOutcomeDetails,
+	}, nil
+}
+
+func (p *PromReporter) TrackObservation(obs committypes.Observation) {
```
mateusz-sekara, line 110: To be implemented later, I want to check metrics first on staging before making a decision whether to keep it or not

pkg/types/ccipocr3/generic_types.go
```diff
@@ -95,6 +95,10 @@ func (s SeqNumRange) String() string {
 	return fmt.Sprintf("[%d -> %d]", s[0], s[1])
 }
 
+func (s SeqNumRange) Length() int {
```
asoliman92, line 98: Thanks! 😄

PR #376, opened 2024-12-17
Use encoded sizes hash [CCIP-4618]

asoliman92:
Truncate observation algorithm to fit max size was encoding whole observation after removing each single message. This operation is CPU heavy and made it very slow even to finish tests in time.

Fix:
Use a hash to hold encoded messages sizes and while removing messages we do simple arithmetic operations to deduce the current size. Still there's encoding happening but on a higher level and less frequently to reduce any snowball effects from small differences because the arithmetic operations are not 100% accurate.


execute/exectypes/observation.go
```diff
@@ -74,6 +89,10 @@ type Observation struct {
 	// execute report.
 	Messages MessageObservations `json:"messages"`
 	Hashes   MessageHashes       `json:"messageHashes"`
+	// EncodedMsgAndTokenDataSizes are determined during the Get Messages phase
+	// It contains the sum of encoded sizes of messages and their token data counterpart that were observed in same
+	// phase.
+	MessageAndTokenDataEncodedSizes EncodedMsgAndTokenDataSizes `json:"messageEncodedSizes"`
```
winder, line 95: This doesn't seem like it needs to be part of the observation.

execute/plugin_functions.go
```diff
@@ -762,7 +767,7 @@ func getConsensusObservation(
 }
 
 // getMessageTimestampMap returns a map of message IDs to their timestamps.
-// cciptypes.Message does not contain a timestamp, so we need to derive the timestamp from the commit data.
+// cciptypes.MessageAndTokenData does not contain a timestamp, so we need to derive the timestamp from the commit data.
```
asoliman92, line 770:
```suggestion
// cciptypes. Message does not contain a timestamp, so we need to derive the timestamp from the commit data.
```

execute/plugin_functions.go
```diff
@@ -762,7 +767,7 @@ func getConsensusObservation(
 }
 
 // getMessageTimestampMap returns a map of message IDs to their timestamps.
-// cciptypes.Message does not contain a timestamp, so we need to derive the timestamp from the commit data.
+// cciptypes. Message does not contain a timestamp, so we need to derive the timestamp from the commit data.
```
asoliman92, line 770:
```suggestion
// cciptypes.Message does not contain a timestamp, so we need to derive the timestamp from the commit data.
```

execute/exectypes/observation.go
asoliman92: Good point 👍

execute/plugin_functions_test.go
```diff
@@ -1556,7 +1558,7 @@ func Test_truncateObservation(t *testing.T) {
 					withData(make([]byte, 100)),
 				),
 			},
-			maxSize: 1789,
+			maxSize: 1819,
```
asoliman92, line 1561: Will revert it back to try after moving the sizes from the observation as suggested above 👍

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		// Truncated all chains. Return obs as is. (it has other data like contract discovery)
+		if len(obs.CommitReports) == 0 {
+			return obs, nil
+		}
+		// Encoding again after doing a full iteration on all chains and removing messages/commits.
+		// That is because using encoded sizes is not 100% accurate and there are some missing bytes in the calculation.
+		encodedObs, err = obs.Encode()
+		if err != nil {
+			return Observation{}, nil
```
0xnogo, line 126: Should we return the `err` here?

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+			// Remove the encoded size of the message and token data.
+			sizeDelta -= op.emptyEncodedSizes.MessageAndTokenData
+			sizeDelta -= 2 * op.emptyEncodedSizes.SeqNumMap
+			sizeDelta -= 4 // for brackets and commas
+			// Remove costly messages
+			for i, costlyMessage := range observation.CostlyMessages {
+				if costlyMessage == msg.Header.MessageID {
+					observation.CostlyMessages = append(observation.CostlyMessages[:i], observation.CostlyMessages[i+1:]...)
+				}
+			}
```
0xnogo, line 168: Have you checked if we have a sort of race condition here? We are modifying the array while looping through it. Might be working well but feels like we doing something wrong. Same for `truncateChain`

execute/exectypes/type_optimizer.go
asoliman92: True 👌🏼. Missed it here.

asoliman92:
```suggestion
			return Observation{}, err
```

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
+package exectypes
```
winder, line 1: could you move this out of the types package?

winder: Maybe a new internal package, since nothing here really needs to be exported beyond the execute.

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+		return obs, nil
+	}
+
+	chains := maps.Keys(obs.CommitReports)
+	sort.Slice(chains, func(i, j int) bool {
+		return chains[i] < chains[j]
+	})
+
+	messageAndTokenDataEncodedSizes := GetEncodedMsgAndTokenDataSizes(obs.Messages, obs.TokenData)
+	// If the encoded obs is too large, start filtering data.
```
winder, line 76:
```suggestion
	// While the encoded obs is too large, continue filtering data.
```

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+	for encodedObsSize > op.maxEncodedSize {
+		for _, chain := range chains {
```
winder, line 78:
```suggestion
		// go through each chain and truncate observations for the final commit report.
		for _, chain := range chains {
```

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+			commits := obs.CommitReports[chain]
+			if len(commits) == 0 {
+				continue
+			}
+			lastCommit := &commits[len(commits)-1]
+			seqNum := lastCommit.SequenceNumberRange.Start()
+
+			for seqNum <= lastCommit.SequenceNumberRange.End() {
```
winder, line 86:
```suggestion

			// Remove messages one by one starting from the last message of the last commit report.
			lastCommit := &commits[len(commits)-1]
			seqNum := lastCommit.SequenceNumberRange.Start()
			for seqNum <= lastCommit.SequenceNumberRange.End() {
```

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+				// Add empty message and token encoded size
+				encodedObsSize += op.emptyEncodedSizes.MessageAndTokenData
+				seqNum++
+				if encodedObsSize <= op.maxEncodedSize {
+					return obs, nil
+				}
+			}
+
+			var sizeDelta int
+			// Reaching here means that all messages in the report are truncated, truncate the last commit
```
winder, line 103:
```suggestion
			// Reaching here means that all messages in the report were removed, remove the entire commit
```

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+				}
+				obs.Messages[chain][seqNum] = cciptypes.Message{}
+				obs.TokenData[chain][seqNum] = NewMessageTokenData()
+				// Subtract the removed message and token size
+				encodedObsSize -= messageAndTokenDataEncodedSizes[chain][seqNum]
+				// Add empty message and token encoded size
+				encodedObsSize += op.emptyEncodedSizes.MessageAndTokenData
+				seqNum++
+				if encodedObsSize <= op.maxEncodedSize {
+					return obs, nil
```
winder, line 98: The comment below about the estimate not being 100% accurate makes me wonder if you need to double check before returning.

I think if you were to break out of the `range chains` (with a `goto` I guess?), then you would double-check the result, get a new `encodedObsSize`, and remove one or more additional messages.

execute/exectypes/type_optimizer.go
```diff
@@ -0,0 +1,209 @@
...
+				}
+			}
+
+			var sizeDelta int
+			// Reaching here means that all messages in the report are truncated, truncate the last commit
+			obs, sizeDelta = op.truncateLastCommit(obs, chain)
+
+			encodedObsSize += sizeDelta
```
winder, line 106: nit: rather than returning `delta`, maybe return `bytesTruncated`. That way you can subtract rather than adding a negative.

execute/plugin.go
```diff
@@ -114,7 +115,8 @@ func NewPlugin(
 			reportingCfg.OracleID,
 			destChain,
 		),
-		observer: metricsReporter,
+		observer:             metricsReporter,
+		observationOptimizer: exectypes.NewObservationOptimizer(maxObservationLength),
```
winder, line 119: nit: I'd personally avoid making this service. It seems like there's a single function, so you could make it pure and pass the observation length into the `TruncateObservation` function.

winder (commented):
Looks pretty good, mostly nits. Just one small problem in the truncate function that should be fixed.

execute/exectypes/type_optimizer.go
asoliman92: I think this would have been caught in CI tests which runs a lot of times and detect races. I assume that means it's working fine. WDYT @winder ?

execute/plugin.go
asoliman92: I made it like that because we do something similar with reports so maybe in the future extracting an interface with `Truncate` method and any other optimizations we can think about.

asoliman92: No strong opinion though

winder: Thats a good point that we might want to have multiple implementations. Technically, you already have 2.

Closures are a nice way to sneak configuration into simple handler functions. It's a pattern widely used in a go http middleware libraries:

```
type TruncateExecObservationHandler func(observation) observation

func MakeEncodeEstimateTruncateHandler(maxLength int) TruncateExecObservationHandler {
    return func(observation) observation{
        HashEstimateTruncator(maxLength, observation)
    }
}

func EncodeTruncateHandler(observation) observation {
    // the old function which just has the right signature
}
```

execute/exectypes/type_optimizer.go
winder: Not sure about a race, but @0xnogo is right that something is wrong. Two problems I see:
1. It only works if you're only removing one item. After removing the first, `i` would be offset to an incorrect value.
2. You cannot remove the final value, there would be an out of bounds panic because `i+1` is the index after the last element.

I usually copy them into a new slice to avoid dealing with array details:
```
			var newCostlyMessages []Message
			for _, costlyMessage := range observation.CostlyMessages {
				if costlyMessage != msg.Header.MessageID {
					newCostlyMessages = append(newCostlyMessages, costlyMessage)
				}
			}
			observation.CostlyMessages = newCostlyMessages
```

You could also copy them in place:
```
			i := 0
			for _, costlyMessage := range observation.CostlyMessages {
				if costlyMessage != msg.Header.MessageID {
					observation.CostlyMessages[i] = costlyMessage
					i++
				}
			}
			observation.CostlyMessages = observation.CostlyMessages[:i]
```

execute/plugin.go
asoliman92: Seems like a nice pattern for sure. Let's keep it for next iterations then.

asoliman92: One more thing about using a struct is to add other structs and initialize them like the EmptySizesStruct instead of sending it to all the functions or initializing it multiple times.

execute/exectypes/type_optimizer.go
asoliman92: I added a utility function `RemoveIthElement` and using it now 👍

winder (approved):

PR #386, opened 2024-12-18
Add validation preventing caller PDA to be garbage-collected

toblich:


chains/solana/contracts/programs/ccip-router/src/fee_quoter.rs
```diff
@@ -90,6 +90,12 @@ pub fn wrap_native_sol<'info>(
     amount: u64,
     signer_bump: u8,
 ) -> Result<()> {
+    require!(
+        // guarantee that if caller is a PDA it won't get garbage-collected
```
agusaldasoro, line 94: would it be incorrect to check this always? Even if not a PDA?

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -2785,7 +2785,7 @@ func TestCCIPRouter(t *testing.T) {
 			}
 		})
 
-		t.Run("hen sending a Valid CCIP Message but the user does not have enough funds of the fee token, it fails", func(t *testing.T) {
+		t.Run("When sending a Valid CCIP Message but the user does not have enough funds of the fee token, it fails", func(t *testing.T) {
```
agusaldasoro, line 2788: Should we add a test for this specific case?

toblich: I thought about it initially, but that requires me to use another program as the caller (so that it's a PDA) and guarantee that the balance is just right. It seems like an overkill to create a new program (or extend the `ccip-receiver`) just to test this fairly basic check.

chains/solana/contracts/programs/ccip-router/src/fee_quoter.rs
toblich: TBH, it likely doesn't matter. My personal take is that yes, it would be incorrect, though the impact of it is likely very small.

The rent exemption only matters when accounts hold data. When the owner of that account withdraws all funds, leaving it with a 0 balance, the account is closed and the data is removed. This check aims to prevent that. 

However, if the caller does not hold any data (e.g. it is just a wallet), then there is no reason to prevent them spending their last lamports in invoking ccip_send. Technically speaking, the check about the system program owning it is not exactly the same as checking whether the account holds data, as it is possible to have dataless PDAs, but it is close enough for an edge case and should still protect agains all actual "dataful" accounts being closed (even if we're failing in a corner case for a dataless PDA when we shouldn't).

agusaldasoro: Okay. Agree. There's still a chance that the user's account is closed after calling `ccip_send` and some one else then gets ownership of the same account. Anyways, that's the same behavior that transferring tokens has, so not a concern.

chains/solana/contracts/tests/ccip/ccip_router_test.go
agusaldasoro: Yes, let's avoid having more ccip receiver examples

agusaldasoro (approved):

aalu1418 (approved):

PR #387, opened 2024-12-18
Finality violation detection.

winder:
* Add finality violation detection to the extended contract reader. Every read (via `QueryKey`, `GetLatestValue`,  `BatchGetLatestValues`) calls the new `HealthReport` function and errors if there is a finality violation.
* Update `Extended` interface to be a superset of `ContractReaderFacade`.
* Update factories to wrap all contract readers with the `Extended` reader so that all reads have finality detection.

### DIFF
--- a/pkg/contractreader/extended.go
+++ b/pkg/contractreader/extended.go
@@ -131,17 +131,20 @@ func (e *extendedContractReader) QueryKey(
 	limitAndSort query.LimitAndSort,
 	sequenceDataType any,
 ) ([]types.Sequence, error) {
-	if e.hasFinalityViolation() {
-		return nil, ErrFinalityViolated
-	}
-
-	return e.reader.QueryKey(
+	result, err := e.reader.QueryKey(
 		ctx,
 		contract,
 		filter,
 		limitAndSort,
 		sequenceDataType,
 	)
+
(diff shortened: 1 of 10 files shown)


pkg/contractreader/extended.go
```diff
@@ -89,6 +124,26 @@ func (e *extendedContractReader) getOneBinding(contractName string) (ExtendedBou
...
 
+func (e *extendedContractReader) QueryKey(
+	ctx context.Context,
+	contract types.BoundContract,
+	filter query.KeyFilter,
+	limitAndSort query.LimitAndSort,
+	sequenceDataType any,
+) ([]types.Sequence, error) {
+	if e.hasFinalityViolation() {
+		return nil, ErrFinalityViolated
```
0xnogo, line 135: I am wondering why we want to revert finality violations, is it a product team requirement? 

Also how this is going to be used upstream?

winder: The plugins would fail to produce observations for any chain with a finality violation because the reader starts returning errors. My assumption is that many of these would bubble up through the plugin and generate logs. It might be worth adding some specific checks to the should transmit function as well.

0xnogo (approved):

PR #392, opened 2024-12-19
Don't observe destChain fee

asoliman92:
* Don't observe destChain fee
* Fix test setup


0xnogo (approved):

internal/plugincommon/chain_support.go
```diff
@@ -87,7 +87,7 @@ func (c ccipChainSupport) SupportedChains(oracleID commontypes.OracleID) (mapset
 		return mapset.NewSet[cciptypes.ChainSelector](), fmt.Errorf("error getting supported chains: %w", err)
 	}
 
-	return supportedChains, nil
+	return supportedChains.Clone(), nil
```
asoliman92, line 90: Return a clone from supported chains. Otherwise any operations on the returned set will affect the original set

commit/plugin_e2e_test.go
```diff
@@ -438,11 +438,11 @@ func TestPlugin_E2E_AllNodesAgree_ChainFee(t *testing.T) {
 	nodes := make([]ocr3types.ReportingPlugin[[]byte], len(oracleIDs))
```
asoliman92, line 438: changing the tests to simulate real results as we shouldn't expect fees for the destChain itself.

PR #391, opened 2024-12-19
Solana MCM clear root [NONEVM-1057]

jadepark-dev:
- [x] To be merged after utility refactor https://github.com/smartcontractkit/chainlink-ccip/pull/388

This PR adds the ability to clear root data in MCM through the set_config instruction

## Primary Changes
- Added clear root functionality in set_config instruction
- Added root metadata and expiring root accounts to set_config accounts list
- Added preloading instruction builder helpers for signers and signatures

## Testing
- Added tests for clear root functionality
- Updated existing tests to use new account structure
- Added scenarios testing root clearing and subsequent root setting

## Others
- Timelock preloading ix builder helper clean up - Removed unused parameters

PR #388, opened 2024-12-18
solana: reorganize utility functions

aalu1418:
- [x] reorganize functions
- [x] remove `testing` from exported functions

PR #393, opened 2024-12-19
Solana mcm signers in ConfigSet event [NONEVM-1068]

jadepark-dev:
- [x] To be merged after utility refactor https://github.com/smartcontractkit/chainlink-ccip/pull/388
- [x] To be merged after https://github.com/smartcontractkit/chainlink-ccip/pull/391

This PR resolves the out-of-memory issue that was occurring when emitting ConfigSet events with large numbers of signers in the MCM program.

## Primary Changes
- Modified set_config block scopes to fix memory overflow in ConfigSet event emission
- Adjusted maximum signer limit to prevent OOM(200 -> 180)

## Testing
- Verified ConfigSet events with maximum signer capacity
- Validated event emission across different config scenarios

## Notes
- Maximum signers reduced from 200 to 180 to ensure stable event emission


chains/solana/contracts/programs/mcm/src/constant.rs
```diff
@@ -1,5 +1,5 @@
 // Business-logic constants
-pub const MAX_NUM_SIGNERS: usize = 200; // This has to be below u8 limit (255). Value copied from EVM reference contract
+pub const MAX_NUM_SIGNERS: usize = 180; // maximum number of signers supported
```
agusaldasoro, line 2: I'm curious, is this what fixed the overflow?

agusaldasoro (dismissed):

chains/solana/contracts/programs/mcm/src/constant.rs
toblich: I have the same question :smile:

Besides curiosity, I assume the limit is high enough anyway so probably nothing to worry about, but do you know more or less how many signers we have today? And who should we notify about the overall limit shrinking from 200 to 180? This affects other chains as well as MCMS is "many-chain" multisig

jadepark-dev: @agusaldasoro 😅 with additional block scoping in set_config to save some runtime memory.

@toblich Thank you for checking in! I've been discussing the number of signers/signatures with the deployment automation team and research team - currently, 10-12 signers and 6-8 signatures are in production operation using mcms. So, 180 is pretty enough. 

There will be another discussion with research side about those numbers(among mcm w/ timelock - signers, signatures, supported max transactions) after the holidays.

toblich (approved):

PR #397, opened 2024-12-20
No panic on new request id

dimkouv:
`newRequestID()` might panic in some very rare case scenarios. Add some handling to prevent the panic.

### DIFF
--- a/commit/report.go
+++ b/commit/report.go
@@ -8,9 +8,9 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"golang.org/x/exp/maps"
 
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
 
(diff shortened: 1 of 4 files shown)


commit/merkleroot/rmn/controller.go
```diff
@@ -1072,11 +1073,24 @@ func randomShuffle[T any](s []T) []T {
...
 	_, err := crand.Read(b)
 	if err != nil {
-		panic(err)
+		// fallback to time-based id in the very rare scenario that the random number generator fails
+		lggr.Warnw("failed to generate random request id, falling back to time based",
+			"err", err,
+			"atomicCounter", atomicCounter.Load(),
+		)
+		t := time.Now().UTC().UnixNano()
+		c := atomicCounter.Add(1)
```
winder, line 1092: The end of the timestamp has the most entropy, it would be better to include the counter at the front.

Not sure this is strictly correct, but I'm thinking something like this
```
b := make([]byte, 8)
binary.BigEndian.PutInt64(b, t)
b[0]=byte(c%256)
```

winder (approved):

PR #399, opened 2024-12-20
CCIP-4712: Adding nil check to address CL33-16 finding 

b-gopalswami:
Adding nil check in couple of spots.


pkg/reader/ccip.go
```diff
@@ -1033,6 +1033,9 @@ func (r *ccipChainReader) getFeeQuoterTokenPriceUSD(ctx context.Context, tokenAd
 
 	price := timestampedPrice.Value
 
+	if price == nil {
+		return cciptypes.BigInt{}, fmt.Errorf("LINK token price is nil,  addr: %v", tokenAddr)
```
asoliman92, line 1037: That's not `LINK`

b-gopalswami: Is Link mentioned here is by mistake, then? https://github.com/smartcontractkit/chainlink-ccip/pull/399/files#diff-a16bcafc2684d6be248d96e8fdf5fa77ec115f311c0c73d13f3f942ad0edb0c4L1031

b-gopalswami: Removed Link

winder (approved):
lgtm

PR #400, opened 2024-12-20
CCIP-4710: Adding nil and length check on input value to Ed25519 verify call

b-gopalswami:
Making sure public key is not nil and length is accurate before verifying it.


commit/merkleroot/rmn/crypto.go
```diff
@@ -47,6 +47,12 @@ func verifyObservationSignature(
 	msg := append([]byte(signedObservationPrefix), observationBytesSha256[:]...)
 	msgSha256 := sha256.Sum256(msg)
 
+	if rmnNode.OffchainPublicKey == nil {
+		return fmt.Errorf("node %d has no offchain public key", rmnNode.ID)
+	}
+	if len(*rmnNode.OffchainPublicKey) != ed25519.PublicKeySize {
+		return fmt.Errorf("node %d has an invalid offchain public key", rmnNode.ID)
+	}
```
b-gopalswami, line 55: These are the key changes. As OffchainPublicKey is required, test needs to be updated.

winder (approved):

PR #407, opened 2024-12-26
Add solana checks to make file

agusaldasoro:
The Solana Contracts have many validations regarding format and autogenerated code, this new make command needs to be run only when a file under `chains/solana` was modified and auto fix style issues, and auto generates all needed files.


winder:
Could you put this in a make target? Something like "make solana_checks". The pre commit config would call the target (and maybe do the yubi key thing). We could also call that target as part of CI to ensure these files are always consistent.

agusaldasoro:
> Could you put this in a make target? Something like "make solana_checks". The pre commit config would call the target (and maybe do the yubi key thing).
> 
> We could also call that target as part of CI to ensure these files are always consistent. Add a call to the new target here:
> 
> https://github.com/smartcontractkit/chainlink-ccip/blob/f25773de7c0e600e16da43da9cfe261ad19f255e/.github/workflows/codegen.yml#L34

All Solana Validations are run in https://github.com/smartcontractkit/chainlink-ccip/blob/feat/add-docs/.github/workflows/solana.yml

The pre-commit hook is just for local dev

chains/solana/Makefile
```diff
@@ -44,6 +48,9 @@ format:
...
 	go test -v ./... -json -covermode=atomic -coverpkg=./... -coverprofile=integration_coverage.txt 2>&1 | tee /tmp/gotest.log | gotestloghelper -ci=true -singlepackage=true -hidepassingtests=false -hidepassinglogs=false
 
-.PHONY: generate-idl
-generate-idl:
+.PHONY: build-contracts
+build-contracts:
 	cd ./contracts && anchor build
+
+.PHONY: solana-checks
+solana-checks: clippy anchor-go-gen format gomodtidy lint-go rust-tests go-tests build-contracts
```
toblich, line 56: The `anchor-go-gen` is already building the contracts, could you please remove that redundancy to bring down the time it takes to run this command?
```suggestion
solana-checks: clippy anchor-go-gen format gomodtidy lint-go rust-tests go-tests
```

agusaldasoro: No, I needed it to recreate the IDL, as the tests modify them adding the program_ids

toblich (approved):

PR #409, opened 2024-12-30
CCIP-4709: Len check on observation

b-gopalswami:


winder (approved):

commit/tokenprice/processor.go
```diff
@@ -66,6 +67,9 @@ func (p *processor) Outcome(
 		p.lggr.Debugw("TokenPriceBatchWriteFrequency is set to zero, no prices will be reported")
 		return Outcome{}, nil
 	}
+	if len(aos) == 0 {
+		return Outcome{}, fmt.Errorf("no observations to process")
```
makramkd, line 71: Do we wanna return an error here or just an empty outcome? I don't really see this as an error

winder: You're right, this would be a problem for the discovery process specifically. On second inspection, the chainfee processor (and probably all of them) use the same pattern with no `len(aos)` check.

I think the audit issue is invalid, lets close this PR with no change.

winder (changes requested):
On closer inspection, this change is not necessary and could cause problems during contract discovery.

b-gopalswami:
Closing it as the change is not needed and considering audit feedback as invalid.

PR #410, opened 2024-12-31
CCIP-4703: Adding check for ChainFeeUpdates field

b-gopalswami:
Added missing check for ChainFeeUpdates field.

The review comment asks to check for non-positive value on DataAvailableFee field but I believe that can be zero when it is EVM chain so keeping that check for only non-negative values. Only if it is L2 chain, it should be non-positive.


commit/chainfee/validate_observation.go
```diff
@@ -50,5 +50,15 @@ func (p *processor) ValidateObservation(
 		}
 	}
 
+	for _, update := range obs.ChainFeeUpdates {
+		if update.ChainFee.ExecutionFeePriceUSD == nil || update.ChainFee.ExecutionFeePriceUSD.Cmp(zero) <= 0 {
+			return fmt.Errorf("nil or non-positive %s", "execution fee price")
+		}
+
+		if update.ChainFee.DataAvFeePriceUSD == nil || update.ChainFee.DataAvFeePriceUSD.Cmp(zero) < 0 {
```
b-gopalswami, line 58: I believe this DataAvFeePriceUSD might be zero for EVM chain and shouldn't be zero for L2 chains. Is that right? If at all we can know the given chain is L2, then we can add additional check here.

commit/chainfee/validate_observation.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/pkg/errors"
+
```
makramkd, line 8: Don't use errors.Wrap but fmt.Errorf

commit/chainfee/validate_observation.go
```diff
@@ -21,32 +23,71 @@ func (p *processor) ValidateObservation(
...
+		if update.ChainFee.DataAvFeePriceUSD == nil || update.ChainFee.DataAvFeePriceUSD.Cmp(big.NewInt(0)) < 0 {
+			return fmt.Errorf("nil or negative %s", "data availability fee price")
+		}
+	}
+	return nil
+}
+
+func validateFeeComponents(
+	ao plugincommon.AttributedObservation[Observation],
+) error {
```
makramkd, line 64: Can we have a unit test for this

commit/chainfee/validate_observation.go
```diff
@@ -21,32 +23,71 @@ func (p *processor) ValidateObservation(
...
 	}
 
-	for _, feeComponent := range obs.FeeComponents {
-		if feeComponent.ExecutionFee == nil || feeComponent.ExecutionFee.Cmp(zero) <= 0 {
+	return nil
+}
+
+func validateChainFeeUpdates(
+	ao plugincommon.AttributedObservation[Observation],
+) error {
```
makramkd, line 49: And this

commit/chainfee/validate_observation.go
```diff
@@ -21,32 +23,71 @@ func (p *processor) ValidateObservation(
 		return fmt.Errorf("failed to validate FChain: %w", err)
 	}
 
-	observerSupportedChains, err := p.chainSupport.SupportedChains(ao.OracleID)
-	if err != nil {
-		return fmt.Errorf("failed to get supported chains: %w", err)
+	if err := p.ValidateObservedChains(ao); err != nil {
+		return errors.Wrap(err, "failed to validate observed chains")
```
makramkd, line 27: fmt.Errorf instead of errors.Wrap

commit/chainfee/validate_observation.go
```diff
@@ -21,32 +21,74 @@ func (p *processor) ValidateObservation(
...
 			return fmt.Errorf("nil or negative %s", "data availability fee")
 		}
 	}
+	return nil
+}
 
-	for _, token := range obs.NativeTokenPrices {
-		if token.Int == nil || token.Int.Cmp(zero) <= 0 {
-			return fmt.Errorf("nil or non-positive %s", "execution fee")
+func (p *processor) ValidateObservedChains(
```
makramkd, line 78: This function shouldn't be public since it'll only get called from the chainfee processor

commit/chainfee/validate_observation.go
```diff
@@ -21,32 +21,74 @@ func (p *processor) ValidateObservation(
...
+	ao plugincommon.AttributedObservation[Observation],
+) error {
+	obs := ao.Observation
+	observerSupportedChains, err := p.chainSupport.SupportedChains(ao.OracleID)
+	if err != nil {
+		return fmt.Errorf("failed to get supported chains: %w", err)
+	}
+
```
asoliman92, line 86: We'll need to add extra checks here to make sure that all keys in all maps are exactly the same, otherwise we can end up with same lengths but for different chains on different maps.

commit/chainfee/validate_observation.go
```diff
@@ -21,32 +21,74 @@ func (p *processor) ValidateObservation(
...
-			return fmt.Errorf("chain %d is not supported by observer", chain)
+	if err := validateChainFeeUpdates(ao); err != nil {
+		return fmt.Errorf("failed to validate chain fee updates: %w", err)
+	}
+
+	for _, token := range obs.NativeTokenPrices {
+		if token.Int == nil || token.Int.Cmp(zero) <= 0 {
+			return fmt.Errorf("nil or non-positive %s", "execution fee")
 		}
 	}
```
asoliman92, line 41: Add validation for timpestamps making sure they're not zero or more than current utc timestamp.

asoliman92 (changes requested):
Please add validation for timestamps and for keys in maps as mentioned in the comments.

commit/chainfee/validate_observation.go
b-gopalswami: This is an interface method and can't be private.

makramkd: Why is this an interface method?

asoliman92 (approved):
LGTM

PR #412, opened 2025-01-06
TokenInfo validation for Solana

archseer:


tt-cll:
ping @smartcontractkit/ccip-offchain 

we're just looking to merge this to unblock solana

pluginconfig/commit.go
```diff
@@ -44,20 +47,37 @@ type TokenInfo struct {
...
-		return fmt.Errorf("aggregatorAddress must be a valid ethereum address, got %d bytes expected 20", len(decoded))
+	switch a.ChainFamily {
+	case chain_selectors.FamilySolana:
+		decoded, err := base58.Decode(string(a.AggregatorAddress))
+		if err != nil {
+			return fmt.Errorf("aggregatorAddress must be a valid Solana address (i.e. base58 encoded): %w", err)
+		}
+		if len(decoded) != 32 {
+			return fmt.Errorf("aggregatorAddress must be a valid Solana address, got %d bytes expected 32", len(decoded))
+		}
```
makramkd, line 68: Aggregators are only ever deployed on arbitrum and we'll hopefully be switching to Mercury some point thereafter or use keystone feeds. Don't think this makes sense but let me know the use case and we can see what you need.

makramkd (changes requested):

go.mod
```diff
@@ -35,7 +36,6 @@ require (
 	github.com/invopop/jsonschema v0.12.0 // indirect
 	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
-	github.com/mr-tron/base58 v1.2.0 // indirect
```
makramkd, line 38: Kind of concerning that this was even indirect before, looking into the `mod why`

makramkd: Ah I see, looks like its from chain-selectors. Should be OK

tt-cll:
closing because this isn't actually needed. We can safely skip this ctor for solana.

PR #414, opened 2025-01-06
Refactor methods into versioned module [NONEVM-1114]

toblich:
This is the first part of a refactor, more PRs to follow. The goal is to have a versioned module with just logic (no state or external interface) so that it can be more easily upgradeable. 

This PR moves all methods into an `instructions::v1` module, so the `lib.rs` exposes the external interface and just invokes the logic in said module. Upcoming PRs will continue working on other shared files, such as `messages.rs` and `ocr3base.rs` which have a mix of external interfaces and internal logic.


chains/solana/contracts/target/idl/ccip_router.json
```diff
@@ -1857,6 +1857,114 @@
         ]
       }
     },
+    {
+      "name": "LockOrBurnInV1",
```
toblich, line 1861: Note for reviewers: There is no real IDL changes due to this PR, it's just a reordering of some structs that I'm moving to different files so the auto-generated IDL lists them in different order

chains/solana/contracts/programs/ccip-router/src/lib.rs
```diff
@@ -392,18 +293,11 @@ pub mod ccip_router {
...
-        token_admin_registry.pending_administrator = Pubkey::new_from_array([0; 32]);
-        token_admin_registry.lookup_table = Pubkey::new_from_array([0; 32]);
-
-        emit!(AdministratorRegistered {
-            token_mint: mint,
-            administrator: token_admin_registry_admin,
-        });
-
-        Ok(())
+        v1::tokens::register_token_admin_registry_via_get_ccip_admin(
```
agusaldasoro, line 296: nit: rename tokens to token-admin-registry? And then fee_quoter could be a part of it as well

chains/solana/contracts/programs/ccip-router/src/lib.rs
```diff
@@ -18,20 +13,14 @@ use crate::state::*;
...
-mod pools;
-use crate::pools::*;
-
 mod ocr3base;
 use crate::ocr3base::*;
 
-mod fee_quoter;
-use crate::fee_quoter::*;
+mod instructions;
+use crate::instructions::*;
```
agusaldasoro, line 24: Maybe add a comment here saying that no code or logic should be here?

chains/solana/contracts/programs/ccip-router/src/lib.rs
```diff
@@ -18,20 +13,14 @@ use crate::state::*;
...
 
 // Anchor discriminators for CPI calls
 const CCIP_RECEIVE_DISCRIMINATOR: [u8; 8] = [0x0b, 0xf4, 0x09, 0xf9, 0x2c, 0x53, 0x2f, 0xf5]; // ccip_receive
```
agusaldasoro, line 26: Is this var then imported from the instructions crate?

chains/solana/contracts/programs/ccip-router/src/messages.rs
```diff
@@ -1,11 +1,10 @@
-use crate::{
-    BillingTokenConfig, CcipRouterError, DestChain, ReportContext,
-    TOKENPOOL_RELEASE_OR_MINT_DISCRIMINATOR,
-};
+use crate::{BillingTokenConfig, CcipRouterError, DestChain, ReportContext};
 use anchor_lang::prelude::*;
 use ethnum::U256;
 
-use crate::{ocr3base::Ocr3Report, ToTxData, TOKENPOOL_LOCK_OR_BURN_DISCRIMINATOR};
+// TODO this file has to be broken up
```
agusaldasoro, line 5: Public interfaces that should remain the same, for example Solana2Any and Any2Solana should still be part of this module, right? And not part of the instructions crate

chains/solana/contracts/programs/ccip-router/src/instructions/v1/admin.rs
```diff
@@ -0,0 +1,361 @@
+use anchor_lang::prelude::*;
+use anchor_spl::token_interface;
+
+use crate::{
+    AcceptOwnership, AddBillingTokenConfig, AddChainSelector, BillingTokenConfig, CcipRouterError,
```
agusaldasoro, line 5: Is there any cyclic dependency between files?

chains/solana/contracts/programs/ccip-router/src/instructions/v1/messages.rs
```diff
@@ -0,0 +1,58 @@
+use anchor_lang::prelude::*;
```
agusaldasoro, line 1: It looks like there is a typo in the file name, right? This should be something related to pools

toblich: Well, I extracted these from the `messages.rs` so I kept the name. I get that they are related to pools, so I'm happy to move them there if you think that's better, but maintaining the name was the "safer" refactor given that I haven't dived too deep into how those structs are used

chains/solana/contracts/programs/ccip-router/src/messages.rs
toblich: Right, anything that affects the public interface won't be versioned. That's why I have to go through this file with a finer comb and split the code there, as some interface-related structs also have methods that should be upgradeable, so I'll have to make some changes here and there to achieve that. As mentioned in the PR description, I'll do that in a separate PR, to not make this grow indefinitely

chains/solana/contracts/programs/ccip-router/src/lib.rs
toblich: Yes, the new `offramp.rs` file imports this. All the `const`s defined in the `lib.rs` are used in either the onramp or the offramp now, but they are tied to our interface with receivers and tokenpools so the constants themselves shouldn't be versioned.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/admin.rs
toblich: Mostly no. There is only one cycle right now, which is easy to break if you prefer.

The `lib.rs` is the only non-versioned file that imports versioned code. But the versioned code also imports the errors and constants that are listed in the `lib.rs`. This is not a problem in Rust, as I've tested this code locally and it's passing CI, but then again it's trivial for me to move errors and constants to separate files if you'd prefer.

chains/solana/contracts/programs/ccip-router/src/lib.rs
toblich: I'd rather keep fee_quoter separate from the rest of the tokens logic, as the fee calculation will have a lot of code unrelated to general token management. I'll make the rename though :+1:

agusaldasoro (dismissed):

agusaldasoro (dismissed):

agusaldasoro (approved):

PR #415, opened 2025-01-07
Don't exit query when a source chain is not available [CCIP-4759]

asoliman92:


mateusz-sekara (commented):
That one deserves a test, doesn't it?

mateusz-sekara (approved):

PR #419, opened 2025-01-07
solana: allow pools to specify writable accounts

aalu1418:
- [x] store writable accounts in token admin registry
- [x] use writable accounts to build transaction accounts
- [x] validate writable token pool accounts against registry
- [x] validate accounts in lookup table during SetPool
- [x] test cases
  - [x] test cases for incorrect write permissions for pool accounts
  - [x] test cases for incorrect pool accounts for `SetPool`


chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -53,20 +60,91 @@ pub fn register_token_admin_registry_via_owner(
...
+    token_admin_registry.lookup_table = new_pool;
 
-    // TODO: Validate here that the lookup table has everything
+    // set writable indexes
+    token_admin_registry.reset_writable();
+    for ind in writable_indexes {
+        token_admin_registry.set_writable(ind)
+    }
+
+    // validate lookup table contains minium required accounts if not zero address
```
jadepark-dev, line 79: nit: typo
```suggestion
    // validate lookup table contains minimum required accounts if not zero address
```

jadepark-dev (approved):

chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs
```diff
@@ -188,29 +188,48 @@ pub fn validate_and_parse_token_accounts<'info>(
...
-            CcipRouterError::InvalidInputsLookupTableAccounts
-        );
+        {
+            // validate pool addresses
+            let mut expected_keys: Vec<Pubkey> = required_entries.iter().map(|x| x.key()).collect();
+            let mut remaining_keys: Vec<Pubkey> =
+                remaining_accounts.iter().map(|x| x.key()).collect();
+            expected_keys.append(&mut remaining_keys);
+            require!(
+                lookup_table_account.addresses.len() == expected_keys.len(),
```
agusaldasoro, line 211: Do you need this check if you are checking after that the vectors are equal?

chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs
```diff
@@ -188,29 +188,48 @@ pub fn validate_and_parse_token_accounts<'info>(
...
+        {
+            // validate pool address writable
+            let mut expected_is_writable: Vec<bool> =
+                required_entries.iter().map(|x| x.is_writable).collect();
+            let mut remaining_is_writable: Vec<bool> =
+                remaining_accounts.iter().map(|x| x.is_writable).collect();
+            expected_is_writable.append(&mut remaining_is_writable);
+            for (i, is_writable) in expected_is_writable.iter().enumerate() {
+                require!(
+                    token_admin_registry_account.is_writable(i as u8) == *is_writable,
```
agusaldasoro, line 228: I'm not sure I understand this check

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -53,20 +60,91 @@ pub fn register_token_admin_registry_via_owner(
...
 ) -> Result<()> {
+    // set new lookup table
     let token_admin_registry = &mut ctx.accounts.token_admin_registry;
     let previous_pool = token_admin_registry.lookup_table;
-    token_admin_registry.lookup_table = pool_lookup_table;
+    let new_pool = ctx.accounts.pool_lookuptable.key();
+    token_admin_registry.lookup_table = new_pool;
 
-    // TODO: Validate here that the lookup table has everything
+    // set writable indexes
```
agusaldasoro, line 73: What is this doing? Any of the accounts passed can be writable? Like the regular accounts and the additionals?

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -53,20 +60,91 @@ pub fn register_token_admin_registry_via_owner(
 }
 
 pub fn set_pool(
-    ctx: Context<ModifyTokenAdminRegistry>,
+    ctx: Context<SetPoolTokenAdminRegistry>,
```
agusaldasoro, line 63: I'm wondering how this behaves when we'll have to deal with more than one Token Pool per token

chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs
aalu1418: i'll add some more comments - but it basically checks that the provided token pool account is (or is not) writable based on what's configured in the token admin registry (based on index)

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
aalu1418: correct, since these accounts are token pool specific, we need a way to know how to build the transaction offchain

this way provides an onchain record that admins can use to set which token pool accounts are writable

chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs
aalu1418: ah good point - probably not!

agusaldasoro (approved):

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -53,20 +58,87 @@ pub fn register_token_admin_registry_via_owner(
...
+    // indexes can be checked to indicate is_writable for the specific account
+    // this is also used to validate to ensure the correct write permissions are used according the admin
+    token_admin_registry.reset_writable();
+    for ind in writable_indexes {
+        token_admin_registry.set_writable(ind)
+    }
 
-    // TODO: Validate here that the lookup table has everything
+    // validate lookup table contains minimum required accounts if not zero address
+    if new_pool != Pubkey::zeroed() {
```
toblich, line 81: nit: If we use `Pubkey::default()` it returns the same thing as `Pubkey::zeroed()` and can make you remove the dependency on `bytemuck::Zeroable` here
```suggestion
    if new_pool != Pubkey::default() {
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -53,20 +58,87 @@ pub fn register_token_admin_registry_via_owner(
...
+        // deserialize lookup table account
+        let lookup_table_data = &mut &ctx.accounts.pool_lookuptable.data.borrow()[..];
+        let lookup_table_account: AddressLookupTable =
+            AddressLookupTable::deserialize(lookup_table_data)
+                .map_err(|_| CcipRouterError::InvalidInputsLookupTableAccounts)?;
+        require!(
+            lookup_table_account.addresses.len() >= 9,
```
toblich, line 88: Any chance you can link that `9` to a constant? It seems like that value is tied to logic in this file and in `pools.rs`, so it'd be nice if the compiler can check both places stay consistent in case we ever change one of them.

E.g., define a constant, use it here in the check, and use it in an explicit slice type in [this line](https://github.com/smartcontractkit/chainlink-ccip/pull/419/files#diff-47d7a8ac5ffa77e692d1f0db4fd1d601f82cec79acf45ba7fb67e036a0e3f0a1R183)
Said constant could be taken outside of the versioned logic folders if we consider that part of our interface with TokenPools.


Also, if you use `require_gte!` instead of `require!`, the logs will print the actual length in case of error, which can help when troubleshooting failed tests for example

chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs
toblich: Nit: if you use `require_eq!` instead of `require!`, the logs will have more info in case the check fails.

chains/solana/contracts/programs/ccip-router/src/token_context.rs
```diff
@@ -13,6 +13,35 @@ pub struct TokenAdminRegistry {
...
+        match index < 128 {
+            true => self.writable_indexes[0] & 1 << (127 - index) != 0,
+            false => self.writable_indexes[1] & 1 << (255 - index) != 0,
+        }
+    }
+
+    pub fn reset_writable(&mut self) {
+        self.writable_indexes = [0, 0];
+    }
 }
```
toblich, line 45: Up to you whether to leave this here for now or not, but in general I'd rather have functions defined in the versioned logic modules instead of here.

toblich (approved):
LGTM, all my previous comments are optional nitpicks, no need to delay the PR due to them if you'd rather just merge as-is

aalu1418:
> LGTM, all my previous comments are optional nitpicks, no need to delay the PR due to them if you'd rather just merge as-is

i will address these in a follow up PR! they seem pretty straight forward and I don't want any of my changes to block any refactors

PR #421, opened 2025-01-08
Audit fixes - Part 1

dimkouv:
- Remove unused RMNRemote reader interface, the empty implementation and the mocks.
- Improve the logic to better handle scenarios where previously one faulty chain breaks the processing of all the chains.

`core ref: bf3f4cbba69d17e4fa6502bd69d527509e97a6d5`

Core PR: https://github.com/smartcontractkit/chainlink/pull/15860

### DIFF
--- a/commit/merkleroot/query_test.go
+++ b/commit/merkleroot/query_test.go
@@ -90,7 +90,7 @@ func TestProcessor_Query(t *testing.T) {
 				},
 				RMNRemoteCfg: rmnRemoteCfg,
 			},
-			contractAddresses:  contractAddrs,
+			contractAddresses: contractAddrs,
 			cfg: pluginconfig.CommitOffchainConfig{
 				RMNEnabled:           true,
 				RMNSignaturesTimeout: 5 * time.Second,


asoliman92 (approved):

pkg/reader/ccip.go
```diff
@@ -399,22 +399,26 @@ func (r *ccipChainReader) GetExpectedNextSequenceNumber(
...
 		return nil, fmt.Errorf("get source chains config: %w", err)
 	}
 
-	res := make([]cciptypes.SeqNum, 0, len(chains))
+	res := make(map[cciptypes.ChainSelector]cciptypes.SeqNum, len(chains))
 	for _, chain := range chains {
 		cfg, exists := cfgs[chain]
 		if !exists {
-			return nil, fmt.Errorf("source chain config not found for chain %d", chain)
+			r.lggr.Warnw("source chain config not found for chain %d, chain is skipped.", chain)
```
makramkd, line 412: This log is malformed, if you want formatted logs should use `Warnf`, `Warnw` is a structure log so should look like

```golang
r.lggr.Warnw("source chain config not found for chain %d, chain is skipped.", chain)
```

pkg/reader/ccip.go
```diff
@@ -399,22 +399,26 @@ func (r *ccipChainReader) GetExpectedNextSequenceNumber(
...
+			continue
 		}
+
 		if cfg.MinSeqNr == 0 {
-			return nil, fmt.Errorf("minSeqNr not found for chain %d", chain)
+			r.lggr.Errorw("minSeqNr not found for chain %d or is set to 0, chain is skipped.", chain)
```
makramkd, line 417: Ditto here, re: Errorf vs. Errorw

pkg/reader/ccip.go
```diff
@@ -399,22 +399,26 @@ func (r *ccipChainReader) GetExpectedNextSequenceNumber(
 
 func (r *ccipChainReader) NextSeqNum(
 	ctx context.Context, chains []cciptypes.ChainSelector,
-) ([]cciptypes.SeqNum, error) {
+) (map[cciptypes.ChainSelector]cciptypes.SeqNum, error) {
```
makramkd, line 402: Does this method have tests in chainlink? If so needs to be updated there

PR #424, opened 2025-01-08
bump chainlink-common; rm unused PluginFactoryConstructor

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-2709

Requires:
- https://github.com/smartcontractkit/chainlink-common/pull/206

Supports:
- https://github.com/smartcontractkit/chainlink/pull/15495


execute/factory.go
```diff
@@ -244,5 +216,4 @@ func (p PluginFactory) HealthReport() map[string]error {
 }
 
 // Interface compatibility checks.
-var _ core.OCR3ReportingPluginClient = &PluginFactoryConstructor{}
```
jmank88, line 247: This type assertion was incompatible with the common bump, but the type is unused anyways.

mateusz-sekara: Yeah, seems unused. I don't see any references in chainlink repo as well so it should be safe to remove

mateusz-sekara (approved):

silaslenihan (approved):

PR #426, opened 2025-01-09
Allow only price aggregator contract to have multiple bindings [CCIP-4781]

asoliman92:
Allow only price aggregator contract to have multiple bindings, othe contracts should only have one instance at any single time.

core ref: 843e228fba3c5264d34fb5bc2bd972f1576057d8

Core PR: https://github.com/smartcontractkit/chainlink/pull/15854

### DIFF
--- a/pkg/contractreader/extended.go
+++ b/pkg/contractreader/extended.go
@@ -109,9 +109,7 @@ func NewExtendedContractReader(baseContractReader ContractReaderFacade) Extended
 	}
 	// so far this is the only contract that allows multiple bindings
 	// if more contracts are added, this should be moved to a config
-	multiBindAllowed := map[string]bool{
-		consts.ContractNamePriceAggregator: true,
-	}
+	multiBindAllowed := map[string]bool{consts.ContractNamePriceAggregator: true}
 	return &extendedContractReader{
 		reader:                 baseContractReader,
 		contractBindingsByName: make(map[string][]ExtendedBoundContract),
@@ -277,6 +275,13 @@ func (e *extendedContractReader) Bind(ctx context.Context, allBindings []types.B
 				Binding: binding,
 			})
 		} else {
+			if len(e.contractBindingsByName[binding.Name]) > 0 {
+				// Unbind the previous binding
+				err := e.reader.Unbind(ctx, []types.BoundContract{e.contractBindingsByName[binding.Name][0].Binding})
+				if err != nil {
+					return fmt.Errorf("failed to unbind previous binding: %w", err)
+				}
+			}
 			// Override the previous binding
 			e.contractBindingsByName[binding.Name] = []ExtendedBoundContract{{
 				BoundAt: time.Now(),
--- a/pkg/contractreader/extended.go
+++ b/pkg/contractreader/extended.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
 	"sync"
 	"time"
+
+	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
--- a/pkg/contractreader/extended.go
+++ b/pkg/contractreader/extended.go
@@ -107,14 +107,13 @@ func NewExtendedContractReader(baseContractReader ContractReaderFacade) Extended
 	if ecr, ok := baseContractReader.(Extended); ok {
 		return ecr
 	}
-	// so far this is the only contract that allows multiple bindings
-	// if more contracts are added, this should be moved to a config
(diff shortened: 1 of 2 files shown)


pkg/contractreader/extended.go
```diff
@@ -95,17 +97,25 @@ type ExtendedBoundContract struct {
...
 func NewExtendedContractReader(baseContractReader ContractReaderFacade) Extended {
 	// avoid double wrapping
 	if ecr, ok := baseContractReader.(Extended); ok {
 		return ecr
 	}
+	// so far this is the only contract that allows multiple bindings
+	// if more contracts are added, this should be moved to a config
+	multiBindAllowed := map[string]bool{
+		consts.ContractNamePriceAggregator: true,
+	}
```
makramkd, line 114: Inline

pkg/contractreader/extended.go
```diff
@@ -261,10 +271,18 @@ func (e *extendedContractReader) Bind(ctx context.Context, allBindings []types.B
...
-			BoundAt: time.Now(),
-			Binding: binding,
-		})
+		if e.multiBindAllowed[binding.Name] {
+			e.contractBindingsByName[binding.Name] = append(e.contractBindingsByName[binding.Name], ExtendedBoundContract{
+				BoundAt: time.Now(),
+				Binding: binding,
+			})
+		} else {
+			// Override the previous binding
```
makramkd, line 280: This will only override it in this state but not in the contract reader itself, feels like we need to Unbind then Bind there appropriately

asoliman92: That's not needed, we deal only with extended contract reader, and the identifier is the one we use with cr which will only be the latest to add.

pkg/contractreader/extended.go
```diff
@@ -261,10 +269,25 @@ func (e *extendedContractReader) Bind(ctx context.Context, allBindings []types.B
...
-		})
+		if e.multiBindAllowed[binding.Name] {
+			e.contractBindingsByName[binding.Name] = append(e.contractBindingsByName[binding.Name], ExtendedBoundContract{
+				BoundAt: time.Now(),
+				Binding: binding,
+			})
+		} else {
+			if len(e.contractBindingsByName[binding.Name]) > 0 {
+				// Unbind the previous binding
+				err := e.reader.Unbind(ctx, []types.BoundContract{e.contractBindingsByName[binding.Name][0].Binding})
```
makramkd, line 280: Double checking - are we certain there's only one other binding in this situation?

pkg/contractreader/extended.go
```diff
@@ -95,17 +97,23 @@ type ExtendedBoundContract struct {
...
 	if ecr, ok := baseContractReader.(Extended); ok {
 		return ecr
 	}
+	// so far this is the only contract that allows multiple bindings
+	// if more contracts are added, this should be moved to a config
+	multiBindAllowed := map[string]bool{consts.ContractNamePriceAggregator: true}
 	return &extendedContractReader{
 		reader:                 baseContractReader,
 		contractBindingsByName: make(map[string][]ExtendedBoundContract),
+		multiBindAllowed:       multiBindAllowed,
```
makramkd, line 116:
```suggestion
		multiBindAllowed:       map[string]bool{consts.ContractNamePriceAggregator: true},
```

makramkd, line 116:
```suggestion
		multiBindAllowed:       map[string]bool{consts.ContractNamePriceAggregator: true},
```

makramkd (dismissed):

pkg/contractreader/extended.go
asoliman92: Because of the current condition yes there should only be one binding.

makramkd (approved):

pkg/contractreader/extended.go
winder: There could be more if the extended reader was created with a reader that already had bound contracts. But that would just mean there are some dangling bindings, our discovery code would work fine because error checking is done using the `contractBindingsByName` list.

pkg/contractreader/extended.go
```diff
@@ -261,10 +268,25 @@ func (e *extendedContractReader) Bind(ctx context.Context, allBindings []types.B
...
 	for _, binding := range validBindings {
-		e.contractBindingsByName[binding.Name] = append(e.contractBindingsByName[binding.Name], ExtendedBoundContract{
-			BoundAt: time.Now(),
-			Binding: binding,
-		})
+		if e.multiBindAllowed[binding.Name] {
+			e.contractBindingsByName[binding.Name] = append(e.contractBindingsByName[binding.Name], ExtendedBoundContract{
+				BoundAt: time.Now(),
+				Binding: binding,
+			})
```
winder, line 275: style nit: I would be tempted to rewrite this to avoid the `else`. Something like
```
    // unbind previous contract if needed
    if len(e.contractBindingsByName[binding.Name]) > 0 && e.multiBindAllowed[binding.Name] {
        e.reader.Unbind(...)
        e.contractBindingsByName[binding.Name] = nil
    }

    // existing logic here
}
```

winder (approved):
Looks good, one minor suggestions

PR #428, opened 2025-01-09
No more duplicate price updates

dimkouv:
The offRamp contract has a [check](https://github.com/smartcontractkit/chainlink/blob/4a19318efa56c079da53777f82404a1fbb24479a/contracts/src/v0.8/ccip/offRamp/OffRamp.sol#L802) to order the price updates by ignoring the ones from old reports (since reports are not expected to be delivered ordered).
```
if (commitReport.priceUpdates.tokenPriceUpdates.length > 0 || commitReport.priceUpdates.gasPriceUpdates.length > 0)
    {
      uint64 ocrSequenceNumber = uint64(uint256(reportContext[1]));
      if (s_latestPriceSequenceNumber < ocrSequenceNumber) {
```

But the following can happen:
```
ocrSeqNum    reportStatus
1                      done
2                     inflight (tx didn't make it yet)
3                     oracle is processing... and eventually submits the same pricess
```

So on round with ocrSeqNum `3` we might send duplicate prices, which will make it through all checks.
This is considered spam (submitting the same prices more than once).

To prevent that we introduce a new offchain config variable `InflightPriceCheckRetries` which by default is set to 3.
This means that we wait a maximum of 3 rounds for the last price report (both gas prices and token prices to make it through).

So what will happen now when a report is inflight:

Scenario 1: report eventually makes it
```
ocrSeqNum    reportStatus
1                      done
2                     inflight (tx didn't make it yet)
3                     oracle is processing... no prices observed
4                     oracle is processing... prices can be observed (2 made it through the chain)
```

Scenario 2: report never makes it
```
ocrSeqNum    reportStatus
1                      done
2                     inflight (tx didn't make it yet)
3                     inflight (tx didn't make it yet)
4                     inflight (tx didn't make it yet)
5                    oracle is processing... prices can be observed we reached all 3 retries.
```

This allows CCIP protocol to be configured according to chain specific requirements to prevent price updates spam.

`core ref: 2c4db2b2a630b8b36263d15e4b43994bbe66116c`


commit/committypes/types.go
```diff
@@ -54,6 +54,19 @@ type Outcome struct {
 	MerkleRootOutcome merkleroot.Outcome `json:"merkleRootOutcome"`
 	TokenPriceOutcome tokenprice.Outcome `json:"tokenPriceOutcome"`
 	ChainFeeOutcome   chainfee.Outcome   `json:"chainFeeOutcome"`
+	MainOutcome       MainOutcome        `json:"mainOutcome"`
+}
+
+// MainOutcome contains fields produced by the main commit plugin outcome (not of some sub-processor).
+type MainOutcome struct {
```
makramkd, line 61: Haven't read thru full PR yet but any reason this can't be in the processor?

makramkd: Is it dup logic or something else?

pluginconfig/commit.go
```diff
@@ -26,6 +26,7 @@ const (
 	defaultRemoteGasPriceBatchWriteFrequency  = 1 * time.Minute
 	defaultSignObservationPrefix              = "chainlink ccip 1.6 rmn observation"
 	defaultTransmissionDelayMultiplier        = 30 * time.Second
+	defaultInflightPriceCheckRetries          = 3
```
makramkd, line 29: Assuming 1s per round is this really a reasonable # of rounds for a default?

commit/plugin.go
```diff
@@ -384,11 +430,40 @@ func (p *Plugin) Outcome(
...
+	outCtx ocr3types.OutcomeContext,
+	prevOutcome committypes.Outcome,
+	tokenPriceOutcome tokenprice.Outcome,
+	chainFeeOutcome chainfee.Outcome,
+) committypes.MainOutcome {
+	pricesObservedInThisRound := len(tokenPriceOutcome.TokenPrices) > 0 || len(chainFeeOutcome.GasPrices) > 0
+	if pricesObservedInThisRound {
+		return committypes.MainOutcome{
+			InflightPriceOcrSequenceNumber: cciptypes.SeqNum(outCtx.SeqNr),
+			RemainingPriceChecks:           p.offchainCfg.InflightPriceCheckRetries,
```
makramkd, line 449: nit: reorder to have `RemainingPriceChecks` first to match below two MainOutcome instatiations

pluginconfig/commit.go
makramkd: I don't think one block is a reasonable expectation to get a tx in, especially in periods of high traffic.

commit/committypes/types.go
makramkd, line 61: Not a big fan of this name tbh but I don't really have a better suggestion haha

makramkd (approved):

PR #430, opened 2025-01-09
Refactor report builder interfaces.

winder:
This PR contains no functional changes, only refactoring to the execution report builder.

New functional options for `report.NewBuilder` which allow for more dynamic configuration without further modification to the function signature.

Break the message checking functions into multiple `Check` functions which can be added or removed based on requirements for a particular chain. This also allows more complex checks to be added without increasing the overall builder complexity, for example converting the nonce checking to the new scheme allowed me to remove two variables from the builder struct.

### DIFF
--- a/execute/report/builder.go
+++ b/execute/report/builder.go
@@ -18,29 +18,78 @@ type ExecReportBuilder interface {
 	Build() ([]cciptypes.ExecutePluginReportSingleChain, error)
 }
 
+type Option func(erb *execReportBuilder)
+
+func WithMaxGas(maxGas uint64) func(*execReportBuilder) {
+	return func(erb *execReportBuilder) {
+		erb.maxGas = maxGas
+	}
+}
+
+func WithMaxReportSizeBytes(maxReportSizeBytes uint64) Option {
+	return func(erb *execReportBuilder) {
+		erb.maxReportSizeBytes = maxReportSizeBytes
+	}
+}
+
+func MaxMessages(maxMessages uint64) Option {
+	return func(erb *execReportBuilder) {
+		erb.maxMessages = maxMessages
+		panic("not implemented")
+	}
+}
+
+// WithExtraMessageCheck adds additional message checks to the default ones.
+func WithExtraMessageCheck(check Check) Option {
+	return func(erb *execReportBuilder) {
+		erb.checks = append(erb.checks, check)
+	}
+}
+
+func newBuilderInternal(
+	logger logger.Logger,
+	hasher cciptypes.MessageHasher,
+	encoder cciptypes.ExecutePluginCodec,
+	estimateProvider cciptypes.EstimateProvider,
+	destChainSelector cciptypes.ChainSelector,
+	options ...Option,
+) *execReportBuilder {
+	defaultChecks := []Check{
+		CheckIfPseudoDeleted(),
+		CheckAlreadyExecuted(),
+		CheckTokenData(),
+		CheckTooCostly(),
+	}
+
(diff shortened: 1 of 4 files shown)


execute/report/builder.go
```diff
@@ -18,29 +18,81 @@ type ExecReportBuilder interface {
 	Build() ([]cciptypes.ExecutePluginReportSingleChain, error)
 }
 
+// Option that can be passed to the builder.
+type Option func(erb *execReportBuilder)
```
asoliman92, line 22: Do you think it's good to return err to check for any issues while calling an Option?

asoliman92: It's okay leaving it as is, `Check` is doing what I was thinking about better 😁

execute/report/report.go
```diff
@@ -145,136 +147,184 @@ const (
...
-		}
-		if _, ok := b.expectedNonce[execReport.SourceChain][sender]; !ok {
-			b.expectedNonce[execReport.SourceChain][sender] = chainNonces[sender] + 1
+// CheckNonces compares message nonces against the input to determine message validity. A cache is maintained within
+// the check in order to increment nonces for subsequent messages from the same sender.
+func CheckNonces(sendersNonce map[cciptypes.ChainSelector]map[string]uint64) Check {
+	// temporary map to store state between nonce checks for this round.
+	expectedNonce := make(map[cciptypes.ChainSelector]map[string]uint64)
+	return func(lggr logger.Logger, msg cciptypes.Message, idx int, report exectypes.CommitData) (messageStatus, error) {
+		if msg.Header.Nonce != 0 {
```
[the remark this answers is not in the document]
winder: I'll add a comment here explaining that.

execute/report/report.go
```diff
@@ -145,136 +147,184 @@ const (
...
+			if _, ok := expectedNonce[report.SourceChain]; !ok {
+				// initialize expected nonce if needed.
+				expectedNonce[report.SourceChain] = make(map[string]uint64)
+			}
+			if _, ok := expectedNonce[report.SourceChain][sender]; !ok {
+				expectedNonce[report.SourceChain][sender] = chainNonces[sender] + 1
+			}
+
+			// Check expected nonce is valid for sequenced messages.
+			if msg.Header.Nonce != expectedNonce[report.SourceChain][sender] {
```
asoliman92, line 281: That assumes that messages will be sent to this function in order? Should this be a holding assumption always?

execute/report/builder.go
```diff
@@ -18,29 +18,81 @@ type ExecReportBuilder interface {
...
+	encoder cciptypes.ExecutePluginCodec,
+	estimateProvider cciptypes.EstimateProvider,
+	destChainSelector cciptypes.ChainSelector,
+	options ...Option,
+) *execReportBuilder {
+	defaultChecks := []Check{
+		CheckIfPseudoDeleted(),
+		CheckAlreadyExecuted(),
+		CheckTokenData(),
+		CheckTooCostly(),
```
asoliman92, line 65: No Nonce check?

execute/report/report.go
winder: Yes, if they're sent out of my order the expected nonce will be wrong.

asoliman92 (approved):

execute/report/builder.go
winder: I'll add a comment. I had it here and removed it because it wasn't being used.

asoliman92 (approved):
Looks great!

PR #432, opened 2025-01-09
Refactor: move ramp hashes & validations [NONEVM-1114]

toblich:
_Continues the refactor started in #414 & #418_

Moves logic out of `messages.rs` into the versioned module. I've move the message hashing functions to their corresponding on/off-ramp files directly, and I've put the validation of onramp messages in the versioned `messages.rs` as it's also used by fee_quoter and has many lines of tests.


aalu1418 (commented):
lgtm! looks like just an issue with the rust unit test dependencies

chains/solana/contracts/programs/ccip-router/src/instructions/v1/messages.rs
```diff
@@ -1,9 +1,17 @@
...
+use crate::{
+    BillingTokenConfig, CcipRouterError, DestChain, Solana2AnyMessage, CHAIN_FAMILY_SELECTOR_EVM,
+    TOKENPOOL_LOCK_OR_BURN_DISCRIMINATOR, TOKENPOOL_RELEASE_OR_MINT_DISCRIMINATOR,
+};
 
 use super::pools::ToTxData;
 
+///////////////////
+// Pool Messages //
+///////////////////
```
PabloMansanet, line 13: Nit: These kind of "category comments" always seem like a bit of a smell to me: I think they're hinting at the need of a module. I would use `pub mod pool {}` instead, so the namespace ends up looking like `messages::pool::`. Ditto for other categories.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/messages.rs
```diff
@@ -56,3 +64,230 @@ pub struct LockOrBurnOutV1 {
...
+            PRECOMPILE_SPACE,
+            CcipRouterError::InvalidEVMAddress
+        )
+    };
+
+    Ok(())
+}
+
+#[cfg(test)]
+pub(crate) mod tests {
```
PabloMansanet, line 138: Nit: is `(crate)` an artifact of automated refactoring? (asking because it tends to be).

PabloMansanet (approved):
Looks good to me, I only have a couple minor comments on code organization.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/messages.rs
toblich: It was already there. There are a couple of test data factories that are reused in other tests, so it has to be public within some context. It doesn't make any difference whether that's restricted to the crate or not though, as this already has the `#[cfg(test)]` anyway, so I've changed this to a regular `pub` :+1:

aalu1418 (approved):

agusaldasoro (approved):

PabloMansanet (approved):
LGTM!

PR #439, opened 2025-01-13
[Solana][NONEVM-677] processMessageArgs flow

PabloMansanet:
~Requesting this as a draft for now, before I get started on the unit tests, as I want to run some of the decisions by the team before investing time on them.~ Ready for review now.

From what I can tell, this effectively ports the behaviour present in `processMessageArgs`. However, I've organized it in a significantly different way, because I felt like `processMessageArgs` was quite coupled to EVM tradeoffs and architecture, and it didn't quite make sense to port 1:1 as a function. Namely:

* `processMessageArgs` first performs some fee validation (ensuring a maximum fee in LINK). This appears misplaced and I have made it part of `ccip_send` instead. I suspect this validation is done here in EVM simply because of processMessageArgs being visible from outside the contract, but won't be necessary with the monolithic Solana approach. Doing this required expanding the configuration (and ccip_send context) with the LINK mint address.
* `processMessageArgs` then performs some parsing/processing of byte args. This also doesn't seem necessary in Solana as the client is calling the `ccip_send` instruction using structured data where the extra args are already in `Solana2AnyMessage`, so I've simple excluded this part (ironically, since it's the only behaviour that matches the function's name).
* In the end, `processMessageArgs` performs some validation on the lock/burn output data to construct the dest chain token transfers. Part of this work was already done, but I've ported over remaining checks on data length and on gas overhead values. However, I did *not* port validation over the dest family chain, as this is done before during `get_fee`.

### DIFF
--- a/chains/solana/contracts/programs/ccip-router/src/instructions/v1/mod.rs
+++ b/chains/solana/contracts/programs/ccip-router/src/instructions/v1/mod.rs
@@ -4,7 +4,6 @@
 pub mod admin;
 pub mod offramp;
 pub mod onramp;
-pub mod price_math;
 pub mod token_admin_registry;
 
 /////////////////////
@@ -16,3 +15,4 @@ mod messages;
 mod ocr3base;
 mod ocr3impl;
 mod pools;
+mod price_math;
--- a/chains/solana/contracts/programs/ccip-router/src/instructions/v1/mod.rs
+++ b/chains/solana/contracts/programs/ccip-router/src/instructions/v1/mod.rs
@@ -4,6 +4,7 @@
 pub mod admin;
 pub mod offramp;
 pub mod onramp;
+pub mod price_math;
 pub mod token_admin_registry;
 
 /////////////////////
@@ -15,4 +16,3 @@ mod messages;
 mod ocr3base;
 mod ocr3impl;
 mod pools;
-mod price_math;
--- a/chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
+++ b/chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
@@ -6,7 +6,7 @@ use anchor_spl::token_interface;
 use super::fee_quoter::{fee_for_msg, transfer_fee, wrap_native_sol};
 use super::messages::pools::{LockOrBurnInV1, LockOrBurnOutV1};
 use super::pools::{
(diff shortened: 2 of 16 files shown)


chains/solana/contracts/programs/ccip-router/src/instructions/v1/mod.rs
```diff
@@ -4,6 +4,7 @@
 pub mod admin;
 pub mod offramp;
 pub mod onramp;
+pub mod price_math;
```
PabloMansanet, line 7: I've made this module public as some of the utilities were necessary to call from `onramp`, not just `fee_quoter`, plus they seemed to be mostly utility functions. LMK if that's okay with your refactor @toblich

chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs
```diff
@@ -401,18 +401,18 @@ mod tests {
             ],
         };
 
-        assert_eq!(state.is_writable(0), false);
-        assert_eq!(state.is_writable(128), false);
-        assert_eq!(state.is_writable(255), false);
+        assert!(!state.is_writable(0));
```
PabloMansanet, line 404: Sorry for the noise here, just a drive-by fix of some clippy lints.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -119,6 +119,17 @@ pub fn ccip_send<'info>(
         &per_chain_per_token_config_accounts,
     )?;
 
+    let link_fee = fee.convert(
```
toblich, line 122: Looking at the EVM code, there's also a `feeValueJuels` in the `EVM2AnyRampMessage` struct, which is what ultimately is emitted in the onramp event. I see that in Solana we don't have an equivalent property, but it feels like we should probably add it to the `Solana2AnyRampMessage`, wdyt?

toblich (commented):
Looks good! A few nits related to the refactor, and some small questions

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
PabloMansanet: Yep, there's a ticket for that which I'll tackle after this one: https://smartcontract-it.atlassian.net/browse/NONEVM-1147

PR #440, opened 2025-01-13
Add support for config two on ramp as source chain

agusaldasoro:
Allow configuring none, one or two On Ramp addresses for executing messages.
Changes:
- Modified Source Chain PDA to store a vector of 128 bytes, so it can support address up to 64 size
- Assume Right Padding: the trailing zeros are on the right. For example when storing an address that is 32 bytes then the value stored is `[address, vec![0; 32]].concat()`
- Added on ramp address to the Any2Solana Message, as it can not be inferred (as done on EVM) and it's needed for the hash validation.
- Added unit tests in rust for this logic changes


chains/solana/contracts/programs/ccip-router/src/state.rs
```diff
@@ -67,8 +69,49 @@ pub struct GlobalState {
...
+        let mut on_ramp_bytes = [0u8; 64];
+        let len = on_ramp.len().min(64);
+        on_ramp_bytes[..len].copy_from_slice(&on_ramp[..len]);
+
+        let valid_on_ramps = self.get_on_ramps();
+        if valid_on_ramps.is_err() {
+            return false;
+        }
+        valid_on_ramps.unwrap().contains(&on_ramp_bytes)
+    }
```
toblich, line 114: Could you please move this into the versioned module, and not as a method of an object that belongs to our external interface? :pray: :grimacing:

chains/solana/contracts/programs/ccip-router/src/state.rs
```diff
@@ -67,8 +69,49 @@ pub struct GlobalState {
 #[derive(Clone, AnchorSerialize, AnchorDeserialize, InitSpace, Debug)]
 pub struct SourceChainConfig {
     pub is_enabled: bool, // Flag whether the source chain is enabled or not
-    #[max_len(64)]
-    pub on_ramp: Vec<u8>, // OnRamp address on the source chain
+    #[max_len(128)]
+    pub on_ramp: Vec<u8>, // OnRamp addresses supported from the source chain, each of them has a 64 byte address. So this can hold 2 addresses.
```
toblich, line 73: Any reason no to make this a `Vec<Vec<u8>>` instead? Then you could set `max_len(2,64)` on it. Or maybe even `[[u8;64];2]`, though I know the latter sometimes hits complications with anchor

chains/solana/contracts/programs/ccip-router/src/lib.rs
```diff
@@ -564,6 +564,7 @@ pub mod ccip_router {
     ///     * report_context_byte_words[0]: ConfigDigest
     ///     * report_context_byte_words[1]: 24 byte padding, 8 byte sequence number
     ///     * report_context_byte_words[2]: ExtraHash
+    /// * `on_ramp_address` - The address of the on-ramp program to send the message from
```
toblich, line 567: This doesn't seem to be part of the method arguments, could you please remove this from the docstring? As far as I can tell is just a new field in one of the accounts that's part of Anchor's context (not the report context), right?

chains/solana/contracts/programs/ccip-router/src/lib.rs
```diff
@@ -582,6 +583,7 @@ pub mod ccip_router {
     ///
     /// * `ctx` - The context containing the accounts required for the execution.
     /// * `execution_report` - The execution report containing the message and proofs.
+    /// * `on_ramp_address` - The address of the on-ramp program to send the message from
```
toblich, line 586: Same here

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -668,12 +676,13 @@ pub fn validate_execution_report<'info>(
...
 
-fn hash(msg: &Any2SolanaRampMessage, on_ramp_address: &[u8]) -> [u8; 32] {
+fn hash(msg: &Any2SolanaRampMessage) -> [u8; 32] {
     use anchor_lang::solana_program::hash;
 
     // Calculate vectors size to ensure that the hash is unique
     let sender_size = [msg.sender.len() as u8];
-    let on_ramp_address_size = [on_ramp_address.len() as u8];
+    let trimmed_on_ramp_address = trim_leading_zeros(&msg.on_ramp_address);
+    let on_ramp_address_size = [trimmed_on_ramp_address.len() as u8];
```
toblich, line 685: Related to my question regarding the type used to store the list of configured onramps: Any reason why not just hash the fixed 64 bytes we're allocating per address (with the necessary padding)? Just asking because it seems like changing those types might simplify the code a bit in a few places.

agusaldasoro: I wasn't sure about this, I thought that this way will be more simple for the offchain code. But it makes sense we can force them to fit the 64bytes as well

toblich: If we stick to the type being `Vec<u8>` with `max_len(128)`, I think what you did makes sense. My point is that if we change it to `[[u8; 64]; 2]`, then you can simplify code in a few different places, this being one of them by just always using 64 bytes.

This is all in the context of https://github.com/smartcontractkit/chainlink-ccip/pull/440#discussion_r1913757623

aalu1418 (approved):

aalu1418 (approved):

PR #441, opened 2025-01-13
Prevent exec observation early errors [CCIP-4801]

asoliman92:
Allow observation to continue even when having errors fetching some messages or commit reports.

For token data we don't fail on merge errors, keep in mind that we currently have only usdc token data observer only anyways so hitting this error is unlikely in the near future.

### DIFF
--- a/execute/observation.go
+++ b/execute/observation.go
@@ -171,10 +171,11 @@ func regroup(commitData []exectypes.CommitData) exectypes.CommitObservations {
 	return groupedCommits
 }
 
-// filterCommitReports filters out commit reports that have no messages or have missing messages.
+// intersectCommitReportsAndMessages filters out commit reports that have no messages or have missing messages.
 // as all messages for each report are required to be present in the observation, otherwise merkle proofs will fail.
-func filterCommitReports(msgObs exectypes.MessageObservations, commitObs exectypes.CommitObservations) exectypes.CommitObservations {
+func intersectCommitReportsAndMessages(msgObs exectypes.MessageObservations, commitObs exectypes.CommitObservations) (exectypes.MessageObservations, exectypes.CommitObservations) {
 	filteredCommitReports := make(exectypes.CommitObservations)
+	filteredMsgs := msgObs
 	for srcChain, reports := range commitObs {
 		filteredReports := make([]exectypes.CommitData, 0)
 		// filter out reports that have no messages
@@ -193,6 +194,11 @@ func filterCommitReports(msgObs exectypes.MessageObservations, commitObs exectyp
 			}
 			if valid {
 				filteredReports = append(filteredReports, report)
+			} else {
+				// remove messages that are not valid anymore
+				for seq := report.SequenceNumberRange.Start(); seq <= report.SequenceNumberRange.End(); seq++ {
+					delete(filteredMsgs[srcChain], seq)
+				}
 			}
 		}
 
@@ -201,7 +207,7 @@ func filterCommitReports(msgObs exectypes.MessageObservations, commitObs exectyp
 		}
 
 	}
-	return filteredCommitReports
+	return filteredMsgs, filteredCommitReports
 }
 
 func readAllMessages(
@@ -266,7 +272,7 @@ func (p *Plugin) getMessagesObservation(
 
 	messageObs := readAllMessages(ctx, p.lggr, p.ccipReader, commitReportCache)
 
-	filteredCommitReports := filterCommitReports(messageObs, commitReportCache)
+	messageObs, filteredCommitReports := intersectCommitReportsAndMessages(messageObs, commitReportCache)
 
 	messageTimestamps := getMessageTimestampMap(filteredCommitReports, messageObs)
 
--- a/execute/observation.go
+++ b/execute/observation.go
@@ -171,87 +171,48 @@ func regroup(commitData []exectypes.CommitData) exectypes.CommitObservations {
 	return groupedCommits
 }
 
-// intersectCommitReportsAndMessages filters out commit reports that have no messages or have missing messages.
-// as all messages for each report are required to be present in the observation, otherwise merkle proofs will fail.
-func intersectCommitReportsAndMessages(msgObs exectypes.MessageObservations,
-	commitObs exectypes.CommitObservations,
-) (exectypes.MessageObservations, exectypes.CommitObservations) {
-	filteredCommitReports := make(exectypes.CommitObservations)
-	filteredMsgs := msgObs
-	for srcChain, reports := range commitObs {
-		filteredReports := make([]exectypes.CommitData, 0)
-		// filter out reports that have no messages
-		if _, ok := msgObs[srcChain]; !ok {
-			continue
-		}
-
-		// filter out reports that have missing messages
-		for _, report := range reports {
-			valid := true
-			for seq := report.SequenceNumberRange.Start(); seq <= report.SequenceNumberRange.End(); seq++ {
-				if _, ok := msgObs[srcChain][seq]; !ok {
-					valid = false
-					break
-				}
-			}
-			if valid {
-				filteredReports = append(filteredReports, report)
-			} else {
-				// remove messages that are not valid anymore
-				for seq := report.SequenceNumberRange.Start(); seq <= report.SequenceNumberRange.End(); seq++ {
-					delete(filteredMsgs[srcChain], seq)
-				}
-			}
-		}
-
-		if len(filteredReports) > 0 {
-			filteredCommitReports[srcChain] = filteredReports
-		}
-
-	}
-	return filteredMsgs, filteredCommitReports
-}
-
 func readAllMessages(
 	ctx context.Context,
 	lggr logger.Logger,
 	ccipReader reader.CCIPReader,
-	commitObs exectypes.CommitObservations,
-) exectypes.MessageObservations {
+	commitData []exectypes.CommitData,
+) (exectypes.MessageObservations, exectypes.CommitObservations, map[cciptypes.Bytes32]time.Time) {
 	messageObs := make(exectypes.MessageObservations)
+	availableReports := make(exectypes.CommitObservations)
+	messageTimestamps := make(map[cciptypes.Bytes32]time.Time)
+
+	commitObs := regroup(commitData)
 
 	for srcChain, reports := range commitObs {
 		if len(reports) == 0 {
 			continue
 		}
 
-		ranges, err := computeRanges(reports)
-		if err != nil {
-			lggr.Errorw("unable to compute ranges", "err", err)
-			continue
-		}
-
+		messageObs[srcChain] = make(map[cciptypes.SeqNum]cciptypes.Message)
 		// Read messages for each range.
-		for _, seqRange := range ranges {
-			// TODO: check if srcChain is supported.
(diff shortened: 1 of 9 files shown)


execute/tokendata/observer.go
```diff
@@ -203,14 +204,16 @@ func (c *compositeTokenDataObserver) initTokenDataObservations(
...
+// In case of token data mismatch, it returns an error and the base observation.
 func merge(
 	base exectypes.TokenDataObservations,
 	from exectypes.TokenDataObservations,
 ) (exectypes.TokenDataObservations, error) {
 	for chainSelector, chainObservations := range from {
 		for seq, messageTokenData := range chainObservations {
 			if len(messageTokenData.TokenData) != len(base[chainSelector][seq].TokenData) {
-				return nil, errors.New("token data length mismatch")
+				return base, errors.New("token data length mismatch")
```
mateusz-sekara, line 216: Isn't that an antipattern for Go to return results together with error? How about returning nil here and in parent:

```go
merged, err = merge(tokenDataObservations, tokenData)
if err != nil {
	c.lggr.Error("Error while merging token data", "error", err)
        merged = tokenDataObservations
}
```

execute/observation.go
```diff
@@ -183,15 +186,27 @@ func readAllMessages(
...
+					"seqRange", seqRange,
+					"err", err,
+				)
+				// Fill range with empty messages.
+				for _, seqNum := range seqRange {
+					if _, ok := messageObs[srcChain]; !ok {
+						messageObs[srcChain] = make(map[cciptypes.SeqNum]cciptypes.Message)
+					}
+					messageObs[srcChain][seqNum] = cciptypes.Message{}
+				}
```
mateusz-sekara, line 209: I guess we need more tests covering these corner cases, no?

asoliman92: Yes, I'm still writing them.

execute/observation.go
```diff
@@ -222,27 +237,26 @@ func (p *Plugin) getMessagesObservation(
...
+	messageTimestamps := getMessageTimestampMap(commitReportCache, messageObs)
+
+	if len(messageObs) != len(messageTimestamps) {
+		p.lggr.Errorw("wans't able to get all message timestamps")
 	}
 
 	tkData, err1 := p.tokenDataObserver.Observe(ctx, messageObs)
 	if err1 != nil {
-		return exectypes.Observation{}, fmt.Errorf("unable to process token data %w", err1)
+		p.lggr.Errorw("unable to complete token data processing", "err", err1)
```
mateusz-sekara, line 250: Isn't that a place where we should stop processing? A lack of token-data means we could try executing messages requiring those. That gonna result in the onchain revert and wasting gas, no?

execute/observation.go
```diff
@@ -222,27 +237,26 @@ func (p *Plugin) getMessagesObservation(
...
 	}
+
 	if validateTokenDataObservations(messageObs, tkData) != nil {
 		return exectypes.Observation{}, fmt.Errorf("invalid token data observations")
 	}
 
 	costlyMessages, err := p.costlyMessageObserver.Observe(ctx, messageObs.Flatten(), messageTimestamps)
 	if err != nil {
-		return exectypes.Observation{}, fmt.Errorf("unable to observe costly messageObs %w", err)
+		p.lggr.Errorw("unable to observe costly messages", "err", err)
```
mateusz-sekara, line 259: Similar to the previous point. Doesn't it mean that we are trying to build and observation which would include messages that might be reverted or too costly to be executed?

asoliman92: But one costly message error will then cause no other messages to be executed at all.

execute/observation.go
asoliman92: The function doesn't return an error at all, this is to hold the interface. I can add `MustObserve` to the interface but don't think it's a good way as well. Maybe returning makes sense knowing that we shouldn't actually hit this return.

execute/observation.go
asoliman92: This seems wrong on a second thought. We need messages to have hashes for sure. if a message is not available we need to discard the whole commit report that's missing that message. working on it now.

execute/observation.go
```diff
@@ -169,11 +171,53 @@ func regroup(commitData []exectypes.CommitData) exectypes.CommitObservations {
 	return groupedCommits
 }
 
+// intersectCommitReportsAndMessages filters out commit reports that have no messages or have missing messages.
+// as all messages for each report are required to be present in the observation, otherwise merkle proofs will fail.
+func intersectCommitReportsAndMessages(msgObs exectypes.MessageObservations,
+	commitObs exectypes.CommitObservations,
+) (exectypes.MessageObservations, exectypes.CommitObservations) {
```
mateusz-sekara, line 178: nit, but wouldn't it be more readable if we insert instead of deleting? Just a pseudo code, might not compile

```go
func intersectCommitReportsAndMessages(
	msgObs exectypes.MessageObservations,
	commitObs exectypes.CommitObservations,
) (exectypes.MessageObservations, exectypes.CommitObservations) {
	filteredCommitReports := make(exectypes.CommitObservations)
	filteredMsgs := make(exectypes.MessageObservations)

	for srcChain, reports := range commitObs {
		filteredReports := make([]exectypes.CommitData, 0)

		// Check if there are any messages for the source chain
		if _, ok := msgObs[srcChain]; !ok {
			continue
		}

		if _, exists := filteredMsgs[srcChain]; !exists {
			filteredMsgs[srcChain] = make(map[uint64]map[cciptypes.SeqNum]cciptypes.Message)
		}

		for _, report := range reports {
			// Check if all sequence numbers in the report are valid
			for seq := report.SequenceNumberRange.Start(); seq <= report.SequenceNumberRange.End(); seq++ {
				if msg, ok := msgObs[srcChain][seq]; ok {
					filteredMsgs[srcChain][seq] = msg
				} else {
					break
				}
			}

			// Valid when length is the same
			if len(filteredMsgs[srcChain]) == len(msgObs[srcChain]) {
				filteredReports = append(filteredReports, report)
			}
		}

		if len(filteredReports) > 0 {
			filteredCommitReports[srcChain] = filteredReports
		}
	}

	return filteredMsgs, filteredCommitReports
}
```

execute/observation.go
mateusz-sekara: Great question, we need to ask the product what should be the behaviour here. If we can't compute `costlyMessages` then I'm assuming we are missing critical information, in my opinion it should stop the processing. Imagine the following scenario:

You are observing 3 messages, all sent from chain A to chain B.
After these 3 messages were sent, the destination had a huge gas spike. Right now, executing these 3 messages gonna cost 10x more gas than the fee. We don't know about that, because we failed to compute costly messages. I think we want to be as available as we can, but in this case, I think we don't want to lose money in the first place.

execute/observation.go
```diff
@@ -183,15 +227,21 @@ func readAllMessages(
...
+			continue
 		}
 
 		// Read messages for each range.
 		for _, seqRange := range ranges {
 			// TODO: check if srcChain is supported.
 			msgs, err := ccipReader.MsgsBetweenSeqNums(ctx, srcChain, seqRange)
 			if err != nil {
-				return exectypes.MessageObservations{}, err
+				lggr.Errorw("unable to read messages",
```
mateusz-sekara, line 239: 👍

execute/observation.go
asoliman92: That was my initial approach as well. The thing is that we will still need to delete all messages that were added for a specific report once we find an invalid message. So it didn't matter much and actually having the initial object ready and deleting from was clearer in that case.

execute/observation.go
```diff
@@ -222,40 +272,42 @@ func (p *Plugin) getMessagesObservation(
...
-	messageObs, err := readAllMessages(ctx, p.ccipReader, commitReportCache)
-	if err != nil {
-		return exectypes.Observation{}, err
-	}
+	messageObs := readAllMessages(ctx, p.lggr, p.ccipReader, commitReportCache)
 
-	messageTimestamps, err := getMessageTimestampMap(commitReportCache, messageObs)
-	if err != nil {
-		return exectypes.Observation{}, err
+	messageObs, filteredCommitReports := intersectCommitReportsAndMessages(messageObs, commitReportCache)
```
winder, line 277: Avoid creating a new variable, this way you don't have to worry about mixing them up later in the function.
```suggestion
	messageObs, commitReportCache = intersectCommitReportsAndMessages(messageObs, commitReportCache)
```

execute/observation.go
```diff
@@ -169,11 +171,53 @@ func regroup(commitData []exectypes.CommitData) exectypes.CommitObservations {
...
+					break
+				}
+			}
+			if valid {
+				filteredReports = append(filteredReports, report)
+			} else {
+				// remove messages that are not valid anymore
+				for seq := report.SequenceNumberRange.Start(); seq <= report.SequenceNumberRange.End(); seq++ {
+					delete(filteredMsgs[srcChain], seq)
+				}
```
winder, line 203: I don't think you need to remove the messages, they would have never been added in `readAllMessages`.

For clarity, I think it would make sense to combine this function with `readAllMessages`. 

In `for _, seqRange := range ranges {` you could use a two pointer algorithm to walk through the ranges and reports at the same time. Then return the updated CommitObservations along with the new MessageObservations.

execute/observation.go
winder: This error message could also correlate this to the merkle roots which we can't process, which could be helpful for debugging. Getting the roots would also be part of the suggestion above to combine `intersectCommitReportsAndMessages` with this function.

execute/observation.go
winder: If you don't combine these functions, please remove the empty line between them and add a comment saying something about why messages might be missing.

execute/observation.go
asoliman92: 🤔

execute/plugin_functions.go
```diff
@@ -599,13 +599,13 @@ func getConsensusObservation(
...
 	messages exectypes.MessageObservations,
-) (map[cciptypes.Bytes32]time.Time, error) {
+) map[cciptypes.Bytes32]time.Time {
 	messageTimestamps := make(map[cciptypes.Bytes32]time.Time)
 
 	for chainSel, SeqNumToMsg := range messages {
 		commitData, ok := commitReportCache[chainSel]
 		if !ok {
-			return nil, fmt.Errorf("missing commit data for chain %s", chainSel)
+			continue
```
winder, line 608: Log an error here?
```suggestion
			lggr.Errorf("missing commit data for chain %s, this should not happen if the arguments were initialized correctly") 
```

winder: I wonder if this function could also be folded into getAllMessages to avoid all the special setup

execute/observation.go
```diff
@@ -171,37 +173,46 @@ func regroup(commitData []exectypes.CommitData) exectypes.CommitObservations {
...
 
-		ranges, err := computeRanges(reports)
-		if err != nil {
-			return exectypes.MessageObservations{}, err
-		}
-
+		messageObs[srcChain] = make(map[cciptypes.SeqNum]cciptypes.Message)
 		// Read messages for each range.
-		for _, seqRange := range ranges {
-			// TODO: check if srcChain is supported.
```
mateusz-sekara, line 191: This check is still missing, but the comment is gone, right?

mateusz-sekara (approved):

PR #444, opened 2025-01-14
solana: add length prefixes for timelock operation id hashing

jadepark-dev:
This PR adds length prefixes to timelock operation ID hashing[NONEVM-1152]

## Primary Changes
- Added length prefixes when encoding instruction arrays, accounts, and data for operation ID hashing
- Updated hash operation functions to handle length-prefixed encoding in both contract and client implementations

## Others
- Added test cases to verify hash collision prevention for similar but distinct instruction patterns
- Enhanced error handling for encoding operations in Go implementation


chains/solana/contracts/programs/timelock/src/state/operation.rs
```diff
@@ -43,22 +43,33 @@ impl Operation {
         let total_size = self
             .instructions
             .iter()
-            .map(|ix_data: &InstructionData| ix_data.space())
+            .map(|ix_data: &InstructionData| {
+                // length prefixes for instruction array + accounts array + data
+                4 + 4 + 4 + ix_data.space()
```
agusaldasoro, line 48: what do you think of calling here it recursively to InstructionData::len ?
So you don't have logic specific of the fields here, you have it in the specific struct

agusaldasoro (approved):

chains/solana/contracts/programs/timelock/src/state/operation.rs
jadepark-dev: Thanks! that's more solid. And I fixed misplaced the size of prefixes as well

PR #445, opened 2025-01-14
Exec report builder options: max messages, max reports.

winder:
Add new OCR settings:
* MaxReportMessages: limit the number of messages per single chain report.
* MaxSingleChainReports: limit the number of single chain reports in the final execute report.

Fix a bug which could result in empty commit reports resulting in a mismatch between single chain reports and their associated commit reports.


execute/report/builder.go
```diff
@@ -148,15 +156,37 @@ func (b *execReportBuilder) Add(
...
+	// Check if limiting is required.
+	if b.maxSingleChainReports != 0 && uint64(len(b.execReports)) > b.maxSingleChainReports {
+		b.lggr.Infof(
+			"limiting number of reports to maxReports from %d to %d",
+			len(b.execReports),
+			b.maxSingleChainReports,
+		)
+		b.execReports = b.execReports[:b.maxSingleChainReports]
+		b.commitReports = b.commitReports[:b.maxSingleChainReports]
+	}
```
winder, line 184: Max single chain report limiting during final build.

I did this here because in the future this would generate multiple OCR reports. I'm imagining that would be done by analyzing a bunch of ExecutePluginReportSingleChain.

execute/plugin.go
```diff
@@ -254,16 +253,14 @@ func selectReport(
 			return nil, nil, fmt.Errorf("unable to add report to builder: %w", err)
 		}
 
-		selectedReports = append(selectedReports, commitReports[i])
```
winder, line 257: This line is buggy. It will include commit reports even if the result of `Add` is `ErrEmptyReport`, which leads to a mismatch between the returned `execReports` and `selectedReports` lists.

asoliman92: I don't get how it's going to include the commit report when there's an error? we have an error check before this. and in my other PR I'm doing continue which will also prevent it from reaching this part.

asoliman92: I see you already moved the selectedReports to the build function anyways. I just don't get the bug you're mentioning.

execute/report/builder.go
```diff
@@ -148,15 +156,37 @@ func (b *execReportBuilder) Add(
...
+	b.commitReports = append(b.commitReports, updatedReport)
 
 	return updatedReport, nil
 }
 
-func (b *execReportBuilder) Build() ([]cciptypes.ExecutePluginReportSingleChain, error) {
+func (b *execReportBuilder) Build() (
+	[]cciptypes.ExecutePluginReportSingleChain, []exectypes.CommitData, error,
+) {
+	if len(b.execReports) != len(b.commitReports) {
```
asoliman92, line 167: What happens in case one report had an error while adding it? should we take care of this?

execute/report/report.go
```diff
@@ -302,6 +302,23 @@ func CheckNonces(sendersNonce map[cciptypes.ChainSelector]map[string]uint64) Che
 	}
 }
 
+// checkMessages to get a set of which are ready to execute.
+func (b *execReportBuilder) checkMessages(ctx context.Context, report exectypes.CommitData) (map[int]struct{}, error) {
+	readyMessages := make(map[int]struct{})
+	for i := 0; i < len(report.Messages); i++ {
+		updatedReport, status, err := b.checkMessage(ctx, i, report)
+		if err != nil {
```
asoliman92, line 310: continue instead?

execute/report/report.go
```diff
@@ -401,43 +418,38 @@ func (b *execReportBuilder) buildSingleChainReport(
...
-	validReport, meta, err := b.verifyReport(ctx, finalReport)
-	if err != nil {
-		return cciptypes.ExecutePluginReportSingleChain{},
-			exectypes.CommitData{},
-			fmt.Errorf("unable to verify report: %w", err)
-	} else if validReport {
-		return finalize(finalReport, report, meta)
+		validReport, meta, err := b.verifyReport(ctx, finalReport)
+		if err != nil {
+			return cciptypes.ExecutePluginReportSingleChain{},
```
asoliman92, line 443: same, should we halt or continue and check the rest of reports?

execute/plugin.go
winder: The error is swallowed [by the Add function](https://github.com/smartcontractkit/chainlink-ccip/blob/3ba6bac6203a27189da9c2ecc82b83e0415a17bc/execute/report/builder.go#L143) to allow continuing with other reports in case there is enough space for a smaller message. So we end up hitting this line even when the report isn't really selected.

execute/report/report.go
winder: When there is an error in the check we don't execute any messages from the report. An error here is very unusual and indicates something more significant than a bad message. Currently, the only error happens due to a malformed commit report.

execute/report/report.go
winder: This is called within `Add`, we would still try adding the other reports if one of them has an error.

asoliman92 (approved):

PR #448, opened 2025-01-15
Feat/split receiver

agusaldasoro:
This pull request includes significant changes to the `ccip-router` program in the Solana blockchain, focusing on refactoring the receiver handling by splitting it into `token_receiver` and `logic_receiver`. The changes affect multiple files, including core logic, tests, and documentation.

These changes aim to improve the clarity and functionality of the `ccip-router` program by clearly distinguishing between the logic execution receiver and the token transfer receiver.


chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -538,72 +541,59 @@ fn internal_execute<'info>(
...
-// case: no tokens, but there are remaining_accounts passed in
-// case: tokens, but the first token has a non-zero index (indicating extra accounts before token accounts)
-fn should_execute_messaging(token_indexes: &[u8], remaining_accounts_empty: bool) -> bool {
-    (token_indexes.is_empty() && !remaining_accounts_empty)
-        || (!token_indexes.is_empty() && token_indexes[0] != 0)
+// should_execute_messaging checks if:
+// 1. There is at least one account used for messaging (the first subset of accounts). This is because the first account is the program id to do the CPI
+// 2. AND the logic_receiver has a value different than zeros
+fn should_execute_messaging(logic_receiver: &Pubkey, remaining_accounts_empty: bool) -> bool {
+    !remaining_accounts_empty && *logic_receiver != Pubkey::default()
```
agusaldasoro, line 548: Should we check here for the compute Units as well?

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -5298,10 +5300,10 @@ func TestCCIPRouter(t *testing.T) {
 				message, _ := testutils.CreateNextMessage(ctx, solanaGoClient, t)
 
 				// To make the message go through the validations we need to specify all additional accounts used when executing the CPI
-				message.ExtraArgs.IsWritableBitmap = 2 + 32 + 64 + 128
+				message.ExtraArgs.IsWritableBitmap = ccip.GenerateBitMapForIndexes([]int{0, 1, 5, 6, 7})
 				message.ExtraArgs.Accounts = []solana.PublicKey{
-					config.CcipReceiverProgram,
-					config.ReceiverTargetAccountPDA, // writable (index = 1)
+					config.ReceiverExternalExecutionConfigPDA, // writable (index = 0)
```
agusaldasoro, line 5305: It's weird why the tests did pass before this change 🤔

chains/solana/contracts/programs/ccip-router/src/messages.rs
```diff
@@ -59,10 +59,13 @@ pub struct Any2SVMRampMessage {
...
     pub sender: Vec<u8>,
     pub data: Vec<u8>,
-    // receiver is used as the target for the two main functionalities
-    // token transfers: recipient of token transfers (associated token addresses are validated against this address)
-    // arbitrary messaging: expected account in the declared arbitrary messaging accounts (2nd in the list of the accounts)
-    pub receiver: Pubkey,
+    // In EVM receiver means the address that all the listed tokens will transfer to and the address of the message execution.
+    // In Solana the receiver is split into two:
+    // Logic Receiver is the Program ID of the user's program that will execute the message
+    pub logic_receiver: Pubkey,
```
toblich, line 65: Wdyt about naming this `message_receiver` instead of `logic_receiver`? It feels closer to CCIP terminology in general, e.g. ccip use cases include "pure messaging" (thus `message_receiver`) vs "token transfer" vs "programmable token transfer" (which is the sum of the previous two)

Alternatively, using more of the Solana terminology, we could name it `cpi_receiver`

agusaldasoro: There is actually no term for CCIP to distinguish the receiver between logic and tokens as for EVM they are both the same.

Naming them `logic_receiver` and `token_receiver` makes it impossible to have ambiguity between the two terms. Someone could argue that any CPI call is a message receiver or cpi receiver, those are more generics names.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -537,72 +540,59 @@ fn internal_execute<'info>(
...
-/// * `accounts` - accounts passed via `ctx.remaining_accounts`. expected order is: [program, receiver, ...additional message accounts]
+/// * `logic_receiver` - receiver address from x-chain message, used to validate `remaining_accounts`
+/// * `extra_args_accounts` - arbitrary messaging accounts from the x-chain message, used to validate `accounts`.
+/// * `remaining_accounts` - accounts passed via `ctx.remaining_accounts`. expected order is: [program, receiver, ...additional message accounts]
 fn parse_messaging_accounts<'info>(
     token_indexes: &[u8],
-    receiver: Pubkey,
-    source_accounts: &[Pubkey],
+    logic_receiver: Pubkey,
+    extra_args_accounts: &[Pubkey],
```
PabloMansanet, line 562: I got rid of all of these [here](https://github.com/smartcontractkit/chainlink-ccip/pull/477), just as a heads up to make sure what we've done works well together (I don't see why not: I don't mind rebasing on you if this gets merged first)

PabloMansanet (commented):
Looks sensible to me, but I'm still unsure about duplications between the fields in SVM2AnyMessage and the entries of `remaining_accounts` . Seeing as at least the `logic_receiver` (and potentially also `token_receiver`? Not sure) needs to be in `remaining_accounts`, maybe we should not have any `Pubkey` in `Any2SVMRampMessage` and nip the duplication in the bud. WDYT?

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
agusaldasoro: This change is aming for clarity on the users and the protocol itself, it doesn't add any functionality. We need to establish an abstraction that is super clear for the users/offchain code what to send, and the receiver (token or logic) is the most sensitive part of the protocol, we can't have any missunderstanding or wrong configuration. We could only receive accounts and do all the validations, but this way we are sure that all the nodes have signed those addresses as logic and token receivers.

aalu1418 (approved):

PR #449, opened 2025-01-15
[CCIP-4827] Allow intermittent errors while merging during exec outcome phase. Skip failed ones and continue processing the rest.

asoliman92:
Allow intermittent errors while merging during exec outcome phase. Skip failed ones and continue processing the rest.

We want the plugin to be available and resilient enough that if something is missing for one chain it won't affect all other chains. Same if one observer's observation had issues, we don't need to stop other observations from being processed.

### DIFF
--- a/execute/outcome.go
+++ b/execute/outcome.go
@@ -34,7 +34,10 @@ func (p *Plugin) Outcome(
 		return nil, fmt.Errorf("unable to decode previous outcome: %w", err)
 	}
 
-	decodedAos := decodeAttributedObservations(p.lggr, aos)
+	decodedAos, err := decodeAttributedObservations(aos)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode observations: %w", err)
+	}
 
 	// discovery processor disabled by setting it to nil.
 	if p.discovery != nil {
--- a/execute/outcome.go
+++ b/execute/outcome.go
@@ -34,10 +34,7 @@ func (p *Plugin) Outcome(
 		return nil, fmt.Errorf("unable to decode previous outcome: %w", err)
 	}
 
-	decodedAos, err := decodeAttributedObservations(aos)
-	if err != nil {
-		return nil, fmt.Errorf("unable to decode observations: %w", err)
-	}
+	decodedAos := decodeAttributedObservations(p.lggr, aos)
 
 	// discovery processor disabled by setting it to nil.
 	if p.discovery != nil {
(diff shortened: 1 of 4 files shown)


execute/outcome.go
```diff
@@ -34,10 +34,7 @@ func (p *Plugin) Outcome(
 		return nil, fmt.Errorf("unable to decode previous outcome: %w", err)
 	}
 
-	decodedAos, err := decodeAttributedObservations(aos)
-	if err != nil {
-		return nil, fmt.Errorf("unable to decode observations: %w", err)
-	}
+	decodedAos := decodeAttributedObservations(p.lggr, aos)
```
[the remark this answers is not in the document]
asoliman92: In case a malicious node sends a broken observation. it's gonna be rare here but it's a defensive one anyways.

execute/plugin_functions.go
```diff
@@ -274,7 +283,8 @@ func mergeMessageObservations(
 		for selector, messages := range ao.Observation.Messages {
 			validator, ok := validators[selector]
 			if !ok {
-				return exectypes.MessageObservations{}, fmt.Errorf("no validator for chain %d", selector)
+				lggr.Warnw("no F defined for chain", "chain", selector)
+				continue
```
makramkd, line 287: Under what situation does this occur? A node makes an observation on a chain that has not enough observers?

makramkd: Similar question for the other similar checks

makramkd (changes requested):

execute/plugin_functions.go
asoliman92: Main scenario is if we didn't update home chain to have the FChain for a specific chain or it still have the old cache. We don't want that to stop processing for other chains. That might happen as well if we have a bug in homeChainReader and not all chains are being returned correctly. Still we don't want to halt the whole protocol because of this.

makramkd: > Main scenario is if we didn't update home chain to have the FChain for a specific chain 

Feels like this is something that should be detected in ValidateObservation by pretty much all honest nodes, nodes should not be reporting stuff from chains that are not in CCIPHome. If they are being reported its an invalid observation.

> still have the old cache

The cache should be busted pretty quick, at worst its a few missed rounds, but maybe I'm missing something else here, this doesn't seem like good enough justification?

asoliman92: you're not missing context. it's a question whether we allow the protocol to continue even with some bugs in other places or not.  so even if we shouldn't reach that state. if we reached it, should we stop the protocol completely?

makramkd (approved):

PR #451, opened 2025-01-15
Create single entrypoint for contract building

jlaveracll:


chains/solana/README.md
```diff
@@ -18,7 +18,7 @@
 go install github.com/gagliardetto/anchor-go@v0.2.3
 
 # build contracts + IDL
-anchor build
+make buils-contracts
```
agusaldasoro, line 21: typo:

```suggestion
make build-contracts
```

Check in other places also

chains/solana/Makefile
```diff
@@ -50,7 +50,7 @@ go-tests:
 
 .PHONY: build-contracts
 build-contracts:
-	cd ./contracts && anchor build
+	cd ./contracts && rm -rf ./target && anchor build && cd -
 
 .PHONY: solana-checks
-solana-checks: clippy anchor-go-gen format gomodtidy lint-go rust-tests go-tests build-contracts
+solana-checks: clippy format anchor-go-gen gomodtidy lint-go rust-tests go-tests
```
agusaldasoro, line 56: Please don't alter this line, as:
1. The go code needs first to be generated and formatted after
2. At the end we need to run `build-contracts` again as running the tests in go did add the program id in the idls

Maybe we should add a comment here explaining this

agusaldasoro (changes requested):

chains/solana/Makefile
jlaveracll: ah understood! fixed

chains/solana/README.md
jlaveracll: whoops, fixed!

agusaldasoro (approved):

PR #454, opened 2025-01-15
Refactor: Visibility changes for internal structs & move code around [NONEVM-1114]

toblich:
- Move non-interface struct `Any2SolanaMessage` into versioned module, as it was only used as an internal struct in the offramp. 
- Change versioned impl of interface struct `TokenAdminRegistry` to manage its writable flags into standalone functions in its own module
- Restrict the visibility to `pub(super)` for many internal structs, so that they aren't exposed in our IDL & the gobindings. It seems like Anchor, when generating the IDL, doesn't take into account that some of them are public inside of a private module only, and it just looks at the visibility keywords of the struct itself...
- Move OCR3-related errors & events to the public interface, out of the versioned module. They are emitted/returned under some OCR3-related situations, so they shouldn't be changed as easily as business logic


aalu1418 (approved):
lgtm! 

shouldn't be difficult if we need to republish types in the future

PR #455, opened 2025-01-15
ExtraArgs decoder Interfaces and plumbing

prashantkumar1982:
core ref: 7e1e41db83d63b990be5a04a4d6bd82011e5ca1e


pkg/types/ccipocr3/interfaces.go
```diff
@@ -18,6 +20,10 @@ type MessageHasher interface {
 	Hash(context.Context, Message) (Bytes32, error)
 }
 
+type ExtraDataCodec interface {
+	DecodeExtraData(ExtraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
```
winder, line 24: Might be useful to put the encode function in here as well and the hasher can use it too.

prashantkumar1982: As of now, we don't have a use-case for encoding the entire ExtraArgs.
The hasher encodes only specific fields from the full ExtraArgs, so the hasher actually needs to use this decoded type.

We can add encode later if needed. I suspect before we do it, we may refactor the overall Codec interface and change these anyways.

internal/mocks/extradatadecoder.go
```diff
@@ -0,0 +1,19 @@
+package mocks
```
winder, line 1: Use [.mockery.yml](https://github.com/smartcontractkit/chainlink-ccip/blob/main/.mockery.yaml) to generate a mock.

huangzhen1997 (approved):

winder (approved):

PR #457, opened 2025-01-16
Prevent early errors for some ccipreader functions

asoliman92:
Instead of using error groups while discovering contracts, use workgroup and log errors while continuing to check other contract addresses.

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -1276,9 +1276,12 @@ func (r *ccipChainReader) getOnRampDynamicConfigs(
 			if err != nil {
 				if errors.Is(err, contractreader.ErrNoBindings) {
 					// ErrNoBindings is an allowable error.
-					r.lggr.Infow("unable to lookup source fee quoters (onRamp dynamic config), this is expected during initialization", "err", err)
+					r.lggr.Infow(
+						"unable to lookup source fee quoters (onRamp dynamic config), "+
+							"this is expected during initialization", "err", err)
 				} else {
-					r.lggr.Errorw("unable to lookup source fee quoters (onRamp dynamic config)", "chain", chainSel, "err", err)
+					r.lggr.Errorw("unable to lookup source fee quoters (onRamp dynamic config)",
+						"chain", chainSel, "err", err)
 				}
 				return
 			}
@@ -1338,9 +1341,11 @@ func (r *ccipChainReader) getOnRampDestChainConfig(
 			if err != nil {
 				if errors.Is(err, contractreader.ErrNoBindings) {
 					// ErrNoBindings is an allowable error.
-					r.lggr.Infow("unable to lookup source routers (onRamp dest chain config), this is expected during initialization", "chain", chainSel, "err", err)
+					r.lggr.Infow("unable to lookup source routers (onRamp dest chain config), "+
+						"this is expected during initialization", "chain", chainSel, "err", err)
 				} else {
-					r.lggr.Errorw("unable to lookup source routers (onRamp dest chain config)", "chain", chainSel, "err", err)
+					r.lggr.Errorw("unable to lookup source routers (onRamp dest chain config)",
+						"chain", chainSel, "err", err)
 				}
 				return
 			}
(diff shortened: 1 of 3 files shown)


pkg/reader/ccip.go
```diff
@@ -167,7 +167,8 @@ func (r *ccipChainReader) CommitReportsGTETimestamp(
 				cciptypes.ChainSelector(mr.SourceChainSelector),
 			)
 			if err != nil {
-				return nil, fmt.Errorf("get onRamp address for selector %d: %w", mr.SourceChainSelector, err)
+				r.lggr.Errorw("get onRamp address for selector %d: %w", mr.SourceChainSelector, err)
```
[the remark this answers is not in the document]
asoliman92: That should be fine, we divide the returned report to multiple reports based on roots and their seqNrs. any missing root means that we don't have a report for it to execute later.

https://github.com/smartcontractkit/chainlink-ccip/blob/e16f5fec817ea397ba479f392343986754cc78e8/execute/plugin.go#L131-L144

https://github.com/smartcontractkit/chainlink-ccip/blob/e16f5fec817ea397ba479f392343986754cc78e8/execute/plugin_functions.go#L158-L175

pkg/reader/ccip.go
```diff
@@ -1258,22 +1242,24 @@ type getOnRampDynamicConfigResponse struct {
 func (r *ccipChainReader) getOnRampDynamicConfigs(
 	ctx context.Context,
 	srcChains []cciptypes.ChainSelector,
-) (map[cciptypes.ChainSelector]getOnRampDynamicConfigResponse, error) {
+) map[cciptypes.ChainSelector]getOnRampDynamicConfigResponse {
+	result := make(map[cciptypes.ChainSelector]getOnRampDynamicConfigResponse)
 	if err := validateExtendedReaderExistence(r.contractReaders, srcChains...); err != nil {
-		return nil, err
+		r.lggr.Errorw("validate extended reader existence", "err", err)
```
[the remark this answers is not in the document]
asoliman92: good call. I added a ticket to address this in all other functions later as well.

pkg/reader/ccip.go
```diff
@@ -1307,20 +1294,26 @@ func (r *ccipChainReader) getOnRampDynamicConfigs(
 				"chain", chainSel,
 				"resp", resp)
 			if err != nil {
-				return fmt.Errorf("get onramp dynamic config for chain %d: %w", chainSel, err)
+				if errors.Is(err, contractreader.ErrNoBindings) {
+					// ErrNoBindings is an allowable error.
```
mateusz-sekara, line 1298: Is it? Some context on why noBinding is an allowable error? Ideally in the code

pkg/reader/ccip.go
```diff
@@ -1307,20 +1294,26 @@ func (r *ccipChainReader) getOnRampDynamicConfigs(
...
-			return nil
-		})
+		}(chainSel)
 	}
 
-	if err := eg.Wait(); err != nil {
-		return nil, err
-	}
-	return result, nil
+	wg.Wait()
```
mateusz-sekara, line 1314: nit, but why not have it like it used to be, but logging error here instead? I mean
```go
	if err := eg.Wait(); err != nil {
              lggr.Error("something", err)
        }
        return result, nil
```

pkg/reader/ccip.go
```diff
@@ -1364,23 +1359,24 @@ func (r *ccipChainReader) getOnRampDestChainConfig(
 				&resp,
 			)
 			if err != nil {
-				return fmt.Errorf(
-					"get onramp dest chain config, source chain: %d: %w",
-					chainSel, err,
-				)
+				if errors.Is(err, contractreader.ErrNoBindings) {
+					// ErrNoBindings is an allowable error.
```
mateusz-sekara, line 1363: ditto

pkg/reader/ccip.go
asoliman92: It's just in the log 😁 , this is expected during initialization

pkg/reader/ccip.go
asoliman92: TBH I don't know the behavior when there are multiple errors for this, We want to get all the logs for all the errors hence I used wg and logging all errors for different chains.

mateusz-sekara (approved):

PR #458, opened 2025-01-16
pkg/logutil: ctx-aware logging, add seq nr to all commit logs

makramkd:
Ticket: https://smartcontract-it.atlassian.net/browse/CCIP-4877

In order to build more readable logs, we include the OCR sequence number in the context that is provided by libocr. This means that all downstream components, i.e the ccip reader, processors, etc. will be able to extract the sequence number and include it as a field in their loggers.

### DIFF
--- a/commit/plugin.go
+++ b/commit/plugin.go
@@ -254,12 +254,6 @@ func (p *Plugin) Observation(
 		return nil, fmt.Errorf("decode query: %w", err)
 	}
 
-	lggr = logger.With(
-		lggr,
-		"prevOutcome", prevOutcome,
-		"decodedQuery", decodedQ,
-	)
-
 	merkleRootObs, err := p.merkleRootProcessor.Observation(ctx, prevOutcome.MerkleRootOutcome, decodedQ.MerkleRootQuery)
 	if err != nil {
 		lggr.Errorw("get merkle root processor observation",
--- a/commit/plugin.go
+++ b/commit/plugin.go
@@ -86,7 +86,7 @@ func NewPlugin(
 	}
 
 	chainSupport := plugincommon.NewChainSupport(
-		logutil.WithContext(lggr, "ChainSupport"),
+		logutil.WithComponent(lggr, "ChainSupport"),
 		homeChain,
 		oracleIDToP2pID,
 		reportingCfg.OracleID,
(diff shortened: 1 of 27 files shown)


pkg/logutil/logutil.go
```diff
@@ -28,3 +30,54 @@ func WithPluginConstants(
...
+// It assumes that the given ctx has the sequence number set, if not the field will be set to
+// 0 in the log field.
+func LoggerWithSeqNr(
+	ctx context.Context,
+	lggr logger.Logger,
+	keyvals ...any,
+) logger.Logger {
+	return logger.With(
+		lggr,
+		append(keyvals, SeqNrLoggerKey, GetSeqNr(ctx))...,
```
winder, line 70: Is it useful to have the seqNr field if `GetSeqNr` returns zero?

makramkd: Initially I'd say yeah because it'll let us know where it isn't being correctly piped.

pkg/logutil/logutil.go
```diff
@@ -22,9 +26,69 @@ func WithPluginConstants(
...
+	componentLoggerKey    = "component"
+	pluginLoggerKey       = "plugin"
+	oracleIDLoggerKey     = "oracleID"
+	donIDLoggerKey        = "donID"
+	configDigestLoggerKey = "configDigest"
+
+	// ocrSeqNrKey refers to the OCR sequence number.
+	// Linter complains if this is a plain "string" type.
+	ocrSeqNrKey = ContextKey(ocrSeqNrLoggerKey)
+)
```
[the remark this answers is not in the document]
makramkd: Done

commit/plugin.go
```diff
@@ -250,66 +254,78 @@ func (p *Plugin) Observation(
 		return nil, fmt.Errorf("decode query: %w", err)
 	}
 
+	lggr = logger.With(
+		lggr,
+		"prevOutcome", prevOutcome,
+		"decodedQuery", decodedQ,
```
winder, line 260: I think we should make sure these are logged once at the start of our top level observation functions in an info/verbose log rather than in (potentially) multiple log statements. The outcome could be huge.

winder (approved):
I didn't review every log statement, but what I saw so far it looks like we might need to revisit some of them to tune the verbosity and size.

In general this looks like a solid step forward, so I'm OK with merging even with the possible flaws.

commit/plugin.go
makramkd: Will axe for now.

winder (approved):

PR #459, opened 2025-01-16
[CCIP-4865] fee boosting hardening

asoliman92:
* Add validation for relative boost per hour and use mathslib
* Add validation for fees and costs to make sure that they're positive
* Change ZeroFee/Cost implementations to return constants instead
* Use default 20% relative boosting in case the offchain config wasn't provided

core ref: 5ddb8792bb3b27515d18256597bc3e34e35f48e5


execute/costlymessages/costly_messages.go
```diff
@@ -117,10 +117,16 @@ func (o *observer) Observe(
 		if !ok {
 			return nil, fmt.Errorf("missing fee for message %s", msg.Header.MessageID)
 		}
+		if err := validatePositive(fee); err != nil {
+			return nil, fmt.Errorf("invalid fee for message %s: %w", msg.Header.MessageID, err)
```
makramkd, line 121: What happens if we return an error here? Is the fee boosting logic completely skipped for all msgs? Will those messages still get executed?

asoliman92: it follows current setup for missing fees now. we have a ticket specifically for making the plugin more resilient in case of missing/buggy fee found.

makramkd: Follow up: should we just log and continue here instead of skipping all msgs?

asoliman92: this can cause issues later on. this needs some careful thinking so later when we try actual sending it doesn't have unwanted behavior. That's what the ticket I mentioned needs to address.

makramkd (approved):

PR #460, opened 2025-01-16
[Solana][NONEVM-1147] Denominate ccip_send fees in LINK

PabloMansanet:
Adds some extra little bit of type safety with a wrapper around the anchor-serializable U256 for cross-chain communications.

### DIFF
--- a/chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
+++ b/chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
@@ -6,7 +6,7 @@ use anchor_spl::token_interface;
 use super::fee_quoter::{fee_for_msg, transfer_fee, wrap_native_sol};
 use super::messages::pools::{LockOrBurnInV1, LockOrBurnOutV1};
 use super::pools::{
-    calculate_token_pool_account_indices, interact_with_pool, transfer_token, u64_to_le_u256,
+    calculate_token_pool_account_indices, interact_with_pool, transfer_token,
     validate_and_parse_token_accounts, TokenAccounts, CCIP_LOCK_OR_BURN_V1_RET_BYTES,
 };
 
@@ -192,7 +192,8 @@ pub fn ccip_send<'info>(
         },
         extra_args,
         fee_token: message.fee_token,
-        fee_token_amount: fee.amount,
+        fee_token_amount: fee.amount.into(),
+        fee_value_juels: link_fee.amount.into(),
         token_amounts: vec![SVM2AnyTokenTransfer::default(); token_count],
     };
 
@@ -300,7 +301,7 @@ fn token_transfer(
         source_pool_address,
         dest_token_address: lock_or_burn_out_data.dest_token_address,
         extra_data,
-        amount: u64_to_le_u256(token_amount.amount), // pool on receiver chain handles decimals
+        amount: token_amount.amount.into(), // pool on receiver chain handles decimals
         dest_exec_data,
     })
 }
@@ -370,7 +371,11 @@ fn hash(msg: &SVM2AnyRampMessage) -> [u8; 32] {
         &header_sequence_number,
         &header_nonce,
         &msg.fee_token.to_bytes(),
-        &msg.fee_token_amount.to_be_bytes(),
+        // The cross-chain amounts are encoded in little endian, but
+        // this is irrelevant to the hashing function as long as both
+        // sides agree.
+        &msg.fee_token_amount.to_bytes(),
+        &msg.fee_value_juels.to_bytes(),
         // messaging
         &[msg.receiver.len() as u8],
         &msg.receiver,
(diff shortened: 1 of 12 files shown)


chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -369,7 +370,8 @@ fn hash(msg: &SVM2AnyRampMessage) -> [u8; 32] {
         &header_sequence_number,
         &header_nonce,
         &msg.fee_token.to_bytes(),
-        &msg.fee_token_amount.to_be_bytes(),
+        &msg.fee_token_amount.to_le_bytes(),
```
matYang, line 373: curious why this and fee value juels consistently use le bytes, aren't other places using be bytes?

PabloMansanet: Yeah, this was a bit confusing to me too :)

Essentially, as far as I know the encoding we use for the hashing function is arbitrary and is unrelated to EVM or solana endianness. Since I changed the underlying `fee_token_amount` type, this needed to change as well, but we could use whichever endianness is clearer.

I'll see if I can tweak this to be a bit more clear and/or add comments.

matYang: in that case I think big endian would be preferred, both Solidity and Golang use big endian by default

chains/solana/contracts/programs/ccip-router/src/messages.rs
```diff
@@ -126,7 +129,7 @@ pub struct Any2SVMTokenTransfer {
     // CCIP_LOCK_OR_BURN_V1_RET_BYTES bytes. If more data is required, the TokenTransferFeeConfig.destBytesOverhead
     // has to be set for the specific token.
     pub extra_data: Vec<u8>,
-    pub amount: [u8; 32], // LE encoded u256, any cross-chain token amounts are u256
+    pub amount: CrossChainAmount,
```
toblich, line 132: Based on the chat we just had about endianness with @matYang, I'd still leave the comment here making it explicit that this is little-endian. 

I think it makes sense to keep this LE to be consistent with the onramp token transfer as well, though this may require the offchain code to manually encode the field. Keeping it LE also makes it consistent with Borsh, which is how we handle ints/uints up to 128 bits. There's just no support for u256 out of the box, but I'd rather this remained consistent with how u128 is encoded (LE) and not that we tie this to golang's default (BE) but make it inconsistent with the rest of the numbers.

From [Borsh spec](https://borsh.io/):
>General principles:
> • integers are little endian;

and the following table:

Informal type | Rust EBNF * | Pseudocode
-- | -- | --
Integers | integer_type: ["u8" \| "u16" \| "u32" \| "u64" \| "u128" \| "i8" \| "i16" \| "i32" \| "i64" \| "i128" ] | little_endian(x)
Floats | float_type: ["f32" \| "f64" ] | err_if_nan(x)little_endian(x as integer_type)

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -370,7 +371,11 @@ fn hash(msg: &SVM2AnyRampMessage) -> [u8; 32] {
         &header_sequence_number,
         &header_nonce,
         &msg.fee_token.to_bytes(),
-        &msg.fee_token_amount.to_be_bytes(),
+        // The cross-chain amounts are encoded in little endian, but
+        // this is irrelevant to the hashing function as long as both
+        // sides agree.
+        &msg.fee_token_amount.to_bytes(),
+        &msg.fee_value_juels.to_bytes(),
```
toblich, line 378: agreed, we can use either encoding here (it's transparent to users and only affects our offchain code), whichever you think makes it easier for the offchain side or is most consistent with how we encode other fields is fine by me :+1:

toblich (approved):

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
toblich: Per https://github.com/smartcontractkit/chainlink-ccip/pull/460#discussion_r1922517197, it's completely fine by me if you want to change this to BE here

matYang (approved):

chains/solana/contracts/programs/ccip-router/src/messages.rs
PabloMansanet: Is it OK to keep the comment in the type definition? It's the first thing you see when looking at `CrossChainAmount`, so I feel like it looks a bit cleaner than repeating it on every usage of the type (removing that noise was part of the reason why I extracted it into a type).

PabloMansanet:
Thank you for the review guys: I'll merge as is and roll any comment improvements / potential endianness tweaks over to the next PR to keep things moving.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
PabloMansanet: After your explanation [here](https://github.com/smartcontractkit/chainlink-ccip/pull/460#discussion_r1922559601) I'd rather stay in LE to be internally consistent: Better to stay close to the solana/borsh assumptions I think.

PR #463, opened 2025-01-16
solana: align with keccak hashing

aalu1418:


chains/solana/utils/ccip/ccip_messages.go
```diff
@@ -19,7 +19,7 @@ import (
 var leafDomainSeparator = [32]byte{}
 
 func HashCommitReport(ctx [3][32]byte, report ccip_router.CommitInput) ([]byte, error) {
-	hash := sha256.New()
+	hash := sha3.NewLegacyKeccak256()
```
agusaldasoro, line 22: why legacy?

aalu1418: it's the one that MCMS also uses - the `sha3` package only exports keccak hashing via the legacy function

https://pkg.go.dev/golang.org/x/crypto/sha3#NewLegacyKeccak256

agusaldasoro (approved):

PR #464, opened 2025-01-16
CCIP-4873: Add RMN client metrics

rstout:
Adds client-side metrics for individual RMN calls. These calls are asynchronous and use RageP2P. To support this functionality, we introduce state to store inflight requests in this change. The inflight request store maps request IDs to the timestamp the request was made and the node ID of the RMN node the request was sent to.

When we get a response back for a specific request ID, we look it up in the inflight request store and mark the latency of the request. On a timeout of the entire RMN report flow, we iterate through the inflight request store and mark metrics for all inflight requests that have not received a response.

Metrics are tagged with the node ID, which allows us to correlate slow or failed requests with specific nodes.

### DIFF
--- a/commit/metrics/prom.go
+++ b/commit/metrics/prom.go
@@ -1,9 +1,12 @@
 package metrics
 
 import (
+	"strconv"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	sel "github.com/smartcontractkit/chain-selectors"
+
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 
 	"github.com/smartcontractkit/chainlink-ccip/commit/chainfee"
@@ -32,6 +35,7 @@ const (
 )
 
 var (
+	RequestLatencyBucketsMilliseconds     = []float64{5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000}
 	promMerkleProcessorObservationDetails = promauto.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "ccip_commit_merkle_processor_observation_components_sizes",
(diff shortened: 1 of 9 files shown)


commit/metrics/prom.go
```diff
@@ -153,6 +176,17 @@ func (p *PromReporter) TrackMerkleOutcome(outcome merkleroot.Outcome, state stri
 	}
 }
 
+func (p *PromReporter) TrackRmnReport(latency float64, destChainSelector uint64, success bool) {
+	destChainSelectorStr := strconv.FormatUint(destChainSelector, 10)
+	successStr := strconv.FormatBool(success)
+	p.merkleProcessorRmnReportHistogram.WithLabelValues(destChainSelectorStr, successStr).Observe(latency)
+}
+
+func (p *PromReporter) TrackRmnRequest(method string, latency float64, nodeID uint64, err string) {
```
carte7000, line 185: I wondering if this can cause an issue with the label cardinality if we label with the err. Unless the errors are bounded by used a know set of error code we risk have an unbounded label here?

carte7000 (approved):

commit/metrics/prom.go
rstout: Currently the set of errors is bounded, they're all statically defined (as opposed to created from converting arbitrary errors to string), so there's little risk of cardinality issues.

commit/metrics/prom.go
```diff
@@ -46,6 +49,22 @@ var (
...
+		},
+		[]string{"dest_chain_selector", "success"},
+	)
+	promRmnControllerRmnRequestLatency = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "ccip_commit_rmn_controller_rmn_request_latency",
+			Help:    "This metric tracks the client-observed latency of a single RMN request",
+			Buckets: prometheus.DefBuckets,
+		},
+		[]string{"method", "node_id", "error"},
```
mateusz-sekara, line 66: Other metrics have camelCase metrics, let's be consistent there

commit/metrics/prom.go
```diff
@@ -46,6 +49,22 @@ var (
 		},
 		[]string{"chainID", "state", "type"},
 	)
+	promMerkleProcessorRmnReportLatency = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "ccip_commit_merkle_processor_rmn_report_latency",
+			Help:    "This metric tracks the client-observed latency of building an full RMN report with signatures",
+			Buckets: prometheus.DefBuckets,
+		},
+		[]string{"dest_chain_selector", "success"},
```
mateusz-sekara, line 58: I would consider transforming `dest_chain_selector` to the destChainID. DestChainSelectors are very hard to reason about - they require an additional step to transform to chainID which are human-readable. Also other metrics rely on chain ids instead of selectors

commit/metrics/prom.go
```diff
@@ -46,6 +49,22 @@ var (
 		},
 		[]string{"chainID", "state", "type"},
 	)
+	promMerkleProcessorRmnReportLatency = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "ccip_commit_merkle_processor_rmn_report_latency",
+			Help:    "This metric tracks the client-observed latency of building an full RMN report with signatures",
+			Buckets: prometheus.DefBuckets,
```
mateusz-sekara, line 56: Most of the other metrics rely on buckets in millis instead of seconds. Mixing seconds with millis might lead to some confusion across devs. Ideally we should follow `prometheus.DefBuckets` everywhere, but considering that we have millis everywhere I would preferably follow a single standard.

https://github.com/smartcontractkit/chainlink/blob/78d548d6101afd7fc82c2568a3665b4b57d5b1eb/core/services/ocr2/plugins/promwrapper/plugin.go#L65
https://github.com/smartcontractkit/chainlink/blob/78d548d6101afd7fc82c2568a3665b4b57d5b1eb/core/services/functions/listener.go#L74
https://github.com/smartcontractkit/chainlink/blob/78d548d6101afd7fc82c2568a3665b4b57d5b1eb/common/txmgr/broadcaster.go#L50

mateusz-sekara (approved):

PR #470, opened 2025-01-17
[CCIP-4908] pkg/logutil: add ocr phase to log and ctx

makramkd:
Ticket: https://smartcontract-it.atlassian.net/browse/CCIP-4908

Add the OCR phase to the ctx and logger so that its always logged by components.


pkg/logutil/logutil.go
```diff
@@ -22,10 +22,22 @@ const (
...
+
+// OCR phase values, all loggers should use these to standardize output.
+const (
+	PhaseQuery          = "qry"
+	PhaseObservation    = "obs"
+	PhaseOutcome        = "otcm"
+	PhaseReports        = "rprt"
+	PhaseShouldAccept   = "sacc"
+	PhaseShouldTransmit = "strn"
 )
```
winder, line 41: Not sure about the shorthand. Consider a type with long/short forms.

makramkd: Logic behind the shorthand was to not contribute too much to log payload sizes

winder (approved):

PR #472, opened 2025-01-17
support for decoding destexecdata

huangzhen1997:
Add new function to the `extradatacodec` interface for parsing message `destexecdata`. The destExecData is used when Any2EVM or SVM2Any messages for storing uint32 dest chain gas prices in bytes. Because ABI uses big-endian encoding while Borsch uses little-endian encoding.

Corresponding change for chainlink repo: https://github.com/smartcontractkit/chainlink/pull/16016


pkg/types/ccipocr3/interfaces.go
```diff
@@ -21,7 +21,8 @@ type MessageHasher interface {
 }
 
 type ExtraDataCodec interface {
-	DecodeExtraData(ExtraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
+	DecodeExtraArgs(extraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
```
prashantkumar1982, line 24: nit: add comment on which exact field on the Message struct does this work on.

pkg/types/ccipocr3/interfaces.go
```diff
@@ -21,7 +21,8 @@ type MessageHasher interface {
...
+	DecodeDestExecData(destExecData Bytes, sourceChainSelector ChainSelector) (Bytes, error)
```
prashantkumar1982, line 25: To what format should this function decode to?
The input is []byte, and output is []byte, so its confusing.

Also, since this is []byte, it means theoretically a family can encode even a struct here, even though today EVM and SVM just use uint32. So I think, to handle any generic struct, we likely need the format similar to the ExtraArgsDecoded.

All the fields here in the generic Any2Any Message ideally should be in a format that is chain agnostic. So we can't have these in either source/destination specific coding.

Lets confirm with CCIP team if they agree here, or have other suggestions.

prashantkumar1982, line 25: nit: suggestion: function name rename to decodeTokenAmountDestExecData

makramkd:
> The destExecData is used when Any2EVM or SVM2Any messages for storing uint32 dest chain gas prices in bytes.

Hey @huangzhen1997 this doesn't make sense to me, why would we do this in ccip? I'm not aware of this as a feature or functionality.

pkg/types/ccipocr3/interfaces.go
makramkd: We need comments on this entire struct, please add doc comments so this is understandable to people who maintain this codebase

pkg/types/ccipocr3/interfaces.go
```diff
@@ -21,7 +21,8 @@ type MessageHasher interface {
 }
 
 type ExtraDataCodec interface {
-	DecodeExtraData(ExtraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
+	DecodeExtraArgs(extraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
+	DecodeTokenAmountDestExecData(destExecData Bytes, sourceChainSelector ChainSelector) (Bytes, error)
```
makramkd, line 25: Please add a comment here on what this is

makramkd (changes requested):

pkg/types/ccipocr3/interfaces.go
makramkd, line 25: Ah I see, I believe this is referring to : https://github.com/smartcontractkit/chainlink/blob/99860e4df73602187a926df4cc65c3919232d8f6/contracts/src/v0.8/ccip/libraries/Internal.sol#L230-L232

Lets make that very clear here that this is what this interface is doing.

pkg/reader/ccip.go
```diff
@@ -367,11 +367,19 @@ func (r *ccipChainReader) MsgsBetweenSeqNums(
...
+		msg.Message.ExtraArgsDecoded, err = r.extraDataCodec.DecodeExtraArgs(msg.Message.ExtraArgs, sourceChainSelector)
 		if err != nil {
 			return nil, fmt.Errorf("failed to decode ExtraArgs: %w", err)
 		}
 		msg.Message.Header.OnRamp = onRampAddress
+
+		for i, token := range msg.Message.TokenAmounts {
+			token.DestExecData, err = r.extraDataCodec.DecodeTokenAmountDestExecData(token.DestExecData, sourceChainSelector)
+			if err != nil {
+				return nil, fmt.Errorf("failed to decode DestExecData: %w", err)
```
makramkd, line 379:
```suggestion
				return nil, fmt.Errorf("failed to decode token amount destExecData (%v): %w", tokenAmount, err)
```

pkg/reader/ccip.go
```diff
@@ -367,11 +367,19 @@ func (r *ccipChainReader) MsgsBetweenSeqNums(
...
 		}
 
-		msg.Message.ExtraArgsDecoded, err = r.extraDataCodec.DecodeExtraData(msg.Message.ExtraArgs, sourceChainSelector)
+		msg.Message.ExtraArgsDecoded, err = r.extraDataCodec.DecodeExtraArgs(msg.Message.ExtraArgs, sourceChainSelector)
 		if err != nil {
 			return nil, fmt.Errorf("failed to decode ExtraArgs: %w", err)
 		}
 		msg.Message.Header.OnRamp = onRampAddress
+
+		for i, token := range msg.Message.TokenAmounts {
```
makramkd, line 376:
```suggestion
		for i, tokenAmount := range msg.Message.TokenAmounts {
```

pkg/types/ccipocr3/interfaces.go
winder: Could you ask the onchain team? They confirmed the top level ExtraArgs is an object. Is this an object also?

At a minimum, I expect that this type should be `any` or `map[string]any`.

pkg/types/ccipocr3/interfaces.go
huangzhen1997: Yes

huangzhen1997:
Not sure how to fix the CI test, it's using a different Chainlink repo I provided in PR description.

pkg/types/ccipocr3/interfaces.go
```diff
@@ -21,7 +21,10 @@ type MessageHasher interface {
 }
 
 type ExtraDataCodec interface {
-	DecodeExtraData(ExtraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
+	// DecodeExtraArgs reformat bytes into a chain agnostic map[string]any representation for extra args
+	DecodeExtraArgs(extraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
+	// DecodeTokenAmountDestExecData provide special treatment for DestExecData in token message depends on source chain
```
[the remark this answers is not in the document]
huangzhen1997: I'm waiting for feedback [here](https://chainlink-core.slack.com/archives/C0779R3AH8R/p1737499687872009) so I will have more details to provide in comment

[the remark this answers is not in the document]
makramkd: See the thread, posted a message

pkg/reader/ccip.go
```diff
@@ -367,11 +367,23 @@ func (r *ccipChainReader) MsgsBetweenSeqNums(
...
-		msg.Message.ExtraArgsDecoded, err = r.extraDataCodec.DecodeExtraData(msg.Message.ExtraArgs, sourceChainSelector)
+		msg.Message.ExtraArgsDecoded, err = r.extraDataCodec.DecodeExtraArgs(msg.Message.ExtraArgs, sourceChainSelector)
 		if err != nil {
 			return nil, fmt.Errorf("failed to decode ExtraArgs: %w", err)
 		}
+
 		msg.Message.Header.OnRamp = onRampAddress
+
+		for i, tokenAmount := range msg.Message.TokenAmounts {
+			tokenAmount.DestExecDataDecoded, err = r.extraDataCodec.DecodeTokenAmountDestExecData(
```
prashantkumar1982, line 378: nit: maybe you can directly do this:
msg.Message.TokenAmounts[i].DestExecDataDecoded, err = ...

prashantkumar1982 (approved):

prashantkumar1982 (approved):

jmank88 (approved):

skudasov (approved):

prashantkumar1982 (dismissed):

winder (approved):
lgtm

prashantkumar1982 (approved):

PR #473, opened 2025-01-19
Fee quoter review and nits

matYang:


chains/solana/contracts/programs/ccip-router/src/instructions/v1/fee_quoter.rs
```diff
@@ -32,7 +33,6 @@ pub const ANY_2_EVM_MESSAGE_FIXED_BYTES: U256 = U256::new(32 * 14);
 pub const ANY_2_EVM_MESSAGE_FIXED_BYTES_PER_TOKEN: U256 = U256::new(32 * ((2 * 3) + 3));
 
 pub fn fee_for_msg(
-    _dest_chain_selector: u64,
```
PabloMansanet, line 35: No idea why I kept this around 😅

PabloMansanet (dismissed):
LGTM pending confirmation from others about the endianness encoding situation.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/fee_quoter.rs
```diff
@@ -77,7 +77,8 @@ pub fn fee_for_msg(
     let execution_gas = gas_limit
         + U256::new(dest_chain.config.dest_gas_overhead as u128)
         + U256::new(message.data.len() as u128)
-            * U256::new(dest_chain.config.dest_gas_per_payload_byte as u128)
+            * (U256::new(dest_chain.config.dest_gas_per_payload_byte as u128)
+                + network_fee.transfer_bytes_overhead)
         + network_fee.transfer_gas;
```
toblich, line 83: Qq, I don't see in [this part of the solidity contract](https://github.com/smartcontractkit/ccip/blob/583f07c85a72bbd241c644e3094561d4969a4b11/contracts/src/v0.8/ccip/FeeQuoter.sol#L590-L591) something equivalent to this, is it maybe implemented in some branch that I should be looking at instead of `ccip-develop`?

My understanding with this change is that the math now looks like the following abbreviated formula:

```ExecGas = GasLimit + DestGasOverhead + MsgLen * (DestGasPerByte + TransferBytesOverhead) + TransferGas```

But my understanding from reading the attribute names here is that the newly added `network_fee.transfer_bytes_overhead` (`TransferBytesOverhead` in the previous formula) is a number of bytes, not gas, right? So wouldn't it have to be added to the `MsgLen` instead of the `DestGasPerByte`? Thus bringing the formula to

```ExecGas = GasLimit + DestGasOverhead + (MsgLen + TransferBytesOverhead) * DestGasPerByte + TransferGas```

File: chains/solana/contracts/programs/ccip-router/src/instructions/v1/fee_quoter.rs [hunk 3]
```diff
@@ -249,10 +250,10 @@ pub struct UnpackedDoubleU224 {
 impl From<&TimestampedPackedU224> for UnpackedDoubleU224 {
     fn from(packed: &TimestampedPackedU224) -> Self {
         let mut u128_buffer = [0u8; 16];
-        u128_buffer[2..16].clone_from_slice(&packed.value[14..]);
-        let high = u128::from_be_bytes(u128_buffer);
-        u128_buffer[2..16].clone_from_slice(&packed.value[..14]);
+        u128_buffer[2..16].clone_from_slice(&packed.value[14..]); // @TODO: byte alignment good?
```
toblich, line 253: Great catch that the high/low were getting swapped given that inputs are big-endian, thanks! :+1: 

I think the `TODO` can be removed, the byte alignment seems fine as far as I can tell
```suggestion
        u128_buffer[2..16].clone_from_slice(&packed.value[14..]);
```

toblich (commented):
The endianness-related changes look good to me, but one question regarding the change of the fee execution_gas formula

chains/solana/contracts/programs/ccip-router/src/instructions/v1/fee_quoter.rs
matYang: the latest impl of FeeQuoter.sol had been moved to chainlink repo: https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/ccip/FeeQuoter.sol#L621

but yeah good catch, made a typo here, meant to be ` (U256::new(message.data.len() as u128) + network_fee.transfer_bytes_overhead)
            * U256::new(dest_chain.config.dest_gas_per_payload_byte as u128)`

PabloMansanet: Good catch @toblich :)

PabloMansanet (approved):

aalu1418 (approved):

PR #476, opened 2025-01-20
CCIP-4574 Use Any instead of OR

dhaidashenko:
core ref: 1cc219caca76662777aedb9ebdb2bfe9d1bfa5a3


ilija42 (dismissed):

mateusz-sekara (commented):
USDC integration tests are failing, could that be because query doesn't work as it is supposed to?

mateusz-sekara (approved):

ilija42 (approved):

PR #477, opened 2025-01-21
[Solana][NONEVM-1028] Deduplicate accounts in execution

PabloMansanet:


agusaldasoro (changes requested):
We need to ensure that the hash of the message is the same offchain and onchain. If saving space for those values in the extra args message, it means that the offchain code will need to hash the message with those values in, and then remove them to send the message through the contract. Then the program will need to do the same but the other way out, adding them on the hash.
If you feel that the saved space is worth it we could do this change, but if not it feels like a possible UX concern of the protocol, as all the nodes will need to sign a message with those values in, but then the message sent through is different.

PabloMansanet:
CC @aalu1418 after our sync discussion yesterday on how to proceed with this 🙂

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -572,13 +567,16 @@ fn should_execute_messaging(logic_receiver: &Pubkey, remaining_accounts_empty: b
 ///
 /// # Arguments
 /// * `token_indexes` - start indexes of token pool accounts, used to determine ending index for arbitrary messaging accounts
-/// * `logic_receiver` - receiver address from x-chain message, used to validate `remaining_accounts`
 /// * `extra_args_accounts` - arbitrary messaging accounts from the x-chain message, used to validate `accounts`.
```
agusaldasoro, line 570:
```suggestion
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -572,13 +567,16 @@ fn should_execute_messaging(logic_receiver: &Pubkey, remaining_accounts_empty: b
...
-/// * `remaining_accounts` - accounts passed via `ctx.remaining_accounts`. expected order is: [program, receiver, ...additional message accounts]
+/// * `remaining_accounts` - accounts passed via `ctx.remaining_accounts`. expected order is: [logic_receiver, token_receiver, ...additional message accounts]
```
agusaldasoro, line 571:
```suggestion
/// * `remaining_accounts` - accounts passed via `ctx.remaining_accounts`. expected order is: [logic_receiver, ...additional message accounts]
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -572,13 +567,16 @@ fn should_execute_messaging(logic_receiver: &Pubkey, remaining_accounts_empty: b
...
+///
+/// # Return values
+//  * `logic_receiver` - In EVM receiver means the address that all the listed tokens will transfer to and the address of the message execution.
+//     In Solana the receiver is split into two:
+//     * Logic Receiver (returned here) is the Program ID of the user's program that will execute the message.
+//     * Token Receiver (not present here, part of Any2SVMRampMessage), is the address which the ATA will be calculated from.
```
agusaldasoro, line 577: This method does not include anything related to token transfers
```suggestion
/// # Return values
//  * `logic_receiver` is the Program ID of the user's program that will execute the message.
//  * `msg_accounts` the list of accounts used for the arbitrary execution
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -593,31 +591,18 @@ fn parse_messaging_accounts<'info>(
...
-                is_writable(source_bitmap, (i) as u8) == current_acc.is_writable,
-                CcipRouterError::InvalidInputs
-            );
-        }
+    // Validate that the bitmap corresponds to the individual writable flags
+    for (i, acc) in msg_accounts.iter().enumerate().skip(1) {
+        require!(
+            is_writable(source_bitmap, i as u8) == acc.is_writable,
+            CcipRouterError::InvalidInputs
+        );
```
agusaldasoro, line 602: This validation should be in the hash, here we should trust 100% on the remaining accounts

chains/solana/contracts/programs/ccip-router/src/messages.rs
```diff
@@ -39,14 +39,12 @@ pub struct ExecutionReportSingleChain {
 pub struct SVMExtraArgs {
     pub compute_units: u32,
     pub is_writable_bitmap: u64,
```
agusaldasoro, line 41:
```suggestion
    pub is_writable_bitmap: u64, // part of the message to avoid calculating it onchain
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -877,16 +859,13 @@ mod tests {
             extra_args: SVMExtraArgs {
                 compute_units: 1000,
                 is_writable_bitmap: 1,
-                accounts: vec![
```
agusaldasoro, line 880: I'd definitely add the accounts to the hash test as well so we can still have the comparision between go and rust.
Probably the `hash` method should receive the accounts as another param, so it can serialize it here?

agusaldasoro (changes requested):
Let's keep the serialization of the accounts inside the hash, with this change the hash values should remain the same.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
PabloMansanet: Hmm, I see. This test was preexistent though, was it incorrect before?

Another way to think about it: Is it possible that this information is simply redundant? Do we need to pass the bitmap separately when that information is already encoded in the accs? (which we'll make part of the hash anyway).

PabloMansanet: (discussed in daily)

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -562,8 +561,13 @@ fn get_token_accounts_for<'a>(
 // should_execute_messaging checks if:
 // 1. There is at least one account used for messaging (the first subset of accounts). This is because the first account is the program id to do the CPI
 // 2. AND the logic_receiver has a value different than zeros
```
agusaldasoro, line 563:
```suggestion
// There is at least one account used for messaging (the first subset of accounts). This is because the first account is the program id to do the CPI
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -562,8 +561,13 @@ fn get_token_accounts_for<'a>(
...
-fn should_execute_messaging(logic_receiver: &Pubkey, remaining_accounts_empty: bool) -> bool {
-    !remaining_accounts_empty && *logic_receiver != Pubkey::default()
+fn should_execute_messaging<'a>(
+    remaining_accounts: &'a [AccountInfo<'a>],
+    token_indices: &[u8],
+) -> bool {
+    // The first entry in the accounts corresponds to a token, which means there is no logic receiver
+    let only_tokens = token_indices.first().map(|i| *i == 0).unwrap_or_default();
```
agusaldasoro, line 569: does the `unwrap_or_default` defaults to false?

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -572,13 +576,13 @@ fn should_execute_messaging(logic_receiver: &Pubkey, remaining_accounts_empty: b
...
 /// # Arguments
 /// * `token_indexes` - start indexes of token pool accounts, used to determine ending index for arbitrary messaging accounts
-/// * `logic_receiver` - receiver address from x-chain message, used to validate `remaining_accounts`
-/// * `extra_args_accounts` - arbitrary messaging accounts from the x-chain message, used to validate `accounts`.
-/// * `remaining_accounts` - accounts passed via `ctx.remaining_accounts`. expected order is: [program, receiver, ...additional message accounts]
+/// * `remaining_accounts` - accounts passed via `ctx.remaining_accounts`. expected order is: [logic_receiver, ...additional message accounts]
+///
+/// # Return values
+//  * `logic_receiver` is the Program ID of the user's program that will execute the message.
+//  * `msg_accounts` the remaining list of accounts used for the arbitrary execution//
```
agusaldasoro, line 583:
```suggestion
//  * `msg_accounts` the remaining list of accounts used for the arbitrary execution
```

agusaldasoro (commented):
Excelent work!

Can we add tests (as I don't think we have that already covered) for:
- Sending the correct accounts but a wrong bitmap for writable
- Hashing some accounts different than the ones sent in the transaction

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
```diff
@@ -465,22 +463,17 @@ fn internal_execute<'info>(
     let message = Any2SVMMessage {
         message_id: execution_report.message.header.message_id,
         source_chain_selector: execution_report.source_chain_selector,
-        sender: execution_report.message.sender,
-        data: execution_report.message.data,
+        sender: execution_report.message.sender.clone(),
+        data: execution_report.message.data.clone(),
```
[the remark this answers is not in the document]
PabloMansanet: That's just because moving the `verify_merkle_root` later in the function causes a borrow of a partially moved value: without the clone, part of `execution_report` would've moved out here.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/offramp.rs
PabloMansanet: Yep it does! So if there are no token indices, this would be false.

PabloMansanet:
@agusaldasoro addressed your review comments and added the tests you mentioned, PTAL when you have a moment :)

aalu1418 (approved):

agusaldasoro (approved):
Awesome work!

PabloMansanet:
@agusaldasoro @aalu1418 resolved conflicts, feel free to merge right away when green :)

toblich (approved):

PR #481, opened 2025-01-21
solana: validate evm extra args

aalu1418:
validate passed in extra args against:
* max_per_msg_gas_limit
* enforce_out_of_order

### DIFF
--- a/chains/solana/contracts/tests/ccip/ccip_router_test.go
+++ b/chains/solana/contracts/tests/ccip/ccip_router_test.go
@@ -56,6 +56,7 @@ func TestCCIPRouter(t *testing.T) {
 	require.NoError(t, gerr)
 
 	var nonceEvmPDA solana.PublicKey
+	var nonceSvmPDA solana.PublicKey
 
 	// billing
 	type AccountsPerToken struct {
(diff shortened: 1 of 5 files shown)


jadepark-dev (approved):

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -2374,6 +2379,78 @@ func TestCCIPRouter(t *testing.T) {
 			require.Equal(t, uint64(3), ccipMessageSentEvent.Message.Header.Nonce)
 		})
 
+		t.Run("When gasLimit is too high, it fails", func(t *testing.T) {
```
PabloMansanet, line 2382: Echoing @toblich's comment on a PR of mine a while ago, this kind of validation can be captured with a Rust unit test rather than the comparatively heavier Go integration test, which I think will help us with iteration time (it's easy to run a single Rust test/group of tests, while the Go suite is more coupled and order-dependent).

I don't have a strong opinion and I wouldn't stop the merge on this, but I'd mildly prefer this to be a Rust unit test instead :).

PabloMansanet (approved):
LGTM! Just one minor comment on whether to use unit or integration tests to capture this behaviour.

chains/solana/contracts/tests/ccip/ccip_router_test.go
aalu1418: ah good point - i will need to do a bit of refactoring here anyway to support multiple types so I will move the tests in the next PR!

PR #482, opened 2025-01-22
Ensure no gaps in observed messages [CCIP-4929]

asoliman92:


execute/observation.go
```diff
@@ -221,6 +222,22 @@ func readAllMessages(
...
+	msgMap := make(map[cciptypes.SeqNum]struct{})
+	for _, msg := range msgs {
+		msgMap[msg.Header.SequenceNumber] = struct{}{}
+	}
+	// Check for gaps in sequence numbers.
+	for i := numberRange.Start(); i <= numberRange.End(); i++ {
+		if _, ok := msgMap[i]; !ok {
+			return true
+		}
+	}
```
makramkd, line 237: q: isn't it sufficient to check that `msg[i+1].SequenceNumber == msg[i].SequenceNumber+1`?

execute/observation.go
```diff
@@ -221,6 +222,22 @@ func readAllMessages(
 	return messageObs, availableReports, messageTimestamps
 }
 
+// gapInSequenceNumbers returns true if there is a gap messages sequence numbers.
+func gapInSequenceNumbers(msgs []cciptypes.Message, numberRange cciptypes.SeqNumRange) bool {
+	// Add msgs to a map to check for duplicates.
+	msgMap := make(map[cciptypes.SeqNum]struct{})
+	for _, msg := range msgs {
+		msgMap[msg.Header.SequenceNumber] = struct{}{}
```
makramkd, line 230: Did you intend to do a dup check here? don't see it being done

execute/plugin_functions.go
```diff
@@ -113,6 +113,20 @@ func validateObservedSequenceNumbers(
 					return fmt.Errorf("executed message %d not in observed range %v", seqNum, data.SequenceNumberRange)
 				}
 			}
+
+			// No gap in messages sequence numbers observed
+			// When there are no messages observed then it's okay to proceed as GetCommitReports state don't
+			// observe messages but only the merkle roots and sequence number ranges.
+			if len(data.Messages) != 0 && data.SequenceNumberRange.Length() != len(data.Messages) {
```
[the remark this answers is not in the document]
asoliman92: I mentioned it in the comment above. Different states have different values. GetCommitReports state doesn't have any messages but have seqNrs

execute/observation.go
asoliman92: Initially yes, not gonna do it as we already check for messages length in the main function and then check that all seqNrs are observed here.

execute/observation.go
makramkd: I suggest we have a standard "gap checking" algorithm, seems like this is a variation of the other one that doesn't actually do a gap check but a subset check, i.e it checks that `set(seqNrRange) ⊆ set(msgs.SeqNr)` which is not exactly the same.

asoliman92: checking the length of messages should be enough as a previous condition to the current algorithm. Which I do in the earlier function. Let me extract the length into this one reuse it.

PR #488, opened 2025-01-22
solana: support any extra args on ccip_send

aalu1418:
- [x] extra arg handling
- [x] address validation
- [x] test case fixes


chains/solana/contracts/programs/ccip-router/README.md
```diff
@@ -17,10 +17,9 @@ Collapsed Router + OnRamp + OffRamp Contracts Implementation for CCIP in SVM.
```
agusaldasoro, line 17: Is this readme still valid? I remember we used to write all the TODOs here, but I don't know if that's still the case

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
```diff
@@ -0,0 +1,56 @@
...
+        let mut buffer = EVM_EXTRA_ARGS_V2_TAG.to_be_bytes().to_vec();
+        let mut data = self.try_to_vec().unwrap();
+
+        buffer.append(&mut data);
+        buffer
+    }
+    pub fn default_config(cfg: &DestChainConfig) -> EVMExtraArgsV2 {
+        EVMExtraArgsV2 {
+            gas_limit: cfg.default_tx_gas_limit as u128,
+            ..Default::default()
```
agusaldasoro, line 28: What does this line do? Doesn't the DestChainConfig have the default value for OOO as well?

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+}
+
+#[derive(Clone, AnchorSerialize, AnchorDeserialize, Default)]
+pub struct SVMExtraArgsV1 {
+    pub compute_units: u32,
```
agusaldasoro, line 35: Are we doing something with this value? Or is this just for the offramp to know if they may need to increase the compute units?

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub account_is_writable_bitmap: u64,
+    pub allow_out_of_order_execution: bool,
```
agusaldasoro, line 37: I think that this field can be removed? As it always will be out of order

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub token_receiver: [u8; 32],
+    pub accounts: Vec<[u8; 32]>,
```
agusaldasoro, line 39: Is this correct? Shouldn't this be part of the transaction in additional accounts?

chains/solana/contracts/programs/ccip-router/src/instructions/v1/fee_quoter.rs
```diff
@@ -38,7 +38,7 @@ pub fn fee_for_msg(
     fee_token_config: &BillingTokenConfig,
     additional_token_configs: &[Option<BillingTokenConfig>],
     additional_token_configs_for_dest_chain: &[PerChainPerTokenConfig],
-) -> Result<SVMTokenAmount> {
+) -> Result<(SVMTokenAmount, Vec<u8>, bool)> {
```
agusaldasoro, line 41: Can the result be a struct instead of a tuple? It will be more easy to read and follow. Also, can you add docs here explaining what the response of this message?

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -307,24 +305,87 @@ fn token_transfer(
...
 
-    if let Some(gas_limit) = extra_args.gas_limit {
-        gas_limit.clone_into(&mut result_args.gas_limit)
+// process_extra_args returns serialized extraArgs, gas_limit, allow_out_of_order_execution
+// it calls the chain-specific extra args validation logic
+pub fn process_extra_args(
+    dest_config: &DestChainConfig,
+    extra_args: &[u8],
+    message_contains_tokens: bool,
+) -> Result<(Vec<u8>, u128, bool)> {
```
agusaldasoro, line 315: Can the result be a struct instead of a tuple? It will be more easy to read and follow.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -307,24 +305,87 @@ fn token_transfer(
...
+    require_gte!(extra_args.len(), 4, CcipRouterError::InvalidInputs);
+
+    let tag: [u8; 4] = extra_args[..4].try_into().unwrap();
+    let mut data = &extra_args[4..];
+
+    match u32::from_be_bytes(dest_config.chain_family_selector) {
+        CHAIN_FAMILY_SELECTOR_EVM => parse_and_validate_evm_extra_args(dest_config, tag, &mut data),
+        CHAIN_FAMILY_SELECTOR_SVM => {
+            parse_and_validate_svm_extra_args(dest_config, tag, &mut data, message_contains_tokens)
```
agusaldasoro, line 324: Are we doing this twice? https://github.com/smartcontractkit/chainlink-ccip/pull/488/files#diff-35c2605216d31b008c6dc84b39a597abf74b4be60ca2d0e43ed443d749b64145R152

chains/solana/contracts/programs/ccip-router/README.md
aalu1418: not sure - but i can go through and remove things since it's probably out of date

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
aalu1418: not quite, the DestChain config has `EnforceOutOfOrder` to make sure that the user sets out of order - otherwise, we don't validate

and on the EVM side, if `EnforceOutOfOrder = true` then it requires the user to explicitly set it rather than use a default

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
aalu1418: correct, this would be for the offchain side

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
aalu1418: this aligns with the SVM ExtraArgs that are defined in the solidity contracts since this is just for the ccipSend side

(also see previous comment about requiring users to explicitly allow ooo processing even if it's required)

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
aalu1418: i should add a comment that these ExtraArgs are only for SVM -> XXX on the onramp (it has no connection to the offramp)

so the extra accounts would be passed in explicitly for the remote SVM chain

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
aalu1418: the switch case is defined twice, yes

one function is for parsing/validating extra args, the other is for validating the receiver address

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
agusaldasoro: so we need it here for the hashing?

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
agusaldasoro: Yes, now I get OOO

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
aalu1418: this would be for sending from SVM -> SVM, but all extraArgs should be encoded as `bytes` during hashing

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
```diff
@@ -32,11 +35,11 @@ impl EVMExtraArgsV2 {
 
 #[derive(Clone, AnchorSerialize, AnchorDeserialize, Default)]
 pub struct SVMExtraArgsV1 {
```
agusaldasoro, line 37:
```suggestion
// Extra Args of message when SVM -> SVM
#[derive(Clone, AnchorSerialize, AnchorDeserialize, Default)]
pub struct SVMExtraArgsV1 {
```

chains/solana/contracts/programs/ccip-router/src/extra_args.rs
```diff
@@ -10,8 +13,8 @@ pub const SVM_EXTRA_ARGS_V1_TAG: u32 = 0x1f3b3aba;
 
 #[derive(Clone, AnchorSerialize, AnchorDeserialize, Default)]
 pub struct EVMExtraArgsV2 {
```
agusaldasoro, line 15:
```suggestion
// Extra Args of message when SVM -> EVM 
#[derive(Clone, AnchorSerialize, AnchorDeserialize, Default)]
pub struct EVMExtraArgsV2 {
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -306,13 +310,21 @@ fn token_transfer(
 // Helpers //
 /////////////
 
+// ProcessedExtraArgs contains serialized extra args and unpacks commonly used parameters
+#[derive(Debug)]
+pub struct ProcessedExtraArgs {
```
agusaldasoro, line 315: Is this only for EVM should it declare it in the type?

aalu1418: no it's chain agnostic - it has the serialized bytes, and two parameters that are checked for all chains (gas_limit + ooo)

PR #489, opened 2025-01-23
solana: onchain multiple tokens test

aalu1418:
- [x] ccipSend
- [x] commit + execute


chains/solana/utils/common/transactions.go
```diff
@@ -175,6 +175,7 @@ func sendTransactionWithLookupTables(ctx context.Context, rpcClient *rpc.Client,
 
 	txsig, err := rpcClient.SendTransactionWithOpts(ctx, tx, rpc.TransactionOpts{SkipPreflight: skipPreflight, PreflightCommitment: rpc.CommitmentProcessed})
 	if err != nil {
+		fmt.Println(tx) // debugging if tx errors
```
toblich, line 178: Doesn't this pollute logs too much for tests that expect errors?

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -2738,98 +2780,177 @@ func TestCCIPRouter(t *testing.T) {
...
+					Receiver: validReceiverAddress[:],
+					Data:     []byte{4, 5, 6},
+					TokenAmounts: []ccip_router.SVMTokenAmount{
+						{
+							Token:  token0.Mint.PublicKey(),
+							Amount: 1,
+						},
+						{
+							Token:  token1.Mint.PublicKey(),
+							Amount: 1,
```
toblich, line 2897: nit: I generally try not to repeat values in tests, so that you can implicitly test that they are not getting swapped with no additional effort, so I'd suggest using different amounts in each token 🙂

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -5037,108 +5158,213 @@ func TestCCIPRouter(t *testing.T) {
...
+					message.Data = []byte{}
+					message.TokenReceiver = config.ReceiverExternalExecutionConfigPDA
+					message.TokenAmounts = []ccip_router.Any2SVMTokenTransfer{{
+						SourcePoolAddress: []byte{1, 2, 3},
+						DestTokenAddress:  token0.Mint.PublicKey(),
+						Amount:            ccip_router.CrossChainAmount{LeBytes: tokens.ToLittleEndianU256(1)},
+					}, {
+						SourcePoolAddress: []byte{4, 5, 6},
+						DestTokenAddress:  token1.Mint.PublicKey(),
+						Amount:            ccip_router.CrossChainAmount{LeBytes: tokens.ToLittleEndianU256(1)},
```
toblich, line 5287: nit: same as before regarding amounts 🙂

toblich (approved):
LGTM, just a couple of nitpick comments

chains/solana/utils/common/transactions.go
aalu1418: i don't think so? it would be an RPC error which is a different method - everything else would not expect errors on the initial RPC call, it would expect onchain reverts

toblich (approved):

PR #490, opened 2025-01-23
Pectra EIP-7623 support

matYang:
Support the new calldata fee model introduced in EIP 7623

Reference implementation in Solidity:
https://github.com/smartcontractkit/chainlink/pull/15904

Full context:
https://chainlink-core.slack.com/archives/C02PBKHCP28/p1735750292998669


rstout (dismissed):

chains/solana/contracts/programs/ccip-router/src/instructions/v1/fee_quoter.rs
```diff
@@ -74,27 +74,48 @@ pub fn fee_for_msg(
...
+    // Calculate calldata gas cost while accounting for EIP-7623 variable calldata gas pricing
+    // This logic works for EVMs post Pectra upgrade, while being backwards compatible with pre-Pectra EVMs.
+    let calldata_gas = if calldata_length
+        > U256::new(dest_chain.config.dest_gas_per_payload_byte_threshold as u128)
+    {
+        U256::new(dest_chain.config.dest_gas_per_payload_byte_base as u128)
+            * U256::new(dest_chain.config.dest_gas_per_payload_byte_threshold as u128)
+            + (calldata_length
+                - U256::new(dest_chain.config.dest_gas_per_payload_byte_threshold as u128))
+                * U256::new(dest_chain.config.dest_gas_per_payload_byte_high as u128)
```
PabloMansanet, line 89: Nit: I know this follows the reference EVM implementation verbatim, but I do wonder if it would be possible to find meaningful names for these individual calculations, so we can extract them into variables.

Currently this nested calculation is a bit difficult to parse visually, so it would be good to give each factor a variable.

PabloMansanet (approved):
I've taken a look at the EVM implementation and it seems to match 1:1, so on that note, 👍. I do think we could do with a bit of clarity by extracting variables from the calculation, but I wouldn't stop the merge on it!

chains/solana/contracts/programs/ccip-router/src/instructions/v1/fee_quoter.rs
toblich: +1, it looks like it matches EVM but I also find it somewhat hard to read as-is

toblich (approved):

winder (approved):

PR #491, opened 2025-01-23
CCIP-4951 Reduce polling by RMNHome background routine

mateusz-sekara:
core ref: 08ed8f5153ffb07bd9b0b7feb2f2a2d23a84ac0a

RMNChainPoller is created on every OCR3 initialization [chainlink-ccip/commit/factory.go](https://github.com/smartcontractkit/chainlink-ccip/blob/23551e14342ad197dcd9aadb3d0c344e8d331891/commit/factory.go#L165) 

Although there is always a single `RMNHome.sol` contract (similar to CCIPHome), we keep spawning background routines for every Commit job, which frequently poll the same contract. Unfortunately, OCR3Config allows defining RMNHomeAddres per chain so we can't follow a similar pattern as CCIPHomeChain.

Therefore, we need to implement our own singleton (per address) to support multiple RMNHomes

### DIFF
--- a/pkg/reader/helpers.go
+++ b/pkg/reader/helpers.go
@@ -16,7 +16,7 @@ import (
 const (
 	// HomeChainPollingInterval is the interval at which the home chain is polled for updates.
 	// It should be used by RMNHome and CCIPHome to poll the home chain for updates.
-	// Ethereum was selected for the home chain for CCIP, therefore pooling more frequent
+	// Ethereum was selected for the home chain for CCIP, therefore polling more frequent
 	// than block time doesn't bring any value.
 	// We selected 15 seconds for simplicity, but this could be extended even further as
 	// we accept some delay when fetching the configuration updates.
--- a/pkg/reader/helpers.go
+++ b/pkg/reader/helpers.go
@@ -3,6 +3,7 @@ package reader
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
@@ -10,6 +11,17 @@ import (
 	typeconv "github.com/smartcontractkit/chainlink-ccip/internal/libs/typeconv"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/contractreader"
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+)
+
+const (
+	// HomeChainPollingInterval is the interval at which the home chain is polled for updates.
+	// It should be used by RMNHome and CCIPHome to poll the home chain for updates.
+	// Ethereum was selected for the home chain for CCIP, therefore pooling more frequent
+	// than block time doesn't bring any value.
+	// We selected 15 seconds for simplicity, but this could be extended even further as
+	// we accept some delay when fetching the configuration updates.
+	// It's advised to use wrap that interval with some jitter to avoid congestion.
+	HomeChainPollingInterval = 15 * time.Second
 )
 
 // bindable is a helper interface to represent all different types of contract readers.
--- a/pkg/reader/rmn_home.go
+++ b/pkg/reader/rmn_home.go
@@ -97,7 +97,7 @@ func GetRMNHomePoller(
 	hexEncodedAddr := " 0x" + hex.EncodeToString(rmnHomeAddress)
 	key := fmt.Sprintf("%s-%s", rmnHomeChainSelector.String(), hexEncodedAddr)
 
-	instance, ok := instances[key]
+	instance, ok := getInstance(key)
 	if ok {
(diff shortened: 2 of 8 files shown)


pkg/reader/rmn_home.go
```diff
@@ -69,12 +75,80 @@ type rmnHomePoller struct {
 	pollingDuration      time.Duration // How frequently the poller fetches the chain configs
 }
 
-func NewRMNHomePoller(
+// GetRMNHomePoller returns a rmnHomePoller instance if it already exists, else creates a new one.
+// Returned rmnHomePoller is always started, therefore user of that API doesn't need to care about
+// concurrency and accidentally starting the same poller multiple times.
```
makramkd, line 80: This is a small nit but I wonder if we can return an interface that doesn't have `Start`?

pkg/reader/rmn_home.go
```diff
@@ -69,12 +75,80 @@ type rmnHomePoller struct {
...
+	}
+
+	rmnHomeReader := newRMNHomePoller(
+		contractReader,
+		rmnHomeBoundContract,
+		logutil.WithComponent(lggr, "RMNHomePoller"),
+		pollingInterval,
+	)
+
+	if err := rmnHomeReader.Start(ctx); err != nil {
```
makramkd, line 125: How will closing happen? I think right now every plugin is closing their own RMN home poller instance

makramkd: A sort of dirty approach is to use https://pkg.go.dev/runtime#SetFinalizer if we don't have clear ownership semantics - otherwise maybe in the chainlink node delegate?

pkg/reader/rmn_home.go
makramkd: Also maybe no `Close` either

pkg/reader/rmn_home.go
mateusz-sekara: Oh shoot, we need some sort of counting references here. Therefore close only when every commit holding reference is closed

mateusz-sekara: Btw, the bug is already there, because we never close `rmnHomeReader`

makramkd: Re: counting references thats what `SetFinalizer` does, might be less error prone

pkg/reader/rmn_home_poller_test.go
```diff
@@ -0,0 +1,475 @@
+package reader
```
mateusz-sekara, line 1: no changes here, only file rename

pkg/reader/helpers.go
```diff
@@ -12,6 +13,17 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
+const (
+	// HomeChainPollingInterval is the interval at which the home chain is polled for updates.
+	// It should be used by RMNHome and CCIPHome to poll the home chain for updates.
+	// Ethereum was selected for the home chain for CCIP, therefore pooling more frequent
```
makramkd, line 19:
```suggestion
	// Ethereum was selected for the home chain for CCIP, therefore polling more frequent
```

pkg/reader/rmn_home_poller.go
```diff
@@ -0,0 +1,386 @@
...
+			SourceChainF:   homeFMap,
+			ConfigDigest:   versionedConfig.ConfigDigest,
+			OffchainConfig: versionedConfig.DynamicConfig.OffchainConfig,
+		}
+	}
+	return rmnHomeConfigs
+}
+
+// IsNodeObserver checks if a node is an observer for the given source chain
+func IsNodeObserver(sourceChain SourceChain, nodeIndex int, totalNodes int) (bool, error) {
```
makramkd, line 362: This seems to only be used internally in this package, we can make it package-private

pkg/reader/rmn_home_poller.go
```diff
@@ -0,0 +1,386 @@
...
+
+func (r *rmnHomePoller) Ready() error {
+	return r.sync.Ready()
+}
+
+func (r *rmnHomePoller) HealthReport() map[string]error {
+	f := r.failedPolls.Load()
+	if f >= maxFailedPolls {
+		err := fmt.Errorf("polling failed %d times in a row (maximum allowed: %d)", f, maxFailedPolls)
+		r.sync.SvcErrBuffer.Append(err)
```
makramkd, line 199: I wonder if this err buffer needs to be cleared before appending? Could this be a potential memory leak situation?

makramkd (approved):
🚀

pkg/reader/rmn_home_poller.go
makramkd: Actually no need, the `Healthy` call below will call `Flush` on the buffer and clear it

makramkd (approved):

0xAustinWang (approved):

PR #492, opened 2025-01-23
[Solana][NONEVM-1107] Expand remote pool addresses to array

PabloMansanet:
Support for remote pool upgrades with in-flight messages.


chains/solana/contracts/programs/token-pool/src/lib.rs
```diff
@@ -512,13 +522,32 @@ pub struct ChainConfig {
 
 #[derive(InitSpace, Clone, AnchorSerialize, AnchorDeserialize)]
 pub struct RemoteConfig {
-    #[max_len(64)]
-    pub pool_address: Vec<u8>,
-    #[max_len(64)]
-    pub token_address: Vec<u8>,
+    #[max_len(32)]
```
PabloMansanet, line 525: I chose this arbitrarily, please sanity check @aalu1418 or @agusaldasoro

agusaldasoro: Ideally this vector should be unbounded

PabloMansanet: If I understand the Solana documentation correctly, this wouldn't be limiting: We'd just need to call `realloc` when they need to grow past this initial reserved size.

PabloMansanet: I've made it unbounded now after our discussions :)

aalu1418 (approved):

PR #493, opened 2025-01-23
Observe FChain in exec plugin and use it in outcome phase [CCIP-4876]

asoliman92:
core ref: 08ed8f5153ffb07bd9b0b7feb2f2a2d23a84ac0a

### DIFF
--- a/execute/outcome.go
+++ b/execute/outcome.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/consensus"
+
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 
 	"github.com/smartcontractkit/chainlink-ccip/execute/internal"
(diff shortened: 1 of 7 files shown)


execute/outcome.go
```diff
@@ -66,10 +68,17 @@ func (p *Plugin) Outcome(
...
+			observedFChains[chain] = append(observedFChains[chain], f)
+		}
 	}
+	// consensus on the fChain map uses the role DON F value
+	// because all nodes can observe the home chain.
+	donThresh := consensus.MakeConstantThreshold[cciptypes.ChainSelector](consensus.TwoFPlus1(p.reportingCfg.F))
+	fChain := consensus.GetConsensusMap(lggr, "fChain", observedFChains, donThresh)
 	_, ok := fChain[p.destChain] // check if the destination chain is in the FChain.
 	if !ok {
 		return ocr3types.Outcome{}, fmt.Errorf("destination chain %d is not in FChain", p.destChain)
```
makramkd, line 84: Can we move all this logic to `getConsensusObservation` since that's technically part of getting the consensus observation?

PR #495, opened 2025-01-23
Fix MinObservation to make observations per observer unique

asoliman92:
This is still work in progress.

### DIFF
--- a/execute/plugin_functions.go
+++ b/execute/plugin_functions.go
@@ -3,6 +3,7 @@ package execute
 import (
 	"errors"
 	"fmt"
+	"github.com/smartcontractkit/libocr/commontypes"
 	"sort"
 
 	mapset "github.com/deckarep/golang-set/v2"
@@ -311,7 +312,7 @@ func mergeMessageObservations(
 			}
 			// Add reports
 			for _, msg := range messages {
-				validator.Add(msg)
+				validator.Add(msg, ao.OracleID)
 			}
 		}
 	}
@@ -358,7 +359,7 @@ func mergeCommitObservations(
 			}
 			// Add reports
 			for _, commitReport := range commitReports {
-				validator.Add(commitReport)
+				validator.Add(commitReport, ao.OracleID)
 			}
 		}
 	}
@@ -407,14 +408,14 @@ func mergeMessageHashes(
 					validators[selector][seqNr] =
 						consensus.NewMinObservation[cciptypes.Bytes32](consensus.FPlus1(f), nil)
 				}
-				validators[selector][seqNr].Add(hash)
+				validators[selector][seqNr].Add(hash, ao.OracleID)
 			}
 
 		}
 	}
 
-	for selector, seqNrs := range validators {
-		for seqNum, validator := range seqNrs {
+	for selector, seqNumValidator := range validators {
+		for seqNum, validator := range seqNumValidator {
 			if hashes := validator.GetValid(); len(hashes) == 1 {
 				results[selector][seqNum] = hashes[0]
 			}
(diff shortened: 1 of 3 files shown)


execute/plugin.go
```diff
@@ -218,6 +218,7 @@ func getPendingExecutedReports(
 	return groupedCommits, fullyExecuted, nil
 }
 
+//nolint:gocyclo
```
asoliman92, line 221:
```suggestion
```

internal/plugincommon/consensus/oracle_min_observation.go
```diff
@@ -0,0 +1,68 @@
...
+
+type observersCounter[T any] struct {
+	data      T
+	observers map[commontypes.OracleID]struct{}
+}
+
+// OracleMinObservation provides a way to ensure a minimum number of observations for
+// some piece of data have occurred. It maintains an internal cache and provides a list
+// of valid or invalid data points.
+type OracleMinObservation[T any] interface {
```
makramkd, line 20: This API doc comment should outline why this would be used instead of `MinObservation`. I think you can say that `OracleMinObservation` ensures that duplicate observations from the same observer are only counted once? TBH this should be the default behavior, I'm not even sure we should have a separate MinObservation and OracleMinObservation.

cc @winder

internal/plugincommon/consensus/oracle_min_observation.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// NewOracleMinObservation constructs a concrete MinObservation object. The
+// supplied idFunc is used to generate a uniqueID for the type being observed.
+func NewOracleMinObservation[T any](minThreshold Threshold, idFunc func(T) [32]byte) OracleMinObservation[T] {
+	if idFunc == nil {
+		idFunc = func(data T) [32]byte {
+			return sha3.Sum256([]byte(fmt.Sprintf("%v", data)))
+		}
+	}
```
winder, line 41: Thinking out loud, but could this version be implemented by overloading the ID func?
```
	if idFunc == nil {
		idFunc = func(oracle int, data T) [32]byte {
			return sha3.Sum256([]byte(fmt.Sprintf("%d_%v", oracle, data)))
		}
	}
	idFunc = func(oracle int, data T) [32]byte {
		return sha3.Sum256([]byte(fmt.Sprintf("%d_%v", oracle, idFunc(data))))
	}
```

internal/plugincommon/consensus/oracle_min_observation.go
```diff
@@ -0,0 +1,68 @@
...
+// some piece of data have occurred. It maintains an internal cache and provides a list
+// of valid or invalid data points.
+type OracleMinObservation[T any] interface {
+	Add(data T, oracleID commontypes.OracleID)
+	GetValid() []T
+}
+
+// minObservation is a helper object to filter data based on observation counts.
+// It keeps track of all inputs, determines if they are consistent
+// with one another, and whether they meet the required count threshold.
```
makramkd, line 27: This seems like a copy pasta comment, perhaps update to indicate that it implements `OracleMinObservation`

internal/plugincommon/consensus/oracle_min_observation.go
```diff
@@ -0,0 +1,68 @@
...
+type oracleMinObservation[T any] struct {
+	minObservation Threshold
+	cache          map[cciptypes.Bytes32]*observersCounter[T]
+	idFunc         func(T) [32]byte
+}
+
+// NewOracleMinObservation constructs a concrete MinObservation object. The
+// supplied idFunc is used to generate a uniqueID for the type being observed.
```
makramkd, line 35: This doc comment can be improved by removing redundancies

```suggestion
// NewOracleMinObservation returns an OracleMinObservation with the
// supplied idFunc is used to generate a uniqueID for the type being observed.
// If idFunc is nil a default is used.
```

internal/plugincommon/consensus/oracle_min_observation.go
asoliman92: thought about this yes and tried it,  was't straightforward to use from the callers' though. don't remember the specific reason for it now.

winder (approved):

PR #497, opened 2025-01-24
param all PDA access

tt-cll:
We want to re-use all PDA fetching logic as part of deployments/tooling. If we parameterize the program IDs as part of every fetch, we can import and re-use this elsewhere.

### DIFF
--- a/chains/solana/utils/ccip/ccip_transactions.go
+++ b/chains/solana/utils/ccip/ccip_transactions.go
@@ -5,11 +5,9 @@ import (
 
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4/ecdsa"
-	"github.com/gagliardetto/solana-go"
 
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/contracts/tests/config"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/gobindings/ccip_router"
-	"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/common"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/eth"
 )
 
@@ -34,28 +32,4 @@ func SignCommitReport(ctx [3][32]byte, report ccip_router.CommitInput, baseSigne
 		sigs = append(sigs, [65]byte(baseSig))
 	}
 	return sigs, nil
-}
-
-func GetSourceChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
-	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("source_chain_state"), chainSelectorLE}, ccipRouterProgram)
-	return p, err
-}
-
-func GetDestChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
-	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("dest_chain_state"), chainSelectorLE}, ccipRouterProgram)
-	return p, err
-}
-
-func GetCommitReportPDA(chainSelector uint64, root [32]byte, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
-	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("commit_report"), chainSelectorLE, root[:]}, ccipRouterProgram)
-	return p, err
-}
-
-func GetNoncePDA(chainSelector uint64, user solana.PublicKey, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
-	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("nonce"), chainSelectorLE, user.Bytes()}, ccipRouterProgram)
-	return p, err
 }
--- a/chains/solana/utils/ccip/ccip_transactions.go
+++ b/chains/solana/utils/ccip/ccip_transactions.go
@@ -36,26 +36,26 @@ func SignCommitReport(ctx [3][32]byte, report ccip_router.CommitInput, baseSigne
 	return sigs, nil
 }
 
-func GetSourceChainStatePDA(chainSelector uint64) (solana.PublicKey, error) {
+func GetSourceChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
 	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("source_chain_state"), chainSelectorLE}, config.CcipRouterProgram)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("source_chain_state"), chainSelectorLE}, ccipRouterProgram)
 	return p, err
 }
 
-func GetDestChainStatePDA(chainSelector uint64) (solana.PublicKey, error) {
+func GetDestChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
 	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("dest_chain_state"), chainSelectorLE}, config.CcipRouterProgram)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("dest_chain_state"), chainSelectorLE}, ccipRouterProgram)
 	return p, err
 }
 
-func GetCommitReportPDA(chainSelector uint64, root [32]byte) (solana.PublicKey, error) {
+func GetCommitReportPDA(chainSelector uint64, root [32]byte, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
 	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("commit_report"), chainSelectorLE, root[:]}, config.CcipRouterProgram)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("commit_report"), chainSelectorLE, root[:]}, ccipRouterProgram)
 	return p, err
 }
 
-func GetNoncePDA(chainSelector uint64, user solana.PublicKey) (solana.PublicKey, error) {
+func GetNoncePDA(chainSelector uint64, user solana.PublicKey, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
 	chainSelectorLE := common.Uint64ToLE(chainSelector)
-	p, _, err := solana.FindProgramAddress([][]byte{[]byte("nonce"), chainSelectorLE, user.Bytes()}, config.CcipRouterProgram)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("nonce"), chainSelectorLE, user.Bytes()}, ccipRouterProgram)
 	return p, err
 }
--- a/chains/solana/utils/common/pda.go
+++ b/chains/solana/utils/common/pda.go
@@ -1,49 +1,40 @@
 package common
 
 import (
-	"encoding/binary"
-
 	"github.com/gagliardetto/solana-go"
 )
 
-func FindConfigPDA(programID solana.PublicKey) (solana.PublicKey, uint8, error) {
-	return solana.FindProgramAddress([][]byte{[]byte("config")}, programID)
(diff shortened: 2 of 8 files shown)


chains/solana/utils/ccip/ccip_transactions.go
```diff
@@ -35,27 +33,3 @@ func SignCommitReport(ctx [3][32]byte, report ccip_router.CommitInput, baseSigne
 	}
 	return sigs, nil
 }
-
-func GetSourceChainStatePDA(chainSelector uint64) (solana.PublicKey, error) {
```
tt-cll, line 39: needed to move this out to prevent an import cycle

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,71 @@
+package common
```
agusaldasoro, line 1: Having this file is excelent, as it represents all the PDAs that we need, but this should be inside ccip folder, right? The idea of the common folder is to have only functions that are shared between different programs

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,71 @@
...
+
+func FindExternalExecutionConfigPDA(programID solana.PublicKey) (solana.PublicKey, uint8, error) {
+	return solana.FindProgramAddress([][]byte{[]byte("external_execution_config")}, programID)
+}
+
+func FindExternalTokenPoolsSignerPDA(programID solana.PublicKey) (solana.PublicKey, uint8, error) {
+	return solana.FindProgramAddress([][]byte{[]byte("external_token_pools_signer")}, programID)
+}
+
+func FindCounterPDA(programID solana.PublicKey) (solana.PublicKey, uint8, error) {
```
agusaldasoro, line 25: This is not part of the ccip-router program, it's part of an specific receiver. Let's remove from this file and keep here only the ccip router related PDAs

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,71 @@
...
+
+func FindTokenAdminRegistryPDA(programID, mint solana.PublicKey) (solana.PublicKey, uint8, error) {
+	return solana.FindProgramAddress([][]byte{[]byte("token_admin_registry"), mint.Bytes()}, programID)
+}
+
+func FindFeeBillingTokenConfigPDA(programID, mint solana.PublicKey) (solana.PublicKey, uint8, error) {
+	return solana.FindProgramAddress([][]byte{[]byte("fee_billing_token_config"), mint.Bytes()}, programID)
+}
+
+func FindCcipTokenpoolBillingPDA(programID, mint solana.PublicKey, chainSelector uint64) (solana.PublicKey, uint8, error) {
```
agusaldasoro, line 41:
```suggestion
func FindCcipTokenpoolBillingPDA(programID, chainSelector uint64, mint solana.PublicKey) (solana.PublicKey, uint8, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,71 @@
...
+	return solana.FindProgramAddress([][]byte{[]byte("ccip_tokenpool_billing"), binary.LittleEndian.AppendUint64([]byte{}, chainSelector), mint.Bytes()}, programID)
+}
+
+func FindCcipTokenpoolChainconfigPDA(programID, mint solana.PublicKey, chainSelector uint64) (solana.PublicKey, uint8, error) {
```
agusaldasoro, line 45:
```suggestion
func FindCcipTokenpoolChainconfigPDA(programID, chainSelector uint64, mint solana.PublicKey) (solana.PublicKey, uint8, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,71 @@
...
+	return solana.FindProgramAddress([][]byte{[]byte("ccip_tokenpool_chainconfig"), binary.LittleEndian.AppendUint64([]byte{}, chainSelector), mint.Bytes()}, programID)
+}
+
+func GetSourceChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```
agusaldasoro, line 49: Please modify all methods to respect the same format, the params of them should be in the same order as they are serialized in the seeds, and also they all should be `ccipRouterProgram` not just `programID`

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,71 @@
...
+	chainSelectorLE := Uint64ToLE(chainSelector)
```
agusaldasoro, line 50: You could use this method in the above methods for all chainSelector params as well right?

agusaldasoro (commented):
Thanks for doing this abstraction, just a few comments regarding:

1. Keep the methods to CCIP Router PDAs, so they should be in a ccip folder, they shouldn't be used for other programs.
2. Keep all the functions with the same format: same order of params and seeds, same names for the program id, etc.

chains/solana/utils/common/pda.go
tt-cll: Unfortunately, it can't be in the `ccip` folder because that introduces a circular dependency between it and the `config` folder. For example, https://github.com/smartcontractkit/chainlink-ccip/blob/e42d6d0c44e98f3c5ec16d38757f065468808554/chains/solana/contracts/tests/config/ccip_config.go#L38 would need to import `ccip` to resolve these methods after we declare them as functions, but then `ccip` would need to import `config` to get to the addresses and PDAs.

Happy to structure this in whatever way seems intuitive, but I think if the PDAs are standalone it will guarantee no import issues.

chains/solana/utils/common/pda.go
tt-cll: Yes good call.

chains/solana/utils/common/pda.go
tt-cll: I'm finding it tricky to move this somewhere else. The config populates a receiver variable using this logic that is used in the router tests as well as the ccip module. https://github.com/smartcontractkit/chainlink-ccip/blob/b1377ff2ab115033db14c4b88e213c9677a37fe2/chains/solana/utils/ccip/ccip_messages.go#L108

I'd rather just leave it here at the bottom and mark it as receiver code.

chains/solana/utils/common/pda.go
agusaldasoro: Okay, let's create a different folder then, something like `ccip-state`. As this is not part of commons. Commons should only have code to interact with any Program

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+
+func FindExternalTokenPoolsSignerPDA(ccipRouterProgram solana.PublicKey) (solana.PublicKey, uint8, error) {
+	return solana.FindProgramAddress([][]byte{[]byte("external_token_pools_signer")}, ccipRouterProgram)
+}
+
+func FindFeeBillingSignerPDA(ccipRouterProgram solana.PublicKey) (solana.PublicKey, uint8, error) {
+	return solana.FindProgramAddress([][]byte{[]byte("fee_billing_signer")}, ccipRouterProgram)
+}
+
+func FindTokenAdminRegistryPDA(mint, ccipRouterProgram solana.PublicKey) (solana.PublicKey, uint8, error) {
```
agusaldasoro, line 27:
```suggestion
func FindTokenAdminRegistryPDA(ccipRouterProgram solana.PublicKey, mint solana.PublicKey) (solana.PublicKey, uint8, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+	return solana.FindProgramAddress([][]byte{[]byte("token_admin_registry"), mint.Bytes()}, ccipRouterProgram)
+}
+
+func FindFeeBillingTokenConfigPDA(mint, ccipRouterProgram solana.PublicKey) (solana.PublicKey, uint8, error) {
```
agusaldasoro, line 31:
```suggestion
func FindFeeBillingTokenConfigPDA(ccipRouterProgram solana.PublicKey, mint solana.PublicKey) (solana.PublicKey, uint8, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+	return solana.FindProgramAddress([][]byte{[]byte("fee_billing_token_config"), mint.Bytes()}, ccipRouterProgram)
+}
+
+func FindCcipTokenpoolBillingPDA(chainSelector uint64, mint, ccipRouterProgram solana.PublicKey) (solana.PublicKey, uint8, error) {
```
agusaldasoro, line 35:
```suggestion
func FindCcipTokenpoolBillingPDA(chainSelector uint64, mint solana.PublicKey, ccipRouterProgram solana.PublicKey) (solana.PublicKey, uint8, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+	chainSelectorLE := Uint64ToLE(chainSelector)
+	return solana.FindProgramAddress([][]byte{[]byte("ccip_tokenpool_billing"), chainSelectorLE, mint.Bytes()}, ccipRouterProgram)
+}
+
+func GetSourceChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```
agusaldasoro, line 40:
```suggestion
func FindSourceChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+	chainSelectorLE := Uint64ToLE(chainSelector)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("nonce"), chainSelectorLE, user.Bytes()}, ccipRouterProgram)
+	return p, err
+}
+
+// Test Code
+
+func FindCounterPDA(ccipReceiverProgram solana.PublicKey) (solana.PublicKey, uint8, error) {
+	return solana.FindProgramAddress([][]byte{[]byte("counter")}, ccipReceiverProgram)
+}
```
agusaldasoro, line 68: Remove this from here, and leave that in the tests
```suggestion
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+	return p, err
+}
+
+func GetCommitReportPDA(chainSelector uint64, root [32]byte, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
+	chainSelectorLE := Uint64ToLE(chainSelector)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("commit_report"), chainSelectorLE, root[:]}, ccipRouterProgram)
+	return p, err
+}
+
+func GetNoncePDA(chainSelector uint64, user, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```
agusaldasoro, line 58:
```suggestion
func FindNoncePDA(chainSelector uint64, user solana.PublicKey, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+	return p, err
+}
+
+func GetDestChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
+	chainSelectorLE := Uint64ToLE(chainSelector)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("dest_chain_state"), chainSelectorLE}, ccipRouterProgram)
+	return p, err
+}
+
+func GetCommitReportPDA(chainSelector uint64, root [32]byte, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```
agusaldasoro, line 52:
```suggestion
func FindCommitReportPDA(chainSelector uint64, root [32]byte, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```

chains/solana/utils/common/pda.go
```diff
@@ -0,0 +1,68 @@
...
+	return solana.FindProgramAddress([][]byte{[]byte("ccip_tokenpool_billing"), chainSelectorLE, mint.Bytes()}, ccipRouterProgram)
+}
+
+func GetSourceChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
+	chainSelectorLE := Uint64ToLE(chainSelector)
+	p, _, err := solana.FindProgramAddress([][]byte{[]byte("source_chain_state"), chainSelectorLE}, ccipRouterProgram)
+	return p, err
+}
+
+func GetDestChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```
agusaldasoro, line 46:
```suggestion
func FindDestChainStatePDA(chainSelector uint64, ccipRouterProgram solana.PublicKey) (solana.PublicKey, error) {
```

agusaldasoro (changes requested):
A little bit more of feedback to keep the abstraction consistent

chains/solana/utils/common/pda.go
tt-cll: Sorry but didn't we want the params to be consistent with how they're added to the seed? https://github.com/smartcontractkit/chainlink-ccip/pull/497#discussion_r1928660427

agusaldasoro (dismissed):

agusaldasoro (approved):

PR #498, opened 2025-01-24
[CCIP-4938] Use map instead of slice in token price observations and outcome

asoliman92:
Use map instead of slice in token price observations and outcome

This is to make by design duplicate resistent so that we don't have issues in consensus later.

Fixed TokenPrice outcome timestamp consensus as it wasn't checking that it has enough observations before.

### DIFF
--- a/commit/factory.go
+++ b/commit/factory.go
@@ -39,7 +39,7 @@ const (
 	// Estimated maximum number of priced tokens that the Commit DON supports.
 	// This value does not indicate a system limitation but just an estimation to properly tune the OCR parameters.
 	// The value can be adjusted as needed.
-	estimatedMaxNumberOfPricedTokens = 10_000
+	estimatedMaxNumberOfPricedTokens = 14_445
 
 	// maxQueryLength is set to twice the maximum size of a theoretical merkle root processor query
 	// that assumes estimatedMaxNumberOfSourceChains source chains and
@@ -57,7 +57,7 @@ const (
 
 	// maxReportLength is set to an estimate of a maximum report size
 	// check factory_test for the calculation
-	maxReportLength = 993_982
+	maxReportLength = 128_2900
 
 	// maxReportCount is set to 1 because the commit plugin only generates one report per round.
 	maxReportCount = 1
--- a/commit/plugin_e2e_test.go
+++ b/commit/plugin_e2e_test.go
@@ -274,8 +274,11 @@ func TestPlugin_E2E_AllNodesAgree_TokenPrices(t *testing.T) {
 			prevOutcome: committypes.Outcome{},
 			mockPriceReader: func(m *readerpkg_mock.MockPriceReader) {
 				m.EXPECT().
-					// tokens need to be ordered, plugin checks all tokens from commit offchain config
-					GetFeedPricesUSD(mock.Anything, []ccipocr3.UnknownEncodedAddress{arbAddr, ethAddr}).
+					GetFeedPricesUSD(mock.Anything, mock.MatchedBy(func(tokens []ccipocr3.UnknownEncodedAddress) bool {
+						expectedTokens := mapset.NewSet(arbAddr, ethAddr)
+						actualTokens := mapset.NewSet(tokens...)
+						return expectedTokens.Equal(actualTokens)
+					})).
 					Return(ccipocr3.TokenPriceMap{
 						arbAddr: ccipocr3.NewBigInt(arbPrice),
 						ethAddr: ccipocr3.NewBigInt(ethPrice),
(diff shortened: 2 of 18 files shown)


commit/tokenprice/validate_observation.go
```diff
@@ -20,34 +18,26 @@ func (p *processor) ValidateObservation(
 		return fmt.Errorf("failed to validate FChain: %w", err)
 	}
 
-	observerSupportedChains, err := p.chainSupport.SupportedChains(ao.OracleID)
-	if err != nil {
-		return fmt.Errorf("failed to get supported chains: %w", err)
-	}
-
-	for chain := range obs.FChain {
```
asoliman92, line 28: This is wrong. FChain is observed from homeChain so all nodes can read them without issues.

commit/tokenprice/outcome.go
```diff
@@ -32,6 +31,13 @@ func (p *processor) getConsensusObservation(
 			fmt.Errorf("no consensus value for fDestChain, destChain: %d", p.destChain)
 	}
 
+	if consensus.LtTwoFPlusOne(fDestChain, len(aggObs.Timestamps)) {
```
asoliman92, line 34: wasn't checking that we have enough observations for timestamps before

commit/tokenprice/outcome.go
```diff
@@ -103,17 +106,14 @@ func (p *processor) selectTokensForUpdate(
...
 		}
 	}
 
 	// sort the token prices by tokenID
-	sort.Slice(tokenPrices, func(i, j int) bool {
-		return tokenPrices[i].TokenID < tokenPrices[j].TokenID
-	})
+	//sort.Slice(tokenPrices, func(i, j int) bool {
+	//	return tokenPrices[i].TokenID < tokenPrices[j].TokenID
+	//})
```
makramkd, line 116: Delete?

commit/tokenprice/processor.go
```diff
@@ -81,6 +81,11 @@ func (p *processor) Outcome(
 		"tokenPrices", tokenPriceOutcome,
 	)
 
+	if len(tokenPriceOutcome) == 0 {
+		lggr.Debugw("No token prices to report")
+		return Outcome{}, nil
+	}
+
 	out := Outcome{TokenPrices: tokenPriceOutcome}
 	p.metricsReporter.TrackTokenPricesOutcome(out)
```
makramkd, line 90: Do we wanna track the empty outcome?

commit/tokenprice/validate_observation.go
```diff
@@ -20,34 +18,21 @@ func (p *processor) ValidateObservation(
 		return fmt.Errorf("failed to validate FChain: %w", err)
 	}
 
-	observerSupportedChains, err := p.chainSupport.SupportedChains(ao.OracleID)
-	if err != nil {
-		return fmt.Errorf("failed to get supported chains: %w", err)
-	}
```
makramkd, line 26: Isn't the validation here just that `ao.OracleID`:
* can read the feed chain if they submitted feed prices
* can read the dest chain if they submitted fee quoter prices

?

commit/tokenprice/processor.go
asoliman92: Will be very noisy to track them. I think the debug should be enough. Also previous conditions that checks and returns empty outcome don't track it as well.

commit/factory.go
```diff
@@ -57,7 +57,7 @@ const (
 
 	// maxReportLength is set to an estimate of a maximum report size
 	// check factory_test for the calculation
-	maxReportLength = 993_982
+	maxReportLength = 128_2900
```
winder, line 60: Why did this increase?

commit/plugin_e2e_test.go
```diff
@@ -267,10 +261,10 @@ func TestPlugin_E2E_AllNodesAgree_TokenPrices(t *testing.T) {
...
 	tokensToQuery := maps.Keys(params.offchainCfg.TokenInfo)
 	sort.Slice(tokensToQuery, func(i, j int) bool { return tokensToQuery[i] < tokensToQuery[j] })
-	orderedTokenPrices := make([]ccipocr3.TokenPrice, 0, len(tokensToQuery))
-	for _, token := range tokensToQuery {
-		orderedTokenPrices = append(orderedTokenPrices, tokenPriceMap[token])
-	}
+	//orderedTokenPrices := make([]ccipocr3.TokenPrice, 0, len(tokensToQuery))
+	//for _, token := range tokensToQuery {
+	//	orderedTokenPrices = append(orderedTokenPrices, tokenPriceMap[token])
+	//}
```
winder, line 267: This can be deleted

commit/tokenprice/observation.go
```diff
@@ -81,17 +80,10 @@ func (p *processor) ObserveFeedTokenPrices(ctx context.Context, lggr logger.Logg
 	if err != nil {
```
winder, line 80: There is a sort 3 lines up. It's no longer needed because you're returning a map instead of a slice.

winder (approved):
Nice improvements. More resilient, efficient and fewer lines. :+1:

commit/tokenprice/validate_observation.go
makramkd: Just confirming that the above would be part of the validation right?

asoliman92: Yes, just currently it's not used. using in the pr I'm working on next. Currently as is it's not useful and wasn't contributing and actually was introducing a bug later with FChain check.

winder (approved):

makramkd (dismissed):

winder (approved):

PR #501, opened 2025-01-24
Multiple offramps support according to CCIP minor version [NONEVM-1200]

toblich:
This PR differentiates the offramp between CCIP minor versions, so that a message _must_ be executed invoking the same offramp that it was committed in. 

⚠️ Interface changes:
- Split `add_chain_selector` into two methods, one for the source chain (versioned) and one for the dest chain (unversioned, as we can't change PDAs in the onramp transparently to the end user).
- Offramp methods require passing in a `ccip_version` parameter
- Seeds for some PDAs now include the `ccip_version` in them

### TODOs (may be separate PRs to not make this one grow even more)
- [ ] Add support to reset the onramp seq nr when versions are bumped
- [ ] Add support to rollback the onramp version state


chains/solana/contracts/programs/ccip-router/src/context.rs
```diff
@@ -317,20 +331,6 @@ pub struct UpdateConfigCCIPRouter<'info> {
...
-        mut,
-        seeds = [CONFIG_SEED],
-        bump,
-        constraint = valid_version(config.load()?.version, MAX_CONFIG_V) @ CcipRouterError::InvalidInputs, // validate state version
-    )]
-    pub config: AccountLoader<'info, Config>,
-    #[account(address = config.load()?.owner @ CcipRouterError::Unauthorized)]
-    pub authority: Signer<'info>,
-    pub system_program: Program<'info, System>,
-}
```
toblich, line 332: This seemed to be dead code

chains/solana/contracts/programs/ccip-router/src/context.rs
```diff
@@ -29,8 +29,8 @@ pub fn uninitialized(v: u8) -> bool {
 }
 
 // Fixed seeds - different contexts must use different PDA seeds
-pub const DEST_CHAIN_STATE_SEED: &[u8] = b"dest_chain_state";
-pub const SOURCE_CHAIN_STATE_SEED: &[u8] = b"source_chain_state";
+pub const DEST_CHAIN_SEED: &[u8] = b"dest_chain_state";
+pub const SOURCE_CHAIN_SEED: &[u8] = b"source_chain_state";
```
toblich, line 33: I'm removing the `state` bit of the names here as they hold config & state. I left the seeds untouched to no affect offchain code that may already be using those values though. If you think it's ok to change the seed values and just notify people, I'll happily do it.

toblich:
We decided against this in the end. We'll break the offramp into a separate program altogether.

PR #504, opened 2025-01-27
Refactor token admin registry to match EVM implementation

agusaldasoro:
Refactor the claim ownership to follow a 2 step process:
- propose admin
- accept admin
- transfer admin

Changes:
- [x] Added Propose Admin
- [x] Make TokenAdminRegistry init_if_needed to support overriding a pending administrator
- [x] Modified mint as param to be a Mint Account so it is validated by Anchor
- [x] Refactor all Token Admin Registry to be compliance with EVM implementation


chains/solana/contracts/programs/ccip-router/src/event.rs
```diff
@@ -148,3 +128,27 @@ pub struct OwnershipTransferred {
     pub from: Pubkey,
     pub to: Pubkey,
 }
+
+/////////////////////////////////
+// Token Admin Registry Events //
+/////////////////////////////////
```
PabloMansanet, line 134: Nit: This is a nit that I had for tobi as well: I personally don't like "header comments" like this, particularly in Rust, as they're a poor substitute for inline modules. It's easy for the comment to go out of sync with the code, and for the block covered by the comment to become unclear.

Instead, I'd prefer `pub mod token_admin_registry`, which makes it clearer when importing (i.e. events::token_admin_registry::PoolSet`.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -5,74 +5,120 @@ use solana_program::{address_lookup_table::state::AddressLookupTable, log::sol_l
...
+    token_admin_registry: &mut Account<'_, crate::TokenAdminRegistry>,
+    new_admin: Pubkey,
+) -> Result<()> {
+    require_neq!(
+        new_admin,
+        Pubkey::new_from_array([0; 32]),
+        CcipRouterError::InvalidTokenAdminRegistryInputsZeroAddress
+    );
+
+    // TODO: Revert if administrator != 0 AlreadyRegistered
```
PabloMansanet, line 53: I feel like this TODO should be addressed before merging (should only be a `require` with a custom error, right?) because otherwise the method signature is confusing: it claims to only set the pending administrator but it's also deleting the current one.

PabloMansanet: Side note: Maybe this could be named/commented to clarify this is meant to be used for initialization only: otherwise it might be tempting to erroneously call this from the `transfer` function. Maybe calling this "initial_set_pending_administrator" would help.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -5,74 +5,120 @@ use solana_program::{address_lookup_table::state::AddressLookupTable, log::sol_l
...
+    token_admin_registry.administrator = new_admin;
     token_admin_registry.pending_administrator = Pubkey::new_from_array([0; 32]);
-    token_admin_registry.lookup_table = Pubkey::new_from_array([0; 32]);
 
-    emit!(AdministratorRegistered {
-        token_mint,
-        administrator,
+    emit!(AdministratorTransferred {
+        token: token_mint,
+        new_admin,
```
PabloMansanet, line 101: Should this event have an `old_admin` entry? I see that most other storage change events reflect the inbound and outbound values, but I'm not sure if it's convention or coincidence.

chains/solana/contracts/programs/ccip-router/src/lib.rs
```diff
@@ -44,6 +44,10 @@ pub mod ccip_router {
 
     use super::*;
 
+    //////////////////////////
+    /// Initialization Flow //
+    //////////////////////////
```
PabloMansanet, line 49: Ditto on "section comments" vs modules.

chains/solana/contracts/programs/ccip-router/src/token_context.rs
```diff
@@ -23,47 +23,50 @@ pub struct TokenAdminRegistry {
...
+pub struct RegisterTokenAdminRegistryByOwner<'info> {
     #[account(
         seeds = [CONFIG_SEED],
         bump,
         constraint = valid_version(config.load()?.version, MAX_CONFIG_V) @ CcipRouterError::InvalidInputs,
     )]
     pub config: AccountLoader<'info, Config>,
     #[account(
-        init,
+        init, // TODO: change to init_if_needed
```
PabloMansanet, line 59: If the purpose of changing to `init_if_needed` is to be able to override a pending administrator, would it be a bit cleaner to just implement a `override_pending_administrator` instruction? WDYT?

I read that `init_if_needed` is a security concern, and I do think it results in slightly more confusing flow than the alternative of well delimited initialization and "editing" instructions, but I don't know if there are any constraints that I'm not seeing.

PabloMansanet (changes requested):
Looking good! :)

I only have a couple small things that I would prefer to see addressed: A TODO that I feel warrants being done sooner than later, and the issue with section comments.

chains/solana/contracts/programs/ccip-router/src/token_context.rs
agusaldasoro: Mmm, I guess you are right. I tried to keep it as similar as possible to EVM, that's why I thought of using an `init_if_needed`, but as this functionality should be used only to fix a missconfiguration it shouldn't be part of the common use case. It's safer and more clear to have a separate `override_pending_administrator` message.

chains/solana/contracts/programs/ccip-router/src/event.rs
agusaldasoro: Good point here, using modules as you did for seeds makes everything more clear. I need to figure out if Anchor auto gen code will work the same way anywas 🤔

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
agusaldasoro: Oh, yes sorry, this will be handled before merging. Just wanted a quick review to save time. 
I'm going to create a different message for pending admin override, so that check will be there and this internal function can be renamed to reflect the initialization as well. Thanks

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
agusaldasoro: I copied the same format as EVM events: https://github.com/smartcontractkit/chainlink/blob/16f1529856a575c8d2091a16033a8d0371408d96/contracts/src/v0.8/ccip/tokenAdminRegistry/TokenAdminRegistry.sol#L28

chains/solana/contracts/programs/ccip-router/src/event.rs
agusaldasoro: Worked! 🙌

chains/solana/contracts/programs/ccip-router/src/lib.rs
agusaldasoro: Here we can not use modules, as the go bindings do not work 😢

chains/solana/contracts/programs/ccip-router/src/event.rs
agusaldasoro: I refactored all events

chains/solana/contracts/programs/ccip-router/src/lib.rs
PabloMansanet: Ah well, what can we do!

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,151 @@
...
+// Overrides the pending administrator for the given token
+pub fn ccip_admin_override_pending_administrator(
+    ctx: Context<OverridePendingTokenAdminRegistryByCCIPAdmin>,
+    token_admin_registry_admin: Pubkey,
+) -> Result<()> {
+    let token_admin_registry = &mut ctx.accounts.token_admin_registry;
+    token_admin_registry.pending_administrator = token_admin_registry_admin;
+
+    Ok(())
+}
```
aalu1418, line 44: when is this expected to be called?

it seems like this can be called any time which allows the ccip admin or token owner to override the pending admin at any time

my understanding is that once an `administrator` is set, then the token owner and ccip admin can no longer change anything

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,151 @@
...
+    let token_mint = ctx.accounts.mint.key().to_owned();
+    let token_admin_registry = &mut ctx.accounts.token_admin_registry;
+
+    init_with_pending_administrator(token_admin_registry, token_mint, token_admin_registry_admin)?;
+
+    Ok(())
+}
+
+// Overrides the pending administrator for the given token
+pub fn owner_override_pending_administrator(
```
aalu1418, line 60: same question here

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
aalu1418: so maybe there needs to be one more constraint added to the contexts?

aalu1418 (commented):
overall looks good - just a question on permissions

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
agusaldasoro: Oh, good catch! I forgot to add a check for `administrator == 0`. As this method is intended to be used when a missconfiguration happened. Will add a test for that!

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,165 @@
...
+    AcceptAdminRoleTokenAdminRegistry, CcipRouterError, ModifyTokenAdminRegistry,
+    SetPoolTokenAdminRegistry,
 };
 
 const MINIMUM_TOKEN_POOL_ACCOUNTS: usize = 9;
 
-pub fn register_token_admin_registry_via_get_ccip_admin(
-    ctx: Context<RegisterTokenAdminRegistryViaGetCCIPAdmin>,
-    mint: Pubkey, // should we validate that this is a real token program?
+// Proposes an administrator for the given token as pending administrator
```
PabloMansanet, line 22: Nit: Would be good to have these be doc comments (/// instead of //) for the `cargo doc` users.

PabloMansanet (dismissed):
Looks good to me! 👏

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
agusaldasoro, line 22:
```suggestion
/// Proposes an administrator for the given token as pending administrator
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,165 @@
...
 ) -> Result<()> {
+    let token_mint = ctx.accounts.mint.key().to_owned();
     let token_admin_registry = &mut ctx.accounts.token_admin_registry;
+
+    init_with_pending_administrator(token_admin_registry, token_mint, token_admin_registry_admin)?;
+
+    Ok(())
+}
+
+// Overrides the pending administrator for the given token
```
agusaldasoro, line 35:
```suggestion
/// Overrides the pending administrator for the given token
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,165 @@
...
+        Pubkey::new_from_array([0; 32]),
+        CcipRouterError::InvalidTokenAdminRegistryProposedAdmin
+    );
+
+    token_admin_registry.pending_administrator = token_admin_registry_admin;
+
+    Ok(())
+}
+
+// Proposes an administrator for the given token as pending administrator
```
agusaldasoro, line 53:
```suggestion
/// Proposes an administrator for the given token as pending administrator
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,165 @@
...
+) -> Result<()> {
+    let token_mint = ctx.accounts.mint.key().to_owned();
+    let token_admin_registry = &mut ctx.accounts.token_admin_registry;
+
+    init_with_pending_administrator(token_admin_registry, token_mint, token_admin_registry_admin)?;
+
+    Ok(())
+}
+
+// Overrides the pending administrator for the given token
```
agusaldasoro, line 66:
```suggestion
/// Overrides the pending administrator for the given token
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,165 @@
...
+        current_admin: Pubkey::new_from_array([0; 32]),
+        new_admin,
     });
 
     Ok(())
 }
 
-pub fn register_token_admin_registry_via_owner(
-    ctx: Context<RegisterTokenAdminRegistryViaOwner>,
+// Transfers the administrator role for a token to a new address with a 2-step process
```
agusaldasoro, line 110:
```suggestion
/// Transfers the administrator role for a token to a new address with a 2-step process
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,165 @@
...
+    emit!(events::AdministratorTransferRequested {
+        token: token_mint,
+        current_admin: token_admin_registry.administrator,
+        new_admin,
+    });
+
+    Ok(())
+}
 
+// Accepts the admin role for a token
```
agusaldasoro, line 129:
```suggestion
/// Accepts the admin role for a token
```

chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs
```diff
@@ -1,77 +1,165 @@
...
-pub fn set_pool(
-    ctx: Context<SetPoolTokenAdminRegistry>,
-    mint: Pubkey,
-    writable_indexes: Vec<u8>,
-) -> Result<()> {
-    // set new lookup table
+// Sets the lookup table for pool for a token.
+// TODO: Add a test with look up table as 0 address
+// Setting the lookup table to address(0) effectively delists the token from CCIP.
+// Setting the lookup table to any other address enables the token on CCIP.
```
agusaldasoro, line 151:
```suggestion
/// Sets the lookup table for pool for a token.
/// TODO: Add a test with look up table as 0 address
/// Setting the lookup table to address(0) effectively delists the token from CCIP.
/// Setting the lookup table to any other address enables the token on CCIP.
```

PabloMansanet (approved):

aalu1418 (approved):

aalu1418 (approved):

PR #505, opened 2025-01-27
Extended CommitReportInfo to include PriceUpdates

silaslenihan:
Resolves dependencies: 
- https://github.com/smartcontractkit/chainlink-solana/pull/1009

### DIFF
--- a/pkg/types/ccipocr3/plugin_commit_types.go
+++ b/pkg/types/ccipocr3/plugin_commit_types.go
@@ -72,6 +72,7 @@ type CommitReportInfo struct {
 	// RemoteF Max number of faulty RMN nodes; f+1 signers are required to verify a report.
 	RemoteF     uint64            `json:"remoteF"`
 	MerkleRoots []MerkleRootChain `json:"merkleRoots"`
+	TokenPrices []TokenPrice      `json:"tokenPrices"`
 }
 
 func (cri CommitReportInfo) Encode() ([]byte, error) {
--- a/commit/report.go
+++ b/commit/report.go
@@ -64,6 +64,7 @@ func (p *Plugin) Reports(
 		repInfo = cciptypes.CommitReportInfo{
 			RemoteF:     outcome.MerkleRootOutcome.RMNRemoteCfg.FSign,
 			MerkleRoots: rep.MerkleRoots,
+			TokenPrices: rep.PriceUpdates.TokenPriceUpdates,
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


pkg/types/ccipocr3/plugin_commit_types.go
```diff
@@ -66,12 +66,22 @@ type PriceUpdates struct {
...
 }
 
+type CommitReportArgs struct {
+	ReportContext [2][32]byte
+	Report        []byte
+	Rs            [][32]byte
+	Ss            [][32]byte
+	RawVs         [32]byte
+	Info          CommitReportInfo
+}
```
winder, line 76: Please remove this. It's OCR/CW code for the transmitter.

silaslenihan: Is there some other common place we can add this outside of chainlink core? I want to be able to import this into `chainlink-solana` to run some tests

makramkd: What kind of tests? All integration tests are currently in chainlink. If this is transmitter related, it should be in chainlink and the test should be there as well

pkg/types/ccipocr3/plugin_execute_types.go
```diff
@@ -18,6 +18,12 @@ type ExecutePluginReportSingleChain struct {
 	ProofFlagBits       BigInt        `json:"proofFlagBits"`
 }
 
+type ExecuteReportArgs struct {
```
makramkd, line 21: Ditto on earlier comment

makramkd (changes requested):

pkg/types/ccipocr3/plugin_commit_types.go
silaslenihan: I wrote some tests specifically to test the Solana ChainWriter's ability to receive the contract transmitter args and encode / transform them for execution on chain https://github.com/smartcontractkit/chainlink-solana/blob/da75f6d7734685f31e757b191725a77cf0094120/pkg/solana/chainwriter/chain_writer_test.go#L697-L752

winder: I'd like to see a generic type as part of the CW that we could use:
```
type ChainWriterOCRArgs struct {
	ReportContext [2][32]byte
	Report        []byte
	Rs            [][32]byte // omit empty?
	Ss            [][32]byte // omit empty?
	RawVs         [32]byte // omit empty?
	// Info is a special field used by CW to dynamically accept chain specific data.
	Info          any
}
```

silaslenihan: Ok, removed this from this PR

winder (approved):

PR #509, opened 2025-01-28
Fix exec plugin issue related to filtering executed msgs from reports

dimkouv:
## Issue
Wrong implementation in the Execute Plugin when filtering out the executed messages

## Description of the issue
In the execute/plugin_functions.go file, the filterOutExecutedMessages function exists which is used by the Observation function, through a series of calls, during the first round of the plugin. In this round, the Commit Plugin transmitted reports which have pending messages for execution are fetched. The filterOutExecutedMessages function is used in the last step to exclude the messages that have already been executed.

However, the way in which the execute/plugin.go::getPendingExecutedReports is implemented does not actually comply with the implementation of the filterOutExecutedMessages function which is called by the former.

More specifically, the aforementioned function gets as input an array with sequence number ranges. These ranges, though, are trivial as they only contain a single sequence number each (e.g. [N,N]). This is enforced by the implementation of the ccipReader.ExecutedMessageRanges function which calculates them.

The filterOutExecutedMessages function, though, is implemented like if larger ranges were expected.

## Solution
Replace `ExecutedMessageRanges` function with `ExecutedMessages` which returns a slice of the executed messages in the provided sequence numbers range. Update all the execution plugin related logic that called this function. This leads to making the calculations very simple without having to deal with complex sequence number ranges of 1 element each.

> The issue is described in more detail in `M16`

## Dependencies

Chainlink PR: https://github.com/smartcontractkit/chainlink/pull/16093

`core ref: 7d543d6436c3f41653fc399411b445c3983e1c96`

### DIFF
--- a/execute/plugin.go
+++ b/execute/plugin.go
@@ -144,9 +144,12 @@ func getPendingExecutedReports(
 			continue
 		}
 
+		lggr.Debugw("grouped reports", "selector", selector, "reports", reports, "count", len(reports))
 		sort.Slice(reports, func(i, j int) bool {
 			return reports[i].SequenceNumberRange.Start() < reports[j].SequenceNumberRange.Start()
 		})
+		// todo: remove this logs after investigating whether the sorting above can be safely removed
+		lggr.Debugw("sorted reports", "selector", selector, "reports", reports, "count", len(reports))
 
 		ranges, err := computeRanges(reports)
 		if err != nil {


pkg/reader/ccip.go
```diff
@@ -285,13 +285,13 @@ func (r *ccipChainReader) ExecutedMessageRanges(
...
 
-	executed := make([]cciptypes.SeqNumRange, 0)
+	executed := make([]cciptypes.SeqNum, 0)
 	for _, item := range iter {
 		stateChange, ok := item.Data.(*ExecutionStateChangedEvent)
 		if !ok {
 			return nil, fmt.Errorf("failed to cast %T to executionStateChangedEvent", item.Data)
 		}
-		executed = append(executed, cciptypes.NewSeqNumRange(stateChange.SequenceNumber, stateChange.SequenceNumber))
+		executed = append(executed, stateChange.SequenceNumber)
```
asoliman92, line 294: much better like that 👍

execute/plugin.go
```diff
@@ -156,20 +157,25 @@ func getPendingExecutedReports(
 			return nil, err
 		}
 
-		var executedMessages []cciptypes.SeqNumRange
+		executedMessageSet := mapset.NewSet[cciptypes.SeqNum]()
 		for _, seqRange := range ranges {
-			executedMessagesForRange, err2 := ccipReader.ExecutedMessageRanges(ctx, selector, dest, seqRange)
+			executedMessagesForRange, err2 := ccipReader.ExecutedMessages(ctx, selector, dest, seqRange)
```
asoliman92, line 162: it's going to be even simpler if `ExecutedMessages` returns a Set directly.

asoliman92 (approved):

execute/plugin.go
asoliman92: They're non blockers. approved. I prefer the set or map[T]struct{} for these kind of stuff normally.

asoliman92 (approved):

execute/plugin.go
```diff
@@ -127,20 +128,12 @@ func getPendingExecutedReports(
 	ts time.Time,
 	lggr logger.Logger,
 ) (exectypes.CommitObservations, error) {
-	latestReportTS := time.Time{}
-	commitReports, err := ccipReader.CommitReportsGTETimestamp(ctx, dest, ts, 1000)
+	commitReports, err := ccipReader.CommitReportsGTETimestamp(ctx, dest, ts, 1000) // todo: configurable limit
```
makramkd, line 131: I thought we had a limit in the offchain config

makramkd: Seems not. In <= v1.5 we had a const in the code somewhere, I think that might also be fine vs. the magic 1K. Lets create a ticket in the performance backlog?

execute/plugin.go
```diff
@@ -151,25 +144,30 @@ func getPendingExecutedReports(
 			continue
 		}
 
+		sort.Slice(reports, func(i, j int) bool {
+			return reports[i].SequenceNumberRange.Start() < reports[j].SequenceNumberRange.Start()
+		})
```
makramkd, line 149: Not 100% sure but isn't this kind of sort already being done by the `CommitReportsGTETimestamp` method? Generally, performance wise, I think its better to do those kinds of ops as part of the DB query especially if the query needs to use a sorted index anyway.

If its _not_ being done there I think we should create a ticket to do it there + have the sorted order enforced via tests + godoc on the method.

cc @0xAustinWang @mateusz-sekara

execute/plugin.go
```diff
@@ -151,25 +144,30 @@ func getPendingExecutedReports(
...
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("compute report ranges: %w", err)
 		}
 
-		var executedMessages []cciptypes.SeqNumRange
+		executedMessageSet := mapset.NewSet[cciptypes.SeqNum]()
 		for _, seqRange := range ranges {
-			executedMessagesForRange, err2 := ccipReader.ExecutedMessageRanges(ctx, selector, dest, seqRange)
+			executedMessagesForRange, err2 := ccipReader.ExecutedMessages(ctx, selector, dest, seqRange)
```
makramkd, line 158: Just another comment on this re: performance - DB-query wise, you'd probably just wanna do one query to cover all the different contiguous sub-ranges. This right now does a query on every single sub-range separately, which makes DB I/O quite heavy for this function/loop and probably much slower than it should be.

cc: @0xAustinWang @mateusz-sekara

asoliman92 (approved):

PR #510, opened 2025-01-28
[Solana][NONEVM-1201] Readability pass: context.rs

PabloMansanet:
First of many readability passes, this one focused on `context.rs`

* Removed redundant comments, and "obvious" comments re-stating the code.
* Expanded some unclear ones.
* Turned comments into docstrings for `cargo doc`.
* Fixed an inconsistency with the `SEED` constant names, where some had the `SEED` prefix and some didn't, by namespacing instead.
* Made spacing between elements consistent.
* Made comment style consistent (generally preferring above VS inline, never commenting between attribute and definition, etc).


agusaldasoro (approved):
Amazing!

PR #518, opened 2025-01-29
Exec: commit root caching

winder:
Add a `CommitsRootCache` which allows the execute plugin to remember what has already been executed. This enables it to skip RPC calls associated with known-executed commit roots.

In addition, a snooze button was added which allows us to snooze commit roots associated with cursed chains.

### DIFF
--- a/execute/internal/cache/commit_root_cache.go
+++ b/execute/internal/cache/commit_root_cache.go
@@ -1,0 +1,122 @@
+package cache
+
+import (
+	"sync"
+	"time"
+
+	"github.com/patrickmn/go-cache"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+
+	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+)
+
+const (
+	// EvictionGracePeriod defines how long after the messageVisibilityInterval a root is still kept in the cache
+	EvictionGracePeriod = 1 * time.Hour
+	// CleanupInterval defines how often roots cache is scanned to evict stale roots
+	CleanupInterval = 30 * time.Minute
+)
+
+// CommitsRootsCache keeps track of commit roots (and messages?) that are eligible for execution.
+//
+// This cache is used when:
+//   - a commit root is fully executed and we want to skip it in the future. This would be called
+//     in the GetCommitReports phase after checking if each SeqNr is executed (or if the Txm state is finalized).
+//   - remember the oldest pending commit root to limit the database scan to only the unfinalized part of the chain.
+type CommitsRootsCache interface {
+	CanExecute(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32) bool
+	MarkAsExecuted(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32)
+	Snooze(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32)
+}
+
+func NewCommitRootsCache(
+	lggr logger.Logger,
+	messageVisibilityInterval time.Duration,
+	rootSnoozeTime time.Duration,
+) CommitsRootsCache {
+	return internalNewCommitRootsCache(
+		lggr,
+		messageVisibilityInterval,
+		rootSnoozeTime,
+		CleanupInterval,
+		EvictionGracePeriod,
+	)
+}
+
+func internalNewCommitRootsCache(
+	lggr logger.Logger,
+	messageVisibilityInterval time.Duration,
+	rootSnoozeTime time.Duration,
+	cleanupInterval time.Duration,
+	evictionGracePeriod time.Duration,
+) *commitRootsCache {
+	snoozedRoots := cache.New(rootSnoozeTime, cleanupInterval)
+	executedRoots := cache.New(messageVisibilityInterval+evictionGracePeriod, cleanupInterval)
+
+	return &commitRootsCache{
+		lggr:                      lggr,
+		rootSnoozeTime:            rootSnoozeTime,
+		executedRoots:             executedRoots,
+		snoozedRoots:              snoozedRoots,
+		messageVisibilityInterval: messageVisibilityInterval,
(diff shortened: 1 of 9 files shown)


execute/internal/cache/commit_root_cache.go
```diff
@@ -0,0 +1,122 @@
...
+func internalNewCommitRootsCache(
+	lggr logger.Logger,
+	messageVisibilityInterval time.Duration,
+	rootSnoozeTime time.Duration,
+	cleanupInterval time.Duration,
+	evictionGracePeriod time.Duration,
+) *commitRootsCache {
+	snoozedRoots := cache.New(rootSnoozeTime, cleanupInterval)
+	executedRoots := cache.New(messageVisibilityInterval+evictionGracePeriod, cleanupInterval)
+
```
0xnogo, line 56: I wonder if we want to add a cache size limits here? We probably don't expect to have millions of insertions, but wondering if you thought about it.

execute/internal/cache/commit_root_cache.go
```diff
@@ -0,0 +1,122 @@
...
+	CanExecute(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32) bool
+	MarkAsExecuted(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32)
+	Snooze(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32)
+}
+
+func NewCommitRootsCache(
+	lggr logger.Logger,
+	messageVisibilityInterval time.Duration,
+	rootSnoozeTime time.Duration,
+) CommitsRootsCache {
```
0xnogo, line 37: nit - we could have validation on `messageVisibilityInterval` and `rootSnoozeTime` but I believe only exec plugin is supposed to use this constructor, so misusage is very unlikely.

execute/internal/cache/commit_root_cache.go
mateusz-sekara: Adding limits to the cache size might lead to an unexpected behaviour with not having all the items required and relying on some partial state. I think time retention is enough and gonna protect well against memory leakage.
 
Simple napkin math:
* per dest load = 1 msg / 10 seconds, 
* pessimistic batching, always 1 merkle root per 1 message
* permissionlessExecWindow (aka messageVisibilityInterval) 24 hours (it's 8 on mainnet)

14400 merkle roots stored in the cache per dest chain
merkleroot is 32 bytes which gives us ~450kB per dest chain (pessimistic)

execute/plugin.go
```diff
@@ -144,6 +156,17 @@ func getPendingExecutedReports(
 			continue
 		}
 
+		// Filter out reports that cannot be executed (snoozed).
+		// TODO: filter func instead of 'canExecute'?
+		var filtered []exectypes.CommitData
+		for _, commitReport := range reports {
```
mateusz-sekara, line 162: The original `commit_roots.go` also has a nice trick with a sliding window. Therefore we were limiting a number of database logs fetched from the db to memory by narrowing the lookup window based on the latest finalized and executed Commit Report. Not sure if applicable here, but I wanted to raise that as a follow-up perf improvement.

execute/plugin.go
```diff
@@ -60,9 +60,13 @@ type Plugin struct {
...
 	lggr                  logger.Logger
 
-	observationOptimizer optimizers.ObservationOptimizer
 	// state
 	contractsInitialized bool
+	// this cache remembers commit root details to optimize DB lookups.
+	commitRootsCache cache.CommitsRootsCache
+
+	// TODO: remove this, it can be transient. Also its logger is never initialized.
+	observationOptimizer optimizers.ObservationOptimizer
```
mateusz-sekara, line 69: Can't we remove it as a part of this PR? If not, could please create a ticket? I think we easily forget about these TODOs in the code

execute/plugin.go
```diff
@@ -144,6 +156,17 @@ func getPendingExecutedReports(
 			continue
 		}
 
+		// Filter out reports that cannot be executed (snoozed).
+		// TODO: filter func instead of 'canExecute'?
+		var filtered []exectypes.CommitData
+		for _, commitReport := range reports {
+			if !canExecute(commitReport.SourceChain, commitReport.MerkleRoot) {
```
mateusz-sekara, line 163: Might be worth logging as debug these skipped reports

mateusz-sekara (dismissed):

execute/plugin.go
winder: Yes, thats another great optimization that I've been thinking about. We can take it a step further by adding a "Pending" cache, that way stuck transactions wouldn't prevent us from narrowing the lookup window.

execute/plugin.go
winder: Good idea, added a log.

execute/internal/cache/commit_root_cache.go
winder: What sort of validation were you thinking?

execute/internal/cache/commit_root_cache.go
winder: I believe its 450kB per source chain because the report is a slice of reports per source chain.

In the future we may also support multiple reports per OCR round.

Even with all of that we're only talking about 10s of MBs.

winder: I believe its 450kB per source chain because the report is a slice of reports per source chain.

In the future we may also support multiple reports per OCR round.

Even with all of that we're only talking about 10s of MBs.

winder: Thinking about it more, 1 msg / 10 seconds might be too pessimistic. The commit plugin pauses after each commit report is submitted to wait for transmission, so it could be more than 3 rounds spent per report.

execute/internal/cache/commit_root_cache.go
0xnogo: Thinking about it again, I'm not even sure Duration can be negative. Closing this comment.

mateusz-sekara (approved):

PR #520, opened 2025-01-29
[Solana][NONEVM-1246] Implement allowlist on onramp

PabloMansanet:


chains/solana/contracts/programs/ccip-router/src/state.rs
```diff
@@ -119,6 +119,25 @@ pub struct DestChainConfig {
...
+    pub allow_list_enabled: bool,
+}
+
+impl DestChainConfig {
+    pub fn space(&self) -> usize {
+        Self::INIT_SPACE + self.dynamic_space()
+    }
+
+    pub fn dynamic_space(&self) -> usize {
+        self.allowed_senders.len() * std::mem::size_of::<Pubkey>()
```
agusaldasoro, line 139: Don't arrays have a base size of 4? I remember using `4 + len * size`

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -2138,6 +2141,71 @@ func TestCCIPRouter(t *testing.T) {
 			require.NotNil(t, result)
 		})
 
+		t.Run("When sending with an empty but enabled allowlist, it fails", func(t *testing.T) {
```
agusaldasoro, line 2144: Should we add a test for sending with disabled allowlist as well?

chains/solana/contracts/programs/ccip-router/src/state.rs
```diff
@@ -119,6 +119,25 @@ pub struct DestChainConfig {
...
     pub chain_family_selector: [u8; 4], // Selector that identifies the destination chain's family. Used to determine the correct validations to perform for the dest chain.
+
+    // list of senders authorized to send messages to this destination chain.
+    // Note: The attribute name `max_len` is slightly misleading: it is not in any
+    // way limiting the actual length of the vector during initialization; it just
+    // helps the InitSpace derive macro work out the initial space. We can leave it at
+    // zero and calculate the actual length in the instruction context.
+    #[max_len(0)]
+    pub allowed_senders: Vec<Pubkey>,
+    pub allow_list_enabled: bool,
```
aalu1418, line 130: can we double check the product requirements for `allowed_senders`?

if it's unbounded, then we should allowlist based on a user specific PDA since there is a maximum account size that would bound the list

chains/solana/contracts/programs/ccip-router/src/context.rs
```diff
@@ -312,8 +311,34 @@ pub struct UpdateSourceChainSelectorConfig<'info> {
...
+#[instruction(new_chain_selector: u64, dest_chain_config: DestChainConfig)]
 pub struct UpdateDestChainSelectorConfig<'info> {
+    #[account(
+        mut,
+        seeds = [seed::DEST_CHAIN_STATE, new_chain_selector.to_le_bytes().as_ref()],
+        bump,
+        constraint = valid_version(dest_chain_state.version, MAX_CHAINSTATE_V) @ CcipRouterError::InvalidInputs,
+        realloc = ANCHOR_DISCRIMINATOR + DestChain::INIT_SPACE + dest_chain_config.dynamic_space(),
+        realloc::payer = authority,
+        realloc::zero = false
```
agusaldasoro, line 323: If using false, then we need to ensure zeroing the bytes when decreasing the size of the allowed senders?

chains/solana/contracts/programs/ccip-router/src/state.rs
PabloMansanet: This is captured by `INIT_SPACE`. When specifying a `max_len` of 0, `INIT_SPACE` only considers the vector discriminator (4 bytes), so we only need to provide the actual dynamic space in this method.

I find it a bit annoying that 'max_len' is so strangely named, because it's way more useful than I first thought: It should be called `elements_to_preallocate_at_init` or something 😆.

chains/solana/contracts/tests/ccip/ccip_router_test.go
PabloMansanet: I don't think it's necessary: that's already captured by all the other tests, as the allowlist is disabled by default.

chains/solana/contracts/programs/ccip-router/src/state.rs
PabloMansanet: Checked by @agusaldasoro , we're OK to go with the boundary enforced by maximum account size.

chains/solana/contracts/programs/ccip-router/src/context.rs
agusaldasoro: All the docs that I read recommend using realloc zero to true, or in this doc they recommend using an ALT table instead 

https://www.helius.dev/blog/a-hitchhikers-guide-to-solana-program-security?utm_source=chatgpt.com#account-data-reallocation

agusaldasoro: This module talks about issues when init, but they may be valid for resize as well? https://solana.com/developers/courses/program-security/reinitialization-attacks

PabloMansanet: @agusaldasoro the guide you linked (and the [official documentation](https://docs.rs/anchor-lang/latest/anchor_lang/prelude/struct.AccountInfo.html#method.realloc)) bot specify that `realloc::zero = true` is only necessary in the cases where *a single instruction* can downsize and then upsize in one call. This is impossible in our case as we only reallocate once on entry. This means the bytes are automatically zeroed, and using `realloc::zero = true` would be wasted cycles.

agusaldasoro: Ahhh, okay! that's clear now. Can you please add a comment in code staying that so we don't forget about that in the future? thanks!

chains/solana/contracts/programs/ccip-router/src/state.rs
agusaldasoro: We estimated a max size of 300k which is okay

chains/solana/contracts/programs/ccip-router/src/context.rs
PabloMansanet: Sounds good! :)

agusaldasoro (approved):

aalu1418 (approved):

PR #521, opened 2025-01-29
solana: example receiver

aalu1418:
- [x] rename existing receivers with `test-` prefix
- [x] create new `example-ccip-receiver` as template
- [x] test cases for `example-ccip-receiver`
- [x] integrate `example-ccip-receiver` with `test-ccip-receiver`


chains/solana/gobindings/test_ccip_invalid_receiver/alias.go
aalu1418, line 1: this is not an auto-generated file
this alias file allows resolving the types that are declared in other contracts/packages

chains/solana/contracts/programs/example-ccip-receiver/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+    pub deny: ChainList,
+}
+
+#[derive(InitSpace, Debug, Clone, AnchorSerialize, AnchorDeserialize, Default)]
+pub struct ChainList {
+    pub is_enabled: bool,
+    pub xs: [u64; 20],
+    len: u8,
+}
+arrayvec!(ChainList, u64, u8);
```
PabloMansanet, line 164: I'm not sure if I'm a fan of this arrayvec, both as an abstraction and as its usage here.

As an abstraction it's pretty leaky because there's nothing stopping the user from just inspecting values that are conceptually "not allocated" (i.e. beyond the limit of `len`). Solving this would require making `xs` private, but I can't think of an easy way to enforce that limitation from the macro other than delegating the definition of the type to the macro itself, which leads me to the next issue...

I'm generally not a fan of code-generating macros of this kind as they significantly obscure what's going on. Particularly in this case, where we're writing template code for clients to follow, I think it pays off to be crystal clear. Introducing an opaque codegen macro here can be a pain point for readers of the template, especially if they lack experience with Rust, as they'll likely be unsure why it's needed.

As for what to do instead... After having to do it for two contracts, I haven't found the variable length accounts to be that painful in practice, but I might be missing previous negative experiences by the team. Do you think it would be a problem to simply have a `Vec` here and utilize a `realloc` constraint? I feel like it would have some teaching value for readers of the template, as they're likely to want to learn how to do it.

Alternatively, rather than the `arrayvec` macro, I'd prefer a dead simple array of `Option<u64>` with a basic "fill from the front, empty from the back" behaviour. WDYT?

PabloMansanet (commented):
Couldn't leave a full review as I have to stop for today, but wanted to just leave some thoughts on `arrayvec` for the time being. 

Either way, I'll have to base my offramps auth code in this PR, so I'll be prioritizing reviewing this asap 🙂.

chains/solana/contracts/programs/example-ccip-receiver/src/lib.rs
aalu1418: make sense to me! i will take a look at swapping it out with something that doesn't obscure functionality

PabloMansanet: I can also quickly put together an `arrayvec` just using generics + const generics, so we can just use `ArrayVec<u64, 20>`. I have the nagging suspicion that the go binding generator won't like it thought, and maybe that's why you haven't done that already 😆

aalu1418: i just changed it to a `Vec<u8>` with `realloc` let me know what you think about that

aalu1418: i ended doing a refactor on the chain constraints because the example contracts limit based on source chain + source sender

https://github.com/smartcontractkit/chainlink/blob/6137b5191e486b73d1b698dff650518a696a85f6/contracts/src/v0.8/ccip/applications/external/CCIPReceiver.sol#L54

PabloMansanet (commented):
Looks good! Just a minor comment fix and an open question on test determinism.

chains/solana/contracts/programs/example-ccip-receiver/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+        require_keys_neq!(router, Pubkey::default(), CcipReceiverError::InvalidRouter);
+        require_eq!(self.owner, owner, CcipReceiverError::OnlyOwner);
+        self.router = router;
+        Ok(())
+    }
+}
+
+#[account]
+#[derive(InitSpace, Default, Debug)]
+pub struct ApprovedSender {}
```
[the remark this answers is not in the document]
aalu1418: haha yeah - it's the solana way of unbounded data structures (make it a mappable PDA)

PabloMansanet (approved):

chains/solana/contracts/programs/example-ccip-receiver/src/lib.rs
archseer: > I have the nagging suspicion that the go binding generator won't like it thought, and maybe that's why you haven't done that already

Precisely yeah, we've had to keep the typing plain so that other tooling is able to inspect it

PR #522, opened 2025-01-30
Consistency in CCIP reader interface regarding destination chain

dimkouv:
Currently the reader interface and the main implementation are not consistent.

1. Sometimes the destination chain is provided as an argument and sometimes not.
2. In the implementation sometimes the value passed as an argument is used and sometimes the field of the struct. Even within the same function!
3. Sometimes we have validations whether the provided destination chain matches the struct dest chain and sometimes not.

To make the reader consistent I am removing the dest chain argument from all the methods.

The alternative approach was to provide the destination chain as parameter to every method and remove it from the state, but this seemed to be problematic, due to:
1) the destination chain is never changing, meaning that we always provide the same destination chain.
2) the destination chain might not be available to all the existing callers.
3) the destination chain is required to be provided to the reader constructor since it is also used for other purposes (e.g. offRamp address encoding).

Due to the issues mentioned above it seemed more straightforward and cleaner to go with the approach implemented in this PR.

Any new method added to the interface should follow the same approach.

Chainlink PR: https://github.com/smartcontractkit/chainlink/pull/16139
`core ref: 8dab162d78d38f5307046eb254b88b233772d130`

### DIFF
--- a/pkg/reader/ccip_interface.go
+++ b/pkg/reader/ccip_interface.go
@@ -91,9 +91,9 @@ type CCIPReader interface {
 		seqNumRange cciptypes.SeqNumRange,
 	) ([]cciptypes.SeqNum, error)
 
-	// MsgsBetweenSeqNums reads the provided chains.
-	// Finds and returns ccip messages submitted between the provided sequence numbers.
-	// Messages are sorted ascending based on their timestamp and limited up to the provided limit.
+	// MsgsBetweenSeqNums reads the provided chains, finds and returns ccip messages
+	// submitted between the provided sequence numbers. Messages are sorted ascending based on
+	// their timestamp and limited up to the provided limit.
 	MsgsBetweenSeqNums(
 		ctx context.Context,
 		chain cciptypes.ChainSelector,


pkg/reader/ccip.go
```diff
@@ -88,11 +88,11 @@ func (r *ccipChainReader) WithExtendedContractReader(
 }
 
 func (r *ccipChainReader) CommitReportsGTETimestamp(
-	ctx context.Context, dest cciptypes.ChainSelector, ts time.Time, limit int,
+	ctx context.Context, ts time.Time, limit int,
 ) ([]plugintypes2.CommitPluginReportWithMeta, error) {
 	lggr := logutil.WithContextValues(ctx, r.lggr)
 
-	if err := validateExtendedReaderExistence(r.contractReaders, dest); err != nil {
+	if err := validateExtendedReaderExistence(r.contractReaders, r.destChain); err != nil {
```
asoliman92, line 95: It makes sense now to check this in the constructor and fail the plugin initialization early. It doesn't make sense to have a plugin for a destChain without proper destChain reader anyways.

asoliman92 (approved):

PR #523, opened 2025-01-30
Fix reader panic cases

dimkouv:
While working on another reader related issue I discovered the following issues.

- Missing checks for reader existence which will lead to panic
- Path with an uninitialized address type where we call `nil.Append`

`core ref: 8dab162d78d38f5307046eb254b88b233772d130`


mateusz-sekara (dismissed):

pkg/reader/ccip.go
```diff
@@ -641,6 +641,11 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
 	lggr := logutil.WithContextValues(ctx, r.lggr)
 	feeUpdates := make(map[cciptypes.ChainSelector]plugintypes.TimestampedBig, len(selectors))
 	for _, chain := range selectors {
+		if _, ok := r.contractReaders[r.destChain]; !ok {
```
asoliman92, line 644: This condition should be moved before the loop. It's only checking the destChain not all the chains. if the destChain is not available no price updates can be retrieved. You can use the validate reader like in other ccip reader functions.

asoliman92 (approved):

mateusz-sekara (approved):

PR #526, opened 2025-01-30
Port curve-based gas deviation logic over from chainlink repo, add DisableChainFeeDeviationCheck

ogtownsend:
- This was added to `chainlink` in https://github.com/smartcontractkit/chainlink/commit/d2fa4bcbbfe783844238dcd33cb0813fac06d4d4
- Port it here as well (CCIP-4813)
- Also add `DisableChainFeeDeviationCheck` to `FeeInfo` so we can disable deviation checks on a lane-specific level. Since this param lives in `FeeInfo`, there is a value for each source chain configured for this dest chain.

core ref: efbc0805fc0920f68b214f7fa06db6b4d7b7965e

### DIFF
--- a/commit/chainfee/outcome.go
+++ b/commit/chainfee/outcome.go
@@ -253,9 +253,15 @@ func (p *processor) getGasPricesToUpdate(
 		if feeInfo == nil {
 			continue
 		}
+
 		ci, ok := feeInfo[chain]
 		if !ok {
 			lggr.Warnf("could not find fee info for chain %d", chain)
+			continue
+		}
+
+		if ci.DisableChainFeeDeviationCheck {
+			lggr.Debugf("chain fee deviation check disabled, skipping deviation check for chain %d", chain)
 			continue
 		}
 
--- a/commit/chainfee/outcome.go
+++ b/commit/chainfee/outcome.go
@@ -256,7 +256,7 @@ func (p *processor) getGasPricesToUpdate(
 			continue
 		}
 
-		gasDeviates := p.gasComponentsDeviate(currentChainFee, lastUpdate, ci)
+		gasDeviates := p.gasComponentsDeviate(lggr, chain, currentChainFee, lastUpdate, ci)
 		if gasDeviates {
 			gasPrices = append(gasPrices, cciptypes.GasPriceChain{
 				ChainSel: chain,
@@ -269,6 +269,8 @@ func (p *processor) getGasPricesToUpdate(
 }
 
 func (p *processor) gasComponentsDeviate(
+	lggr logger.Logger,
+	chain cciptypes.ChainSelector,
 	currentChainFee ComponentsUSDPrices,
 	lastUpdate Update,
 	ci pluginconfig.FeeInfo,
@@ -284,6 +286,23 @@ func (p *processor) gasComponentsDeviate(
 			currentChainFee.DataAvFeePriceUSD,
 			lastUpdate.ChainFee.DataAvFeePriceUSD,
 			p.cfg.DataAvNoDeviationThresholdUSDWei.Int,
+		)
+		// Log the deviation check info while we potentially still need to fine-tune the curve.
+		lggr.Debugf(
+			"using curve based deviation check for chain %d - "+
+				"executionFeeDeviates: %t, "+
+				"currentChainFee.executionFeePriceUSD: %s, "+
+				"lastUpdate.executionFeePriceUSD: %s, "+
+				"dataAvFeeDeviates: %t, "+
+				"currentChainFee.dataAvFeePriceUSD: %s, "+
+				"lastUpdate.dataAvFeePriceUSD: %s",
+			chain,
+			executionFeeDeviates,
+			currentChainFee.ExecutionFeePriceUSD.String(),
+			lastUpdate.ChainFee.ExecutionFeePriceUSD.String(),
+			dataAvFeeDeviates,
+			currentChainFee.DataAvFeePriceUSD.String(),
+			lastUpdate.ChainFee.DataAvFeePriceUSD.String(),
 		)
 	} else {
 		executionFeeDeviates = mathslib.Deviates(
--- a/commit/chainfee/outcome.go
+++ b/commit/chainfee/outcome.go
@@ -11,21 +11,12 @@ import (
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
 
 	"github.com/smartcontractkit/chainlink-ccip/internal/libs/mathslib"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
 
 	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon"
 	"github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/consensus"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/logutil"
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
-)
-
-const (
-	// ExecNoDeviationThresholdUSD is the lower bound no deviation threshold for exec gas. If the exec gas price is
-	// less than this value, we should never trigger a deviation. This is set to 10 gwei in USD terms.
-	ExecNoDeviationThresholdUSD = 10e9
-
-	// DataAvNoDeviationThresholdUSD is the lower bound no deviation threshold for DA gas. If the DA gas price is less
-	// than this value, we should never trigger a deviation. This is set to 20 gwei in USD terms.
-	DataAvNoDeviationThresholdUSD = 20e9
 )
 
 func (p *processor) Outcome(
@@ -265,21 +256,8 @@ func (p *processor) getGasPricesToUpdate(
 			continue
 		}
 
-		executionFeeDeviates := mathslib.DeviatesOnCurve(
-			currentChainFee.ExecutionFeePriceUSD,
-			lastUpdate.ChainFee.ExecutionFeePriceUSD,
-			big.NewInt(ExecNoDeviationThresholdUSD),
-			ci.ExecDeviationPPB.Int64(),
-		)
-
-		dataAvFeeDeviates := mathslib.DeviatesOnCurve(
-			currentChainFee.DataAvFeePriceUSD,
-			lastUpdate.ChainFee.DataAvFeePriceUSD,
-			big.NewInt(DataAvNoDeviationThresholdUSD),
-			ci.DataAvailabilityDeviationPPB.Int64(),
-		)
-
-		if executionFeeDeviates || dataAvFeeDeviates {
+		gasDeviates := p.gasComponentsDeviate(currentChainFee, lastUpdate, ci)
+		if gasDeviates {
 			gasPrices = append(gasPrices, cciptypes.GasPriceChain{
 				ChainSel: chain,
 				GasPrice: packedFee,
@@ -288,6 +266,38 @@ func (p *processor) getGasPricesToUpdate(
 	}
 
 	return gasPrices
+}
+
+func (p *processor) gasComponentsDeviate(
+	currentChainFee ComponentsUSDPrices,
+	lastUpdate Update,
+	ci pluginconfig.FeeInfo,
+) bool {
+	var executionFeeDeviates, dataAvFeeDeviates bool
+	if p.cfg.CurveBasedGasDeviationEnabled {
+		executionFeeDeviates = mathslib.DeviatesOnCurve(
+			currentChainFee.ExecutionFeePriceUSD,
+			lastUpdate.ChainFee.ExecutionFeePriceUSD,
+			p.cfg.ExecNoDeviationThresholdUSDWei.Int,
+		)
+		dataAvFeeDeviates = mathslib.DeviatesOnCurve(
+			currentChainFee.DataAvFeePriceUSD,
+			lastUpdate.ChainFee.DataAvFeePriceUSD,
+			p.cfg.DataAvNoDeviationThresholdUSDWei.Int,
+		)
+	} else {
+		executionFeeDeviates = mathslib.Deviates(
+			currentChainFee.ExecutionFeePriceUSD,
+			lastUpdate.ChainFee.ExecutionFeePriceUSD,
+			ci.ExecDeviationPPB.Int64(),
+		)
+		dataAvFeeDeviates = mathslib.Deviates(
+			currentChainFee.DataAvFeePriceUSD,
+			lastUpdate.ChainFee.DataAvFeePriceUSD,
+			ci.DataAvailabilityDeviationPPB.Int64(),
+		)
+	}
+	return executionFeeDeviates || dataAvFeeDeviates
 }
 
 // chainFeeUpdateAggregator aggregates a slice of ChainFeeUpdates into a single Update
--- a/commit/chainfee/outcome.go
+++ b/commit/chainfee/outcome.go
@@ -18,6 +18,16 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
+const (
+	// ExecNoDeviationThresholdUSD is the lower bound no deviation threshold for exec gas. If the exec gas price is
+	// less than this value, we should never trigger a deviation. This is set to 10 gwei in USD terms.
+	ExecNoDeviationThresholdUSD = 10e9
+
+	// DataAvNoDeviationThresholdUSD is the lower bound no deviation threshold for DA gas. If the DA gas price is less
+	// than this value, we should never trigger a deviation. This is set to 20 gwei in USD terms.
+	DataAvNoDeviationThresholdUSD = 20e9
+)
+
 func (p *processor) Outcome(
 	ctx context.Context,
 	_ Outcome,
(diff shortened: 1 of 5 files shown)


commit/chainfee/outcome.go
```diff
@@ -18,6 +18,16 @@ import (
...
 )
 
+const (
+	// ExecNoDeviationThresholdUSD is the lower bound no deviation threshold for exec gas. If the exec gas price is
+	// less than this value, we should never trigger a deviation. This is set to 10 gwei in USD terms.
+	ExecNoDeviationThresholdUSD = 10e9
+
+	// DataAvNoDeviationThresholdUSD is the lower bound no deviation threshold for DA gas. If the DA gas price is less
+	// than this value, we should never trigger a deviation. This is set to 20 gwei in USD terms.
+	DataAvNoDeviationThresholdUSD = 20e9
```
matYang, line 28: can probably turn these into config values

internal/libs/mathslib/calc.go
```diff
@@ -1,9 +1,16 @@
...
 
 import (
+	"math"
 	"math/big"
 )
 
+const (
+	// CurveBasedDeviationPPB is the deviation threshold when writing to Ethereum's PriceRegistry and is the trigger for
+	// using curve-based deviation logic.
+	CurveBasedDeviationPPB = 4e9
```
matYang, line 11: this was a quick & dirty solution on 1.5, for 1.6 it can def become a config

ogtownsend: yup, will update 👍

internal/libs/mathslib/calc.go
```diff
@@ -29,3 +30,46 @@ func CalculateUsdPerUnitGas(sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) *bi
...
+	const exponent = 0.665
+	xPower := math.Pow(x, exponent)
+	threshold := constantFactor / xPower
+
+	// Convert curve output percentage to PPB
+	thresholdPPB := int64(threshold * 1e7)
+	return thresholdPPB
+}
+
+func MergeEpochAndRound(epoch uint32, round uint8) uint64 {
```
asoliman92, line 73: Not used?

ogtownsend: 😱🍝

internal/libs/mathslib/calc.go
```diff
@@ -40,13 +34,7 @@ func CalculateUsdPerUnitGas(sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) *bi
...
-func DeviatesOnCurve(xNew, xOld, noDeviationLowerBound *big.Int, ppb int64) bool {
-	// This is a temporary gating mechanism that ensures we only apply the gas curve deviation logic to eth-bound price
-	// updates. If ppb from config is not equal to 4000000000, do not apply the gas curve.
-	if ppb != CurveBasedDeviationPPB {
-		return Deviates(xOld, xNew, ppb)
-	}
-
+func DeviatesOnCurve(xNew, xOld, noDeviationLowerBound *big.Int) bool {
 	// If xNew < noDeviationLowerBound, Deviates should never be true
 	if xNew.Cmp(noDeviationLowerBound) < 0 {
```
asoliman92, line 39: should we also add an OR condition on xOld?

pluginconfig/commit.go
```diff
@@ -148,6 +148,18 @@ type CommitOffchainConfig struct {
 
 	// ChainFeeAsyncObserverSyncTimeout defines the timeout for a single sync operation (e.g. fetch token prices).
 	ChainFeeAsyncObserverSyncTimeout time.Duration `json:"chainFeeAsyncObserverSyncTimeout"`
+
+	// ExecNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for exec gas. If the exec gas price is
+	// less than this value, we should never trigger a deviation. A value of 5e9 would correspond to 0.000000005 USD.
+	ExecNoDeviationThresholdUSDWei cciptypes.BigInt `json:"execNoDeviationThresholdUSDWei"`
```
asoliman92, line 154: There's a validation and applyDefaults [functions](https://github.com/smartcontractkit/chainlink-ccip/blob/dd0db587bff45ae96b21e8c40b5c937f34910995/pluginconfig/commit.go#L239-L242), can you  add validation and apply the proper defaults for these values?

ogtownsend: ah nice, will add

asoliman92 (approved):

pluginconfig/commit.go
```diff
@@ -213,6 +215,18 @@ func (c *CommitOffchainConfig) applyDefaults() {
 			c.ChainFeeAsyncObserverSyncTimeout = defaultAsyncObserverSyncTimeout
 		}
 	}
+
+	// Only apply defaults if the curve based deviation is disabled. Otherwise, these values should already be set in
```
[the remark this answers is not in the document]
ogtownsend: @asoliman92 well, how are these defaults usually applied? It seems like they're applied AFTER the factory loads the actual config via `offchainConfig, err := pluginconfig.DecodeCommitOffchainConfig(config.OffchainConfig)` and `ApplyDefaultsAndValidate()` is used to fill in any values that the config doesn't include. 

If `CurveBasedGasDeviationEnabled` is true, then `ExecNoDeviationThresholdUSDWei` and `DataAvNoDeviationThresholdUSDWei` should both be set already, in other words, if it's true then we don't want to override their values with defaults. 

If `CurveBasedGasDeviationEnabled` is false, then I set the defaults to avoid a nil pointer just in case someone accidentally uses `DataAv/ExecNoDeviationThresholdUSDWei`.

However, maybe it would be safest to change this to just:
```
if c.ExecNoDeviationThresholdUSDWei.Int == nil {
	c.ExecNoDeviationThresholdUSDWei = cciptypes.BigInt{Int: big.NewInt(defaultExecNoDeviationThresholdUSDWei)}
}

if c.DataAvNoDeviationThresholdUSDWei.Int == nil {
	c.DataAvNoDeviationThresholdUSDWei = cciptypes.BigInt{Int: big.NewInt(defaultDataAvNoDeviationThresholdUSDWei)}
}
```

That way they'll never be nil even if someone accidentally sets `CurveBasedGasDeviationEnabled = true` and forgets to set the two threshold values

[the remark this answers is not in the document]
ogtownsend: updated

asoliman92 (dismissed):

pluginconfig/commit.go
```diff
@@ -170,6 +173,18 @@ type CommitOffchainConfig struct {
...
+	// less than this value, we should never trigger a deviation. A value of 5e9 would correspond to 0.000000005 USD.
+	ExecNoDeviationThresholdUSDWei cciptypes.BigInt `json:"execNoDeviationThresholdUSDWei"`
+
+	// DataAvNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for DA gas. If the DA gas price is
+	// less than this value, we should never trigger a deviation. A value of 5e9 would correspond to 0.000000005 USD.
+	DataAvNoDeviationThresholdUSDWei cciptypes.BigInt `json:"dataAvNoDeviationThresholdUSDWei"`
+
+	// CurveBasedGasDeviationEnabled determines whether or not we should use the deviation curve or the fixed PPB values
+	// when determining whether or not the gas price has deviated.
+	CurveBasedGasDeviationEnabled bool `json:"curveBasedGasDeviationEnabled"`
```
makramkd, line 187: Is this kinda feature flag really necessary? Why not cut over to this approach right away?

pluginconfig/commit.go
```diff
@@ -170,6 +173,18 @@ type CommitOffchainConfig struct {
...
 	// NOTE: this can only be used if RMNEnabled == false.
 	MultipleReportsEnabled bool `json:"multipleReports"`
+
+	// ExecNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for exec gas. If the exec gas price is
+	// less than this value, we should never trigger a deviation. A value of 5e9 would correspond to 0.000000005 USD.
+	ExecNoDeviationThresholdUSDWei cciptypes.BigInt `json:"execNoDeviationThresholdUSDWei"`
+
+	// DataAvNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for DA gas. If the DA gas price is
+	// less than this value, we should never trigger a deviation. A value of 5e9 would correspond to 0.000000005 USD.
+	DataAvNoDeviationThresholdUSDWei cciptypes.BigInt `json:"dataAvNoDeviationThresholdUSDWei"`
```
makramkd, line 183: Can you please comment on when these should be used vs. ExecDeviationPPB and DataAvailabilityDeviationPPB

makramkd: These seem redundant to me and makes it even more confusing to figure out what values to set to get the behavior desired.

makramkd: Also, this value is applied irrespective of source chain? We have a map `FeeInfo` to be able to have deviation values per-source.

pluginconfig/commit.go
```diff
@@ -29,11 +29,14 @@ const (
...
+	defaultExecNoDeviationThresholdUSDWei     = 10e9
+	defaultDataAvNoDeviationThresholdUSDWei   = 20e9
 )
 
 type FeeInfo struct {
-	ExecDeviationPPB             cciptypes.BigInt `json:"execDeviationPPB"`
-	DataAvailabilityDeviationPPB cciptypes.BigInt `json:"dataAvailabilityDeviationPPB"`
+	ExecDeviationPPB              cciptypes.BigInt `json:"execDeviationPPB"`
+	DataAvailabilityDeviationPPB  cciptypes.BigInt `json:"dataAvailabilityDeviationPPB"`
+	DisableChainFeeDeviationCheck bool             `json:"disableChainFeeDeviationCheck"`
```
makramkd, line 39: For consistency's sake
```suggestion
	ChainFeeDeviationDisabled bool             `json:"disableChainFeeDeviationCheck"`
```

internal/libs/mathslib/calc.go
```diff
@@ -29,3 +30,42 @@ func CalculateUsdPerUnitGas(sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) *bi
 	tmp := new(big.Int).Mul(sourceGasPrice, usdPerFeeCoin)
 	return tmp.Div(tmp, big.NewInt(1e18))
 }
+
+// DeviatesOnCurve calculates a deviation threshold on the fly using xNew. For now it's only used for gas price
+// deviation calculation. It's important to make sure the order of xNew and xOld is correct when passed into this
+// function to get an accurate deviation threshold.
+func DeviatesOnCurve(xNew, xOld, noDeviationLowerBound *big.Int) bool {
```
makramkd, line 37: Docstring should at least describe on a high level when this function should be used if there are alternatives (i.e whats the difference b/w this and `Deviates` above?) Is there some wikipedia article or something you can reference here to make readers understand the math below?

internal/libs/mathslib/calc.go
```diff
@@ -29,3 +30,42 @@ func CalculateUsdPerUnitGas(sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) *bi
...
+	// If xNew < noDeviationLowerBound, Deviates should never be true
+	if xNew.Cmp(noDeviationLowerBound) < 0 {
+		return false
+	}
+
+	xNewFloat := new(big.Float).SetInt(xNew)
+	xNewFloat64, _ := xNewFloat.Float64()
```
makramkd, line 44: This code is kind of odd, why not just use `BigFloat` everywhere and have peace of mind when it comes to these large gas price numbers and loss of precision / accuracy that compounds when doing floating point arithmetic?

commit/chainfee/outcome.go
```diff
@@ -243,25 +244,20 @@ func (p *processor) getGasPricesToUpdate(
...
-			ci.ExecDeviationPPB.Int64(),
-		)
-
-		dataAvFeeDeviates := mathslib.Deviates(
-			currentChainFee.DataAvFeePriceUSD,
-			lastUpdate.ChainFee.DataAvFeePriceUSD,
-			ci.DataAvailabilityDeviationPPB.Int64(),
-		)
+		if ci.DisableChainFeeDeviationCheck {
+			lggr.Debugf("chain fee deviation check disabled, skipping deviation check for chain %d", chain)
```
makramkd, line 255: Please use structured logging where possible, so `Debugw(..., "chain", chain)`

commit/chainfee/outcome.go
```diff
@@ -272,6 +268,57 @@ func (p *processor) getGasPricesToUpdate(
...
+			lastUpdate.ChainFee.ExecutionFeePriceUSD,
+			p.cfg.ExecNoDeviationThresholdUSDWei.Int,
+		)
+		dataAvFeeDeviates = mathslib.DeviatesOnCurve(
+			currentChainFee.DataAvFeePriceUSD,
+			lastUpdate.ChainFee.DataAvFeePriceUSD,
+			p.cfg.DataAvNoDeviationThresholdUSDWei.Int,
+		)
+		// Log the deviation check info while we potentially still need to fine-tune the curve.
+		lggr.Debugf(
```
makramkd, line 291: Structured log plz

makramkd (changes requested):

pluginconfig/commit.go
ogtownsend: We haven't had a chance to test it in v1.5 yet since the latest release has been delayed. There's a chance it ends up being more costly than fixed PPB deviation thresholds. cc @matYang

pluginconfig/commit.go
ogtownsend: Would naming them something like `CurveExecNoDeviationThresholdUSDWei` and `CurveDataAvNoDeviationThresholdUSDWei` help clear things up?

internal/libs/mathslib/calc.go
ogtownsend: We have an internal confluence page on it, there's also this public page but I'm not sure it adds much value? https://en.wikipedia.org/wiki/Power_law

I'll add a sentence to clarify the difference this and `Deviates()`

pluginconfig/commit.go
ogtownsend: > Can you please comment on when these should be used vs. ExecDeviationPPB and DataAvailabilityDeviationPPB

Two reasons for why these are separated:
1) To differentiate the two deviation approaches:
    1) `ExecDeviationPPB` and `DataAvailabilityDeviationPPB` should only be used when applying fixed-PPB thresholds 
    2) `ExecNoDeviationThresholdUSDWei` and `DataAvNoDeviationThresholdUSDWei` should only be used when applying the curve deviation logic, though I agree these could probably have better names and/or reference "Curve" in the name 
2) They're used slightly differently:
    - `ExecDeviationPPB` and `DataAvailabilityDeviationPPB` are percentages. If (gasNew - gasOld)/(gasOld) > PPB, then that's a deviation. 
    - `ExecNoDeviationThresholdUSDWei` and `DataAvNoDeviationThresholdUSDWei` are in USD terms. This allows us to specify an exact dollar value at which point we don't apply the curve and return `deviates=false` early. This allows us to say "when the per unit gas price is lower than this 'no deviation' USD amount, we don't care, don't bother with a price update. Even if gasNew is 10,000x gasOld, if it's 1000x of $0.000000000001, then so what. This is needed because while the curve is fit pretty well in the middle range, it doesn't fit as well at the most extreme points (especially the lower ends).

Take sending a message from Eth --> Zircuit as an example, in this case "Eth" is the dest chain from the plugin's perspective. Using some historical zircuit per unit gas price data, within ~20 minutes Zircuit's per unit usdwei gas price went from `50417041346 --> 1044061 --> 51345772224 --> 3839401503438`. 
- `50417041346 --> 1044061` is a 48,000x (would've triggered `ExecDeviationPPB`)
- `1044061 --> 51345772224` is a 49,000x (would've triggered `ExecDeviationPPB`)
- `3839401503438 / 51345772224` is a 75x (would've triggered `ExecDeviationPPB`)
- but `50417041346 --> 1044061 --> 51345772224` == `51345772224 / 50417041346` is just a ~1x
- `1044061` is so negligible so if we had our `ExecNoDeviationThresholdUSDWei = 1000000000 = $0.000000001`, we would've saved two zircuit-caused deviation updates to ethereum

ogtownsend: > Also, this value is applied irrespective of source chain? We have a map FeeInfo to be able to have deviation values per-source.

- I thought we were too concerned about quadratic config / config explosion? If you don't mind, then yes adding them to `FeeInfo` would give us more much more finely-tuned control (this was actually our original thought).
- That said, another way to explain the usage of `Exec/DaNoDeviationThresholdUSDWei` is: `the per unit gas cost at which it becomes more expensive to report the price to this dest chain's price registry than it is to take any loss during ccipSend getFee due to gas inaccuracy`. If you look at it that way, then the param is sort of a function of the dest chain only. And since we were approaching this problem from the perspective of reducing the cost of price updates to a chains FeeQuoter, we made it dest-only to start.
- That said, if we start finding "per unit gas price" can have large impacts across different chains, especially in non-EVMs (solana? cc @matYang) in the future, then the fine tune control might be worth the added complexity.

makramkd: I get that this approach may be more precise, but I believe we're at a point where simplifying the architecture is actually massively better for chain expansion, and therefore better for CCIP, even if there may be short periods of time where the fee calculations are potentially suboptimal and imprecise.

PR #527, opened 2025-01-31
Use GetSourceChainConfig instead of OffRampGetAllSourceChainConfigs

reductionista:
There isn't any efficient way to implement OffRampGetAllSourceChainConfigs on Solana, due to the information being spread across many different Accounts while not having the seeds passed as input params to determine the PDA's for those accounts. GetSourceChainConfig works on both evm & solana

Note:
We could do it using getProgramAccounts, but this requires account indexing to be enabled on the rpc server, doubling memory requirements to run it. We could also get around it by storing a list of the PDA's in an account with a fixed prefix, but this would require us to have a maximum limit on the number of chains we support. If the maximum limit was too high, it would waste costly on-chain storage space without actually using it. Too low, and we'd have to update the smart contracts when we hit the limit.


prashantkumar1982 (approved):
Looks good to me, but I don't have too much context on this codebase.
So please get some approvals from others too, who know this code.

pkg/reader/ccip.go
```diff
@@ -1146,117 +1148,76 @@ func (scc sourceChainConfig) check() (bool /* enabled */, error) {
...
-	err := r.contractReaders[r.destChain].ExtendedGetLatestValue(
-		ctx,
-		consts.ContractNameOffRamp,
-		consts.MethodNameOffRampGetAllSourceChainConfigs,
-		primitives.Unconfirmed,
-		map[string]any{},
-		&resp,
+	results, _, err := r.contractReaders[r.destChain].ExtendedBatchGetLatestValues(
+		ctx, contractreader.ExtendedBatchGetLatestValuesRequest{consts.ContractNameOffRamp: contractBatch},
+		false,
```
winder, line 1182: Previously this was using an `Unconfirmed` confidence level. What level is used with batch get?

reductionista: It's ignored anyway on Solana. Looks like it does default to `Unconfirmed` on EVM, and the possibility of overriding the default looks mostly implemented. Not sure why they didn't expose that feature in the API:
https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/read/batch.go#L132-L140

https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/read/bindings.go#L177

PR #538, opened 2025-01-31
New factory constructors with stable function signature.

winder:
By using a parameter object, we can freely modify the factory inputs without breaking compatibility between chainlink-ccip and chainlink repositories.

core ref: 7ac42f05240ec84385a88cad25f96c9c8d6aaa08


asoliman92 (approved):
can you clean the old functions as well? no need to do them in another PR

PR #540, opened 2025-02-01
Refactor: Break fee quoter into a separate program [NONEVM-1232]

toblich:
Separates Fee Quoter into its own program. The flow of calls now is:
- The offramp makes a single CPI to the FeeQuoter on commits with price updates.
- Users will invoke fee quoter directly to get the fee for a message. 
- The onramp makes the same call to know how much to charge. This same call already includes the parsed extra args, saving 1 cpi. However, it's the onramp itself who makes the transfers and who accumulates the funds, thus preventing the fee quoter to be the one who invokes the token program, and therefore sparing extra depth in the CPI stack (which solana limits to 4).
- To register new billing tokens or update their config, the admin has to invoke fee quoter directly
- To register new destination chains, the admin has to invoke both the onramp and the fee quoter
- To disable a destination chain, the admin has to invoke just fee quoter (not the router). This is similar to how EVM works too.

I've also dropped support to remove a billing token. While it could be done, it would require more coupling of the onramp & fee quoter, as fee quoter manages the lifecycle of the accumulating token account, which is an ATA tied to the router. So, fee quoter cannot close it the way it is today. 

### TODOs (may be done in this PR or a separate one)
- [x] Deduplicate configs for dest chains from onramp and fee quoter. This might even allow us to not have to invoke both to setup a new dest chain
- [ ] Remove dependency on ethnum for the onramp
- [ ] Remove unused errors in FeeQuoter, and fix FQ's error type gobinding generation


chains/solana/contracts/programs/ccip-router/src/instructions/v1/mod.rs
```diff
@@ -10,10 +10,9 @@ pub mod token_admin_registry;
 // Private modules //
 /////////////////////
 mod config;
-mod fee_quoter;
+mod fees;
```
PabloMansanet, line 13: Any particular reason for this rename? Not opposed to it, but it loses a bit of connection with the EVM codebase.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -99,49 +57,22 @@ pub fn ccip_send<'info>(
...
-                accs.token_billing_config,
-                accs.mint.key(),
-                dest_chain_selector,
-            )
-        })
-        .collect::<Result<Vec<_>>>()?;
-
-    let (fee, processed_extra_args) = fee_for_msg(
+    // Fee Quoter validates the message, calculates the fee in both the billing token
+    // and LINK, asserts that it doesn't exceed a maximum, and returns it
```
PabloMansanet, line 61: Nit: Since this comment is just describing what the function does, maybe it would be better as a doc comment of `get_fee_cpi` instead.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -275,32 +215,31 @@ pub fn ccip_send<'info>(
     Ok(())
 }
 
+/////////////
+// Helpers //
+/////////////
+
```
PabloMansanet, line 221: Nit: `pub mod helpers` would be nicer 🙏

chains/solana/contracts/programs/fee-quoter/src/instructions/v1/admin.rs
```diff
@@ -0,0 +1,175 @@
...
+//     ctx.accounts.config.onramp = onramp;
+//     // TODO emit event
+//     Ok(())
+// }
+
+// pub fn update_offramp(ctx: Context<UpdateConfig>, offramp: Pubkey) -> Result<()> {
+//     ctx.accounts.config.offramp = offramp;
+//     // TODO emit event
+//     Ok(())
+// }
```
PabloMansanet, line 51: Commented out code (in case you missed it).

chains/solana/contracts/programs/fee-quoter/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    ///
+    /// * First, the billing token config accounts for each token sent with the message, sequentially.
+    ///   For each token with no billing config account (i.e. tokens that cannot be possibly used as fee
+    ///   tokens, which also have no BPS fees enabled) the ZERO address must be provided instead.
+    /// * Then, the per chain / per token config of every token sent with the message, sequentially
+    ///   in the same order.
+    ///
+    /// # Returns
+    ///
+    /// The fee amount in u64.
```
PabloMansanet, line 204: It now returns a lot more than this: this section should be updated.

PabloMansanet (commented):
Looking good! Very good work 😃.

Just left a few minor comments.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/mod.rs
toblich: Because fee quoter is now a separate program altogether, and this file is doing other things than just invoking it (e.g. wrapping native tokens and actually transferring tokens), I renamed it to prevent confusion. 
The thing that EVM calls "Fee Quoter" is our Fee Quoter program; this submodule of the router is not it. I'm happy to rename it to something else if you have any suggested names 😄

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
```diff
@@ -275,521 +215,239 @@ pub fn ccip_send<'info>(
...
-        nonce_counter_account.counter = 0;
+        Ok(final_nonce)
     }
 
-    // config.enforce_out_of_order checked in `validate_svm2any`
-    let mut final_nonce = 0;
-    if !allow_out_of_order_execution {
-        nonce_counter_account.counter = nonce_counter_account.counter.checked_add(1).unwrap();
-        final_nonce = nonce_counter_account.counter;
+    pub(super) fn hash(msg: &SVM2AnyRampMessage) -> [u8; 32] {
```
agusaldasoro, line 276: Did the hash change or is just moving code?

agusaldasoro (dismissed):

chains/solana/contracts/programs/ccip-router/src/instructions/v1/onramp.rs
toblich: Just moved it to remove an entire submodule after most code belonged to fee quoter

agusaldasoro (approved):

PR #543, opened 2025-02-03
[CCIP-4928] Harden DiscoveryProcessor validation

asoliman92:
* Add validation for FChain in DiscoveryProcessor
* Add tests
* Make it clearer that offRamp configs are only read from DestChain


internal/plugincommon/discovery/processor.go
```diff
@@ -94,6 +99,17 @@ func (cdp *ContractDiscoveryProcessor) ValidateObservation(
...
+	err = plugincommon.ValidateFChain(ao.Observation.FChain)
+	if err != nil {
+		return fmt.Errorf("invalid FChain: %w", err)
+	}
+
+	if !supportedChains.Contains(cdp.dest) {
+		return fmt.Errorf(
+			"oracle %d is not allowed to observe destination chain %s, can't observe any contracts",
+			ao.OracleID, cdp.dest)
+	}
```
winder, line 111: The source chains also have addresses to discover (FeeQuote/Router from onramp)

internal/plugincommon/discovery/processor.go
```diff
@@ -105,18 +121,21 @@ func (cdp *ContractDiscoveryProcessor) ValidateObservation(
 					return fmt.Errorf(
 						"oracle %d is not allowed to observe chain %s for %s", ao.OracleID, chain, contract)
 				}
+				if chain == cdp.dest {
+					continue
+				}
```
winder, line 126: Why? Don't we need the router to find the fee token?

internal/plugincommon/discovery/processor.go
```diff
@@ -105,18 +121,21 @@ func (cdp *ContractDiscoveryProcessor) ValidateObservation(
...
+				_, err = (*cdp.reader).GetContractAddress(consts.ContractNameOnRamp, chain)
+				if err != nil {
+					return fmt.Errorf(
+						"oracle %d must observe onramp contract on chain %s before observing %s",
+						ao.OracleID, chain, contract)
+				}
```
winder, line 132: Why? This seems like an implementation detail which doesn't need to be here.

internal/plugincommon/discovery/processor.go
```diff
@@ -105,18 +121,21 @@ func (cdp *ContractDiscoveryProcessor) ValidateObservation(
...
 		case consts.ContractNameOnRamp,
 			consts.ContractNameOffRamp,
 			consts.ContractNameNonceManager,
 			consts.ContractNameRMNRemote:
-
-			if !supportedChains.Contains(cdp.dest) {
-				return fmt.Errorf(
-					"oracle %d is not allowed to observe contract (%s) on the destination chain %s",
-					ao.OracleID, contract, cdp.dest)
-			}
```
winder, line 119: This should come back, I think moving the check above is wrong.

internal/plugincommon/discovery/processor.go
asoliman92: To actually discover what you linked we need to read offRamp (destChain) to read onRamp from which we then read the feeQuoter and Router from. That said, having role don and each node can read some parts, it can be that the destChain nodes already reports all onRamp addresses so a node without destChain reader can then follow up with the rest of the discoveries.

internal/plugincommon/discovery/processor.go
asoliman92: Agreed.

PR #545, opened 2025-02-03
Replace JSON with Protobuf for OCR related types (Refactoring PR)

dimkouv:
This PR is performing all the plumbing required for a smooth transition from JSON to Protobuf for OCR related types encoding (Query, Observation, Outcome).

1. Create `Codec` interfaces for both Commit and Exec plugins and implement using JSON by copying the existing logic as is.
2. Replace the existing method calls with the codec everywhere in the codebase. `outcome.Encode() -> codec.EncodeOutcome(outcome)`
3. Remove plugin specific code from OCR3Runner. `WARN` The OCR3Runner is a generic component used to simulate OCR3 and make it easy to write unit tests. Being generic means that we should never put plugin specific logic. Some exec related logic was added there which was causing import loops. Also it was very likely for this code to be called by other non-exec plugins leading to errors.

### DIFF
--- a/chains/solana/Makefile
+++ b/chains/solana/Makefile
@@ -56,5 +56,11 @@ ccip-router-contracts-go-tests:
 build-contracts:
 	cd ./contracts && anchor build
 
+# Note on CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+# Sparse index significantly speeds up dependency fetching, supposedly default in 1.70 but anchor build was still using a slow git download
+.PHONY: docker-build-contracts
+docker-build-contracts:
+	docker run --rm -it -v $(shell pwd)/contracts:/workdir -e CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse ${ANCHOR_IMAGE} anchor build
+
 .PHONY: solana-checks
 solana-checks: clippy anchor-go-gen format gomodtidy lint-go rust-tests go-tests build-contracts


execute/exectypes/outcome.go
```diff
@@ -77,12 +74,12 @@ func NewOutcome(
 	selectedCommits []CommitData,
 	report cciptypes.ExecutePluginReport,
 ) Outcome {
-	return newSortedOutcome(state, selectedCommits, report)
+	return NewSortedOutcome(state, selectedCommits, report)
 }
 
-// newSortedOutcome ensures canonical ordering of the outcome.
+// NewSortedOutcome ensures canonical ordering of the outcome.
 // TODO: handle canonicalization in the encoder.
```
winder, line 81: I probably put this TODO here. You made an encoder, I think it makes sense to move the sort there.

asoliman92 (approved):

pkg/ocrtypecodec/commit.go
```diff
@@ -0,0 +1,82 @@
...
+	err := json.Unmarshal(data, &q)
+	return q, err
+}
+
+func (*CommitCodecJSON) EncodeObservation(observation committypes.Observation) ([]byte, error) {
+	encodedObservation, err := json.Marshal(observation)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode Observation: %w", err)
+	}
+	return encodedObservation, nil
```
makramkd, line 48: Can this just be `return json.Marshal(observation)`?

pkg/ocrtypecodec/commit.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (*CommitCodecJSON) EncodeOutcome(outcome committypes.Outcome) ([]byte, error) {
+	// Sort all lists to ensure deterministic serialization
+	outcome.MerkleRootOutcome.Sort()
+	encodedOutcome, err := json.Marshal(outcome)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode Outcome: %w", err)
+	}
+
+	return encodedOutcome, nil
```
makramkd, line 68: Same here, `return json.Marshal(outcome)`. Or is it the error msg that is more useful?

makramkd (commented):
I dunno if we're gonna keep the JSON codec around, but just in case for future PRs, some minor simplifications.

PR #546, opened 2025-02-03
Introducing ExtendedBatchGetLatestValuesGraceful

0xnogo:
This will be needed to batch the config call in the ccip reader as some bindings might not be done when we execute the batch call.


pkg/contractreader/extended.go
```diff
@@ -84,6 +84,21 @@ type Extended interface {
 		ctx context.Context,
 		request ExtendedBatchGetLatestValuesRequest,
 	) (types.BatchGetLatestValuesResult, error)
+
+	// ExtendedBatchGetLatestValuesGraceful performs the same operations as ExtendedBatchGetLatestValues
+	// but handles ErrNoBindings gracefully by:
+	// 1. Skipping contracts with no bindings instead of returning an error
+	// 2. Returning partial results for contracts that do have bindings
+	// 3. Including information about which contracts were skipped due to no bindings
+	ExtendedBatchGetLatestValuesGraceful(
```
winder, line 93: Are we using `ExtendedBatchGetLatestValues` anywhere? Maybe we can add this functionality to the existing function instead of creating a new one.

0xnogo: We are using it in ccip reader in `Nonces()`. I can create a boolean flag considering that the function is only used once. But on the other side we would have to return an empty`SkippedNoBinds`.

PR #547, opened 2025-02-03
Use the root snooze time config.

winder:


pluginconfig/execute.go
```diff
@@ -27,7 +27,7 @@ type ExecuteOffchainConfig struct {
 	// The caching prevents us from issuing another report while one is already in flight.
 	InflightCacheExpiry commonconfig.Duration `json:"inflightCacheExpiry"`
 
-	// RootSnoozeTime is the interval at which we check roots for executable messages.
+	// RootSnoozeTime is the delay duration used when a commit root is snoozed.
```
winder, line 30: It looks like this configuration served a different purpose for the OCR2 plugin. Is that old behavior what we want?

asoliman92 (approved):

pluginconfig/execute.go
asoliman92: CC @mateusz-sekara

makramkd: I think this is just a typo, I believe I ported these over last year, could be a copy pasta.

makramkd (approved):

makramkd (approved):

PR #548, opened 2025-02-03
Add inflight message cache.

winder:
Adds a simple inflight message cache. Filter inflight messages from reports.


execute/internal/cache/inflight_message_cache.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+
+	"github.com/patrickmn/go-cache"
+
+	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+)
+
+type InflightMessageCache struct {
+	// inflight is a cache of messages that are currently in flight. This cache
+	// is used to prevent duplicate reports from being sent for the same message.
```
makramkd, line 13: Can put this comment on the struct instead of the unexported field

execute/plugin.go
```diff
@@ -472,6 +482,16 @@ func (p *Plugin) Reports(
...
 
+	// Update cache.
+	// Note: this could wait until should transmit, but the next round may start before then.
+	// For that reason we update caches here. If there is some sort of failure to transmit,
+	// messages will be delayed until the inflight cache expires and messages are reprocessed.
+	for _, chainReport := range decodedOutcome.Report.ChainReports {
+		for _, message := range chainReport.Messages {
+			p.inflightMessageCache.MarkInflight(chainReport.SourceChainSelector, message.Header.MessageID)
+		}
+	}
```
makramkd, line 493: Reports() and Outcome() are supposed to be pure, I feel like we're straying farther from that and this might lead to bugs due to the OCR protocol assuming they're pure when they're not

makramkd: I think this logic was maybe fine in OCR2 but in OCR3 seems like a no-no, perhaps we can ask research for help on this.

execute/observation.go
```diff
@@ -173,6 +182,8 @@ func (p *Plugin) getCommitReportsObservation(
 		return exectypes.Observation{}, err
 	}
 
+	// TODO: message from fullyExecutedCommits which are in the inflight messages cache could be cleared here.
```
makramkd, line 185: Still a relevant TODO?

winder: Still relevant. We need some extra message data from the `ExecutionStateChangedEvent`.

Issue for this optimization/TODO: [CCIP-5167](https://smartcontract-it.atlassian.net/browse/CCIP-5167)

makramkd (approved):

PR #550, opened 2025-02-04
Source chain specific RMN blessings enable/disable

dimkouv:
https://github.com/smartcontractkit/chainlink/pull/16201/commits

### DIFF
--- a/pkg/types/ccipocr3/plugin_commit_types.go
+++ b/pkg/types/ccipocr3/plugin_commit_types.go
@@ -31,13 +31,13 @@ type CommitPluginReport struct {
 	RMNSignatures []RMNECDSASignature `json:"rmnSignatures"`
 }
 
-// IsEmpty returns true if the CommitPluginReport is empty
+// IsEmpty returns true if the CommitPluginReport is empty.
+// NOTE: A report is considered empty when core fields are missing (MerkleRoots, TokenPrices, GasPriceUpdates).
 func (r CommitPluginReport) IsEmpty() bool {
 	return len(r.BlessedMerkleRoots) == 0 &&
 		len(r.UnblessedMerkleRoots) == 0 &&
 		len(r.PriceUpdates.TokenPriceUpdates) == 0 &&
-		len(r.PriceUpdates.GasPriceUpdates) == 0 &&
-		len(r.RMNSignatures) == 0
+		len(r.PriceUpdates.GasPriceUpdates) == 0
 }
 
 // MerkleRootChain Mirroring https://github.com/smartcontractkit/chainlink/blob/cd5c78959575f593b27fd83d8766086d0c678487/contracts/src/v0.8/ccip/libraries/Internal.sol#L356-L362
--- a/pkg/types/ccipocr3/plugin_commit_types_test.go
+++ b/pkg/types/ccipocr3/plugin_commit_types_test.go
@@ -13,6 +13,12 @@ func TestCommitPluginReport(t *testing.T) {
 	t.Run("is empty", func(t *testing.T) {
 		r := CommitPluginReport{}
 		assert.True(t, r.IsEmpty())
+
+		// If a report only contains signatures it is still considered empty.
+		r = CommitPluginReport{
+			RMNSignatures: make([]RMNECDSASignature, 1),
+		}
+		assert.True(t, r.IsEmpty())
 	})
 
 	t.Run("is not empty", func(t *testing.T) {
@@ -26,11 +32,6 @@ func TestCommitPluginReport(t *testing.T) {
 				TokenPriceUpdates: make([]TokenPrice, 1),
 				GasPriceUpdates:   make([]GasPriceChain, 1),
 			},
-		}
-		assert.False(t, r.IsEmpty())
-
-		r = CommitPluginReport{
-			RMNSignatures: make([]RMNECDSASignature, 1),
 		}
 		assert.False(t, r.IsEmpty())
 


commit/merkleroot/observation.go
```diff
@@ -279,9 +279,23 @@ func (p *Processor) getObservation(
...
+			}
+			lggr.Debugw("fetched RMN-enabled chains from rmnHome", "rmnEnabledChains", rmnEnabledChains)
+		}
+
 		return Observation{
-			MerkleRoots: p.observer.ObserveMerkleRoots(ctx, previousOutcome.RangesSelectedForReport),
-			FChain:      p.observer.ObserveFChain(ctx),
+			MerkleRoots:      p.observer.ObserveMerkleRoots(ctx, previousOutcome.RangesSelectedForReport),
+			FChain:           p.observer.ObserveFChain(ctx),
+			RMNEnabledChains: rmnEnabledChains,
```
makramkd, line 298: q: since this is a home-chain read i think we can treat this like fchain? Why not read it in any state of the plugin?

commit/merkleroot/outcome.go
```diff
@@ -217,7 +217,15 @@ func buildMerkleRootsOutcome(
 			signedRoots.Add(rk)
 		}
 
-		// Only report roots that are present in RMN signatures.
+		if signedRoots.Cardinality() > 0 {
+			parsedSigs, err := rmn.NewECDSASigsFromPB(q.RMNSignatures.Signatures)
+			if err != nil {
+				return Outcome{}, fmt.Errorf("failed to parse RMN signatures: %w", err)
+			}
+			sigs = parsedSigs
```
makramkd, line 225: This weird aliasing is confusing - why not just do `var err2 error` and then

```golang
sigs, err2 = rmn.NewECDSASigsFromPB(q.RMNSignatures.Signatures)
```

commit/merkleroot/outcome.go
```diff
@@ -227,11 +235,15 @@ func buildMerkleRootsOutcome(
...
-			} else {
-				lggr.Infow("Root not signed by RMN, skipping from the report", "root", rk)
+			case !consensusObservation.RMNEnabledChains[root.ChainSel]:
+				lggr.Infow("Root is not signed, but RMN is disabled for the chain, appending to the report", "root", rk)
+				rootsToReport = append(rootsToReport, root)
+			default:
+				lggr.Infow("Root not signed, skipping from the report", "root", rk)
 			}
 		}
 		roots = rootsToReport
```
makramkd, line 249: Ditto here on aliasing, we can't just append to `roots` directly?

makramkd: Oh I just realized we loop through `roots` above ... is it just me or is this entire if block rly confusing and hard to understand?

commit/merkleroot/types.go
```diff
@@ -16,11 +16,12 @@ type Query struct {
 }
 
 type Observation struct {
-	MerkleRoots        []cciptypes.MerkleRootChain     `json:"merkleRoots"`
-	OnRampMaxSeqNums   []plugintypes.SeqNumChain       `json:"onRampMaxSeqNums"`
-	OffRampNextSeqNums []plugintypes.SeqNumChain       `json:"offRampNextSeqNums"`
-	RMNRemoteConfig    rmntypes.RemoteConfig           `json:"rmnRemoteConfig"`
-	FChain             map[cciptypes.ChainSelector]int `json:"fChain"`
+	MerkleRoots        []cciptypes.MerkleRootChain      `json:"merkleRoots"`
+	RMNEnabledChains   map[cciptypes.ChainSelector]bool `json:"rmnEnabledChains"`
```
makramkd, line 20: Is this meant to be just a set or will this have all source chains w/ literally mapped true and false values?

RMNHome will not explicitly state that a source chain is not supported, that source chain would just be omitted.

Maybe would be clearer if this is `map[chainSel]struct{}` to signify that

pkg/reader/ccip.go
```diff
@@ -1118,14 +1132,15 @@ func (r *ccipChainReader) getFeeQuoterTokenPriceUSD(ctx context.Context, tokenAd
...
-type sourceChainConfig struct {
-	Router    []byte // local router
-	IsEnabled bool
-	MinSeqNr  uint64
-	OnRamp    cciptypes.UnknownAddress
+type SourceChainConfig struct {
+	Router                    []byte // local router
+	IsEnabled                 bool
+	MinSeqNr                  uint64
+	IsRMNVerificationDisabled bool
```
makramkd, line 1139: If not done already lets update the ccip reader test in chainlink to test this is read correctly

pkg/reader/ccip.go
```diff
@@ -1632,6 +1647,14 @@ func (r *ccipChainReader) GetOffRampConfigDigest(ctx context.Context, pluginType
 	return resp.OCRConfig.ConfigInfo.ConfigDigest, nil
 }
 
+// GetOffRampSourceChainsConfig returns all the source chains configs including disabled chains.
+//
+//nolint:revive // todo
+func (r *ccipChainReader) GetOffRampSourceChainsConfig(ctx context.
```
makramkd, line 1653: Where is this used

commit/merkleroot/observation.go
RensR: It's cached in `rmnHomeReader` right? I would be OK with making it plugin state but that means the cache eviction also needs to move there which might be harder?

makramkd (approved):

makramkd (approved):

PR #551, opened 2025-02-04
Config cache

0xnogo:
This PR modifies the CCIPChainReader to use a cache layer for static configuration data. Instead of making contract calls for each configuration request, static data like offramp config, RMN config, and fee quoter config is now cached with a configurable refresh period.

The `OnRampGetDynamicConfig` and `OnRampGetDestChainConfig` are called from the src chain, so cannot be batched. We can eventually cache them (probably in an upcoming PR).


pkg/reader/cache.go
```diff
@@ -0,0 +1,440 @@
...
+
+	c.lggr.Infow("Successfully refreshed cache",
+		"chain", chainSel,
+		"refreshDuration", refreshDuration)
+
+	return newData, nil
+}
+
+// prepareBatchRequests creates the batch request for all configurations
+func (c *configCache) prepareBatchRequests() contractreader.ExtendedBatchGetLatestValuesRequest {
```
winder, line 159: I think the main purpose of `ccip.go` is to have these types of query details. Is there a way to avoid moving the query logic?

0xnogo: I initially had the batch call in the `ccip.go` but it's creating a dependency loop as the reader is using the cache and the cache is using the reader. 
If we consider the component as a ConfigPoller, then it makes sense that data fetching is happening in itself, like we are doing with the [`HomePoller`](https://github.com/smartcontractkit/chainlink-ccip/blob/eaa7b7f45aa906aadfeba1a9bd5b05646568b101/internal/reader/home_chain.go#L125-L168) or [`RMNPoller`](https://github.com/smartcontractkit/chainlink-ccip/blob/eaa7b7f45aa906aadfeba1a9bd5b05646568b101/pkg/reader/rmn_home_poller.go#L235-L268)

winder: The big difference is that those have separate contract reader configurations.

pkg/reader/config_poller.go
```diff
@@ -0,0 +1,183 @@
...
+		return ChainConfigSnapshot{}, fmt.Errorf("no contract reader for chain %d", chainSel)
+	}
+
+	chainCache := c.getOrCreateChainCache(chainSel)
+
+	chainCache.RLock()
+	timeSinceLastRefresh := time.Since(chainCache.lastRefresh)
+	if timeSinceLastRefresh < c.refreshPeriod {
+		defer chainCache.RUnlock()
+		c.lggr.Infow("Cache hit",
```
asoliman92, line 95: This will become very noisy quickly, maybe use debug log or metrics with a simple counter per chainSelector

pkg/reader/config_poller.go
```diff
@@ -0,0 +1,183 @@
...
+) (ChainConfigSnapshot, error) {
+	chainCache := c.getOrCreateChainCache(chainSel)
+
+	chainCache.Lock()
+	defer chainCache.Unlock()
+
+	// Double check if another goroutine has already refreshed
+	timeSinceLastRefresh := time.Since(chainCache.lastRefresh)
+	if timeSinceLastRefresh < c.refreshPeriod {
+		c.lggr.Infow("Cache was refreshed by another goroutine",
```
asoliman92, line 119: same as above

pkg/reader/config_poller.go
```diff
@@ -0,0 +1,183 @@
...
+			"chain", chainSel,
+			"timeSinceLastRefresh", timeSinceLastRefresh)
+		return chainCache.data, nil
+	}
+
+	startTime := time.Now()
+	newData, err := c.fetchChainConfig(ctx, chainSel)
+	refreshDuration := time.Since(startTime)
```
asoliman92, line 127: I'd confuse the name with refreshPeriod when reading the logs. Maybe `fetchConfigLatency`?

pkg/reader/ccip.go
```diff
@@ -1577,43 +1522,245 @@ func (r *ccipChainReader) GetLatestPriceSeqNr(ctx context.Context) (uint64, erro
...
+		staticConfig          offRampStaticChainConfig
+		dynamicConfig         offRampDynamicChainConfig
+		rmnRemoteAddress      []byte
+		rmnDigestHeader       rmnDigestHeader
+		rmnVersionConfig      versionedConfig
+		feeQuoterConfig       feeQuoterStaticConfig
+	)
+
+	return contractreader.ExtendedBatchGetLatestValuesRequest{
+		consts.ContractNameOffRamp: {
```
asoliman92, line 1553: worth adding some warning comment that changing the order will fail the processing function. I suggest adding some unit test to make sure that the order doesn't change as well.

pkg/reader/config_poller.go
0xnogo: Good point. Changed it to `Debugw`

pkg/reader/ccip.go
0xnogo: The order here is not important. The processing is happening in a for loop with a switch statement. So even if the order changes, we would still be able to process the batched result correctly.

mateusz-sekara (approved):

PR #552, opened 2025-02-04
[Solana] Offramp allowlist on example receiver

PabloMansanet:
Still needs updating the token pools and the test receiver, but this should be a good POC for the approach.


chains/solana/contracts/programs/ccip-router/src/event.rs
```diff
@@ -119,6 +119,18 @@ pub mod events {
...
+        pub struct AddedAllowedOfframp {
+            pub source_chain_selector: u64,
+            pub offramp: Pubkey,
+        }
+
+        #[event]
+        pub struct RemovedAllowedOfframp {
+            pub source_chain_selector: u64,
+            pub offramp: Pubkey,
+        }
```
toblich, line 133: Nice that you already added events :+1:
I've looked at the [EVM implmentation](https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/ccip/Router.sol#L34-L35) which has similar events, would you mind renaming these to match EVM?

chains/solana/contracts/programs/example-ccip-receiver/src/lib.rs
```diff
@@ -351,6 +372,8 @@ pub enum CcipReceiverError {
     OnlyOwner,
     #[msg("Address is not proposed_owner")]
     OnlyProposedOwner,
+    #[msg("Offramp is not allowed")]
+    OfframpNotAllowed,
```
toblich, line 376: Wdyt about renaming the error to a simple `Unauthorized` or `InvalidCaller`? The way this is phrased right now assumes that the caller is in fact an offramp, and that it just happens to not be allowed. But it could be the case that the caller is just some random (or malicious) user/program and not even an offramp, so something more generic would be appropriate then. 
I've checked and [EVM token pools](https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/ccip/pools/TokenPool.sol#L605) (which do a similar auth flow) define an `CallerIsNotARampOnRouter` error, which feels verbose to me but would also be a possibility (note: token pools get called by both onramp and offramp, unlike the receiver).

toblich (commented):
Looks good 👍 👍 

I've left a couple of tiny comments, but the change looks great.

jadepark-dev (approved):

PR #563, opened 2025-02-05
transfer sol with invoke

matYang:
When transferring native Sol from tx signer to Router, we can do so with `invoke` as long as `from` is the signer of the top level transaction, if I understand it correctly, `system_instruction::transfer` builds the instruction with `from` as signer in the AccountMeta under the hood.

The previous way of calling with PDA did not make sense given PDA is the receiver of the funds, not the sender. My best guess is Anchor saw the mismatch between PDA signer key v.s `from`'s key, and fell back to just using `from` as signer


toblich:
Hey @matYang, not quite. The previous way is explicitly signing the transfer with a PDA from the router, and it was the expectation that the user had previously approved said PDA to transfer a certain amount of funds from that token account (there's an SPL Token method to do that approval, you can take a look at our tests where we do it). 
This newly proposed way requires the owner of the funds to be the one calling CCIP directly and they wouldn't have an easy way of guaranteeing that CCIP only bills them for the right amount...

matYang:
> Hey @matYang, not quite. The previous way is explicitly signing the transfer with a PDA from the router, and it was the expectation that the user had previously approved said PDA to transfer a certain amount of funds from that token account (there's an SPL Token method to do that approval, you can take a look at our tests where we do it). This newly proposed way requires the owner of the funds to be the one calling CCIP directly and they wouldn't have an easy way of guaranteeing that CCIP only bills them for the right amount...

that SPL token transfer flow stays the same further down the file, what's being changed is way to transfer native Sol, afaik there is no priori approval

toblich (approved):

PR #564, opened 2025-02-06
[Solana] Allowlist on token pool, receiver & fee quoter

PabloMansanet:

Similar to the approach taken for the example receiver.

`TokenOfframp` takes a whitelisted offramp as signer, but `TokenOnramp` continues to use the router as signer, as before.

---

On commit & execute, the offramp now takes in additional accounts to be forwarded to the token pool, receiver, and fee quoter. The auth flow now looks like the following:

### For the receiver and the token pool
They only need to configure & trust the router address. Then, on offramp interactions to them (receiver: `ccip_receive` | token pool: mint or release), they receive:
1. The message (as always), which includes the source chain
2. The authority (i.e. the signer of the call), which is a PDA derived from the offramp address. This uses different seeds for the token pool CPI and for the receiver CPI
3. The offramp address
4. A PDA from the router, derived from the offramp address (received in `3`) and the source chain (received in `1`). This must match the expected seeds and be initialized. It is a single PDA allowing the offramp, both for the receiver & the pool, so it's up to each of them to check that the authority in `2` is derived from the offramp address with the expected seeds.

### For the fee quoter
It keeps its own list of allowed "price updaters". For the time being, that's just offramps, but at some point it may include other accepted price update sources. So, the admin registers each caller PDA directly in the offramp. Then, on price updates, the fee quoter expects:
- The authority
- The PDA (from fee quoter) that is derived from the authority and fixed seeds, which must be initialized to signal that the caller was previously allowed by an admin.


chains/solana/contracts/programs/token-pool/src/context.rs
```diff
@@ -71,9 +72,25 @@ pub struct AcceptOwnership<'info> {
 #[instruction(release_or_mint: ReleaseOrMintInV1)]
 pub struct TokenOfframp<'info> {
     // CCIP accounts ------------------------
-    #[account(address = config.ramp_authority @ CcipTokenPoolError::InvalidPoolCaller)]
     pub authority: Signer<'info>,
 
+    /// CHECK PDA of the router program verifying the signer is an allowed offramp.
+    /// If PDA does not exist, the router doesn't allow this offramp
```
agusaldasoro, line 78: Is that PDA closed when removing the Allowed offramp?

chains/solana/contracts/programs/token-pool/src/context.rs
```diff
@@ -71,9 +72,25 @@ pub struct AcceptOwnership<'info> {
...
+    #[account(
+        constraint = {
+        let (pda, _) = Pubkey::find_program_address(
+            &[
+                ALLOWED_OFFRAMP,
+                release_or_mint.remote_chain_selector.to_le_bytes().as_ref(),
```
agusaldasoro, line 84: why is the chain selector part of the PDA seeds?

chains/solana/contracts/programs/token-pool/src/lib.rs
```diff
@@ -77,13 +78,30 @@ pub mod token_pool {
...
+        emit!(RampAuthorityUpdated {
             old_authority,
             new_authority
         });
         Ok(())
     }
 
+    // set_router changes the router program ID. This is used to derive the list
+    // of valid mint/release callers (offramps).
+    pub fn set_router(ctx: Context<SetConfig>, new_router: Pubkey) -> Result<()> {
```
agusaldasoro, line 90: could this method be cyclic in some way? Is that a concern?

agusaldasoro (changes requested):

agusaldasoro:
can we add a test (if not present) of removing an offramp and then checking for it?

chains/solana/contracts/programs/token-pool/src/lib.rs
toblich: Not sure what you mean exactly about `cyclic` in this case 🤔 
This method may only be invoked by the pool owner, not the router. Then, the router address here is used just to later check if the calling offramp is a valid one or not.

I've checked a bit and we already have another field in the pool though called `ramp_authority` that has the onramp address... And that's always the same as the router address, so we could just reuse it (even though in EVM they are separate concepts, here they are pretty much the same thing....)

chains/solana/contracts/programs/example-ccip-receiver/src/lib.rs
```diff
@@ -133,27 +134,38 @@ pub struct Initialize<'info> {
...
+        seeds::program = offramp_program.key(),
+    )]
     pub authority: Signer<'info>,
 
+    /// CHECK offramp program: exists only to derive the allowed offramp PDA
+    /// and the authority PDA. Must be second.
+    pub offramp_program: UncheckedAccount<'info>,
+
+    // PDA to verify that calling offramp is valid. Must be third. It is left up to the implementer to decide
+    // how they want to persist the router address to verify that this is the correct account (e.g. in the top level of
```
matYang, line 151: we should provide recommendation on how they should store the Router address, a top level config makes sense, can we name it `config` as opposed to `state`, technically Router is a static config

chains/solana/contracts/programs/ccip-offramp/src/context.rs
```diff
@@ -374,6 +376,25 @@ pub struct ExecuteReportContext<'info> {
...
+                ALLOWED_OFFRAMP,
+                source_chain.chain_selector.to_le_bytes().as_ref(),
+                offramp.key().as_ref(),
+            ],
+            &reference_addresses.router,
+        );
+        allowed_offramp.key() == pda && allowed_offramp.owner.key() == reference_addresses.router
+        } @ CcipOfframpError::InvalidInputs
+    )]
+    pub allowed_offramp: UncheckedAccount<'info>,
```
matYang, line 396: dumb q, I can't wrap my head around why we need it here, isn't this the offramp program itself?

toblich: Hey matt, the signer is always a PDA of the offramp program, not the program itself. Also, the offramp has a different signer for the receiver and for the token pool. We keep in the router a single list of register offramps (the program itself) and then the receiver & pool check that they are called by a PDA with the right seeds that belongs to an offramp that's allowed by the router.

chains/solana/contracts/programs/token-pool/src/lib.rs
toblich: >And that's always the same as the router address, so we could just reuse it (even though in EVM they are separate concepts, here they are pretty much the same thing....)

I've implemented this change and removed the method (and config field) completely

toblich (approved):
This LGTM, though I've co-authored many of the changes here.

chains/solana/contracts/programs/example-ccip-receiver/src/lib.rs
toblich: Noted as improvement on a follow-up PR :+1:
Given the current deadlines, I'd rather merge this as-is and submit a following PR during this week improving on this.

chains/solana/contracts/programs/ccip-offramp/src/context.rs
agusaldasoro: can you write a comment on that in this file? Explaining the signers that we need

chains/solana/contracts/programs/base-token-pool/src/common.rs
```diff
@@ -68,7 +72,9 @@ impl BaseConfig {
         );
         self.owner = owner;
         self.rate_limit_admin = owner;
-        self.ramp_authority = ramp_authority;
+        self.router = router;
+        (self.onramp_authority, _) =
```
agusaldasoro, line 76: should this be `router_authority`?

agusaldasoro (dismissed):

chains/solana/contracts/programs/base-token-pool/src/common.rs
toblich: As chatted offline, renamed to `router_onramp_authority` 👍

toblich (approved):

PR #570, opened 2025-02-07
[CCIP-5174] Ignore sigs when checking for empty commit report

dimkouv:
https://smartcontract-it.atlassian.net/browse/CCIP-5174


makramkd (approved):
We'll have a follow-up where we check for mismatched roots

PR #571, opened 2025-02-07
Add param to disable available fee checks

ogtownsend:
CCIP-4250


execute/costlymessages/costly_messages.go
```diff
@@ -130,6 +137,11 @@ func (o *observer) Observe(
 		if err := validatePositive(execCost); err != nil {
 			return nil, fmt.Errorf("invalid fee for message %s: %w", msg.Header.MessageID, err)
 		}
+		checkFeeDisabled, exists := o.disableAvailableFeeUsdCheckByChain[msg.Header.SourceChainSelector]
```
matYang, line 140: above this line there is `if err := validatePositive(fee); err != nil {` and `execCost, ok := execCosts[msg.Header.MessageID]`, seems like we should allow for the edge case where fee does not exist or is 0 right?

execute/test_utils.go
```diff
@@ -160,6 +160,7 @@ func (it *IntTest) WithCustomFeeBoosting(
 		it.ccipReader,
 		relativeBoostPerWaitHour,
 		now,
+		false,
```
matYang, line 163: I'm surprised this is not being used in more places, might wanna check other setups, especially smoke tests & tooling

ogtownsend: I don't see any uses in `chainlink` or `-deployments`, though I will be following this up with a tooling PR to add support for the new params once we reach more of an alignment on this approach with others.

execute/costlymessages/costly_messages.go
```diff
@@ -128,7 +139,7 @@ func (o *observer) Observe(
 			return nil, fmt.Errorf("missing exec cost for message %s", msg.Header.MessageID)
 		}
 		if err := validatePositive(execCost); err != nil {
-			return nil, fmt.Errorf("invalid fee for message %s: %w", msg.Header.MessageID, err)
```
ogtownsend, line 131: copy pasta? 👀

0xnogo (approved):

execute/costlymessages/costly_messages.go
```diff
@@ -116,10 +123,14 @@ func (o *observer) Observe(
 
 	costlyMessages := make([]cciptypes.Bytes32, 0)
 	for _, msg := range messages {
-		fee, ok := messageFees[msg.Header.MessageID]
-		if !ok {
-			return nil, fmt.Errorf("missing fee for message %s", msg.Header.MessageID)
+		checkFeeDisabled := o.disableAvailableFeeUsdCheckByChain[msg.Header.SourceChainSelector]
+		fee, feeExists := messageFees[msg.Header.MessageID]
+		if checkFeeDisabled && (!feeExists || fee.Cmp(big.NewInt(0)) >= 0) {
```
matYang, line 128: what's the reason for the && statement here, we are guarding against negative fees?

ogtownsend: right, having a negative fee still seemed like an incorrect input and worth throwing on

ogtownsend: decided to remove this and allow negative fees

execute/costlymessages/costly_messages_test.go
```diff
@@ -96,6 +98,77 @@ func TestCCIPCostlyMessageObserver_Observe(t *testing.T) {
...
+			sourceSelectorsByMessageID: map[ccipocr3.Bytes32]ccipocr3.ChainSelector{
+				b1: 1,
+				b2: 2,
+				b3: 3,
+			},
+			want:    []ccipocr3.Bytes32{b2},
+			wantErr: assert.NoError,
+		},
+		{
+			name:       "disable available fee USD check for one source chain",
```
asoliman92, line 128: Disabled for 2 source chains not 1

execute/costlymessages/costly_messages_test.go
```diff
@@ -96,6 +98,77 @@ func TestCCIPCostlyMessageObserver_Observe(t *testing.T) {
 			want:    []ccipocr3.Bytes32{b2},
 			wantErr: assert.NoError,
 		},
+		{
+			name:       "happy path disable fee check explicitly set to false",
```
asoliman92, line 102: Don't we need to add sourceSelectorsByMessageID in the previous cases as well?

ogtownsend: defaults to false for the other tests but to be clear it would be good to add it, I'll update 👍

asoliman92 (approved):

PR #578, opened 2025-02-10
pkg/reader: CommitReportsGTETimestamp & ExecutedMessages to use unconfirmed

makramkd:
core ref: bf18a8d8a08e9279c79cb0736dfd9eb9b4c09fc7


asoliman92 (approved):

mateusz-sekara (commented):
Can we please have tests for that? To avoid misconfigs like these in the future?

pkg/reader/ccip.go
```diff
@@ -313,7 +315,9 @@ func (r *ccipChainReader) ExecutedMessages(
 					Value:    0,
 					Operator: primitives.Gt,
 				}),
-				query.Confidence(primitives.Finalized),
+				// We don't need to wait for an execute state changed event to be finalized
+				// before we optimistically mark a message as executed.
+				query.Confidence(primitives.Unconfirmed),
```
mateusz-sekara, line 320: Wouldn't that break root snoozing?

makramkd:
Yup @mateusz-sekara this is currently untested (still in draft) need to test and make sure this works.

makramkd:
Some data from running one of the integration tests:

```
before (base case - reading finalized): commit 1m4s exec 2m11s
after (reading unconfirmed - couple of runs): commit (54s 1m2s 1m2s 1m0s) exec (1m26s 1m4s 1m8s 1m14)
```

pkg/reader/ccip.go
makramkd: @winder are you able to chime in here? I can look into it but you might know the answer offhand :-)

0xAustinWang (approved):

mateusz-sekara (approved):

PR #582, opened 2025-02-10
Split offramp initialization into two methods to work around stack size limits

toblich:

For context, the feature `EenyoWx9UMXYKpR8mW5Jmfmy2fRjzUtM7NduYMY8bx33` was enabled by default locally but not in devnet/mainnet. So, with that feature disabled, the offramp initialization method was failing with `Access violation in stack frame 5 at address 0x200005d90 of size 8`, which relates to the stack size exceeding the limit. 

To work around that, this PR splits the initialization into two methods, one just for the config (which is by far the big struct with the OCR config in it), and another one with everything else.

I've also doubled the tests timeout when funding accounts (from 30s to 1m), as CI was failing often due to timeouts in that step.


chains/solana/contracts/programs/ccip-offramp/src/lib.rs
```diff
@@ -65,6 +47,34 @@ pub mod ccip_offramp {
...
+    ) -> Result<()> {
+        let mut config = ctx.accounts.config.load_init()?;
+        require!(config.version == 0, CcipOfframpError::InvalidInputs); // assert uninitialized state - AccountLoader doesn't work with constraint
+        config.version = 1;
+        config.svm_chain_selector = svm_chain_selector;
+        config.enable_manual_execution_after = enable_execution_after;
+        config.owner = ctx.accounts.authority.key();
+        config.ocr3 = [
+            Ocr3Config::new(OcrPluginType::Commit as u8),
+            Ocr3Config::new(OcrPluginType::Execution as u8),
```
archseer, line 73: It seems odd to me that we'd hit the stack limit since `config` is zero copy, perhaps it's these two lines? Does it make any difference if you make the new method `const fn` or if you inline the code here?

Try also directly assigning `config.ocr3[0]` and `config.ocr3[1]`. It must either be this array allocation, or the `Ocr3Config::new` calls

toblich: Hey Blaz, I thought so too, but at first I tried even commenting out the entire code of the method, and it still failed once the feature flag was deactivated. It seems like just the size of the context and the instruction data is already big enough to make it fail. 
Also, based on https://github.com/anza-xyz/agave/issues/1186#issue-2278967670, it seems like the `init` constraint already eats away a decent amount of the stack, meaning that `Can't have more than 3 accounts to init within the same instruction atm.`

chains/solana/contracts/tests/testutils/anchor.go
```diff
@@ -49,7 +49,7 @@ func FundAccounts(ctx context.Context, accounts []solana.PrivateKey, solanaGoCli
 		remaining = unconfirmedTxCount
 
 		time.Sleep(500 * time.Millisecond)
-		if count > 60 {
+		if count > 120 {
```
toblich, line 52: CI flakiness was very annoying, so increasing the timeout here as we've discussed in today's standup

PR #594, opened 2025-02-11
PoC: routing traffic between versiones of the logic module

toblich:


PabloMansanet:
Looks like a reasonable approach to me!

PR #596, opened 2025-02-12
Background observer for merkle roots processor

dimkouv:
Implementing asynchronous observations for merkle roots processor.
This will lead to `just a few ms` merkle root processor observations.

https://github.com/smartcontractkit/chainlink/pull/16347/commits
`core ref: b94888568d55a47bc3c51b195d06f0e8d1264ba8`

### DIFF
--- a/commit/merkleroot/observation.go
+++ b/commit/merkleroot/observation.go
@@ -316,7 +316,7 @@ type Observer interface {
 type asyncObserver struct {
 	lggr                logger.Logger
 	syncObserver        observerImpl
-	cancelFunc          context.CancelFunc
+	cancelFunc          func()
 	mu                  *sync.RWMutex
 	offRampNextSeqNums  []plugintypes.SeqNumChain
 	onRampLatestSeqNums []plugintypes.SeqNumChain
@@ -340,6 +340,12 @@ func newAsyncObserver(lggr logger.Logger, observer observerImpl, tickDur, syncTi
 	ticker := time.NewTicker(tickDur)
 	lggr.Debugw("async observer started", "tickDur", tickDur, "syncTimeout", syncTimeout)
 	o.start(ctx, ticker.C, syncTimeout)
+
+	o.cancelFunc = func() {
+		cf()
+		ticker.Stop()
+	}
+
 	return o
 }
 


commit/merkleroot/observation.go
```diff
@@ -318,22 +291,155 @@ type Observer interface {
...
+func (o *asyncObserver) sync(ctx context.Context, syncTimeout time.Duration) {
+	o.lggr.Debugw("async observer is syncing", "syncTimeout", syncTimeout)
+	ctxSync, cf := context.WithTimeout(ctx, syncTimeout)
+
+	syncOps := []struct {
+		id string
+		op func(context.Context)
+	}{
+		{
+			id: "offRampNextSeqNums",
```
asoliman92, line 368: use some constants better for these

commit/merkleroot/observation.go
```diff
@@ -318,22 +291,155 @@ type Observer interface {
...
+	cf()
+}
+
+// applySyncOp applies the given operation synchronously.
+func (o *asyncObserver) applySyncOp(
+	ctx context.Context, lggr logger.Logger, id string, wg *sync.WaitGroup, op func(ctx context.Context)) {
+	defer wg.Done()
+	tStart := time.Now()
+	o.lggr.Debugw("async observer applying sync operation", "id", id)
+	o.mu.Lock()
```
asoliman92, line 396: nit optimization, I think it's better to have mutex per operation, otherwise they're blocking each other unnecessarily

internal/plugincommon/chain_support.go
```diff
@@ -61,6 +61,10 @@ func (c ccipChainSupport) DestChain() cciptypes.ChainSelector {
 }
 
 func (c ccipChainSupport) KnownSourceChainsSlice() ([]cciptypes.ChainSelector, error) {
+	if c.homeChain == nil {
+		return nil, fmt.Errorf("home chain is nil")
+	}
```
makramkd, line 66: Is this actually happening? Null pointer exceptions on `homeChain`? Feels like there should be a null check in the constructor in that case.

commit/merkleroot/observation.go
```diff
@@ -318,22 +291,155 @@ type Observer interface {
 	ObserveOffRampNextSeqNums(ctx context.Context) []plugintypes.SeqNumChain
 
 	// ObserveLatestOnRampSeqNums observes the latest OnRamp sequence numbers for each configured source chain.
-	// NOTE: Make sure that caller supports the provided destination chain.
-	ObserveLatestOnRampSeqNums(ctx context.Context, destChain cciptypes.ChainSelector) []plugintypes.SeqNumChain
+	// NOTE: Make sure that caller supports the destination chain.
+	ObserveLatestOnRampSeqNums(ctx context.Context) []plugintypes.SeqNumChain
```
makramkd, line 295: Was the destChain arg just not used?

commit/merkleroot/observation.go
```diff
@@ -318,22 +291,155 @@ type Observer interface {
...
+	cf                  context.CancelFunc
+	mu                  *sync.RWMutex
+	offRampNextSeqNums  []plugintypes.SeqNumChain
+	onRampLatestSeqNums []plugintypes.SeqNumChain
+}
+
+// newAsyncObserver creates a new asyncObserver.
+// It fetches the data from the base observer asynchronously and caches the results.
+// It fetches the data every tickDur and uses a timeout of syncTimeout to kill long RPC calls.
+func newAsyncObserver(lggr logger.Logger, observer observerImpl, tickDur, syncTimeout time.Duration) *asyncObserver {
```
makramkd, line 328: Feels like there's quite a bit of components right now that do essentially the same thing - they do some I/O in a background thread and cache state. We already have this for CCIPHome, RMNHome, RMNRemote (I think), and now this component.

Having duplication here is pretty bad because we could have the same bug 3x times instead of just once.

Just trying to see the forest from the trees - feels like we need just a standard `Cache` implementation that is generic enough to handle this (maybe the `State` is just passed in and made generic). I think we had something like this for <= v1.5.

cc @mateusz-sekara

commit/merkleroot/observation.go
```diff
@@ -318,22 +291,155 @@ type Observer interface {
...
+
+	// Close closes the observer and releases any resources.
+	Close() error
+}
+
+// asyncObserver is an Observer implementation that fetches the data asynchronously.
+type asyncObserver struct {
+	lggr                logger.Logger
+	syncObserver        observerImpl
+	cf                  context.CancelFunc
```
makramkd, line 319:
```suggestion
	cancelFunc                  context.CancelFunc
```

`cf` is not sufficiently descriptive on its own to know what it is w/out an editor (i.e reading code on GH)

commit/merkleroot/observation.go
```diff
@@ -318,22 +291,155 @@ type Observer interface {
...
+		},
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(len(syncOps))
+	for _, op := range syncOps {
+		go o.applySyncOp(ctxSync, o.lggr, op.id, wg, op.op)
+	}
+	wg.Wait()
+	cf()
```
makramkd, line 393: Better to `defer cancelFunc()` after you create the ctx

commit/merkleroot/observation.go
makramkd: That feels likelier that you'd make a copy/pasta bug by locking the wrong thing, I've seen this a couple of times just over the last month

asoliman92 (approved):

commit/merkleroot/observation.go
makramkd: Yeh all the above I mentioned are self-syncing as well, just sync periods are slightly different

0xnogo (approved):

commit/merkleroot/observation.go
```diff
@@ -318,22 +291,159 @@ type Observer interface {
...
+	o := &asyncObserver{
+		lggr:                lggr,
+		syncObserver:        observer,
+		cancelFunc:          cf,
+		mu:                  &sync.RWMutex{},
+		offRampNextSeqNums:  make([]plugintypes.SeqNumChain, 0),
+		onRampLatestSeqNums: make([]plugintypes.SeqNumChain, 0),
+	}
+
+	ticker := time.NewTicker(tickDur)
```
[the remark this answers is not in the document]
makramkd: suggestion: Wouldn't it be simpler (and more idiomatic) to create the ticker in the only goroutine where its used?

makramkd (approved):

PR #598, opened 2025-02-12
CCIP-5216 Tracking latencies and simplification in the metrics code

mateusz-sekara:
Simplifying tracking metrics for the processors by:
* reducing number of interfaces and moving stats to the struct 
* reducing number of method in the reporter
* tracking metrics is done in the decorator (see `TrackedProcessor`), so plugin is less polluted with the observability stuff 
* any processor can add tracking of their observations and outcomes as long it's wrapped with TrackedProcessor 

Additionally, tracking errors and latencies for the processor methods

### DIFF
--- a/commit/merkleroot/types.go
+++ b/commit/merkleroot/types.go
@@ -242,17 +242,17 @@ func (p processorState) String() string {
 // MetricsReporter exposes only relevant methods for reporting merkle roots from metrics.Reporter
 type MetricsReporter interface {
 	TrackRmnReport(latency float64, success bool)
-	TrackProcessorLatency(processor string, method string, latency time.Duration)
-	TrackProcessorObservation(processor string, obs plugintypes.Trackable, err error)
-	TrackProcessorOutcome(processor string, out plugintypes.Trackable, err error)
+	TrackProcessorLatency(processor string, method string, latency time.Duration, err error)
+	TrackProcessorObservation(processor string, obs plugintypes.Trackable)
+	TrackProcessorOutcome(processor string, out plugintypes.Trackable)
 }
 
 type NoopMetrics struct{}
 
 func (n NoopMetrics) TrackRmnReport(float64, bool) {}
 
-func (n NoopMetrics) TrackProcessorLatency(string, string, time.Duration) {}
+func (n NoopMetrics) TrackProcessorLatency(string, string, time.Duration, error) {}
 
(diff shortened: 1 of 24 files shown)


commit/merkleroot/types.go
```diff
@@ -10,6 +11,13 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
+const (
+	processorLabel    = "merkleroot"
+	rootsLabel        = "roots"
+	messagesLabel     = "messages"
+	rmnSignatureLabel = "rmnSignatures"
```
0xnogo, line 18: Are the rootsLabel, messagesLabel, rmnSignatureLabel under the processorLabel? Meaning we don't need to name them explicitly as part of the mertkleroot processor?

commit/metrics/prom.go
```diff
@@ -187,72 +129,31 @@ func (p *PromReporter) TrackRmnRequest(method string, latency float64, nodeID ui
...
+		WithLabelValues(p.chainID, processor, method).
+		Observe(float64(latency.Milliseconds()))
 }
 
-func chainFeeObservationMetrics(obs chainfee.Observation) map[string]int {
-	return map[string]int{
-		feeComponentsLabel:     len(obs.FeeComponents),
-		nativeTokenPricesLabel: len(obs.NativeTokenPrices),
-		chainFeeUpdatesLabel:   len(obs.ChainFeeUpdates),
+func (p *PromReporter) TrackProcessorObservation(processor string, obs plugintypes.Trackable) {
```
0xnogo, line 145: nit `TrackProcessorOutcome` and `TrackProcessorObservation` are the same. Do we need to define them separately or can we just pass the counter as input?

mateusz-sekara: You are right, this could be unified

commit/merkleroot/types.go
mateusz-sekara: I don't understand the question (btw these labels were there before, I'm only moving them from one place to another)

mateusz-sekara:
> It is also useful to track the `oracleID` in this metrics.
> Sometimes maybe a specific oracle has issues that contribute to high P99,P90 because the RPC they are using is slow.

These are tracked automatically

0xnogo (approved):

PR #603, opened 2025-02-12
Onramp upgrade & rollback flows [NONEVM-1297]

toblich:
This PR supports emulating a CCIP minor version bump on the onramp. The bump is done per-lane (so, per dest-chain for the onramp). To do the bump, the owner should:
- Update the dest-chain configuration as needed (mostly in fee quoter)
- Invoke the new bump method in the router/onramp, which mostly just resets the sequence number for that lane

Then, to rollback, the owner should:
- Update the dest-chain configuration as needed (mostly in fee quoter)
- Invoke the new rollback method in the router/onramp, which mostly just restores the sequence number for that lane that was there before the last upgrade.

Caveats:
- As upgrades of CCIP minor version are not too often, we only support one consecutive rollback. We don't expect nor support rolling-back twice in a row, without having bumped again at least once in between
- It is up to the admin to update the fee quoter config accordingly with each operation, and bundling all changes in a single tx


chains/solana/contracts/programs/ccip-router/src/lib.rs
```diff
@@ -206,6 +206,34 @@ pub mod ccip_router {
...
+        v1::admin::bump_ccip_version_for_dest_chain(ctx, dest_chain_selector)
+    }
+
+    /// Rolls back the CCIP version for a destination chain.
+    /// This effectively just restores the old version's sequence number of the destination chain state.
+    ///
+    /// # Arguments
+    /// * `ctx` - The context containing the accounts required for the rollback.
+    /// * `dest_chain_selector` - The destination chain selector to rollback the version for.
+    pub fn rollback_ccip_version_for_dest_chain(
```
PabloMansanet, line 230: Seeing as this is the part that will be seen from cargo docs, it may be good to include here an explanation about the consecutive rollback limitation. Especially because the error message you get isn't necessarily too explici.

chains/solana/contracts/programs/ccip-router/src/state.rs
```diff
@@ -20,9 +22,33 @@ pub struct Config {
     pub fee_aggregator: Pubkey, // Allowed address to withdraw billed fees to (will use ATAs derived from it)
 }
 
-#[derive(Clone, AnchorSerialize, AnchorDeserialize, InitSpace, Debug)]
+#[derive(Clone, AnchorSerialize, AnchorDeserialize, InitSpace, Debug, PartialEq, Eq, Copy)]
+pub enum RestoreOnAction {
+    None,
+    Upgrade,
+    Rollback,
+}
```
PabloMansanet, line 30: I don't think there's necessarily anything wrong with this naming, but the logic here takes a moment to understand, so I think comments would be good.

PabloMansanet (approved):
LGTM! Just two minor points on comments/clarity.

chains/solana/contracts/programs/ccip-router/src/instructions/v1/admin.rs
```diff
@@ -172,3 +215,126 @@ fn validate_dest_chain_config(dest_chain_selector: u64, _config: &DestChainConfi
     );
     Ok(())
 }
+
+fn bump_seq_nr(dest_chain_state: &mut DestChainState) -> Result<()> {
```
agusaldasoro, line 219: can you add a comment here explaining why is this needed?

agusaldasoro (approved):

PR #605, opened 2025-02-13
Cache onramp config calls

0xnogo:
Add the OnRampGetDynamicConfig and OnRampGetDestChainConfig in the ConfigPoller. These 2 calls cannot be simply added to the ConfigPoller as it is only batching dest chain requests.


pkg/reader/ccip.go
```diff
@@ -944,16 +943,41 @@ func (r *ccipChainReader) DiscoverContracts(ctx context.Context,
...
+		}
+
+		// Check if we have a reader for this chain
+		if _, exists := r.contractReaders[chain]; !exists {
+			lggr.Debugw("Contract reader not found for chain", "chain", chain)
+			continue
+		}
+
+		// Get cached OnRamp configurations
+		config, err := r.configPoller.GetChainConfig(ctx, chain)
```
mateusz-sekara, line 959: Looping over that might cause a problem until configPoller is fully migrated to the background process, right? When the cache is expired that's gonna create N sequential calls

pkg/reader/ccip.go
```diff
@@ -1599,9 +1506,30 @@ func (r *ccipChainReader) prepareBatchConfigRequests() contractreader.ExtendedBa
...
+	// Only add OnRamp config requests if this is a source chain (not destination chain)
+	if chainSel != r.destChain {
+		requests[consts.ContractNameOnRamp] = []types.BatchRead{
+			{
+				ReadName:  consts.MethodNameOnRampGetDynamicConfig,
+				Params:    map[string]any{},
+				ReturnVal: &onRampDynamicConfig,
+			},
+			{
+				ReadName: consts.MethodNameOnRampGetDestChainConfig,
```
mateusz-sekara, line 1519: Out of curiosity, could we solve that problem by exposing function from solidity like
```solidity
func getDestChainConfigs() // all at once

// or

func getDestChainConfigs(uint64[] destChainSelectors) // pass selectors
```

Contracts are finalized but I guess could be something for 1.7 or 1.6.1?

0xnogo: It worth raising this point to onchain team. We used to use `OffRampGetAllSourceChainConfigs` but due to some incompatibly with solana, we had to move back to a batch call of `GetSourceChainConfig`, so we might have the same issue with onramp configs

https://github.com/smartcontractkit/chainlink-ccip/blob/bb00d60949abbd7c444fe6a02d9ebf2305e3ee51/pkg/reader/ccip.go#L1230-L1243

pkg/reader/ccip.go
0xnogo: As discussed offline, we need to check
* how the plugin is behaving when a new src chain is added
* implement a background poller in order to move as much processing from foreground (impacting OCR) to background

mateusz-sekara: We could use a similar approach to the previous implementation. Run these cache checks in parallel. Therefore we won't make it worse when cache is evicted and configuration needs to be rechecked. Therefore for cases when cache is evicted, behaviour is gonna be exactly the same as the one in old impl

pkg/reader/ccip.go
```diff
@@ -1008,18 +1008,60 @@ func (r *ccipChainReader) DiscoverContracts(ctx context.Context,
...
+
+			// Use mutex to safely update the shared resp
+			mu.Lock()
+			defer mu.Unlock()
+
+			// Add FeeQuoter from dynamic config
+			if len(config.OnRamp.DynamicConfig.DynamicConfig.FeeQuoter) > 0 {
+				resp = resp.Append(
+					consts.ContractNameFeeQuoter,
+					chain,
```
mateusz-sekara, line 1048: Is it thread safe? That value is taken from the for loop, does it require some kind of reinitialization to make sure it's not overridden by the looping thread

0xnogo: I think you right, we need to create a local copy of the loop variable chain

mateusz-sekara (approved):

PR #611, opened 2025-02-13
solana: CCIP admin & observability events

jadepark-dev:
## Primary Changes

- Added `ConfigSet` events for `Router`, `Offramp`, and `Fee Quoter` programs to track configuration updates
- Added `ReferenceAddressesSet` event for Offramp program initialization
- Standardized event emission after all admin configuration changes

## Others
- Removed hardcoded constant `MaxCU = 1_400_000` in favor of using `computebudget.MAX_COMPUTE_UNIT_LIMIT` from SDK
- Made minor code formatting improvements in test assertions


toblich (dismissed):
LGTM in general, though the `version` should ideally not be in the events, and given that it's independent for each struct I also would not extract that to a shared constant

toblich (approved):

PR #612, opened 2025-02-13
Implement routing layer for ccip-router

toblich:
Similar to #604 but for the `ccip-router` now. 

Besides the routing layer and equivalent changes to what was done in #604 for that other program, this also includes:
- Makes the ccip-router `Config` account no longer a `zero_copy` one
- Prevents the default code version to be set to `default`, which would be redundant and unclear. This is also ported to `fee-quoter` in this PR.


chains/solana/contracts/programs/ccip-router/src/instructions/interfaces.rs
```diff
@@ -0,0 +1,151 @@
...
+use crate::messages::SVM2AnyMessage;
+use crate::state::{CodeVersion, DestChainConfig};
+use crate::token_context::{
+    AcceptAdminRoleTokenAdminRegistry, ModifyTokenAdminRegistry,
+    OverridePendingTokenAdminRegistryByCCIPAdmin, OverridePendingTokenAdminRegistryByOwner,
+    RegisterTokenAdminRegistryByCCIPAdmin, RegisterTokenAdminRegistryByOwner,
+    SetPoolTokenAdminRegistry,
+};
+
+pub trait Admin {
```
agusaldasoro, line 16: Love this!

PR #614, opened 2025-02-14
[CCIP-5235] execute: check for already executed in ShouldTransmit

makramkd:
core ref: c0b64a32323da3736f0311d147a0495b2f141bff

Check in ShouldAccept and ShouldTransmit for messages that have already been executed and return `false` for both when we find an already executed message.

This is not currently a perfect strategy since it makes the reports all-or-nothing, but we can refine it in the future. For now, it definitely seems like the same report is being executed verbatim by different oracles in the transmission schedule, so this should definitely fix that.


execute/plugin.go
```diff
@@ -537,9 +544,73 @@ func (p *Plugin) validateReport(
...
+	// TODO: batch these queries? these are all DB reads.
+	// maybe some alternative queries exist.
+	for sourceChainSelector, seqNrRange := range seqNrRangesBySource {
+		executed, err := p.ccipReader.ExecutedMessages(ctx, sourceChainSelector, seqNrRange)
+		if err != nil {
+			return fmt.Errorf("couldn't check if messages already executed: %w", err)
+		}
+
+		if len(executed) > 0 {
+			// Some subset of messages have been executed, we can't accept this report.
```
makramkd, line 583: An alternative to this "all or nothing" logic here is maybe we can have a % of mgs in the report that have been executed, if > 50% or something, then don't transmit, if < 50%, then still worth transmitting. Will leave this kinda logic for another PR, I feel like it'd be something good to test in the high-traffic envs like load tests.

execute/plugin.go
```diff
@@ -537,9 +544,73 @@ func (p *Plugin) validateReport(
...
+	if errors.Is(err, errAlreadyExecuted) {
+		// Some messages in the report have already
+		// been executed, so we don't want to re-execute them.
+		// This gives the exec plugin a chance to remedy the situation
+		// by selecting a different set of messages.
+		return false, decodedReport, nil
+	}
+	if err != nil {
+		// TODO: should we return true here if we couldn't check for already executed messages?
+		return false, decodedReport, fmt.Errorf("checking for already executed messages failed: %w", err)
```
makramkd, line 560: This is basically DB I/O error - either timeout or something else.

execute/plugin.go
```diff
@@ -537,9 +544,98 @@ func (p *Plugin) validateReport(
...
+// the sequence number range of the messages in the report.
+func getSnRangeSetPairsBySource(
+	chainReports []cciptypes.ExecutePluginReportSingleChain,
+) map[cciptypes.ChainSelector]snRangeSetPair {
+	seqNrRangesBySource := make(map[cciptypes.ChainSelector]snRangeSetPair)
+	for _, chainReport := range chainReports {
+		cmpr := func(a, b cciptypes.Message) int {
+			return cmp.Compare(a.Header.SequenceNumber, b.Header.SequenceNumber)
+		}
+		minMsg := slices.MinFunc(chainReport.Messages, cmpr)
```
mateusz-sekara, line 619: MinFunc/MaxFunc gonna panic when a slice is empty. Probably either using a different method or just adding length check should do the job. I recommend adding a testcase for it

mateusz-sekara (approved):

execute/plugin.go
makramkd: Oh I guess that makes sense with this API.

makramkd: Will add a check for message length in the chain report. We check that ChainReport is > 0 length and report is non-nil, I guess its a super-bug if the messages in it are nil, but will add the check anyway to be defensive.

makramkd:
Discussed the above offline, will revisit if load tests point to this as a bottleneck.

PR #619, opened 2025-02-15
typeconv: Implement for Solana

archseer:


internal/libs/typeconv/address.go
```diff
@@ -4,27 +4,56 @@ import (
...
-	return "0x" + hex.EncodeToString(addr)
+	family, err := chainsel.GetSelectorFamily(chainSelector)
+	if err != nil {
+		panic(err)
+	}
+
+	switch family {
+	case chainsel.FamilyEVM:
+		// TODO: not EIP-55. Fix this?
+		return "0x" + hex.EncodeToString(addr)
```
prashantkumar1982, line 23: Just use hexutil.Encode(addr)
It encodes the bytes as a hex string with 0x prefix

Its a geth library from:
"github.com/ethereum/go-ethereum/common/hexutil"

internal/libs/typeconv/address.go
```diff
@@ -4,27 +4,56 @@ import (
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to decode EVM address '%s': %w", addr, err)
+		panic(err)
 	}
 
-	return addrBytes, nil
+	switch family {
+	case chainsel.FamilyEVM:
+		// lower case in case EIP-55 and trim 0x prefix if there
+		addrBytes, err := hex.DecodeString(strings.ToLower(strings.TrimPrefix(addr, "0x")))
```
prashantkumar1982, line 42: common.HexToAddress(addr)
Import "github.com/ethereum/go-ethereum/common"

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.23.3
 
 require (
 	github.com/deckarep/golang-set/v2 v2.6.0
+	github.com/gagliardetto/solana-go v1.12.0
```
winder, line 7: This dependency is a non starter.

internal/libs/typeconv/address.go
winder: @prashantkumar1982 this probably needs to be another service.

winder (changes requested):
This needs to be a new service for encoding and decoding addresses.

archseer:
Replaced by https://github.com/smartcontractkit/chainlink-ccip/pull/647

PR #620, opened 2025-02-16
commit: Only attempt to fetch latestMsgSeqNum from enabled source chains

archseer:
We found this when testing with Solana: the default setup requires two
EVM chains: a home chain and a feeds chain. In the solana setup the feeds
chain has a configured relayer but no lane connecting it (isEnabled: false)

### DIFF
--- a/commit/merkleroot/observation.go
+++ b/commit/merkleroot/observation.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 	"time"
 
-	mapset "github.com/deckarep/golang-set/v2"
 	"golang.org/x/sync/errgroup"
 
 	chainsel "github.com/smartcontractkit/chain-selectors"
@@ -544,21 +543,22 @@ func (o observerImpl) ObserveLatestOnRampSeqNums(ctx context.Context) []pluginty
 		return nil
 	}
 
-	supportedChains, err := o.chainSupport.SupportedChains(o.oracleID)
+	sourceChainsConfig, err := o.ccipReader.GetOffRampSourceChainsConfig(ctx, allSourceChains)
 	if err != nil {
-		lggr.Warnw("call to KnownSourceChainsSlice failed", "err", err)
+		lggr.Errorw("get offRamp source chains config failed", "err", err)
 		return nil
 	}
 
-	sourceChains := mapset.NewSet(allSourceChains...).Intersect(supportedChains).ToSlice()
-	sort.Slice(sourceChains, func(i, j int) bool { return sourceChains[i] < sourceChains[j] })
-
 	mu := &sync.Mutex{}
-	latestOnRampSeqNums := make([]plugintypes.SeqNumChain, 0, len(sourceChains))
+	latestOnRampSeqNums := make([]plugintypes.SeqNumChain, 0, len(sourceChainsConfig))
 
 	wg := &sync.WaitGroup{}
-	wg.Add(len(sourceChains))
-	for _, sourceChain := range sourceChains {
+	for sourceChain, cfg := range sourceChainsConfig {
+		if !cfg.IsEnabled {
+			lggr.Debugw("ObserveLatestOnRampSeqNums source chain is disabled, skipping", "chain", sourceChain)
+			continue
+		}
+		wg.Add(1)
 		go func() {
 			defer wg.Done()
 			latestOnRampSeqNum, err := o.ccipReader.LatestMsgSeqNum(ctx, sourceChain)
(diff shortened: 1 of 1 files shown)


archseer:
There's a couple other spots (`chainfee/observation.go`, `merkleroot/observation.go` `ObserveMerkleRoots`) that might need the same change.

commit/merkleroot/observation.go
```diff
@@ -548,20 +547,22 @@ func (o observerImpl) ObserveLatestOnRampSeqNums(ctx context.Context) []pluginty
...
-
 	mu := &sync.Mutex{}
-	latestOnRampSeqNums := make([]plugintypes.SeqNumChain, 0, len(sourceChains))
+	latestOnRampSeqNums := make([]plugintypes.SeqNumChain, 0, len(sourceChainsConfig))
 	eg := &errgroup.Group{}
 
-	for _, sourceChain := range sourceChains {
+	for sourceChain, cfg := range sourceChainsConfig {
+		if !cfg.IsEnabled {
+			lggr.Infof("ObserveLatestOnRampSeqNums source chain %d is disabled, chain is skipped", sourceChain)
```
matYang, line 562: this might become a bit spammy in info logs, maybe debug level or aggregate disabled chains into 1 line?

commit/merkleroot/observation.go
```diff
@@ -548,20 +547,22 @@ func (o observerImpl) ObserveLatestOnRampSeqNums(ctx context.Context) []pluginty
 		return nil
 	}
 
-	supportedChains, err := o.chainSupport.SupportedChains(o.oracleID)
+	sourceChainsConfig, err := o.ccipReader.GetOffRampSourceChainsConfig(ctx, allSourceChains)
```
matYang, line 550: it does not seem ccipReader. GetOffRampSourceChainsConfig uses caching, I don't have the context on if this is low impact enough to work in the short term, in the long run it makes sense for supported source chains to be cached (&refreshed during chain config updates)

commit/chainfee/observation.go
```diff
@@ -39,6 +39,17 @@ func (p *processor) Observation(
...
+	// filter out disabled chains
+	sourceChainsConfig, err := p.ccipReader.GetOffRampSourceChainsConfig(ctx, supportedChains.ToSlice())
+	if err != nil {
+		return Observation{}, fmt.Errorf("get offRamp source chains config: %w", err)
+	}
+	for selector, config := range sourceChainsConfig {
+		if !config.IsEnabled {
+			supportedChains.Remove(selector)
+		}
+	}
```
winder, line 51: This logic should probably go into `SupportedChains`.
@makramkd do you know if there's a reason we don't check the enabled flag?

archseer: I think it was an optimization, this lookup will need to be cached since it happens within the `Observation`. OCR3 doesn't like it if a round starts taking too long

commit/merkleroot/observation.go
mateusz-sekara: @0xnogo do you mind chiming in?

mateusz-sekara: If this is RPC hotcall that's gonna be made during the Observation then we must not do it like this. It's gonna be a blocker for further release as we are moving away from direct RPC calls from observations. Please see `config_poller` for how the configurations calls are made or consult the approach directly with Noufel

0xnogo: I need to check deeper but `MethodNameGetSourceChainConfig` looks like a good candidate for caching. We will have to figure out a way to pass the list of chain selectors without breaking the `prepareBatchConfigRequests` method, but it should be feasible. We used to have the AllConfigs version of it cached, but this was removed due to some limitation on Solana implementation.

commit/chainfee/observation.go
archseer: I'll revert this and see if reports are still getting generated.

archseer: I remember now, we need https://github.com/smartcontractkit/chainlink-ccip/pull/667 merged first

commit/merkleroot/observation.go
```diff
@@ -544,21 +543,22 @@ func (o observerImpl) ObserveLatestOnRampSeqNums(ctx context.Context) []pluginty
...
 	if err != nil {
-		lggr.Warnw("call to KnownSourceChainsSlice failed", "err", err)
+		lggr.Errorf("get offRamp source chains config: %w", err)
```
makramkd, line 548: please use structured logging instead for consistency & easier log querying

```golang
lggr.Errorw("get offRamp source chains config failed", "err", err)
```

makramkd (approved):

PR #621, opened 2025-02-17
reader/ccip: Skip fetching GetRMNRemoteConfig,  on chains that don't support it

archseer:
This isn't a full fix because ObserveOffRampNextSeqNums still attempts to use RMNRemote.

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -837,6 +837,11 @@ func (r *ccipChainReader) GetRMNRemoteConfig(ctx context.Context) (rmntypes.Remo
 		return rmntypes.RemoteConfig{}, fmt.Errorf("get chain config: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/reader/ccip.go
```diff
@@ -870,12 +875,27 @@ func (r *ccipChainReader) GetRmnCurseInfo(
 		return nil, fmt.Errorf("validate dest=%d extended reader existence: %w", r.destChain, err)
 	}
 
+	config, err := r.configPoller.GetChainConfig(ctx, r.destChain)
+	if err != nil {
+		return nil, fmt.Errorf("get chain config: %w", err)
+	}
+
+	if config.Offramp.StaticConfig.RmnRemote == nil {
```
matYang, line 883: this won't be needed soon* given Solana onchain is now committed to building RMNRemote for curse right?

PR #622, opened 2025-02-17
solana: expose anchor ts bindings[NONEVM-1212]

jadepark-dev:
This PR includes the TypeScript interfaces generated during compilation, which were previously ignored. By unignoring and adding them to the repository, we ensure proper type support for interacting with the program. These interfaces are already generated during program compilation, so no additional setup is required. This change ensures they are available in the repository for teams that need them.

## Changes
- Unignored and added the types folder containing the generated TypeScript interfaces.

## Example
https://www.anchor-lang.com/docs/clients/typescript#client-program
```typescript
import { Program, AnchorProvider, setProvider } from "@coral-xyz/anchor";
import { useAnchorWallet, useConnection } from "@solana/wallet-adapter-react";
import type { HelloAnchor } from "./idlType"; // generated typescript binding
import idl from "./idl.json";
 
const { connection } = useConnection();
const wallet = useAnchorWallet();
 
const provider = new AnchorProvider(connection, wallet, {});
setProvider(provider);
 
// now the program instance has full type support
export const program = new Program(idl as HelloAnchor, {
  connection,
});
```


agusaldasoro (changes requested):
We also need to add the CI check that compares between the added files and generating them again through a diff. That check must be in the commit history.

jadepark-dev:
@agusaldasoro I think the check is already on the ci? It had failed with outdated ts bindings.
https://github.com/smartcontractkit/chainlink-ccip/actions/runs/13370624058/job/37338133898?pr=622

agusaldasoro (dismissed):
Sorry! I didn't know

winder (approved):

agusaldasoro (approved):

PR #625, opened 2025-02-18
fix: Token price only reports were missing data in report info

archseer:
Fixes a mistake from https://github.com/smartcontractkit/chainlink-ccip/pull/505, without this change price update reports would contain empty report info.


winder (changes requested):
This fix has been incorporated in #638

archseer:
There's a second change here, adding gas price update information to the header since it was missing

PR #631, opened 2025-02-19
[Solana][readability] PluginType type safety

PabloMansanet:
Minor type safety/readability improvement. Since zero_copy serialization of enums isn't possible, we stored the plugin type as a u8 in many places, but it can be made a bit safer with an additional type wrapper.


chains/solana/contracts/programs/ccip-offramp/src/instructions/interfaces.rs
```diff
@@ -95,7 +99,7 @@ pub trait Admin {
     fn set_ocr_config(
         &self,
         ctx: Context<SetOcrConfig>,
-        plugin_type: u8, // OcrPluginType, u8 used because anchor tests did not work with an enum
+        plugin_type: OcrPluginType,
```
agusaldasoro, line 102: Is this a breaking change?  If yes, then we need to inform tooling team to update their interactions as well.

agusaldasoro (dismissed):

agusaldasoro (approved):

PR #633, opened 2025-02-19
Misc. Fixes

asoliman92:
Misc. fixes

### DIFF
--- a/commit/chainfee/validate_observation.go
+++ b/commit/chainfee/validate_observation.go
@@ -19,7 +19,6 @@ func (p *processor) ValidateObservation(
 	ao plugincommon.AttributedObservation[Observation],
 ) error {
 	obs := ao.Observation
-	zero := big.NewInt(0)
 
 	if obs.IsEmpty() {
 		return nil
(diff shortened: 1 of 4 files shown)


commit/chainfee/validate_observation.go
```diff
@@ -67,7 +66,7 @@ func validateChainFeeUpdates(
 		if update.ChainFee.DataAvFeePriceUSD == nil || update.ChainFee.DataAvFeePriceUSD.Cmp(big.NewInt(0)) < 0 {
 			return fmt.Errorf("nil or negative %s", "data availability fee price")
 		}
-		if update.Timestamp.IsZero() {
+		if update.Timestamp.IsZero() || update.Timestamp.After(time.Now().UTC()) {
```
makramkd, line 69: Maybe better as a separate condition + error msg?

pkg/types/ccipocr3/common_types.go
```diff
@@ -176,5 +176,9 @@ func (b BigInt) IsEmpty() bool {
 }
 
 func (b BigInt) IsPositive() bool {
-	return b.Int != nil && b.Int.Cmp(big.NewInt(0)) > 0
+	return Positive(b.Int)
+}
+
+func Positive(b *big.Int) bool {
+	return b != nil && b.Cmp(big.NewInt(0)) > 0
```
makramkd, line 183: This really worth the refactor?

asoliman92: Not really, was planning on using it everywhere we do the conditions for bit.Int across the code and ditched it for using it only here and doing the rest of the refactor later.

PR #634, opened 2025-02-19
More validations in report validations

asoliman92:


commit/report.go
```diff
@@ -154,6 +154,30 @@ func (p *Plugin) validateReport(
...
+			return false, cciptypes.CommitPluginReport{}, nil
+		}
+	}
+
+	for _, root := range decodedReport.BlessedMerkleRoots {
+		if root.SeqNumsRange.Start() > root.SeqNumsRange.End() {
+			lggr.Warnw("invalid seqNumsRange", "root", root)
+			return false, cciptypes.CommitPluginReport{}, nil
+		}
+	}
```
makramkd, line 169: Can we merge this loop with the one above?

commit/report.go
```diff
@@ -154,6 +154,30 @@ func (p *Plugin) validateReport(
 		return false, cciptypes.CommitPluginReport{}, fmt.Errorf("decode report info: %w", err)
 	}
 
+	for _, root := range decodedReport.BlessedMerkleRoots {
```
makramkd, line 157: Do we need similar checks for `UnblessedMerkleRoots`?

asoliman92: Can they revert on chain? If not then I think Blessed is enough?

makramkd: They can indeed

makramkd (approved):

PR #635, opened 2025-02-19
add address codec interface

huangzhen1997:
core ref: ff61f3c72f2f7688e38d7fc32c58915084c812a0
https://github.com/smartcontractkit/chainlink-ccip/pull/619#discussion_r1960494733


pkg/types/ccipocr3/interfaces.go
```diff
@@ -20,11 +20,9 @@ type MessageHasher interface {
...
 }
 
-type ExtraDataCodec interface {
-	// DecodeExtraArgs reformat bytes into a chain agnostic map[string]any representation for extra args
-	DecodeExtraArgs(extraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
-	// DecodeTokenAmountDestExecData reformat bytes to chain-agnostic map[string]any for tokenAmount DestExecData field
-	DecodeTokenAmountDestExecData(destExecData Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
+type AddressCodec interface {
+	AddressBytesToString([]byte) (string, error)
+	AddressStringToBytes(string) ([]byte, error)
```
winder, line 25: We typically decode addresses based on the source chain, so I think this interface needs to include the chain selector.

You can also use the `UnknownAddress` type to make sure we pass in the correct type. Changing the type would probably let you remove many spots that cast the UnknownAddress type to bytes.

pkg/types/ccipocr3/interfaces.go
```diff
@@ -20,11 +20,9 @@ type MessageHasher interface {
 	Hash(context.Context, Message) (Bytes32, error)
 }
 
-type ExtraDataCodec interface {
-	// DecodeExtraArgs reformat bytes into a chain agnostic map[string]any representation for extra args
-	DecodeExtraArgs(extraArgs Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
-	// DecodeTokenAmountDestExecData reformat bytes to chain-agnostic map[string]any for tokenAmount DestExecData field
-	DecodeTokenAmountDestExecData(destExecData Bytes, sourceChainSelector ChainSelector) (map[string]any, error)
+type AddressCodec interface {
+	AddressBytesToString([]byte) (string, error)
```
prashantkumar1982, line 24: We would need to also pass in the ChainSelector param, I think.
All the ccip code which calls this interface may not know or have this codec instance tied to a chain in my opinion.

prashantkumar1982 (dismissed):

winder (dismissed):

prashantkumar1982 (approved):

winder (approved):

PR #637, opened 2025-02-20
[Solana][NonEVM-1309] Fix address trimming

PabloMansanet:


chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/config.rs
```diff
@@ -1,34 +1,16 @@
...
-    // filter out empty values
-    if !first_item.is_empty() {
-        on_ramps.push(first_item);
-    }
-    if !second_item.is_empty() {
-        on_ramps.push(second_item);
-    }
-
-    on_ramps
+pub fn get_on_ramps(config: &SourceChainConfig) -> impl Iterator<Item = &[u8; 64]> {
```
agusaldasoro, line 3: can you add a comment here on why we are doing this?

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/config.rs
```diff
@@ -1,34 +1,16 @@
...
-    let end = input
-        .iter()
-        .rposition(|&byte| byte != 0)
-        .map_or(0, |pos| pos + 1);
-    &input[..end]
+    if on_ramp.len() > 64 {
+        return false;
+    }
+    let prefix_matches = |long: &[u8], short: &[u8]| &long[0..short.len()] == short;
+    let suffix_is_zero = |a: &[u8], start: usize| a[start..].iter().all(|b| *b == 0);
```
agusaldasoro, line 12: can you add a comment here explaining this two lines?

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/config.rs
```diff
@@ -1,34 +1,16 @@
...
+    get_on_ramps(config).any(|o| prefix_matches(o, on_ramp) && suffix_is_zero(o, on_ramp.len()))
```
toblich, line 13: This seems like it will probably work fine in practice, though now the following case could still happen:
- Configured router address: `0xAAAAAAAAA0000000` (plus more zeroes at the end)

All the following passed in router addresses to validate  will be successful, even though in theory only one is the true one:
- `0xAAAAAAAAA0000000`
- `0xAAAAAAAAA00000`
- `0xAAAAAAAAA000`
- ...

Just a suggestion of an alternative implementation to this could be, instead of storing `pub on_ramp: [[u8; 64]; 2]` in our state, doing something like the following:

```rs
// this struct still has a fixed size, so we don't need to realloc or anything like that
pub struct OnRampAddress {
  pub padded_address: [u8; 64],
  pub len: u8,
}

#[derive(Clone, AnchorSerialize, AnchorDeserialize, InitSpace, Debug)]
    // ... other props here

    pub on_ramp: [OnRampAddress; 2],
}

```

And then you can just take the original `len` into account when validating if some untrusted onramp address is in fact valid or not.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/config.rs
PabloMansanet: Yep, will do 🙂

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/config.rs
PabloMansanet: 🤔 I can't think of how I'd write those comments in a way that doesn't simply reiterate what the code says. If you think the code is unclear, that's fair: I can try to rewrite it in a less dense way rather than try to back it up with a comment.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/config.rs
PabloMansanet: Your first solution definitely looks good, it's basically a mini-arrayvec :) will do that.

PabloMansanet: On second thought: Why not just use a `Vec` instead for storage? Is it a dynamic length/realloc concern?

We are using vectors in many places and I feel like we have enough experience with `realloc` to prefer it to somewhat fragile `arrayvec` implementations. Especially in this case when we'll be comparing the data to a `vec` later on.

PabloMansanet: Hmm yeah, bothering tooling with this is probably reason enough to do the bespoke arrayvec.

archseer: Was just about to propose this: we **should** use `{ len: u8, data: [u8; 64] }`.

We've already ran into issues with this padding in offchain tooling, since the addresses would have to be trimmed in a source-chain specific way which is really difficult to do in CR. For the time being we've left padded the addresses rather than right padding them (since that's the behaviour supported by EVM: `0x00ff` == `0xff`, but `0xff` and `0xff00` are clearly not the same)

archseer (changes requested):

archseer (approved):

PR #638, opened 2025-02-20
Add multiple commit reports and commit report splitting.

winder:
core ref: e4baf61cbfe27a2df0142e80e250cafd1c9fcbd4
core ref pr: https://github.com/smartcontractkit/chainlink/pull/16626

Add options to truncate and split commit reports so that a single merkle root is included in each report.

Manually integration tested these options with `Test_CCIPBatching_MultiSource`, updated config with:
```
		testhelpers.WithOCRConfigOverride(func(params *v1_6.CCIPOCRParams) {
			params.CommitOffChainConfig.MultipleReports = true
			params.CommitOffChainConfig.MaxMerkleRootsPerReport = 1
		}),
```

I see in the logs that multiple reports are built and tests pass.

[Followup ticket for integration test.](https://smartcontract-it.atlassian.net/browse/CCIP-5329)


commit/report.go
```diff
@@ -98,6 +84,184 @@ func (p *Plugin) Reports(
...
+	}
+
+	for _, r := range outcome.MerkleRootOutcome.RootsToReport {
+
+		if outcome.MerkleRootOutcome.RMNEnabledChains[r.ChainSel] {
+			blessedMerkleRoots = append(blessedMerkleRoots, r)
+		} else {
+			unblessedMerkleRoots = append(unblessedMerkleRoots, r)
+		}
+
```
winder, line 170: numRoots++

commit/merkleroot/outcome.go
```diff
@@ -293,6 +293,7 @@ func checkForReportTransmission(
 	consensusObservation consensusObservation,
 ) Outcome {
 	offRampUpdated := false
+	// TODO: does this need to be updated to ensure all seqNum's are updated instead of just one?
```
winder, line 296: @dimkouv it looks like a single updated offramp allows us to transition states. If we were to continue with checking all all seqNums would that work with multiple reports?

commit/report.go
```diff
@@ -22,73 +24,57 @@ import (
...
 		sort.Slice(allRoots, func(i, j int) bool { return allRoots[i].ChainSel < allRoots[j].ChainSel })
 		repInfo = cciptypes.CommitReportInfo{
-			RemoteF:     outcome.MerkleRootOutcome.RMNRemoteCfg.FSign,
+			RemoteF:     rmnRemoteFSign,
 			MerkleRoots: allRoots,
-			TokenPrices: rep.PriceUpdates.TokenPriceUpdates,
 		}
 	}
+	// in case of a price-only update, add prices regardless of outcome type.
+	repInfo.PriceUpdates = rep.PriceUpdates
```
winder, line 70: Pulled in from https://github.com/smartcontractkit/chainlink-ccip/pull/621, this type is used by the contract writer for Solana.

commit/report.go
```diff
@@ -133,41 +315,44 @@ func (p *Plugin) validateReport(
 	lggr logger.Logger,
 	seqNr uint64,
 	r ocr3types.ReportWithInfo[[]byte],
-) (bool, cciptypes.CommitPluginReport, error) {
```
winder, line 136: The error type can now be used to distinguish between an invalid report and a validation error. This way we can pass additional context about why an error has occurred.

commit/merkleroot/outcome.go
makramkd: The plugin only ever operates on a single offramp, there is no multiple-offramp scenario.

commit/report.go
```diff
@@ -22,73 +24,57 @@ import (
...
-		"roots", outcome.MerkleRootOutcome.RootsToReport,
-		"tokenPriceUpdates", outcome.TokenPriceOutcome.TokenPrices,
-		"gasPriceUpdates", outcome.ChainFeeOutcome.GasPrices,
-		"rmnSignatures", outcome.MerkleRootOutcome.RMNReportSignatures,
-	)
-
+// buildOneReport is the common logic for building a report. Different report building
+// algorithms can reassemble reports by selecting which blessed and unblessed merkle
+// roots to include in the report, and which price updates and rmn signatures to use.
+func buildOneReport(
```
makramkd, line 30: Is there space for nicer/cleaner abstractions? e.g

```golang
type ReportBuilder interface {
  // Build returns a tuple of (reports, reportInfos) that can be encoded and transmitted onchain.
  // usage:
  // reports, infos, err := builder.Build(ctx)
  // if err != nil { ... }
  // reportsEncoded, infosEncoded, err := encodeReportsAndInfos(reports, infos)
  // if err != nil { ... }
  // reportsPlus := []ocr3types.ReportPlus[...]{...}
  Build(ctx context.Context) ([]cciptypes.CommitPluginReport, []cciptypes.CommitReportInfo, error)
}

// multiSourceSingleReport batches commit roots and prices from multiple sources
// into a single report. The destination chain must support multiple roots and multiple
// price updates in the same report in order for this builder to work.
type multiSourceSingleReport struct { ... }

// multiSourceMultiReport creates a separate report for each (source, commit root)
// pair. Prices are included on a first-come first-serve basis. If the destination chain
// is limited in that it only supports a single root per report
type multiSourceMultiReport struct { ... }

// potential future ones
// separateRootsAndPricesMultiReport separates roots and prices into separate commit
// reports. 
type separateRootsAndPricesMultiReport struct { ... }
```

commit/report.go
```diff
@@ -22,73 +24,57 @@ import (
...
+	lggr logger.Logger,
+	reportCodec cciptypes.CommitPluginCodec,
+	transmissionSchedule *ocr3types.TransmissionSchedule,
+	outcomeType merkleroot.OutcomeType,
+	blessedMerkleRoots []cciptypes.MerkleRootChain,
+	unblessedMerkleRoots []cciptypes.MerkleRootChain,
+	rmnSignatures []cciptypes.RMNECDSASignature,
+	rmnRemoteFSign uint64,
+	priceUpdates cciptypes.PriceUpdates,
+) (*ocr3types.ReportPlus[[]byte], error) {
```
makramkd, line 41: Are there unit tests for this?

commit/report.go
```diff
@@ -98,6 +84,189 @@ func (p *Plugin) Reports(
...
+		return nil, err
+	}
+	if report != nil {
+		return []ocr3types.ReportPlus[[]byte]{*report}, nil
+	}
+	return nil, nil
+}
+
+// buildMultipleReports builds many reports of with at most maxMerkleRootsPerReport roots.
+func buildMultipleReports(
```
makramkd, line 144: Can we have a unit test for this

commit/report.go
```diff
@@ -98,6 +84,189 @@ func (p *Plugin) Reports(
...
+		if report != nil {
+			reports = append(reports, *report)
+		}
+	}
+
+	return reports, nil
+}
+
+// buildTruncatedReport returns a single truncated report.
+func buildTruncatedReport(
```
makramkd, line 232: Ditto here re: unit tests

commit/report.go
```diff
@@ -109,17 +278,30 @@ func (p *Plugin) Reports(
...
-	return []ocr3types.ReportPlus[[]byte]{
-		{
-			ReportWithInfo: ocr3types.ReportWithInfo[[]byte]{
-				Report: encodedReport,
-				Info:   encodedInfo,
-			},
-			TransmissionScheduleOverride: transmissionSchedule,
-		},
-	}, nil
+	// The Solana report has some special limitations. Those limitations can be used if the offchain
```
makramkd, line 283: Incomplete comment

commit/report.go
```diff
@@ -109,17 +278,30 @@ func (p *Plugin) Reports(
...
+		if p.offchainCfg.MultipleReports {
+			lggr.Infow("building multiple reports")
+			reports, err = buildMultipleReports(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
+		} else {
+			lggr.Infow("building truncated reports")
+			reports, err = buildTruncatedReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
+		}
+	} else {
+		reports, err = buildStandardReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome)
+	}
```
makramkd, line 295: Can this kinda logic be in a helper?

commit/report.go
```diff
@@ -98,6 +84,189 @@ func (p *Plugin) Reports(
...
+		}
+		if report != nil {
+			reports = append(reports, *report)
+		}
+	}
+
+	return reports, nil
+}
+
+// buildTruncatedReport returns a single truncated report.
```
makramkd, line 231: Its not clear what a truncated report is, we seem to trim the slice, but no reason as to why we're doing that is written far as I can tell

commit/report.go
```diff
@@ -323,13 +517,13 @@ func (p *Plugin) ShouldTransmitAcceptedReport(
 ) (bool, error) {
 	ctx, lggr := logutil.WithOCRInfo(ctx, p.lggr, seqNr, logutil.PhaseShouldTransmit)
 
-	valid, decodedReport, err := p.validateReport(ctx, lggr, seqNr, r)
-	if err != nil {
+	decodedReport, err := p.validateReport(ctx, lggr, seqNr, r)
+	if errors.Is(err, plugincommon.ErrValidationError) {
```
makramkd, line 521: Are these new changes covered by tests?

execute/plugin.go
```diff
@@ -453,6 +453,10 @@ func (p *Plugin) Reports(
 		return nil, fmt.Errorf("unable to encode report: %w", err)
 	}
 
+	if len(decodedOutcome.Report.ChainReports) == 0 {
+		lggr.Warn("empty report", "report", decodedOutcome.Report)
+		return nil, nil
+	}
```
makramkd, line 459: Should this check be moved to right after we decode the outcome on line 449?

internal/plugincommon/validations.go
```diff
@@ -15,3 +16,40 @@ func ValidateFChain(fChain map[cciptypes.ChainSelector]int) error {
 	}
 	return nil
 }
+
+var ErrValidationError = errors.New("report validation errors")
+var ErrInvalidReport = errors.New("invalid report")
```
makramkd, line 21: Two things: naming nit and godoc
```suggestion
// ErrReportValidation is used when the validation being performed in order to validate a report has errored.
// This does not necessarily mean that the report is valid or invalid.
var ErrReportValidation = errors.New("report validation error")
// ErrInvalidReport is used when a report is found to be invalid.
var ErrInvalidReport = errors.New("invalid report")
```

internal/plugincommon/validations.go
```diff
@@ -15,3 +16,40 @@ func ValidateFChain(fChain map[cciptypes.ChainSelector]int) error {
...
+
+// NewErrValidatingReport is returned when the report could not be validated due to an error.
+func NewErrValidatingReport(err error) error {
+	return &errWrappedValidatingReport{err: err}
+}
```
makramkd, line 26: Do we need a separate struct? I thought `fmt.Errorf` works just fine with `errors.Is`?

pkg/types/ccipocr3/plugin_commit_types.go
```diff
@@ -74,7 +74,7 @@ type CommitReportInfo struct {
 	// RemoteF Max number of faulty RMN nodes; f+1 signers are required to verify a report.
 	RemoteF     uint64            `json:"remoteF"`
 	MerkleRoots []MerkleRootChain `json:"merkleRoots"`
-	TokenPrices []TokenPrice      `json:"tokenPrices"`
+	PriceUpdates
```
makramkd, line 77: Isn't this a breaking change? Or there's a minimal effect to it?

pluginconfig/commit.go
```diff
@@ -146,6 +146,12 @@ type CommitOffchainConfig struct {
 
 	// ChainFeeAsyncObserverSyncTimeout defines the timeout for a single sync operation (e.g. fetch token prices).
 	ChainFeeAsyncObserverSyncTimeout time.Duration `json:"chainFeeAsyncObserverSyncTimeout"`
+
+	// MaxRootsPerReport is the maximum number of roots to include in a single report.
+	MaxMerkleRootsPerReport uint64 `json:"maxRootsPerReport"`
+
+	// MultipleReports is a flag to enable/disable multiple reports per round.
+	MultipleReports bool `json:"multipleReports"`
```
makramkd, line 154: Following up from feedback from other around the config docs, lets add some more info on when these would be used. Also tiny rename to make boolean vars more obvious.

```suggestion
	// MaxRootsPerReport is the maximum number of roots to include in a single report.
	// Set this to 1 for destination chains that cannot process more than one commit root per report (e.g, Solana)
	// When set to 0, there is no maximum. <-- Is this the case?
	MaxMerkleRootsPerReport uint64 `json:"maxRootsPerReport"`

	// MultipleReports is a flag to enable/disable multiple reports per round.
	// This is typically set to true on chains that have `MaxMerkleRootsPerReport` equal to 1,
	// in order to remedy throughput issues of the commit plugin.
	// NOTE: this can only be set if RMNEnabled == false at the moment.
	MultipleReportsEnabled bool `json:"multipleReportsEnabled"`
```

makramkd:
Can we add an integration test w/ these settings applied in chainlink?

internal/plugincommon/validations.go
winder: `errors.Is` only works with error constants, you need the struct so that `Unwrap` returns a constant. I think the alternative would be to return `fmt.Errorf("error reason: %w", ErrValidatingReport)`.

pkg/types/ccipocr3/plugin_commit_types.go
winder: Pretty sure it's non-breaking because the embedded type would expose `tokenPrices` at the same json path.

pluginconfig/commit.go
winder: Good call, updated.

pluginconfig/commit.go
```diff
@@ -146,6 +146,12 @@ type CommitOffchainConfig struct {
 
 	// ChainFeeAsyncObserverSyncTimeout defines the timeout for a single sync operation (e.g. fetch token prices).
 	ChainFeeAsyncObserverSyncTimeout time.Duration `json:"chainFeeAsyncObserverSyncTimeout"`
+
+	// MaxRootsPerReport is the maximum number of roots to include in a single report.
+	MaxMerkleRootsPerReport uint64 `json:"maxRootsPerReport"`
```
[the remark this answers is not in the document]
winder: In general, I think it's good to avoid boolean flags if there's a real configuration that could be used instead. [Lots of discussion about the pattern here](https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior). Maybe I should also use `MaxReports` instead of `MultipleReportsEnabled`.

winder:
> ```
> 	maxRoots := p.offchainCfg.MaxMerkleRootsPerReport
> 	if !p.offchainCfg.RMNEnabled && maxRoots != 0 {
> 		if p.offchainCfg.MultipleReports {
> 			lggr.Infow("building multiple reports")
> 			reports, err = buildMultipleReports(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
> 		} else {
> 			lggr.Infow("building truncated reports")
> 			reports, err = buildTruncatedReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
> 		}
> 	} else {
> 		reports, err = buildStandardReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome)
> 	}
> ```
> 
> There are three report building functions
> 
>     1. multiple reports
> 
>     2. truncated report
> 
>     3. standard report
> 
> 
> What is the truncated report used for? I thought we only wanted to be able to build multiple reports with one root per report and the price updates on the first report.

It could be useful for debugging to have truncated reports instead of multiple reports.

commit/report.go
winder: Yes, there are many tests for the report validation function. Many of them failed until I got this logic just right.

commit/report.go
winder: It's just more flexibility for the NonEVMs if they need it.

commit/merkleroot/outcome.go
winder: Isn't there a seqNum for each source chain? `OffRampNextSeqNums` is `map[cciptypes.ChainSelector]cciptypes.SeqNum`. If we split reports across 5 source chains don't we need to wait until all 5 reports transmit?

commit/report.go
winder: This seems like a lot of extra code for what is essentially the same abstraction. I've already implemented 3 pure functions, this is the same thing but the functions are attached to an interface.

I could have the pure functions implement a function type but then there are unused parameters in some of the implementations.

```
diff --git a/commit/report.go b/commit/report.go
index 4fcaaced..0d6bc504 100644
--- a/commit/report.go
+++ b/commit/report.go
@@ -24,6 +24,51 @@ import (
 )
 
+type ReportBuilder interface {
+	// Build returns a tuple of (reports, reportInfos) that can be encoded and transmitted onchain.
+	// usage:
+	// reports, infos, err := builder.Build(ctx)
+	// if err != nil { ... }
+	// reportsEncoded, infosEncoded, err := encodeReportsAndInfos(reports, infos)
+	// if err != nil { ... }
+	// reportsPlus := []ocr3types.ReportPlus[...]{...}
+	Build(ctx context.Context) ([]cciptypes.CommitPluginReport, []cciptypes.CommitReportInfo, error)
+}
+
+// multiSourceSingleReport batches commit roots and prices from multiple sources
+// into a single report. The destination chain must support multiple roots and multiple
+// price updates in the same report in order for this builder to work.
+type multiSourceSingleReport struct {
+	lggr         logger.Logger
+	reportCodec  cciptypes.CommitPluginCodec
+	transmission *ocr3types.TransmissionSchedule
+	outcome      committypes.Outcome
+}
+
+// multiSourceMultiReport creates a separate report for each (source, commit root)
+// pair. Prices are included on a first-come first-serve basis. If the destination chain
+// is limited in that it only supports a single root per report
+type multiSourceMultiReport struct {
+	lggr                    logger.Logger
+	reportCodec             cciptypes.CommitPluginCodec
+	transmissionSchedule    *ocr3types.TransmissionSchedule
+	outcome                 committypes.Outcome
+	maxMerkleRootsPerReport uint64
+}
+
+type multiSourceTruncatedReport struct {
+	lggr                    logger.Logger
+	reportCodec             cciptypes.CommitPluginCodec
+	transmissionSchedule    *ocr3types.TransmissionSchedule
+	outcome                 committypes.Outcome
+	maxMerkleRootsPerReport uint64
+}
+
+// potential future ones
+// separateRootsAndPricesMultiReport separates roots and prices into separate commit
+// reports.
+//type separateRootsAndPricesMultiReport struct { ... }
+
 // buildOneReport is the common logic for building a report. Different report building
 // algorithms can reassemble reports by selecting which blessed and unblessed merkle
 // roots to include in the report, and which price updates and rmn signatures to use.
@@ -95,19 +140,16 @@ func buildOneReport(
 	}, nil
 }
 
-func buildStandardReport(
+// Build builds a single report with all the roots and prices.
+func (m multiSourceSingleReport) Build(
 	ctx context.Context,
-	lggr logger.Logger,
-	reportCodec cciptypes.CommitPluginCodec,
-	transmission *ocr3types.TransmissionSchedule,
-	outcome committypes.Outcome,
 ) ([]ocr3types.ReportPlus[[]byte], error) {
 
 	blessedMerkleRoots := make([]cciptypes.MerkleRootChain, 0)
 	unblessedMerkleRoots := make([]cciptypes.MerkleRootChain, 0)
 
-	for _, r := range outcome.MerkleRootOutcome.RootsToReport {
-		if outcome.MerkleRootOutcome.RMNEnabledChains[r.ChainSel] {
+	for _, r := range m.outcome.MerkleRootOutcome.RootsToReport {
+		if m.outcome.MerkleRootOutcome.RMNEnabledChains[r.ChainSel] {
 			blessedMerkleRoots = append(blessedMerkleRoots, r)
 		} else {
 			unblessedMerkleRoots = append(unblessedMerkleRoots, r)
@@ -115,20 +157,20 @@ func buildStandardReport(
 	}
 
 	priceUpdates := cciptypes.PriceUpdates{
-		TokenPriceUpdates: outcome.TokenPriceOutcome.TokenPrices.ToSortedSlice(),
-		GasPriceUpdates:   outcome.ChainFeeOutcome.GasPrices,
+		TokenPriceUpdates: m.outcome.TokenPriceOutcome.TokenPrices.ToSortedSlice(),
+		GasPriceUpdates:   m.outcome.ChainFeeOutcome.GasPrices,
 	}
 
 	report, err := buildOneReport(
 		ctx,
-		lggr,
-		reportCodec,
-		transmission,
-		outcome.MerkleRootOutcome.OutcomeType,
+		m.lggr,
+		m.reportCodec,
+		m.transmission,
+		m.outcome.MerkleRootOutcome.OutcomeType,
 		blessedMerkleRoots,
 		unblessedMerkleRoots,
-		outcome.MerkleRootOutcome.RMNReportSignatures,
-		outcome.MerkleRootOutcome.RMNRemoteCfg.FSign,
+		m.outcome.MerkleRootOutcome.RMNReportSignatures,
+		m.outcome.MerkleRootOutcome.RMNRemoteCfg.FSign,
 		priceUpdates,
 	)
 	if err != nil {
@@ -140,14 +182,9 @@ func buildStandardReport(
 	return nil, nil
 }
 
-// buildMultipleReports builds many reports of with at most maxMerkleRootsPerReport roots.
-func buildMultipleReports(
+// Build builds many reports of with at most maxMerkleRootsPerReport roots.
+func (m multiSourceMultiReport) Build(
 	ctx context.Context,
-	lggr logger.Logger,
-	reportCodec cciptypes.CommitPluginCodec,
-	transmissionSchedule *ocr3types.TransmissionSchedule,
-	outcome committypes.Outcome,
-	maxMerkleRootsPerReport uint64,
 ) ([]ocr3types.ReportPlus[[]byte], error) {
 	var reports []ocr3types.ReportPlus[[]byte]
 
@@ -156,11 +193,11 @@ func buildMultipleReports(
 	unblessedMerkleRoots := make([]cciptypes.MerkleRootChain, 0)
 
 	priceUpdates := cciptypes.PriceUpdates{
-		TokenPriceUpdates: outcome.TokenPriceOutcome.TokenPrices.ToSortedSlice(),
-		GasPriceUpdates:   outcome.ChainFeeOutcome.GasPrices,
+		TokenPriceUpdates: m.outcome.TokenPriceOutcome.TokenPrices.ToSortedSlice(),
+		GasPriceUpdates:   m.outcome.ChainFeeOutcome.GasPrices,
 	}
 
-	for _, r := range outcome.MerkleRootOutcome.RootsToReport {
+	for _, r := range m.outcome.MerkleRootOutcome.RootsToReport {
 
 		// TODO: Support RMN.
 		/*
@@ -173,13 +210,13 @@ func buildMultipleReports(
 		unblessedMerkleRoots = append(unblessedMerkleRoots, r)
 		numRoots++
 
-		if numRoots == maxMerkleRootsPerReport {
+		if numRoots == m.maxMerkleRootsPerReport {
 			report, err := buildOneReport(
 				ctx,
-				lggr,
-				reportCodec,
-				transmissionSchedule,
-				outcome.MerkleRootOutcome.OutcomeType,
+				m.lggr,
+				m.reportCodec,
+				m.transmissionSchedule,
+				m.outcome.MerkleRootOutcome.OutcomeType,
 				blessedMerkleRoots,
 				unblessedMerkleRoots,
 				nil, // no RMN for partial reports.
@@ -207,10 +244,10 @@ func buildMultipleReports(
 	if numRoots > 0 {
 		report, err := buildOneReport(
 			ctx,
-			lggr,
-			reportCodec,
-			transmissionSchedule,
-			outcome.MerkleRootOutcome.OutcomeType,
+			m.lggr,
+			m.reportCodec,
+			m.transmissionSchedule,
+			m.outcome.MerkleRootOutcome.OutcomeType,
 			blessedMerkleRoots,
 			unblessedMerkleRoots,
 			nil, // no RMN for partial reports.
@@ -228,21 +265,22 @@ func buildMultipleReports(
 	return reports, nil
 }
 
-// buildTruncatedReport returns a single truncated report.
-func buildTruncatedReport(
+// Build returns a single truncated report.
+func (m multiSourceTruncatedReport) Build(
 	ctx context.Context,
-	lggr logger.Logger,
-	reportCodec cciptypes.CommitPluginCodec,
-	transmissionSchedule *ocr3types.TransmissionSchedule,
-	outcome committypes.Outcome,
-	maxMerkleRootsPerReport uint64,
 ) ([]ocr3types.ReportPlus[[]byte], error) {
-	reports, err := buildMultipleReports(ctx, lggr, reportCodec, transmissionSchedule, outcome, maxMerkleRootsPerReport)
+	reports, err := multiSourceMultiReport{
+		lggr:                    m.lggr,
+		reportCodec:             m.reportCodec,
+		transmissionSchedule:    m.transmissionSchedule,
+		outcome:                 m.outcome,
+		maxMerkleRootsPerReport: m.maxMerkleRootsPerReport,
+	}.Build(ctx)
 	if err != nil {
 		return nil, err
 	}
 	if len(reports) > 1 {
-		lggr.Warnf("buildTruncatedReport: truncating report %d -> 1", len(reports))
+		m.lggr.Warnf("buildTruncatedReport: truncating report %d -> 1", len(reports))
 		return reports[:1], nil
 	}
 
@@ -280,18 +318,36 @@ func (p *Plugin) Reports(
 
 	var reports []ocr3types.ReportPlus[[]byte]
 
-	// The Solana report has some special limitations. Those limitations can be used if the offchain
+	// These options were added to allow for more flexibility around report building. For example Solana
+	// only supports a single merkle root per report.
 	maxRoots := p.offchainCfg.MaxMerkleRootsPerReport
 	if !p.offchainCfg.RMNEnabled && maxRoots != 0 {
 		if p.offchainCfg.MultipleReportsEnabled {
 			lggr.Infow("building multiple reports")
-			reports, err = buildMultipleReports(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
+			reports, err = multiSourceMultiReport{
+				lggr:                    lggr,
+				reportCodec:             p.reportCodec,
+				transmissionSchedule:    transmissionSchedule,
+				outcome:                 outcome,
+				maxMerkleRootsPerReport: maxRoots,
+			}.Build(ctx)
 		} else {
 			lggr.Infow("building truncated reports")
-			reports, err = buildTruncatedReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
+			reports, err = multiSourceTruncatedReport{
+				lggr:                    lggr,
+				reportCodec:             p.reportCodec,
+				transmissionSchedule:    transmissionSchedule,
+				outcome:                 outcome,
+				maxMerkleRootsPerReport: maxRoots,
+			}.Build(ctx)
 		}
 	} else {
-		reports, err = buildStandardReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome)
+		reports, err = multiSourceSingleReport{
+			lggr:         lggr,
+			reportCodec:  p.reportCodec,
+			transmission: transmissionSchedule,
+			outcome:      outcome,
+		}.Build(ctx)
 	}
 
 	if err != nil {
```

winder: Here's the function type version:
```
diff --git a/commit/plugin.go b/commit/plugin.go
index 5136af10..14b2a9e5 100644
--- a/commit/plugin.go
+++ b/commit/plugin.go
@@ -48,6 +48,7 @@ type Plugin struct {
 	ccipReader        readerpkg.CCIPReader
 	tokenPricesReader readerpkg.PriceReader
 	reportCodec       cciptypes.CommitPluginCodec
+	reportBuilder     reportBuilderFunc
 	// Don't use this logger directly but rather through logutil\.WithContextValues where possible
 	lggr                logger.Logger
 	homeChain           reader.HomeChain
@@ -159,6 +160,21 @@ func NewPlugin(
 		reporter,
 	)
 
+	var reportBuilder reportBuilderFunc
+
+	// These options were added to allow for more flexibility around report building. For example Solana
+	// only supports a single merkle root per report.
+	maxRoots := offchainCfg.MaxMerkleRootsPerReport
+	if !offchainCfg.RMNEnabled && maxRoots != 0 {
+		if offchainCfg.MultipleReportsEnabled {
+			reportBuilder = buildMultipleReports
+		} else {
+			reportBuilder = buildTruncatedReport
+		}
+	} else {
+		reportBuilder = buildStandardReport
+	}
+
 	return &Plugin{
 		donID:               donID,
 		oracleID:            reportingCfg.OracleID,
@@ -178,6 +194,7 @@ func NewPlugin(
 		discoveryProcessor:  discoveryProcessor,
 		metricsReporter:     reporter,
 		ocrTypeCodec:        ocrtypecodec.DefaultCommitCodec,
+		reportBuilder:       reportBuilder,
 	}
 }
 
diff --git a/commit/report.go b/commit/report.go
index 9762f43c..c7ed66d2 100644
--- a/commit/report.go
+++ b/commit/report.go
@@ -24,6 +24,15 @@ import (
 )
 
+type reportBuilderFunc func(
+	ctx context.Context,
+	lggr logger.Logger,
+	reportCodec cciptypes.CommitPluginCodec,
+	transmissionSchedule *ocr3types.TransmissionSchedule,
+	outcome committypes.Outcome,
+	maxMerkleRootsPerReport uint64,
+) ([]ocr3types.ReportPlus[[]byte], error)
+
 // buildOneReport is the common logic for building a report. Different report building
 // algorithms can reassemble reports by selecting which blessed and unblessed merkle
 // roots to include in the report, and which price updates and rmn signatures to use.
@@ -101,8 +110,8 @@ func buildStandardReport(
 	reportCodec cciptypes.CommitPluginCodec,
 	transmission *ocr3types.TransmissionSchedule,
 	outcome committypes.Outcome,
+	_ uint64,
 ) ([]ocr3types.ReportPlus[[]byte], error) {
-
 	blessedMerkleRoots := make([]cciptypes.MerkleRootChain, 0)
 	unblessedMerkleRoots := make([]cciptypes.MerkleRootChain, 0)
 
@@ -278,22 +287,8 @@ func (p *Plugin) Reports(
 	lggr.Debugw("transmission schedule override",
 		"transmissionSchedule", transmissionSchedule, "oracleIDToP2PID", p.oracleIDToP2PID)
 
-	var reports []ocr3types.ReportPlus[[]byte]
-
-	// These options were added to allow for more flexibility around report building. For example Solana
-	// only supports a single merkle root per report.
 	maxRoots := p.offchainCfg.MaxMerkleRootsPerReport
-	if !p.offchainCfg.RMNEnabled && maxRoots != 0 {
-		if p.offchainCfg.MultipleReportsEnabled {
-			lggr.Infow("building multiple reports")
-			reports, err = buildMultipleReports(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
-		} else {
-			lggr.Infow("building truncated reports")
-			reports, err = buildTruncatedReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
-		}
-	} else {
-		reports, err = buildStandardReport(ctx, lggr, p.reportCodec, transmissionSchedule, outcome)
-	}
+	reports, err := p.reportBuilder(ctx, lggr, p.reportCodec, transmissionSchedule, outcome, maxRoots)
 
 	if err != nil {
 		return nil, fmt.Errorf("error while building reports: %w", err)
```

commit/report.go
winder: added tests.

commit/report.go
winder: done.

commit/report.go
winder: I added an abstraction and am now initializing this in the constructor. When we get to this point we'll simply call the handler function to build a report w/o needing to know which one we'll have.

commit/merkleroot/outcome.go
makramkd: > If we split reports across 5 source chains don't we need to wait until all 5 reports transmit?

Ideally, no. It would be better if the plugin can build on the chain(s) whose report(s) transmitted while waiting on other chains.

commit/plugin.go
```diff
@@ -159,6 +160,21 @@ func NewPlugin(
...
+	var reportBuilder reportBuilderFunc
+
+	// These options were added to allow for more flexibility around report building. For example Solana
+	// only supports a single merkle root per report.
+	maxRoots := offchainCfg.MaxMerkleRootsPerReport
+	if !offchainCfg.RMNEnabled && maxRoots != 0 {
+		if offchainCfg.MultipleReportsEnabled {
+			reportBuilder = buildMultipleReports
+		} else {
+			reportBuilder = buildTruncatedReport
```
makramkd, line 172: So this is enabled if:
* RMNEnabled == false
* MaxMerkleRootsPerReport > 0
* MultipleReportsEnabled == false
So we want to have potentially multiple roots per report, but only a single report (based on the code), isn't that just a standard report?

Or do we want to special case this on MaxMerkleRootsPerReport == 1? Cuz if you set that to 2, you'd get multiple roots per report.

commit/report.go
makramkd: I'd rather not add flexibility just for the sake of it because its more code to maintain, understand, test, etc. Do we have test runs where this truncated approach works and the multiple reports doesn't, or vice versa? cc @archseer

makramkd (commented):
Can't find the convo on slack but I think @archseer can test this for Solana, he has some test case working E2E for messaging, can we run that locally just to verify this gives us what we want? I also 2nd the truncated report thing, if this isn't going to be used, lets remove it to keep things simple.

commit/plugin.go
winder: >  So we want to have potentially multiple roots per report, but only a single report (based on the code), isn't that just a standard report?

It's not a standard report, this lets you have a single truncated report. Without this feature using `MaxMerkleRootsPerReport` without `MultipleReportsEnabled` will be a configuration error.

commit/merkleroot/outcome.go
winder: > > If we split reports across 5 source chains don't we need to wait until all 5 reports transmit?
> 
> Ideally, no. It would be better if the plugin can build on the chain(s) whose report(s) transmitted while waiting on other chains.

That seems like an unnecessary optimization. AFAIK all transmits should take about the same amount of time, and may even be included in the same block. What your suggesting would require the observation phase to use the same logic we have in the checkForReportTransmission phase.

commit/report.go
winder: This is the code we're talking about:
```
	reports, err := buildMultipleReports(ctx, lggr, reportCodec, transmissionSchedule, outcome, maxMerkleRootsPerReport)
	if err != nil {
		return nil, err
	}
	if len(reports) > 1 {
		lggr.Warnf("buildTruncatedReport: truncating report %d -> 1", len(reports))
		return reports[:1], nil
	}

	return reports, nil
```

I  don't understand the hangup here. It was easier to handle the case when `MultiplReportsEnabled == false` and `MaxMerkleRootsPerReport != 0` than it was to decide how to handle the error.

At this point the code is written and tested.

I'm not sure why we've spent so much time discussing this point when the wait for transmission function is apparently broken.

commit/merkleroot/rmn/controller.go
```diff
@@ -17,7 +17,7 @@ import (
 
 	mapset "github.com/deckarep/golang-set/v2"
 	"golang.org/x/exp/maps"
-	rand2 "golang.org/x/exp/rand"
+	rand2 "golang.org/x/exp/rand" //nolint:staticcheck // ignoring deprecated golang.org/x/exp/rand rule for now.
```
winder, line 20: Not sure how this slipped through to go1.24 PR. I wasn't able to find a seed option in the suggested math/rand/v2 library.

commit/merkleroot/outcome.go
```diff
@@ -316,7 +328,8 @@ func checkForReportTransmission(
 		}
 	}
 
-	if offRampUpdated {
+	// All pending sources have been updated, we can move to the next state.
+	if len(pendingSources) == 0 {
```
winder, line 332: @dimkouv could you double check the logic implemented above?

1. Initialize a set based on all chain selectors from `RootsToReport`. (this may already be broken if there are multiple roots for a single source chain selector.
2. While looping over previous and new seqNums, delete source selectors from the set. The idea being we'll have an empty set at the end if all source selectors have been updated.
3. If there is only 1 report, the logic is short-circuited similar to before.
4. At the end, if there are 0 remaining sources, then transmission must have completed.

commit/merkleroot/outcome.go
makramkd: Yeah, its an optimization, which is why I prefaced with "ideally" :-)

commit/merkleroot/outcome.go
winder: > Added two comments above. We'll also need a follow-up to keep making progress with the source chains that got their reports transmitted and not keep them blocked due to other source chains.

There's a `ReportTransmissionCheckAttempts` limit, I think that should prevent good transmissions from being completely blocked by a bad one. But if a bad report keeps recurring every plugin cycle would be slowed down at the checkReportTransmission phase.

Do you have a solution in mind? I'm not sure what to put in a ticket for this case.

makramkd (approved):

PR #639, opened 2025-02-20
[Solana] Optional merkle_root on commit accepted event

PabloMansanet:


chains/solana/contracts/programs/ccip-offramp/src/event.rs
```diff
@@ -5,9 +5,7 @@ use crate::state::MessageExecutionState;
 
 #[event]
 pub struct CommitReportAccepted {
-    // Will be zero'd (MerkleRoot::Default() in case the commit report included no
-    // merkle root. This is to circumvent a Go deserialization issue.)
-    pub merkle_root: MerkleRoot,
+    pub merkle_root: Option<MerkleRoot>,
```
agusaldasoro, line 8: We need to communicate this change with the following teams:
- Chain Reader
- Log Poller
- Observability

chains/solana/utils/common/event.go
```diff
@@ -69,3 +74,54 @@ func ParseMultipleEvents[T any](logs []string, event string, shouldPrint bool) (
...
+
+// Redefined to avoid a circular dependency
+type EventCommitReportAccepted struct {
+	Discriminator [8]byte
+	Report        *ccip_offramp.MerkleRoot `bin:"optional"`
+	PriceUpdates  ccip_offramp.PriceUpdates
+}
+
+// This event uses bespoke parsing, as it contains an `Optional` field which cannot be unmarshalled through `bin.UnmarshalBorsh`
+func ParseEventCommitReportAccepted(logs []string, event string, obj *EventCommitReportAccepted) error {
```
agusaldasoro, line 86: can you add a comment here on why we need to do this differently?

agusaldasoro (approved):

chains/solana/contracts/programs/ccip-offramp/src/event.rs
archseer: I've requested this change to fix decoding the report in CR/offchain so LGTM 👍🏻

PR #643, opened 2025-02-21
Snoozing unfinalized fully executed roots

0xnogo:
Modified execution tracking to handle reorgs more safely:

Executed + finalized messages: marked as permanently executed
Executed + unfinalized messages: temporarily snoozed
Added tests to verify different execution states

This ensures we don't reprocess finalized messages while allowing recovery from reorgs.

`core ref: fee74468b299bc7a2695826549f7d12394a8666f`

### DIFF
--- a/execute/tracked.go
+++ b/execute/tracked.go
@@ -85,7 +85,7 @@ func withTrackedMethod[T any](
 
 	p.reporter.TrackLatency(state, method, latency, err)
 	p.lggr.Debugw("tracking exec latency",
-		"state",
+		"state", state,
 		"method", method,
 		"latency", latency,
 	)


winder (commented):
Looks like you're on the right track :+1:

execute/plugin.go
```diff
@@ -205,32 +213,57 @@ func getPendingExecutedReports(
...
-		sort.Slice(executedMessages, func(i, j int) bool { return executedMessages[i] < executedMessages[j] })
+		// Get unfinalized messages by taking the difference
+		unfinalizedMsgSet := allExecutedMsgSet.Difference(finalizedMsgSet)
 
-		// Remove fully executed reports.
-		// Populate executed messages on the reports.
-		var executedCommits []exectypes.CommitData
-		groupedCommits[selector], executedCommits = combineReportsAndMessages(reports, executedMessages)
-		fullyExecuted = append(fullyExecuted, executedCommits...)
+		finalizedMessages := finalizedMsgSet.ToSlice()
```
mateusz-sekara, line 242: nit, but this could be extracted to a seperate helper function
```go

function toSortedSlice(set mapset.Set) []whatever {
                messages := set.ToSlice()
		sort.Slice(messages, func(i, j int) bool { return messages[i] < messages[j] })
                 return messages
}
```

execute/plugin.go
```diff
@@ -205,32 +213,57 @@ func getPendingExecutedReports(
...
-			return nil, nil, fmt.Errorf("compute report ranges: %w", err)
+			return nil, nil, nil, fmt.Errorf("compute report ranges: %w", err)
 		}
 
-		executedMessageSet := mapset.NewSet[cciptypes.SeqNum]()
+		// Get both finalized and unfinalized executed messages
+		finalizedMsgSet := mapset.NewSet[cciptypes.SeqNum]()
+		allExecutedMsgSet := mapset.NewSet[cciptypes.SeqNum]()
+
 		for _, seqRange := range ranges {
```
mateusz-sekara, line 223: Maybe not a problem for 6 chains, but we are sequentially looping over IO operations and that might slow the plugin down

mateusz-sekara: On the second thought, could that be replaced with 2 queries instead of 2 * len(ranges)?
Sounds like something that could bet done like this, no?

```go
allMessages, err := ccipReader.ExecutedMessages(ctx, selector, range1 or range2 or .... or rangeN, uncorfirmed)

finalized, err := ccipReader.ExecutedMessages(ctx, selector, range1 or range2 or .... or rangeN, finalized)
```

If that makes sense, can we have a ticket for that? I believe it should not be part of this PR. Of course chain reader, needs to support that types of queries

execute/observation.go
```diff
@@ -193,10 +210,16 @@ func (p *Plugin) getCommitReportsObservation(
 
 	// If fully executed reports are detected, mark them in the cache.
 	// This cache will be re-initialized on each plugin restart.
-	for _, fullyExecutedCommit := range fullyExecutedCommits {
+	for _, fullyExecutedCommit := range fullyFinalizedExecutedCommits {
```
mateusz-sekara, line 213: 👏

execute/observation.go
```diff
@@ -143,6 +143,23 @@ func (p *Plugin) getCurseInfo(ctx context.Context, lggr logger.Logger) (*reader.
...
+//   - MarkAsExecuted removes the commit root from the cache entirely to prevent reprocessing
+//   - These messages will never be considered for execution again
+//
+// 2. For unfinalized executions:
+//   - When messages are executed but not yet finalized, they are temporarily snoozed
+//   - Snoozing adds the commit root to a snooze cache with a TTL
+//   - If a reorg occurs and invalidates the execution, the messages become available again after the snooze period
+//   - This prevents duplicate executions while allowing recovery from reorgs
+//
+// 3. For cursed chains:
```
mateusz-sekara, line 159: Out of curiosity, why not remove cursed chains from the processing? It sounds like snoozing roots is not the proper use case for cursed chains, no?

0xnogo: Ideally they should be removed from the `CommitReportsGTETimestamp` either as part of the DB query or in the processing happening after.

It seems like we have the `sourceChainSelector` in the contract `MerkleRoot` struct.

@winder any opinion on that maybe?

mateusz-sekara (dismissed):

execute/observation.go
winder: I don't have a strong opinion. It was simple to add at the time and seemed to make sense.

I'm not sure about filtering it from the `CommitReportsGTETimestamp` call, if thats the way then we wouldn't have a record of these commit roots in the log.

execute/observation.go
```diff
@@ -143,6 +143,23 @@ func (p *Plugin) getCurseInfo(ctx context.Context, lggr logger.Logger) (*reader.
...
+// 1. For finalized executions:
+//   - When messages are executed with finality (on destChain), they are permanently marked as executed
+//   - MarkAsExecuted removes the commit root from the cache entirely to prevent reprocessing
+//   - These messages will never be considered for execution again
+//
+// 2. For unfinalized executions:
+//   - When messages are executed but not yet finalized, they are temporarily snoozed
+//   - Snoozing adds the commit root to a snooze cache with a TTL
+//   - If a reorg occurs and invalidates the execution, the messages become available again after the snooze period
+//   - This prevents duplicate executions while allowing recovery from reorgs
```
winder, line 157: Could you add a comment to the "InflightCacheExpiry" definition in pluginconfig/execute.go with details about how the parameter needs to be tuned according to desired behavior during a reorg?

execute/observation.go
```diff
@@ -193,10 +210,16 @@ func (p *Plugin) getCommitReportsObservation(
...
 	// If fully executed reports are detected, mark them in the cache.
 	// This cache will be re-initialized on each plugin restart.
-	for _, fullyExecutedCommit := range fullyExecutedCommits {
+	for _, fullyExecutedCommit := range fullyFinalizedExecutedCommits {
 		p.commitRootsCache.MarkAsExecuted(fullyExecutedCommit.SourceChain, fullyExecutedCommit.MerkleRoot)
 	}
 
+	// If fully executed reports are detected, snooze them in the cache.
+	// This cache will be re-initialized on each plugin restart.
+	for _, fullyExecutedCommit := range fullyUnfinalizedExecutedCommits {
```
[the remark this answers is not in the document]
winder: or even `fullyExecutedFinalized` / `fullyExecutedUnfinalized` like you have below

execute/observation.go
0xnogo: I'll create a ticket for that then

execute/observation.go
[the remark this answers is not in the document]
0xnogo: sure

execute/observation.go
0xnogo: Yes

mateusz-sekara (approved):

PR #647, opened 2025-02-21
replace typeconv with addresscodec

huangzhen1997:
core ref: 01f9124cb96c8c6aa82ab20d76dccb065a58e9e7


pkg/reader/ccip.go
```diff
@@ -159,16 +163,39 @@ func (r *ccipChainReader) CommitReportsGTETimestamp(
...
+	ctx context.Context, ts time.Time, limit int, ev *CommitReportAcceptedEvent,
+) ([]types.Sequence, error) {
+	return r.contractReaders[r.destChain].ExtendedQueryKey(
 		ctx,
 		consts.ContractNameOffRamp,
 		query.KeyFilter{
 			Key: consts.EventNameCommitReportAccepted,
 			Expressions: []query.Expression{
 				query.Timestamp(uint64(ts.Unix()), primitives.Gte),
-				// We don't need to wait for the commit report accepted event to be finalized
```
archseer, line 170: Why not keep this comment?

commit/merkleroot/rmn/controller.go
```diff
@@ -400,7 +398,7 @@ func (c *controller) sendObservationRequests(
 			fixedDestLaneUpdateRequests = append(fixedDestLaneUpdateRequests, &rmnpb.FixedDestLaneUpdateRequest{
 				LaneSource: &rmnpb.LaneSource{
 					SourceChainSelector: request.LaneSource.SourceChainSelector,
-					OnrampAddress:       typconv.KeepNRightBytes(request.LaneSource.OnrampAddress, 20),
```
prashantkumar1982, line 403: Good catch. KeepNRightBytes() was just slipping off address bytes.
@winder : This was a bug which would break any chain whose address was longer than 20 bytes.

@huangzhen1997 : Could you please also remove the function KeepNRightBytes()? Or better yet, just delete the file typeconv/address.go

commit/merkleroot/rmn/controller.go
```diff
@@ -660,9 +658,7 @@ func (c *controller) validateSignedObservationResponse(
 		}
 
 		// todo: The original updateReq contains abi encoded onRamp address, the one in the RMN response
-		// is 20 bytes evm address. This is chain specific and should be handled in a chain specific way.
```
prashantkumar1982, line 663: This whole todo seems suspect to me.
>> "he original updateReq contains abi encoded onRamp address"

@winder : Could you check this todo and code? It was wrong to clip the address to 20 bytes, but I didn't understand it completely.

I think @huangzhen1997's change is good, but want CCIP to review this closely.

commit/merkleroot/rmn/controller.go
```diff
@@ -660,9 +658,7 @@ func (c *controller) validateSignedObservationResponse(
 		}
 
 		// todo: The original updateReq contains abi encoded onRamp address, the one in the RMN response
```
prashantkumar1982, line 660: nit: remove the todo comment once CCIP team reviews this

execute/observation.go
```diff
@@ -359,7 +358,12 @@ func (p *Plugin) getFilterObservation(
 		}
 
 		for _, msg := range commitReport.Messages {
-			sender := typeconv.AddressBytesToString(msg.Sender[:], uint64(commitReport.SourceChain))
+			sender, err := p.addrCodec.AddressBytesToString(msg.Sender[:], commitReport.SourceChain)
+			if err != nil {
+				lggr.Errorw("unable to convert sender address to string", "err", err, "sender address", msg.Sender)
+				continue
```
prashantkumar1982, line 364: I would prefer to return error here, since something seriously wrong would cause the addresscodec to fail. By continuing, we are just silently ignoring the failure

@winder : Plz confirm

commit/merkleroot/outcome.go
```diff
@@ -223,10 +222,21 @@ func filterRootsBasedOnRmnSigs(
...
+	addressCodec cciptypes.AddressCodec,
 ) []cciptypes.MerkleRootChain {
 	// Create a set of signed roots for quick lookup.
 	signedRoots := mapset.NewSet[rootKey]()
 	for _, laneUpdate := range signedLaneUpdates {
+		addrStr, err := addressCodec.AddressBytesToString(
+			laneUpdate.LaneSource.OnrampAddress,
+			cciptypes.ChainSelector(laneUpdate.LaneSource.SourceChainSelector))
+		if err != nil {
+			lggr.Errorw("can't convert Onramp address to string",
```
prashantkumar1982, line 234: Change to critical error on all failures of the addressCodec

prashantkumar1982 (commented):
Also delete the typeconv functions, so nobody can accidentally use them:
https://github.com/smartcontractkit/chainlink-ccip/blob/56cfc31489a3e02c830673756c25256dd6d791a4/internal/libs/typeconv/address.go#L9-L28

commit/merkleroot/outcome.go
huangzhen1997: The chainlink-common logger doesn't have critical

commit/merkleroot/rmn/controller.go
winder: Git blame suggests that this was done intentionally.
> * convert onRampAddr from 32bytes to 20bytes in rmn reqs
 
@dimkouv how can we support generic addresses here?

execute/observation.go
winder: We need to be careful about where errors are returned or a malicious NOP could crash the system by supplying bad observations.

Skipping over bad messages like this the usual approach.

commit/merkleroot/rmn/controller.go
huangzhen1997: We need to keep this to pass RMN test. This should be addressed and removed

prashantkumar1982 (approved):

winder (approved):

PR #651, opened 2025-02-24
Fix empty rmn sigs checks

dimkouv:
`core ref: cf4fa1063b6d10015c3b8b8a1613206259408fc4`

### DIFF
--- a/rmn_offchain.proto
+++ b/rmn_offchain.proto
@@ -1,0 +1,102 @@
+syntax="proto3";
+
+package rmn_offchain;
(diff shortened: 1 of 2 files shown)


makramkd (dismissed):

0xnogo (dismissed):

rmn_offchain.proto
makramkd, line 1: Lint is failing due to this delete, i wonder if we can do this in a followup to unblock

0xnogo (approved):

makramkd (approved):

PR #653, opened 2025-02-24
CCIP-5261 Don't require source chain selectors to fetch Curse Info

mateusz-sekara:
SourceChainSelectors are not required when fetching Curse information.  We can initialize sourceChain selectors based on the response from RMNRemote without knowing the supported chain selectors. 

This is the prerequisite to cache CurseInfo in the followup properly.


pkg/reader/ccip.go
```diff
@@ -868,32 +865,26 @@ func (r *ccipChainReader) GetRmnCurseInfo(
...
 		CursedDestination: cursedSubjectsSet.Contains(GlobalCurseSubject) ||
 			cursedSubjectsSet.Contains(chainSelectorToBytes16(destChainSelector)),
 		GlobalCurse: cursedSubjectsSet.Contains(GlobalCurseSubject),
 	}
 
-	for _, ch := range sourceChainSelectors {
-		chainSelB16 := chainSelectorToBytes16(ch)
-		lggr.Debugf("checking if chain %d is cursed after casting it to 16 bytes: %v", ch, chainSelB16)
-		curseInfo.CursedSourceChains[ch] = cursedSubjectsSet.Contains(chainSelB16)
+	for _, cursedSource := range cursedSubjectsSet.ToSlice() {
```
mateusz-sekara, line 884: This is the only change. Instead of converting chainSelector to uint64, I'm doing the opposite - read RMN response and cast to chain selectors

mateusz-sekara: Interesting that none of the tests catch that :P

mateusz-sekara: Sure, renamed

PR #660, opened 2025-02-25
[Solana][NONEVM-1304] RMNRemote contract + manual cursing

PabloMansanet:
core ref: 8ffd73b7971d5cf6529363e55996dc3e1d271a82

Implements the RMNRemote contract as defined in the [CCIP RMN Cursing Analysis](https://docs.google.com/document/d/1v7z3qrKDvSOKAQuPPAiZjc_bDH4_DlXwK8pqWLxP2ss/edit?usp=sharing) document, as well as CPI calls from all interested callers + integration tests for them.

There are some minor deviations and learnings from the original doc, which I summarize here:

* The doc uses the terminology `is_cursed` to refer to the instruction that reverts when the chain is cursed, whether by subject or globally. I believe this naming is misleading as it hints at a boolean return (especially in a Rust context) so I've renamed it as `verify_uncursed`, which more accurately reflects that a revert will happen if the invariant doesn't hold.
* I hit several Solana limits as I was developing this, which I solved in a few different ways:
	* The `example_ccip_sender` contract hit the heap limit from too many sequential CPIs, since it was already making quite a few. Solved by implementing a custom allocator for that example crate in particular. We might consider extending this solution if we find similar issues elsewhere.
	* Several offramp contracts would've hit TxSize limits if the RMNRemote PDAs weren't added to the offramp lookup table, so I added them there.
	* Initializing the offramp hit stack size limits, which I solved by making the preexisting `ReferenceAddresses` struct (which now also includes the RMNRemote contract address) `zero_copy`.


chains/solana/contracts/programs/rmn-remote/src/state.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub proposed_owner: Pubkey,
+    pub default_code_version: CodeVersion,
+}
+
+#[account]
+#[derive(InitSpace, Debug)]
+pub struct Curses {
+    #[max_len(0)]
+    pub cursed_subjects: Vec<CurseSubject>,
+}
```
archseer, line 71: Why not store curses as PDAs? With the seed being the subject. If the account doesn't exist, or it's internal state is `0/false`, no curse. If it exists & state `1/true`, it's cursed

PabloMansanet: @agusaldasoro and @toblich have more context on why this was ruled out.

toblich: @archseer after discussions with Product and CCIP-EVM yesterday, we decided that we'll stick with the single PDA and the CPI.

In a little bit more detail, there are two somewhat-separated discussions:
- Single PDA for all curses vs 1 PDA per cursed subject
- CPI vs loading PDA(s) directly

## Single PDA vs 1 PDA : 1 Curse 
On the EVM side, token pools already only check for cursed chains, so if we ever add new types of subject to curse by, it will already require upgrading token pools there, which token admins have to do (not just us). However, it is transparent to the end user making the ccip-send.

To achieve something similar in Solana, given that we do want to maintain some flexibility of subject types, we have to make sure that all curses are in a single PDA. That way, regardless of how many types of curses we add, it should never require any change on the list of accounts that the end user includes in the tx.

Notably also, we never expect to have really many curses at once, so the single PDA should also not be too big. For the most part we expect there to be 0 curses at a given point in time, and only temporarily have 1 to a few curses at the same time. If any curse were to become "permanent", we would just remove that chain completely, not leave a permanent curse. I get that if we ever really add a completely unthought of type of subject this could change, but no one could really come up with a case where that could happen.

## CPI vs loading PDA directly
It is easier for us to upgrade the logic if we rely on CPIs. However, we can't guarantee what the token admin implements in the pool, so they might load the PDA directly if they want, or they could even skip RMN completely. Even on the EVM side, the pool admin might just not include the RMN check in their implementation. So, we obviously also do the check on our ramps, but other than that we just "strongly recommend" that pools make the CPI, but it's up to them to do whatever they like. 
If we figure out that we're eating away too many CUs, we could shave some off by making our ramps load the PDA directly, though that seems like an optimization that we can make later on and not something that we _need_ to implement right now.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -587,6 +604,24 @@ mod execution_state {
...
+        config: rmn_remote_config,
+        curses: rmn_remote_curses,
+    };
+    let cpi_context = CpiContext::new(rmn_remote, cpi_accounts);
+    rmn_remote::cpi::verify_not_cursed(
+        cpi_context,
+        CurseSubject::from_chain_selector(chain_selector),
+    )?;
+    Ok(())
+}
```
toblich, line 623: This seems to be repeated with the helper in the `commit.rs` file of the offramp, could you please dedupe them?

chains/solana/contracts/programs/ccip-offramp/src/lib.rs
```diff
@@ -171,6 +172,7 @@ pub mod ccip_offramp {
         router: Pubkey,
         fee_quoter: Pubkey,
         offramp_lookup_table: Pubkey,
+        rmn_remote: Pubkey,
```
toblich, line 175: Whoops, I just saw that the docstring for this method was outdated (even before this PR). Could you please update it?

chains/solana/contracts/programs/example-ccip-sender/Cargo.toml
```diff
@@ -13,10 +13,12 @@ no-entrypoint = []
...
-default = []
+default = ["custom-heap"]
+custom-heap = []
 
 [dependencies]
 solana-program = "1.17.25" # pin solana to 1.17
 anchor-lang = { version = "0.29.0", features = [] }
 anchor-spl = "0.29.0"
 fee_quoter = { version = "0.1.0-dev", path = "../fee-quoter", features = ["no-entrypoint"]}
+smalloc = "0.1.2"
```
toblich, line 24: I'm reviewing this allocator's internals. In the meantime however, have you checked if there is any allocator that is more commonly used in the community? This repo doesn't seem to have many eyes on it...
https://github.com/oxfeeefeee/smalloc

I may even prefer the more-trivial approach of https://solana.stackexchange.com/questions/7311/handling-out-of-memory-issues-in-heap personally. Any thoughts? Maybe @archseer also wants to chime in here?

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -2725,6 +2770,344 @@ func TestCCIPRouter(t *testing.T) {
...
+			result = testutils.SendAndFailWith(ctx, t, solanaGoClient, []solana.Instruction{ix}, user, config.DefaultCommitment, []string{"Error Code: GloballyCursed"})
+			require.NotNil(t, result)
+
+			ix, err = rmn_remote.NewVerifyNotCursedInstruction(
+				svmCurse,
+				config.RMNRemoteCursesPDA,
+				config.RMNRemoteConfigPDA,
+			).ValidateAndBuild()
+			require.NoError(t, err)
+			result = testutils.SendAndFailWith(ctx, t, solanaGoClient, []solana.Instruction{ix}, user, config.DefaultCommitment, []string{"Error Code: GloballyCursed"})
```
toblich, line 2845: Please add the new errors for RMN to utils/ccip/ccip_errors.go`

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -2725,6 +2770,344 @@ func TestCCIPRouter(t *testing.T) {
...
+				globalCurse,
+				config.RMNRemoteCursesPDA,
+				config.RMNRemoteConfigPDA,
+			).ValidateAndBuild()
+			require.NoError(t, err)
+			result = testutils.SendAndConfirm(ctx, t, solanaGoClient, []solana.Instruction{ix}, legacyAdmin, config.DefaultCommitment)
+			require.NotNil(t, result)
+		})
+
+		t.Run("ccip_send fails with a cursed destination", func(t *testing.T) {
```
toblich, line 3049: Could you please add 1 test where there _is_ a curse on a subject but the send is for another chain, so it still succeeds?

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -439,6 +444,9 @@ func TestCCIPRouter(t *testing.T) {
 				config.OfframpBillingSignerPDA,
 				config.FeeQuoterProgram,
 				config.FqConfigPDA,
+				config.RMNRemoteProgram,
+				config.RMNRemoteCursesPDA,
+				config.RMNRemoteCursesPDA,
```
toblich, line 449: Should one of these be the RMN Config PDA instead of the duplicating the Curses one? 😬

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -7461,11 +8196,15 @@ func TestCCIPRouter(t *testing.T) {
...
 					config.OfframpTokenPoolsSignerPDA,
+					config.RMNRemoteProgram,
+					config.RMNRemoteCursesPDA,
+					config.RMNRemoteConfigPDA,
 				)
 
 				tokenMetas, addressTables, err := tokens.ParseTokenLookupTable(ctx, solanaGoClient, token0, token0.User[receiver.PublicKey()])
 				require.NoError(t, err)
 				raw.AccountMetaSlice = append(raw.AccountMetaSlice, tokenMetas...)
+				maps.Copy(addressTables, offrampLookupTable) // commonly used ccip addresses - required otherwise tx is too large
```
toblich, line 8207: I think that maybe by fixing the lookup table (see previous comment) you may be able to remove this

chains/solana/contracts/programs/base-token-pool/src/common.rs
```diff
@@ -460,9 +487,34 @@ pub fn validate_release_or_mint(
...
+    rmn_remote: AccountInfo<'info>,
+    rmn_remote_config: AccountInfo<'info>,
+    rmn_remote_curses: AccountInfo<'info>,
+    chain_selector: u64,
+) -> Result<()> {
+    let cpi_accounts = rmn_remote::cpi::accounts::InspectCurses {
+        config: rmn_remote_config,
+        curses: rmn_remote_curses,
+    };
+    let cpi_context = CpiContext::new(rmn_remote, cpi_accounts);
```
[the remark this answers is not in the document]
PabloMansanet: Hmmm... Good question. I think I personally prefer the explicit difference, as it makes it clear that it's a permissionless call, while using `new_with_signer` could mislead a reader into thinking it's required.

chains/solana/contracts/tests/ccip/ccip_router_test.go
PabloMansanet: Ah right I forgot these don't get autogenerated 😓

chains/solana/contracts/tests/ccip/ccip_router_test.go
PabloMansanet: Do you think it could suffice with the "adding chain selector curses" test? I added that test precisely to verify that case: The SVM curse is set and the EVM one is verified and passes.

I felt like adding a successful ccip_send (which is heavy) was not worth the thin layer of logic being tested (essentially that we're passing the correct chain selector to `verify_not_cursed`), since we already tested that the CPI behaves correctly in the face of multiple curses.

chains/solana/contracts/tests/ccip/ccip_router_test.go
PabloMansanet: Totally 😅

chains/solana/contracts/programs/rmn-remote/src/state.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    pub proposed_owner: Pubkey,
+    pub default_code_version: CodeVersion,
+}
+
+#[account]
+#[derive(InitSpace, Debug)]
+pub struct Curses {
+    #[max_len(0)]
+    pub cursed_subjects: Vec<CurseSubject>,
```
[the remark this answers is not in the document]
PabloMansanet: Removed the struct :) And yeah, happy to add that test later.

chains/solana/contracts/programs/ccip-offramp/src/context.rs
```diff
@@ -389,9 +391,34 @@ pub struct CommitReportContext<'info> {
...
         bump,
-        seeds::program = reference_addresses.fee_quoter,
+        seeds::program = reference_addresses.load()?.fee_quoter,
     )]
     pub fee_quoter_config: UncheckedAccount<'info>,
+
+    ////////////////////
+    // RMN Remote CPI //
+    ////////////////////
+    /// CHECK: This is the account for the RMN Remote program
```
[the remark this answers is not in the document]
PabloMansanet: Resolved in sync.

agusaldasoro (approved):

PR #664, opened 2025-02-26
Narrow commit report query with timestamp of latest finalized report

0xnogo:
Added timestamp optimization to the CommitsRootsCache to narrow the search window for commit reports. By storing and using the latest finalized report's timestamp, we can avoid scanning the entire message visibility interval. This reduces database load and improves performance, especially for chains with many commit reports.

As a reference, the v1.5 implementation: https://github.com/smartcontractkit/chainlink/blob/develop/core/services/ocr2/plugins/ccip/internal/cache/commit_roots.go


execute/internal/cache/commit_root_cache.go
```diff
@@ -78,6 +88,9 @@ type commitRootsCache struct {
 	// messageVisibilityInterval). We keep executed roots there to make sure we don't accidentally try to reprocess
 	// already executed CommitReport
 	executedRoots *cache.Cache
+	// latestFinalizedTimestamp tracks the timestamp of the latest finalized commit root
+	// to optimize database queries by starting from this timestamp
+	latestFinalizedTimestamp time.Time
```
makramkd, line 93: Kind of not sure what to name this, `latestFinalizedFullyExecutedRoot` maybe is clear enough though it is a bit verbose. I think `latestFinalizedTimestamp` is definitely a bit ambiguous.
```suggestion
	// latestFinalizedFullyExecutedRoot is the timestamp of the latest finalized commit root that has been 
	// fully executed.
	// This is in order to optimize database queries by reducing the lookback time range.
	latestFinalizedFullyExecutedRoot time.Time
```

execute/internal/cache/commit_root_cache.go
```diff
@@ -120,3 +133,39 @@ func (r *commitRootsCache) isExecuted(key string) bool {
 	_, executed := r.executedRoots.Get(key)
 	return executed
 }
+
+func (r *commitRootsCache) GetTimestampToQueryFrom() time.Time {
+	r.cacheMu.Lock()
+	defer r.cacheMu.Unlock()
```
makramkd, line 139: This is a minor nit, but I see that `cacheMu` is a `RWLock` yet I don't see any `RLock` calls, maybe we can just simplify to a `sync.Mutex` instead.

mateusz-sekara (dismissed):

execute/internal/cache/commit_root_cache.go
0xnogo: Good point.

makramkd (approved):

execute/internal/cache/commit_root_cache.go
winder: I think we should keep business logic outside of the cache. This logic should be a simple helper function in observation.go

PR #667, opened 2025-02-26
[Solana][NONEVM-1367] Ignore unregistered token price updates

PabloMansanet:


chains/solana/contracts/programs/fee-quoter/src/instructions/v1/prices.rs
```diff
@@ -60,10 +61,16 @@ fn apply_token_price_update<'info>(
     token_update: &TokenPriceUpdate,
     token_config_account_info: &'info AccountInfo<'info>,
 ) -> Result<()> {
+    // Token is uninitialized, so we silently ignore it.
+    if token_config_account_info.owner == &system_program::ID {
+        return Ok(());
```
archseer, line 66: Nit: maybe we should emit an event so we can detect that an update was ignored?

archseer (approved):

PR #668, opened 2025-02-26
commit/tokenprice: add async tokenprice observer

makramkd:
Add an asynchronous token price `observer` implementation that follows what is being done in `merkleroot` and `chainfee`.

In the future, we'd want to potentially explore having these caches on a node-level rather than a plugin level to get even more performance gains.

The token price integration test couldn't be added because it requires docker, in a follow up will fix that and add it to this CI.


commit/chainfee/observation_async.go
makramkd, line 1: The changes here are pretty much all logging related, I wanted to add contextual logging where possible to be able to visualize rounds better.

commit/tokenprice/observation.go
```diff
@@ -54,26 +59,58 @@ func (p *processor) ObserveFChain(lggr logger.Logger) map[cciptypes.ChainSelecto
 	return fChain
 }
 
-func (p *processor) ObserveFeedTokenPrices(ctx context.Context, lggr logger.Logger) cciptypes.TokenPriceMap {
-	if p.tokenPriceReader == nil {
+type observer interface {
```
makramkd, line 62: Intentionally written to be as close as possible to the existing async observers

.github/workflows/ccip-integration-test.yml
```diff
@@ -29,6 +29,10 @@ jobs:
             name: "Messaging Test"
           - cmd: cd integration-tests/smoke/ccip && go test ccip_batching_test.go -timeout 12m -test.parallel=2 -count=1 -json
             name: "Batching Test"
+          - cmd: cd integration-tests/smoke/ccip && go test ccip_gas_price_updates_test.go -timeout 12m -test.parallel=2 -count=1 -json
+            name: "Gas Price Updates Test"
+          - cmd: cd integration-tests/smoke/ccip && go test ccip_token_price_updates_test.go -timeout 12m -test.parallel=2 -count=1 -json
+            name: "Token Price Updates Test"
```
makramkd, line 35: Added for additional coverage.

commit/tokenprice/observation.go
0xnogo: Why not creating a new file for the async observer? We would match the style used for chainfee. but this is nit

0xnogo (approved):
lgtm

makramkd:
https://github.com/smartcontractkit/chainlink/pull/16595

makramkd:
Token prices test passes in CL. Merging.

PR #672, opened 2025-02-27
CCIP-5261 Fetching curse together with configuration

mateusz-sekara:


execute/observation.go
```diff
@@ -124,11 +124,11 @@ func (p *Plugin) Observation(
 	return p.ocrTypeCodec.EncodeObservation(observation)
 }
 
-func (p *Plugin) getCurseInfo(ctx context.Context, lggr logger.Logger) (*reader.CurseInfo, error) {
+func (p *Plugin) getCurseInfo(ctx context.Context, lggr logger.Logger) (reader.CurseInfo, error) {
```
makramkd, line 127: Switching to value type just a preference / style thing or does that help somehow?

pkg/reader/ccip.go
```diff
@@ -1796,38 +1791,57 @@ func (r *ccipChainReader) processRMNProxyResults(results []types.BatchReadResult
...
+	RMNRemoteConfig,
+	CurseInfo,
+	error,
+) {
 	config := RMNRemoteConfig{}
 
-	if len(results) != 2 {
-		return RMNRemoteConfig{}, fmt.Errorf("expected 2 RMN remote results, got %d", len(results))
+	if len(results) != 3 {
+		return RMNRemoteConfig{}, CurseInfo{}, fmt.Errorf("expected 2 RMN remote results, got %d", len(results))
```
makramkd, line 1802:
```suggestion
		return RMNRemoteConfig{}, CurseInfo{}, fmt.Errorf("expected 3 RMN remote results, got %d", len(results))
```

execute/observation.go
mateusz-sekara: It's taken from the cache and value type gives us "copy" instead of relying on reference to something that could be wiped out in the meantime. It's just safer

makramkd: Ah I see, got it, makes sense.

makramkd (approved):

PR #677, opened 2025-02-27
Added back extraDataCodec interface

silaslenihan:


archseer:
Since this is just a workaround for now and the interface isn't used anywhere in this repo, can you just include it in your PR in core?

PR #678, opened 2025-02-27
Fix how fee quoter Pubkey is obtained

toblich:
The old way was reading the ID from the crate, which would result in whatever was used there during compilation. Given that we compile with different IDs per environment (devnet/mainnet), there's no guarantee that referencing the crate value will hold the right value the env. So, it should be read from the configuration instead.


PabloMansanet (approved):
Nice catch, makes sense to me 👍

PR #679, opened 2025-02-28
Remove wrong validation

dimkouv:
The dest chain selector in the 'gas prices' context is not the plugin's dest chain selector but some source chain selector, making this validation wrong.

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -116,6 +116,8 @@ type TokenPriceUpdate struct {
 }
 
 type GasPriceUpdate struct {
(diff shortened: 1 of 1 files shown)


pkg/reader/ccip.go
```diff
@@ -116,6 +116,8 @@ type TokenPriceUpdate struct {
 }
 
 type GasPriceUpdate struct {
+	// DestChainSelector is the chain that the gas price is for (some plugin source chain).
```
mateusz-sekara, line 119: 👏

PR #682, opened 2025-02-28
execute/report: validate nonce ordering in report

makramkd:
See https://smartcontract-it.atlassian.net/browse/CCIP-5369 for details.

core ref: 8ffd73b7971d5cf6529363e55996dc3e1d271a82


execute/report/report.go
```diff
@@ -349,11 +349,68 @@ func (b *execReportBuilder) checkMessage(
...
+		} else if latestNonce == 0 {
+			// this is a bug in this code, should never happen since we check if msg.Header.Nonce
+			// is zero before this check.
+			return fmt.Errorf("assumption violation: bug in the code, latestNonce is zero when it should never be")
+		} else if msg.Header.Nonce != latestNonce+1 {
+			// we've seen this sender before and the msg.Header.Nonce is not the expected value,
+			// it should be latestNonce + 1.
+			return fmt.Errorf(
+				"skipped nonce detected for sender %s: nonce in report %d != last nonce seen %d",
+				sender, msg.Header.Nonce, latestNonce)
```
makramkd, line 392: I guess this can technically also be incorrectly ordered nonce if the `Messages` array is incorrectly built, I can do an additional check and tweak this error msg.

makramkd (commented):
Still need to add a log somewhere indicating that a message was added to the batch.

execute/report/report.go
```diff
@@ -430,6 +487,10 @@ func (b *execReportBuilder) buildSingleChainReport(
 	}
 
 	// Unless there is a message limit, attempt to build a report for executing all ready messages.
+	// It is possible that the report produced here is invalid for some reason, such as
+	// report size or gas usage.
+	// In that case, we will execute the loop below to iteratively build a report
+	// with fewer messages until we find a valid report.
```
makramkd, line 493: I'm not sure why we have these as separate loops and why we have this `if` condition. It seems to me that the loop below will eventually fit everything into one report if it can. Is this an optimization @winder ?

execute/report/report.go
```diff
@@ -475,8 +536,23 @@ func (b *execReportBuilder) buildSingleChainReport(
 			finalReport = finalReport2
 			meta = meta2
 
+			b.lggr.Infow(
+				"message added to report",
```
makramkd, line 540: Added log here @mateusz-sekara please review

makramkd: Inspo https://github.com/smartcontractkit/chainlink/blob/a4b88ff2696dc2bc342547223c5b299909304f30/core/services/ocr2/plugins/ccip/ccipexec/batching.go#L406

execute/report/report.go
```diff
@@ -349,11 +349,68 @@ func (b *execReportBuilder) checkMessage(
 	return result, ReadyToExecute, nil
 }
 
+// we cannot have "skipped" nonces since these messages will be skipped onchain.
+// they will also end up in the inflight cache and "snoozed" until they
+// expire from the inflight cache TTL.
```
winder, line 354: Is there a bug in the `CheckNonces` function?

execute/report/report.go
```diff
@@ -349,11 +349,68 @@ func (b *execReportBuilder) checkMessage(
...
 // verifyReport is a final step to ensure the encoded message meets our exec criteria.
 func (b *execReportBuilder) verifyReport(
 	ctx context.Context,
 	execReport ccipocr3.ExecutePluginReportSingleChain,
 ) (bool, validationMetadata, error) {
+	err := checkSkippedNonces(b.addressCodec, execReport)
+	if err != nil {
+		b.lggr.Infow("invalid report, skipped nonce detected", "err", err)
+		return false, validationMetadata{}, nil
+	}
```
winder, line 412: This breaks the `Check` abstraction. Technically, All knowledge of `Nonces` are embedded in the `CheckNonces` function.

This verify function mainly exists to decide how many of the ready messages should be added to the final report.

If we want a redundant check, it should go into `plugin.go:validateReport`. That way we verify reports from other nodes as well.

mateusz-sekara (dismissed):

makramkd:
Summary of offline discussion: there are potentially various ways to re-use CheckNonces in the report building process, but we're punting that to a follow up since this is so high prio.

winder (approved):

PR #684, opened 2025-03-01
Optimize CommitReportsGTETimestamp queries with commit roots tracking

0xnogo:
This PR enhances the CommitsRootsCache to optimize database queries by tracking the earliest unexecuted root timestamp.

## Problem
When Root1 at 10:30am and Root3 at 10:50am are executed but Root2 at 10:40am remains unexecuted, the current implementation would use the latest executed timestamp (from Root3) for subsequent queries. This would miss Root2 entirely since it's before that timestamp.

## Solution
The cache now tracks the earliest unexecuted root timestamp and uses it to optimize query windows. It:
- Always queries from at least the message visibility window
- Uses the earliest unexecuted root's timestamp when it's after the visibility window
- Ensures roots are never missed, even when executed out of order

## Example
```
Root1 (10:30am) - executed
Root2 (10:40am) - unexecuted
Root3 (10:50am) - executed
Subsequent query: From 10:50am → misses Root2

After:
Root1 (10:30am) - executed
Root2 (10:40am) - unexecuted
Root3 (10:50am) - executed
Subsequent query: From 10:40am → finds Root2
```


execute/internal/cache/commit_root_cache_test.go
```diff
@@ -0,0 +1,681 @@
...
+					"Expected query timestamp to be after visibility window")
+			} else {
+				assert.Equal(t, tt.expectedQueryTimestamp, queryTimestamp,
+					"Query timestamp should match expected value")
+			}
+		})
+	}
+}
+
+func TestCommitRootsCache_UpdateEarliestUnexecutedRoot(t *testing.T) {
```
mateusz-sekara, line 107: Can we have a test with multiple reports with the same timestamps? (some executed some not) I gues this is a valid case in which multiple reports got included into the same block (more likely on slower chains)

execute/observation.go
```diff
@@ -167,9 +167,15 @@ func (p *Plugin) getCommitReportsObservation(
 	lggr logger.Logger,
 	observation exectypes.Observation,
 ) (exectypes.Observation, error) {
-	// TODO: set fetchFrom to the oldest pending commit report.
-	// TODO: or, cache commit reports so that we don't need to fetch them again.
-	fetchFrom := time.Now().Add(-p.offchainCfg.MessageVisibilityInterval.Duration()).UTC()
+	// Calculate the message visibility window
+	messageVisibilityWindow := time.Now().Add(-p.offchainCfg.MessageVisibilityInterval.Duration()).UTC()
```
mateusz-sekara, line 171: I guess this not needed, because commitRootsCache already knows that number

execute/internal/cache/commit_root_cache.go
```diff
@@ -39,28 +49,24 @@ func NewCommitRootsCache(
 		lggr,
 		messageVisibilityInterval,
 		rootSnoozeTime,
-		CleanupInterval,
```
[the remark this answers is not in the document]
0xnogo: Good catch, it was actually flagged by the linter so I removed them.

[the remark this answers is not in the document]
0xnogo: This was making testing more handy, hence i'll leave it as the key point of this feature is to battle test the edge cases

execute/internal/cache/commit_root_cache_test.go
0xnogo: added in `Edge Case: Multiple reports with same timestamp, some executed`

execute/internal/cache/commit_root_cache_test.go
```diff
@@ -0,0 +1,920 @@
...
+
+// Mock time provider for testing
+type mockTimeProvider struct {
+	mock.Mock
+}
+
+func (m *mockTimeProvider) Now() time.Time {
+	args := m.Called()
+	return args.Get(0).(time.Time)
+}
```
makramkd, line 45: Why not gen this in mockery.yaml?

execute/internal/cache/commit_root_cache.go
```diff
@@ -8,9 +8,22 @@ import (
...
+// TimeProvider interface to make time testable
+type TimeProvider interface {
+	Now() time.Time
+}
+
+// RealTimeProvider provides actual current time
+type RealTimeProvider struct{}
+
+func (r *RealTimeProvider) Now() time.Time {
+	return time.Now()
```
makramkd, line 24: Are we using local time or UTC time?

execute/internal/cache/commit_root_cache.go
```diff
@@ -128,3 +168,70 @@ func (r *commitRootsCache) isExecuted(key string) bool {
...
+			if earliestTimestamp.IsZero() || report.Timestamp.Before(earliestTimestamp) {
+				earliestTimestamp = report.Timestamp
+			}
+		}
+	}
+
+	// If we found an earliest timestamp, update our tracking
+	if !earliestTimestamp.IsZero() {
+		if r.earliestUnexecutedRoot.IsZero() ||
+			!earliestTimestamp.Equal(r.earliestUnexecutedRoot) {
```
makramkd, line 230: Is it possible that `earliestTimestamp` is actually earlier than `earliestUnexecutedRoot`?

execute/internal/cache/commit_root_cache.go
0xnogo: As `messageVisibilityWindow` is UTC, better to have this one too as UTC. Good catch

execute/internal/cache/commit_root_cache.go
```diff
@@ -39,28 +61,41 @@ func NewCommitRootsCache(
...
 		messageVisibilityInterval,
 		rootSnoozeTime,
-		CleanupInterval,
-		EvictionGracePeriod,
+		&RealTimeProvider{},
+	)
+}
+
+// For testing - create with custom time provider
+func NewCommitRootsCacheWithTimeProvider(
```
makramkd, line 69: Since this is for testing only, lets name it `xxxTestOnlyCommitsRootCache` which makes it private and we won't have to worry about someone outside this package using it.

execute/internal/cache/commit_root_cache.go
makramkd: I guess `!Equal(...)` can mean either two things: the timestamp is earlier or the timestamp is later. Do we want to update in both cases, or only if the timestamp is earlier than the current one we have?

execute/internal/cache/commit_root_cache.go
0xnogo: We already have an internal constructor `internalNewCommitRootsCache`. I created `NewCommitRootsCacheWithTimeProvider` explicitly to help me test in the plugin_test.go and be flexible with time.

makramkd: Makes sense, my comment is just to make it private and clear that its testing only `with the xxxTestOnly` prefix

execute/internal/cache/commit_root_cache.go
0xnogo: The only case that really matter is `earliestTimestamp.After(r.earliestUnexecutedRoot)`. After each round we are trying to move the needle to the right in order to reduce the window of research. The case where `earliestTimestamp` is `BEFORE` `earliestUnexecutedRoot` is not possible as we always track the newest earliest.

I was able to spot x main edge cases:
* root in order exec: each round we will set the newly earliest unexecuted root. So that next round we start from it.
* root ooo exec: here again as long as we consider the earliest unexecuted root, we should not have any miss
* re-org: a root that has been executed can return in an unexecuted state after a re-org. This would means that the `earliestTimestamp` could be updated to a later timestamp. For this case, we need to take the union of both the unexecuted roots and the unfinalized executed roots. If a re-org happens, the earliest root would still take into consideration the unfinalized, so we'll query until they are finalized.

This is covered by these 2:
* a processing of `UpdateEarliestUnexecutedRoot` on the union of the unexecuted roots and the unfinalized executed roots
* updating the earliestUnexecutedRoot with `earliestTimestamp.After(r.earliestUnexecutedRoot)`

execute/internal/cache/commit_root_cache.go
```diff
@@ -28,6 +41,15 @@ type CommitsRootsCache interface {
...
 	MarkAsExecuted(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32)
 	Snooze(source ccipocr3.ChainSelector, merkleRoot ccipocr3.Bytes32)
+
+	// GetTimestampToQueryFrom returns the timestamp that should be used when querying for commit reports.
+	// It will return the message visibility window or the earliest unexecuted root timestamp,
+	// whichever is later, to optimize the query window.
+	GetTimestampToQueryFrom() time.Time
+
+	// UpdateEarliestUnexecutedRoot updates the earliest unexecuted root timestamp
+	// based on the remaining unexecuted reports.
```
makramkd, line 51:
```suggestion
	// based on the remaining unexecuted and executed but not finalized reports.
```

execute/observation.go
```diff
@@ -233,12 +235,43 @@ func (p *Plugin) getCommitReportsObservation(
 		}
 	}
 
+	// Update the earliest unexecuted root based on remaining reports
+	combinedReports := p.buildCombinedReports(groupedCommits, fullyExecutedUnfinalized)
+	p.commitRootsCache.UpdateEarliestUnexecutedRoot(combinedReports)
```
makramkd, line 240: nit: can be inlined

```suggestion
	p.commitRootsCache.UpdateEarliestUnexecutedRoot(p.buildCombinedReports(groupedCommits, fullyExecutedUnfinalized))
```

execute/observation.go
```diff
@@ -233,12 +235,43 @@ func (p *Plugin) getCommitReportsObservation(
...
+
 	observation.CommitReports = groupedCommits
 
 	// TODO: truncate grouped to a maximum observation size?
 	return observation, nil
 }
 
+// buildCombinedReports creates a combined map for updating the earliest unexecuted root
+func (p *Plugin) buildCombinedReports(
```
makramkd, line 249: nit: can be a free function

```suggestion
// buildCombinedReports creates a combined map for updating the earliest unexecuted root
func buildCombinedReports(
```

execute/internal/cache/commit_root_cache.go
```diff
@@ -128,3 +168,69 @@ func (r *commitRootsCache) isExecuted(key string) bool {
...
+	for _, reports := range remainingReports {
+		for _, report := range reports {
+			if earliestTimestamp.IsZero() || report.Timestamp.Before(earliestTimestamp) {
+				earliestTimestamp = report.Timestamp
+			}
+		}
+	}
+
+	// If we found an earliest timestamp, update our tracking
+	if !earliestTimestamp.IsZero() {
```
makramkd, line 228: I feel like this simplifies the nested if with multiple conditions logic: `earliestTimestamp` _should_ get filled in the loop, so a check like:

```golang
if earliestTimestamp.IsZero() {
  return
}
```

Right after the loop makes sense, basically checks the loop logic (there should be a timestamp if len(reports) > 0).

Then we can do:

```golang
if earliestTimestamp.After(r.earliestUnexecutedRoot) {
  // if earliestUnexecutedRoot is zero, and earliestTimestamp is non-zero (checked above)
  // then r.earliestUnexecutedRoot.IsZero() is an unnecessary check, since nonzero time is always
  // greater than zero time (there is no such thing as a negative time.Time)
  r.lggr.Debugw("Updating earliest unexecuted root",
				"oldTimestamp", r.earliestUnexecutedRoot,
				"newTimestamp", earliestTimestamp)
  r.earliestUnexecutedRoot = earliestTimestamp
}
```

0xnogo: Yes sure

makramkd (approved):

PR #686, opened 2025-03-04
Caching GetSourceChainConfig

0xnogo:
Adding `GetSourceChainConfig` as part of the config poller


pkg/reader/ccip.go
```diff
@@ -1294,65 +1294,23 @@ func (r *ccipChainReader) getOffRampSourceChainsConfig(
...
-				return nil, fmt.Errorf("invalid result type from GetSourceChainConfig for chainSelector=%d: %w", chainSel, err)
-			}
-
+	// Filter out disabled chains if needed
+	if !includeDisabled {
+		for chain, cfg := range configs {
 			enabled, err := cfg.check()
-			if err != nil {
-				return nil, fmt.Errorf("source chain config check for chain %d failed: %w", chainSel, err)
+			if err != nil || !enabled {
```
matYang, line 1307: is it safe to ignore the err case here?

matYang (dismissed):

pkg/reader/ccip.go
0xnogo: I'm not sure but it's preferable to stick to the original behavior. Good catch, thanks.

pkg/reader/config_poller.go
```diff
@@ -103,6 +123,137 @@ func (c *configPoller) GetChainConfig(
...
+				chainsToFetch = append(chainsToFetch, chain)
+			} else {
+				// Use cached version
+				result[chain] = entry.config
+			}
+		}
+	}
+
+	// If all chains are in cache and fresh, return them
+	if len(chainsToFetch) == 0 {
```
matYang, line 172: wouldn't it be better to always do a non-blocking return, if cache refresh needs to happen, it happens async in the background, so this func is never blocking during observation?  E.g set a in progress flag in cache, kick off RefreshSourceChainConfigs in background,

0xnogo: We have a task to make the cache poller running in the [background](https://smartcontract-it.atlassian.net/browse/CCIP-5228). If a cache refresh needs to happen (every 30s), only a few rounds will be impacted.

pkg/reader/config_poller.go
```diff
@@ -176,8 +327,85 @@ func (c *configPoller) fetchChainConfig(
...
+			ReadName: consts.MethodNameGetSourceChainConfig,
+			Params: map[string]any{
+				"sourceChainSelector": chain,
+			},
+			ReturnVal: new(SourceChainConfig),
+		})
+	}
+
+	// Execute batch request
+	results, _, err := reader.ExtendedBatchGetLatestValues(
```
mateusz-sekara, line 361: We would be able to do a single call if `GetSourceChainConfig` accepted a list of sourceChainSelectors instead of a single sourceChainSelector, right? 

If so then could be a nice improvement to the next version of the contracts. wdyt?

pkg/reader/config_poller.go
```diff
@@ -103,6 +123,137 @@ func (c *configPoller) GetChainConfig(
...
+	// Merge the new configs with existing cached results
+	for chain, config := range newConfigs {
+		result[chain] = config
+	}
+
+	return result, nil
+}
+
+// RefreshSourceChainConfigs forces a refresh of source chain configurations
+func (c *configPoller) RefreshSourceChainConfigs(
```
mateusz-sekara, line 202: I don't see any usage of that. Is it needed for a future background worker?

pkg/reader/config_poller.go
```diff
@@ -103,6 +123,137 @@ func (c *configPoller) GetChainConfig(
...
+			"error", err,
+			"fetchConfigLatency", fetchConfigLatency)
+		return nil, fmt.Errorf("fetch source chain configs: %w", err)
+	}
+
+	// Update the cache with new configs
+	chainCache.Lock()
+
+	// Initialize the map if needed
+	if chainCache.sourceChainConfigs == nil {
```
mateusz-sekara, line 235: How about doing that in the constructor?

pkg/reader/config_poller.go
```diff
@@ -176,8 +327,85 @@ func (c *configPoller) fetchChainConfig(
...
+	validSourceChains := make([]cciptypes.ChainSelector, 0, len(sourceChains))
+
+	for _, chain := range sourceChains {
+		if chain == destChain {
+			continue
+		}
+
+		validSourceChains = append(validSourceChains, chain)
+		contractBatch = append(contractBatch, types.BatchRead{
+			ReadName: consts.MethodNameGetSourceChainConfig,
```
mateusz-sekara, line 352: Out of curiosity, why don't we call `getAllSourceChainConfigs` instead?

pkg/reader/config_poller.go
```diff
@@ -176,8 +327,85 @@ func (c *configPoller) fetchChainConfig(
...
+				validSourceChains, results)
+		}
+
+		for i, chain := range validSourceChains {
+			v, err := readResult[i].GetResult()
+			if err != nil {
+				c.lggr.Errorw("Failed to get source chain config",
+					"chain", chain,
+					"error", err)
+				return nil, fmt.Errorf("GetSourceChainConfig for chainSelector=%d failed: %w", chain, err)
```
mateusz-sekara, line 388: Is it okay to bubble up the error? Shouldn't we just log it, keep the config initialized for this particular chainSelector but keep continue working?

pkg/reader/config_poller.go
0xnogo: It's being used in the `GetSourceChainConfigs`. It's basically called when we have configs to refresh/missing.

pkg/reader/config_poller.go
0xnogo: This was done to avoid accessing the `sourceChainConfigs` map in the for-loop after it. I don't think it can be done in the constructor as we need to have the `sourceChains` to be fetched.

pkg/reader/config_poller.go
0xnogo: It's a valid point and that was the original implementation I had in mind. But after checking the current code, it seems like we are failing completely. If upstream, the system is assuming that we should return config for the exact same src chains asked, it might cause some issues.

pkg/reader/config_poller.go
0xnogo: If Solana can have the same logic it would be great add up.

pkg/reader/config_poller.go
0xnogo: Sorry i confused it with the other `if chainCache.sourceChainConfigs == nil {`

But I'd favor this lazy initialization as it's more appropriate:
1. create basic chain caches for any chain that's needed
2. oly initialize the `sourceChainConfigs` map when a chain is actually used as a destination chain

pkg/reader/config_poller.go
```diff
@@ -103,6 +123,137 @@ func (c *configPoller) GetChainConfig(
 	return c.RefreshChainConfig(ctx, chainSel)
 }
 
+// GetSourceChainConfigs retrieves cached source chain configurations
+func (c *configPoller) GetSourceChainConfigs(
```
mateusz-sekara, line 127: How are we going to run that in background if sourceChainSelectors are provided by the caller? How do sourceChainSelectors change over time?

mateusz-sekara (approved):

pkg/reader/config_poller.go
0xnogo: With the current approach, we are refreshing the cache on-demand when a caller requests information.
It's going to be challenging to implement the bg poller as we need to know which chain selectors to poll (vs dynamically provided).

Few options at the top of my mind:
* track requested chains and poll them in bg: new chains selectors would still cause an initial blocking refresh but the others will be part of the bg refresh
* Find a way to discover all the src chains: not sure how feasible it is but that would avoid the dependency on the input selectors
* some sort of lazy initialization with background refresh:
 - Keep the initial on-demand refresh for new chain selectors
 - Once a chain selector has been requested, add it to a background refresh list
 - A background task periodically refreshes items in this list

PR #687, opened 2025-03-04
add new function to get consensus on sequence numbers

0xAustinWang:
Updated to use the same approach that we had in 1.5
https://github.com/smartcontractkit/chainlink/blob/5156cfe5ce0f8e060255b39e6b6ce2160da4a3b5/core/services/ocr2/plugins/ccip/ccipcommit/ocr2.go#L329

====
https://chainlink-core.slack.com/archives/C06S76UEBK8/p1739432015202169

SequenceNumbers consensus algorithm
What we currently do for [onRamp seqNums](https://github.com/smartcontractkit/chainlink-ccip/blob/ab5458b5c6a8241f00b2803f3acdc30a2c67ffa5/commit/merkleroot/outcome.go#L368)
```
consensus.GetConsensusMap(lggr, "OnRamp Max Seq Nums", aggObs.OnRampMaxSeqNums, twoFChainPlus1)
// The consensus item for a given chain is the item with the
// most observations that was observed at least fChain times.
```
But imagine that we have a constant stream of seqNums, then it's very unlikely that at least 2f+1 oracles are seeing the same seqNum. e.g.:
```
Oracle1: 13
Oracle2: 15
Oracle3: 16
Oracle4: 15
```

### DIFF
--- a/internal/plugincommon/consensus/consensus.go
+++ b/internal/plugincommon/consensus/consensus.go
@@ -102,7 +102,6 @@ func Median[T any](vals []T, less func(T, T) bool) T {
 // [honest 1, honest 1, honest 2, malicious 2], in this case we pick 2, but it's not enough to be able
 // to build a report since the first 2 honest nodes are unaware of message 2.
 // In an observation of [1, 2, 3, 4], we would pick 1, which is the most conservative choice.
-// https://github.com/smartcontractkit/chainlink/blob/5156cfe5ce0f8e060255b39e6b6ce2160da4a3b5/core/services/ocr2/plugins/ccip/ccipcommit/ocr2.go#L329
 func GetOrderedConsensus[K comparable, T cmp.Ordered](
 	lggr logger.Logger,
 	objectName string,
--- a/internal/plugincommon/consensus/consensus.go
+++ b/internal/plugincommon/consensus/consensus.go
@@ -1,6 +1,7 @@
 package consensus
 
 import (
+	"cmp"
 	"sort"
 	"time"
 
@@ -31,16 +32,17 @@ func GetConsensusMap[K comparable, T any](
 			items = minObservations.GetValid()
 			if len(items) != 1 {
 				// TODO: metrics
-				lggr.Debugf("failed to reach consensus on a %s's for key %+v "+
-					"because no single item was observed more than the expected min (%d) times, "+
-					"all observed items: %v",
-					objectName, key, minThresh, items)
+				lggr.Debugw("could not reach consensus due to not enough observations meeting the minimum threshold",
+					"objectName", objectName,
+					"key", key,
+					"minThreshold", minThresh,
+					"items", items)
 			} else {
 				consensus[key] = items[0]
 			}
 		} else {
 			// TODO: metrics
-			lggr.Warnf("getConsensus(%s): min not found for chain %d", objectName, key)
+			lggr.Warnw("min threshold not found defined", "objectName", objectName, "key", key)
 		}
 	}
 	return consensus
@@ -60,7 +62,9 @@ func GetConsensusMapAggregator[K comparable, T any](
 
 	for key, values := range items {
 		if thresh, ok := f.Get(key); !ok || len(values) < int(thresh) {
-			lggr.Debugf("could not reach consensus on %s for key %v", objectName, key)
+			lggr.Debugw("could not reach consensus in consensusMapAggregator",
+				"objectName", objectName,
+				"key", key)
 			continue
 		}
 		consensus[key] = agg(values)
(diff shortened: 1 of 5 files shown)


internal/plugincommon/consensus/consensus_test.go
```diff
@@ -57,6 +57,91 @@ func Test_fChainConsensus(t *testing.T) {
 	}
 }
 
+func Test_SeqNumConsensus(t *testing.T) {
```
mateusz-sekara, line 60: Can we have a couple more edge cases? for instance
* F=1
* F=len(slice)

mateusz-sekara (approved):
Left some minor comment, aside from that looking good

internal/plugincommon/consensus/consensus.go
```diff
@@ -83,6 +83,37 @@ func Median[T any](vals []T, less func(T, T) bool) T {
 	return valsCopy[len(valsCopy)/2]
 }
 
+// GetMaxSqNrWithConsensus returns the greatest sequence number for each chain that meets the min obs threshold
+// We do this by sorting the array in ascending order, then returning the value of item[minObs]
+func GetMaxSqNrWithConsensus[K comparable](
```
makramkd, line 88: This isn't seq nr specific right? Can work for any `V comparable`, since you just need to `sort.Slice` something?

internal/plugincommon/consensus/consensus.go
```diff
@@ -83,6 +83,37 @@ func Median[T any](vals []T, less func(T, T) bool) T {
...
+				continue
+			}
+
+			sort.Slice(items, func(i, j int) bool {
+				return items[i] > items[j]
+			})
+
+			result[key] = items[int(minThresh)-1]
+		} else {
+			lggr.Warnf("getConsensus(%s): min not found for chain %d", objectName, key)
```
makramkd, line 111: Please use structured logging

internal/plugincommon/consensus/consensus_test.go
```diff
@@ -57,6 +57,91 @@ func Test_fChainConsensus(t *testing.T) {
...
+				cciptypes.ChainSelector(1): 2,
+				cciptypes.ChainSelector(2): 2,
+				cciptypes.ChainSelector(4): 3,
+			},
+		},
+	}
+
+	for _, scenario := range testCases {
+		t.Run(scenario.name, func(t *testing.T) {
+			minObs := MakeConstantThreshold[cciptypes.ChainSelector](Threshold(scenario.f))
```
makramkd, line 138: Lets make this injectable into the test case. We should unit test this function in the ways it is expected to be called, which right now is 2FChain + 1

internal/plugincommon/consensus/consensus_test.go
```diff
@@ -57,6 +57,91 @@ func Test_fChainConsensus(t *testing.T) {
...
+			result := GetMaxSqNrWithConsensus(lggr, "fChain", scenario.inputMap, minObs)
+			assert.Equal(t, scenario.expectedOutput, result)
```
makramkd, line 140: Prefer `require` over `assert` for easier test debugging

internal/plugincommon/consensus/consensus.go
```diff
@@ -83,6 +83,37 @@ func Median[T any](vals []T, less func(T, T) bool) T {
...
+	minObs MultiThreshold[K]) map[K]cciptypes.SeqNum {
+	result := make(map[K]cciptypes.SeqNum)
+
+	for key, items := range itemsByKey {
+		if minThresh, exists := minObs.Get(key); exists {
+			if len(items) < int(minThresh) {
+				lggr.Debugf("failed to reach consensus on a %s's for key %+v "+
+					"because no single item was observed more than the expected min (%d) times, "+
+					"all observed items: %v",
+					objectName, key, minThresh, items)
```
makramkd, line 101: Structured log

internal/plugincommon/consensus/consensus.go
makramkd: Also lets make the message a bit clearer: `min threshold not found for key %v`, don't forget that `K` is generic, so `%d` is not necessarily correct.

makramkd (commented):
Good start,

makramkd:
> I think you need to change the consensus to require at least f+1 (f is the max allowed num of faults) occurences of a seq num and adjust the condition for no consensus.

We can do this but only after ensuring there are at least 2Fchain + 1 total observations for that chain.

internal/plugincommon/consensus/consensus_test.go
```diff
@@ -57,6 +57,91 @@ func Test_fChainConsensus(t *testing.T) {
...
+				cciptypes.ChainSelector(1): 1,
+			},
+		},
+		{
+			name: "single chain with minimal seqNum",
+			f:    3,
+			inputMap: map[cciptypes.ChainSelector][]cciptypes.SeqNum{
+				cciptypes.ChainSelector(1): {1, 1, 1, 1, 1, 1, 2, 3, 3},
+			},
+			expectedOutput: map[cciptypes.ChainSelector]cciptypes.SeqNum{
```
asoliman92, line 85: So with only one observer observing 2 we consider it correct? Don't think that should be the case. A malicious node can break things like that.

internal/plugincommon/consensus/consensus_test.go
0xAustinWang: I updated the test case to accept `f` rather than 2f+1. This is more similar to the way we were doing it in 1.5

internal/plugincommon/consensus/consensus_test.go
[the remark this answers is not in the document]
0xAustinWang: discussed offline, will align with what 1.5 is doing

0xAustinWang:
@makramkd 
> We can do this but only after ensuring there are at least 2Fchain + 1 total observations for that chain.



I think we do that here: 
https://github.com/smartcontractkit/chainlink-ccip/blob/70aa154fb469d98cfd6f4c807c4ca036fae329e5/commit/merkleroot/observation.go#L41

internal/plugincommon/consensus/consensus_test.go
0xAustinWang: Updated the test cases, but in this scenario, the nodes that observed 3 have also observed 2. This means that 3 observers have seen seqnum 2

makramkd:
> I think we do that here, but i have it in the function as well

That's the Role DON quorum, which is not the same as 2FChain + 1 from the subcommittee.

internal/plugincommon/consensus/consensus.go
```diff
@@ -83,6 +87,58 @@ func Median[T any](vals []T, less func(T, T) bool) T {
...
+// cause the maxSeqNum < minSeqNum errors
+// - If an honest oracle reports sorted_max[thresholds] which happens to be stale i.e. that oracle
+// has a delayed view of the source chain, then we simply lose a little bit of throughput.
+// - If we were to pick sorted_max[-thresholds]
+// i.e. the maximum honest node view (a more "aggressive" setting in terms of throughput),
+// then an adversary can continually send high values e.g. imagine we have observations from all 4 nodes
+// [honest 1, honest 1, honest 2, malicious 2], in this case we pick 2, but it's not enough to be able
+// to build a report since the first 2 honest nodes are unaware of message 2.
+// In an observation of [1, 2, 3, 4], we would pick 1, which is the most conservative choice.
+func GetOrderedConsensus[K comparable, T cmp.Ordered](
```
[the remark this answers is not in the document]
0xAustinWang: I agree 😆 I also didn't want it to be too long

PR #695, opened 2025-03-07
execute: checkAlreadyExecuted checks if all are exec'd

makramkd:
Previous behavior of this function was that it would prevent a report
from being transmitted if it detected that a single message in the
report was executed. The logic behind this was to prevent re-executions
that would just waste gas onchain.

However, an edge case has been uncovered that involves out of order
transmissions. Since the transmission protocol is async and could take
longer than a round duration, we could have transmissions that get
fumbled and only some of the messages executed while others are skipped
due to incorrect nonces.

In order to address this, checkAlreadyExecuted will only return an error
and prevent report acceptance / transmission if _all_ the messages in
the report are already executed, making it a completely redundant
transmission. However, if some have not yet been executed, it will try
again and the hope is that eventually we'd get rid of the nonce errors
in the transmission schedule.

### DIFF
--- a/execute/observation.go
+++ b/execute/observation.go
@@ -17,7 +17,6 @@ import (
 	dt "github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery/discoverytypes"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/logutil"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/reader"
-	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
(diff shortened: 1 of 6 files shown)


execute/observation.go
```diff
@@ -17,7 +17,6 @@ import (
 	dt "github.com/smartcontractkit/chainlink-ccip/internal/plugincommon/discovery/discoverytypes"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/logutil"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/reader"
-	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
```
makramkd, line 20: fixed double import

PR #696, opened 2025-03-07
CCIP-5109 Refactor tokendata

bukata-sa:
core ref: 6a481920beb027bdb083f182ac22df66747a3dc7
PR: https://github.com/smartcontractkit/chainlink/pull/16453/commits

This PR proposes tokendata refactoring:
* Split config parts into different structs, embed them and provide custom un/marshalling functions - existing configs are still compatible with these changes
* Moving observer to separated package for dependency management
* Moving http client to separated package
* Moving common parts to token_data.go file in the root of tokendata package


execute/plugin_tokendata_test.go
```diff
@@ -87,31 +100,32 @@ func Test_USDC_Transfer(t *testing.T) {
 	require.Len(t, outcome.CommitReports, 1)
 
 	// Round 3 - Filter
-	// Messages 102-104 are executed, 105 doesn't have token data ready
+	// Messages 102-104,106,108 are executed, 105 and 107 don't have token data ready
 	outcome = runRoundAndGetOutcome(ctx, ocrTypeCodec, t, runner)
 	require.NoError(t, err)
-	require.Len(t, outcome.Report.ChainReports, 1)
+	assert.Len(t, outcome.Report.ChainReports, 1)
```
mateusz-sekara, line 106: Why these changes? `require` will return immeidately but `assert` will continue test execution. I think when number of reports and messages don't match there is no point in continuing

execute/plugin_tokendata_test.go
```diff
@@ -87,31 +100,32 @@ func Test_USDC_Transfer(t *testing.T) {
 	require.Len(t, outcome.CommitReports, 1)
 
 	// Round 3 - Filter
-	// Messages 102-104 are executed, 105 doesn't have token data ready
+	// Messages 102-104,106,108 are executed, 105 and 107 don't have token data ready
```
mateusz-sekara, line 103: 108? I don't see that assertion in tests, seems like comment is wrong, isn't it?

execute/test_utils.go
```diff
@@ -349,13 +353,43 @@ func newConfigurableAttestationServer(responses map[string]string) *Configurable
 	}
 
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		for url, response := range c.responses {
-			if strings.Contains(r.RequestURI, url) {
-				_, err := w.Write([]byte(response))
-				if err != nil {
-					panic(err)
+		if r.Method == http.MethodGet {
```
mateusz-sekara, line 356: Wouldn't it be better to have httpServer per provider? I mean separate one for USDC and separate one for LBTC?

execute/test_utils.go
```diff
@@ -349,13 +353,43 @@ func newConfigurableAttestationServer(responses map[string]string) *Configurable
...
+					_, err := w.Write([]byte(response))
+					if err != nil {
+						panic(err)
+					}
+					return
 				}
-				return
+			}
+		}
+		if r.Method == http.MethodPost {
```
mateusz-sekara, line 367: Hmm, I don't see any post http calls in the business logic

execute/tokendata/http/http_test.go
```diff
@@ -106,76 +103,9 @@ func Test_HTTPClient_Get(t *testing.T) {
 				}))
 			},
 			timeout:            50 * time.Millisecond,
-			expectedError:      ErrTimeout,
+			expectedError:      tokendata.ErrTimeout,
 			expectedStatusCode: http.StatusRequestTimeout,
 		},
-		{
-			name: "200 but attestation response contains error",
```
mateusz-sekara, line 113: Why are these tests removed?

execute/tokendata/observer/observer.go
```diff
@@ -65,13 +64,14 @@ func NewConfigBasedCompositeObservers(
...
-				encoder, readers, addrCodec)
+			observer, err := usdc.NewUSDCTokenDataObserver(ctx, lggr, destChainSelector,
+				*c.USDCCCTPObserverConfig,
+				encoder.EncodeUSDC, readers, addrCodec)
 			if err != nil {
 				return nil, fmt.Errorf("create USDC/CCTP token observer: %w", err)
 			}
 
-			if c.USDCCCTPObserverConfig.NumWorkers == 0 {
+			if c.USDCCCTPObserverConfig.IsForeground() {
```
mateusz-sekara, line 74: Is it even used anywhere? Do we still need it?

pluginconfig/token_test.go
```diff
@@ -1,21 +1,24 @@
...
 
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
-func Test_TokenDataObserver_Unmarshall(t *testing.T) {
+func Test_TokenDataObserver_Unmarshal(t *testing.T) {
 	baseJSON := `{
 					  %s
 					  "batchGasLimit": 1,
+					  "relativeBoostPerWaitHour": 1,
```
mateusz-sekara, line 21: That property was removed afaik. Why did you bring it back?

pluginconfig/token_test.go
```diff
@@ -84,19 +94,24 @@ func Test_TokenDataObserver_Unmarshall(t *testing.T) {
...
 					USDCCCTPObserverConfig: &USDCCCTPObserverConfig{
+						AttestationConfig: AttestationConfig{
+							AttestationAPI:         "http://localhost:8080",
+							AttestationAPITimeout:  commonconfig.MustNewDuration(time.Second),
+							AttestationAPIInterval: commonconfig.MustNewDuration(500 * time.Millisecond),
+							AttestationAPICooldown: commonconfig.MustNewDuration(10 * time.Minute),
+						},
+						WorkerConfig: WorkerConfig{
+							NumWorkers:              10,
+							CacheExpirationInterval: commonconfig.MustNewDuration(5 * time.Second),
```
mateusz-sekara, line 105: Do we really need that level of configuration?

pluginconfig/token_test.go
```diff
@@ -84,19 +94,24 @@ func Test_TokenDataObserver_Unmarshall(t *testing.T) {
 					Type:    "usdc-cctp",
 					Version: "1.0",
 					USDCCCTPObserverConfig: &USDCCCTPObserverConfig{
+						AttestationConfig: AttestationConfig{
+							AttestationAPI:         "http://localhost:8080",
+							AttestationAPITimeout:  commonconfig.MustNewDuration(time.Second),
+							AttestationAPIInterval: commonconfig.MustNewDuration(500 * time.Millisecond),
+							AttestationAPICooldown: commonconfig.MustNewDuration(10 * time.Minute),
```
mateusz-sekara, line 101: Do we have a cooldown for Lombard API?

execute/tokendata/http/http.go
```diff
@@ -16,18 +15,13 @@ import (
...
+
 	"github.com/smartcontractkit/chainlink-ccip/pkg/logutil"
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
 const (
-	apiVersion      = "v1"
-	attestationPath = "attestations"
-
-	// defaultCoolDownDurationSec defines the default time to wait after getting rate limited.
```
mateusz-sekara, line 27: Could we preserve this comment somewhere close to the cooldown values? I think it was very valuable to better understand why this value is needed

execute/tokendata/token_data.go
```diff
@@ -0,0 +1,181 @@
...
+				float64(60 * time.Minute),
+			},
+		},
+		[]string{"token"},
+	)
+)
+
+type AttestationStatus struct {
+	// ID is the identifier of the message that the attestation was fetched for
+	ID cciptypes.Bytes
```
mateusz-sekara, line 71: If this is the same as `MessageID`, let's make that consistent with other usages and rename to `MessageID`

```go
type RampMessageHeader struct {
	// MessageID is a unique identifier for the message, it should be unique across all chains.
	// It is generated on the chain that the CCIP send is requested (i.e. the source chain of a message).
	MessageID Bytes32 `json:"messageId"`
        /// ...
```

execute/tokendata/usdc/attestation.go
```diff
@@ -54,48 +66,49 @@ type AttestationEncoder func(context.Context, cciptypes.Bytes, cciptypes.Bytes)
...
 //	  2 -> AttestationStatus{Error: ErrNotRead, Attestation: nil, MessageHash: nil}
-type AttestationClient interface {
-	Attestations(
-		ctx context.Context,
-		msgs map[cciptypes.ChainSelector]map[reader.MessageTokenID]cciptypes.Bytes,
-	) (map[cciptypes.ChainSelector]map[reader.MessageTokenID]AttestationStatus, error)
-}
-
-type sequentialAttestationClient struct {
+//type AttestationClient interface {
```
mateusz-sekara, line 69: Why comment out this interface? If not needed please remove, don't leave commented out code unless there is a good reason to do so

mateusz-sekara: This interface was introduced to allow smarter processing for Attestations APIs which allow batching. I would expect that Lombard implementation would reimplement that and get all attestations within a single call, no?

mateusz-sekara: Ahh, it's moved to a different file. Please remove it from here and make sure godoc is also moved with the code

pluginconfig/token.go
```diff
@@ -63,113 +68,189 @@ func (t TokenDataObserverConfig) WellFormed() error {
...
-func (t TokenDataObserverConfig) IsUSDC() bool {
+func (t *TokenDataObserverConfig) IsUSDC() bool {
 	return t.Type == USDCCCTPHandlerType
 }
 
-const USDCCCTPHandlerType = "usdc-cctp"
+// MarshalJSON is a custom JSON marshaller for TokenDataObserverConfig.
+// It constructs raw map based on provided type. Custom marshaller is needed because default golang marshaller
+// doesn't marshal clashing fields of pointer embeddings even if only one pointer is present and rest are set to nil
+func (t *TokenDataObserverConfig) MarshalJSON() ([]byte, error) {
```
mateusz-sekara, line 88: Why do we need to MarshallJSON? Config is only read from JSON, never persisted, no?

pluginconfig/token.go
```diff
@@ -63,113 +68,189 @@ func (t TokenDataObserverConfig) WellFormed() error {
...
 
-type USDCCCTPObserverConfig struct {
-	Tokens         map[cciptypes.ChainSelector]USDCCCTPTokenConfig `json:"tokens"`
-	AttestationAPI string                                          `json:"attestationAPI"`
+// UnmarshalJSON is a custom JSON unmarshaller for TokenDataObserverConfig.
+// It first reads top-level fields, then allocates the correct embedded config pointer.
+// (only USDCCCTPObserverConfig for now) before finally unmarshalling into that pointer.
+// Custom unmarshaller is needed because default golang marshaller doesn't unmarshal clashing fields
+// (when they appear beside USDC) of pointer embeddings
+func (t *TokenDataObserverConfig) UnmarshalJSON(data []byte) error {
```
mateusz-sekara, line 110: What did exactly stop it from working? We intentionally kept everything under a single type to make deserialization simpler https://github.com/smartcontractkit/chainlink-ccip/pull/117

It was inspired by the TX types in Algorand, @winder should be able to point to the origin in their codebase

execute/tokendata/token_data.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+var (
+	ErrDataMissing     = errors.New("token data missing")
+	ErrNotReady        = errors.New("token data not ready")
+	ErrRateLimit       = errors.New("token data API is being rate limited")
+	ErrTimeout         = errors.New("token data API timed out")
+	ErrUnknownResponse = errors.New("unexpected response from attestation API")
+
+	promAttestationDurations = promauto.NewHistogramVec(
```
mateusz-sekara, line 33: Could you let me know why you moved them here? To be honest, I think the previous approach had better isolation. `observability` stuff was moved to a separate file and therefore not pollute common code. These `cacheExpirations` and `prom` objects add unnecessary cognitive load here

mateusz-sekara (changes requested):

mateusz-sekara (changes requested):
Please make sure that all integration tests are passing first. Also changes to the offchain configuration (jsons) might break existing changes, so please double check we don't introduce any regression on any env with that

execute/tokendata/http/http_test.go
bukata-sa: it is no longer part of http test, it is part of usdc attestation test. Updated it there to be more similiar to this

execute/tokendata/token_data.go
bukata-sa: No, it is messageID in terms of circle attestation api, not ours

execute/tokendata/observer/observer.go
bukata-sa: I don't know, it can be part of config. I also think that we don't need it, so dropped

execute/test_utils.go
bukata-sa: can do this in next PR when introduce LBTC

pluginconfig/token_test.go
mateusz-sekara: I mean, if you plan to use AttestationConfig across different token providers (for instance Lombard) and Lombard doesn't need cooldown then what's the point of having that field in this generic struct? If this AttestationConfig is meant to be only for USDC then it should be fine.

I guess we can refine it in the PR in which we add lbtc support

pluginconfig/token.go
winder: It's an interesting point. I guess the [code here](https://github.com/algorand/go-algorand/blob/55011f93fddb181c643f8e3f3d3391b62832e7cd/data/transactions/transaction.go#L87) works because embedded types are totally distinct from one another.

If all providers require different configs and they sometimes collide then you need something like these marshal overrides. Alternatively you can prefix the colliding types, or name them rather than embedding.

winder: Since we've already deployed these configs, our options are limited. This might be the best way to deal with the colliding "tokens" type.

pluginconfig/token.go
winder: This is a public type. Hopefully the tooling uses it.

mateusz-sekara (approved):

PR #697, opened 2025-03-09
add debug logs

0xAustinWang:


pkg/reader/ccip.go
```diff
@@ -585,6 +586,7 @@ func (r *ccipChainReader) GetExpectedNextSequenceNumber(
 			sourceChainSelector, r.destChain)
 	}
 
+	r.lggr.Debugw("chain reader returning expected next sequence number", "seqNum", expectedNextSequenceNumber)
```
makramkd, line 589: Please create a logger like the other methods,

```golang
```

So that we have the ocrSeqNr and ocrPhase in the log fields.

pkg/reader/ccip.go
```diff
@@ -552,6 +552,7 @@ func (r *ccipChainReader) LatestMsgSeqNum(
 		return 0, fmt.Errorf("message invalid msg %v: %w", msg, err)
 	}
 
+	lggr.Debugw("chain reader returning latest message sequence number", "seqNum", msg.Message.Header.SequenceNumber)
```
makramkd, line 555: There's a `queried latest message from source` log above that is `Infow` but probably should be `Debugw`. Maybe this can be `Infow`?

makramkd (changes requested):

pkg/reader/ccip.go
```diff
@@ -552,6 +552,7 @@ func (r *ccipChainReader) LatestMsgSeqNum(
 		return 0, fmt.Errorf("message invalid msg %v: %w", msg, err)
 	}
 
+	lggr.Infow("chain reader returning latest message sequence number", "seqNum", msg.Message.Header.SequenceNumber)
```
makramkd, line 555: Ah one more thing, add source chain selector to this log as a field?

pkg/reader/ccip.go
```diff
@@ -585,6 +588,7 @@ func (r *ccipChainReader) GetExpectedNextSequenceNumber(
 			sourceChainSelector, r.destChain)
 	}
 
+	lggr.Debugw("chain reader returning expected next sequence number", "seqNum", expectedNextSequenceNumber)
```
makramkd, line 591: Do we need to add source chain selector as a field?

makramkd (approved):

PR #698, opened 2025-03-10
execute/report: remove unnecessary len check

makramkd:
https://smartcontract-it.atlassian.net/browse/CCIP-4708

The `readyMessages` map always has length > 0, unless `report.Messages` is empty (which is an invariant violation)


makramkd:
Gonna keep this len check to be more defensive, in case `report.Messages` is empty

PR #699, opened 2025-03-10
Make encoders thread safe

asoliman92:


pkg/ocrtypecodec/v1/exec.go
```diff
@@ -24,16 +25,23 @@ type ExecCodec interface {
 }
 
 type ExecCodecProto struct {
-	tr *protoTranslator
+	encodeObsMu     sync.Mutex
+	decodeObsMu     sync.Mutex
+	encodeOutcomeMu sync.Mutex
+	decodeOutcomeMu sync.Mutex
+	tr              *protoTranslator
```
makramkd, line 32: Not sure I see the necessity of this, the encoder has no state access at all.

makramkd (changes requested):

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -50,8 +51,8 @@ func (e *ExecCodecProto) EncodeObservation(observation exectypes.Observation) ([
 		},
 		FChain: e.tr.fChainToProto(observation.FChain),
 	}
-
-	return proto.Marshal(pbObs)
+	encoded, err := proto.Marshal(pbObs)
+	return encoded, err
```
makramkd, line 55: Can you revert this change?

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -75,7 +75,8 @@ func (e *ExecCodecProto) DecodeObservation(data []byte) (exectypes.Observation,
 			Addresses: e.tr.discoveryAddressesFromProto(pbObs.Contracts.ContractNames.Addresses),
 		},
 		FChain: e.tr.fChainFromProto(pbObs.FChain),
-	}, nil
+	}
+	return decoded, nil
```
makramkd, line 79: Ditto here

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -88,8 +89,9 @@ func (e *ExecCodecProto) EncodeOutcome(outcome exectypes.Outcome) ([]byte, error
 			ChainReports: e.tr.chainReportsToProto(outcome.Report.ChainReports),
 		},
 	}
+	encoded, err := proto.MarshalOptions{Deterministic: true}.Marshal(pbObs)
 
-	return proto.MarshalOptions{Deterministic: true}.Marshal(pbObs)
+	return encoded, err
```
makramkd, line 94: Ditto here

execute/optimizers/type_optimizer.go
```diff
@@ -97,7 +97,7 @@ func (op ObservationOptimizer) TruncateObservation(observation exectypes.Observa
 					op.lggr.Errorw("missing message", "seqNum", seqNum, "chain", chain)
 					continue
 				}
-				op.lggr.Debugw("truncating message", "seqNum", seqNum, "chain", chain)
+				//op.lggr.Debugw("truncating message", "seqNum", seqNum, "chain", chain)
```
makramkd, line 100: So are we sure the logger is the cause? Removing these logs is problematic, especially if we suspect issues in the truncation algo

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -109,42 +111,56 @@ func (e *ExecCodecProto) DecodeOutcome(data []byte) (exectypes.Outcome, error) {
...
 }
 
-func (*ExecCodecJSON) DecodeObservation(data []byte) (exectypes.Observation, error) {
+func (e *ExecCodecJSON) DecodeObservation(data []byte) (exectypes.Observation, error) {
 	if len(data) == 0 {
 		return exectypes.Observation{}, nil
 	}
 	obs := exectypes.Observation{}
+	e.mu.Lock()
```
mateusz-sekara, line 143: json.Unmarshall/Marshall must be thread-safe. These locks should not be required

asoliman92: I tried cloning the full object to make sure that we're not trying to encode something that's changing and still got unsafe pointer issues for the encoder.

pkg/ocrtypecodec/v1/exec.go
asoliman92: I left it intentionally in case we want to add smth there later on. Any problems against leaving it as is?

makramkd:
This still relevant @asoliman92 ?

PR #704, opened 2025-03-11
add verified build action

tt-cll:
Add verified build action to be triggered based on a git SHA.

The artifacts will be uploaded as a release to the repo named based on the env and the input SHA

In the future we hope to expand this functionality to also allow for different envs/clusters, by allowing fetching the keys.

Example generated artifact [here](https://github.com/smartcontractkit/chainlink-ccip/releases/tag/solana-artifacts-localtest-b0785c190f700710bcf5bbb0ce6b9b03e86e67ec)


yashnevatia (approved):

.github/workflows/solana-verified-build.yml
```diff
@@ -0,0 +1,46 @@
...
+        tar cfvz artifacts.tar.gz target/deploy/*.so target/idl/*
+    - name: Publish Release
+      uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2.2.1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        tag_name: solana-artifacts-localtest-${{ github.event.inputs.sha }}
+        target_commitish: ${{ github.event.inputs.sha }}
+        files: |
+          chains/solana/contracts/artifacts.tar.gz
+          chains/solana/contracts/target/deploy/*.so
```
tt-cll, line 46: do we need this line? The `.so` files should be included in the tar above, no?

jlaveracll: correct, they do!

I saw in a different project here that we were doing it this way so the plan is to double check with Aaron

jlaveracll: Here: https://github.com/smartcontractkit/chainlink-solana/blob/ab8c7a17937ec4499db0b9e28cd11b3ef8aa0cd6/.github/workflows/contract-release.yml#L39

I don't understand the reason for it, so I wanted to verify that first

jlaveracll: @aalu1418 any recommendation here?

agusaldasoro (approved):

winder (approved):

PR #706, opened 2025-03-12
Sort Exec Outcome by timestamp

asoliman92:


execute/outcome.go
```diff
@@ -103,7 +103,7 @@ func (p *Plugin) Outcome(
 		"outcome", outcome,
 		"numCommitReports", len(outcome.CommitReports),
 		"numChainReports", len(outcome.Report.ChainReports),
-		"numMessages", len(observation.Messages),
+		"numMessages", len(observation.Messages.Flatten()),
```
asoliman92, line 106: same

execute/exectypes/outcome.go
```diff
@@ -104,10 +104,7 @@ func NewSortedOutcome(
 	sort.Slice(
 		pendingCommitsCP,
 		func(i, j int) bool {
-			if pendingCommitsCP[i].SourceChain != pendingCommitsCP[j].SourceChain {
-				return pendingCommitsCP[i].SourceChain < pendingCommitsCP[j].SourceChain
-			}
-			return pendingCommitsCP[i].SequenceNumberRange.Start() < pendingCommitsCP[j].SequenceNumberRange.Start()
+			return pendingCommitsCP[i].Timestamp.Before(pendingCommitsCP[j].Timestamp)
```
makramkd, line 107: This is fine for now, thinking how can we make this much clearer and less cluttery in the future (and also less prone to bugs). The `Test` that tests the timestamp ordering is a good start.

execute/observation.go
```diff
@@ -118,7 +118,7 @@ func (p *Plugin) Observation(
 		"duration", time.Since(tStart),
 		"state", state,
 		"numCommitReports", len(observation.CommitReports),
-		"numMessages", len(observation.Messages))
+		"numMessages", len(observation.Messages.Flatten()))
```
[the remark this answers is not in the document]
makramkd: Thats a good point, though this is a pretty heavy operation (allocates a full slice of `Message` and does `append` all the way till the end) just to get a number. I think having that be

```golang
// Count returns the number of messages in mo, across all chain selectors.
func (mo MessageObservations) Count() (count int) {
  for _, msgs := range mo {
    for _, msg := range msgs {
      count++
    }
  }
  return
}
```

Is probably much faster, but who's counting 😄

execute/plugin_e2e_test.go
```diff
@@ -188,3 +188,79 @@ func TestPlugin_FinalizedUnfinalizedCaching(t *testing.T) {
...
+				makeMsgWithMetadata(100, srcChainA, dstSelector, false),
+				makeMsgWithMetadata(101, srcChainA, dstSelector, false),
+			},
+			timestamp: baseTime.Add(2 * time.Hour),
+			blockNum:  1002,
+		},
+		{
+			messages: []inmem.MessagesWithMetadata{
+				makeMsgWithMetadata(102, srcChainA, dstSelector, false),
+				makeMsgWithMetadata(103, srcChainA, dstSelector, false),
```
makramkd, line 217: This is a nit but the contents of this test data doesn't match with how CCIP works - older reports should have smaller sequence numbers.

Maybe:

```suggestion
				makeMsgWithMetadata(102, srcChainA, dstSelector, false),
				makeMsgWithMetadata(103, srcChainA, dstSelector, false),
			},
			timestamp: baseTime.Add(2 * time.Hour), // more recent
			blockNum:  1002,
		},
		{
			messages: []inmem.MessagesWithMetadata{
				makeMsgWithMetadata(100, srcChainA, dstSelector, false),
				makeMsgWithMetadata(101, srcChainA, dstSelector, false),
```

execute/plugin_e2e_test.go
```diff
@@ -188,3 +188,79 @@ func TestPlugin_FinalizedUnfinalizedCaching(t *testing.T) {
...
+			},
+			timestamp: baseTime,
+			blockNum:  1000,
```
makramkd, line 220: This block num should be later than the one with smaller sequence numbers.

execute/plugin_e2e_test.go
```diff
@@ -188,3 +188,79 @@ func TestPlugin_FinalizedUnfinalizedCaching(t *testing.T) {
...
+		},
+		{
+			// Middle messages
+			messages: []inmem.MessagesWithMetadata{
+				makeMsgWithMetadata(104, srcChainA, dstSelector, false),
+				makeMsgWithMetadata(105, srcChainA, dstSelector, false),
+			},
+			timestamp: baseTime.Add(time.Hour),
+			blockNum:  1001,
```
makramkd, line 229: This should have a block num >= the one with 102 and 103

execute/plugin_e2e_test.go
```diff
@@ -188,3 +188,79 @@ func TestPlugin_FinalizedUnfinalizedCaching(t *testing.T) {
 	report := outcome.CommitReports[0]
 	require.Equal(t, cciptypes.NewSeqNumRange(200, 201), report.SequenceNumberRange)
 }
+
+func TestPlugin_CommitReportTimestampOrdering(t *testing.T) {
+	ctx := tests.Context(t)
+
+	srcChainA := cciptypes.ChainSelector(1)
```
makramkd, line 195: Makes sense to add a test w/ multiple source chains right?

execute/observation.go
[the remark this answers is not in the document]
makramkd: Or who knows, I wonder if Go optimizes the alloc away once it sees the `len` 🤔

execute/report/report.go
```diff
@@ -594,6 +594,11 @@ func (b *execReportBuilder) buildSingleChainReport(
 			}
 		} else {
 			// this message didn't work, continue to the next one
+			b.lggr.Debugw("message did not fit in report, deleting from in-progress report built",
+				"sourceChain", commitData.Messages[i].Header.SourceChainSelector,
+				"messageId", commitData.Messages[i].Header.MessageID,
```
makramkd, line 599:
```suggestion
				"messageID", commitData.Messages[i].Header.MessageID,
```

In order to match the other usages of `messageID` as a log key (see line 580 above)

pkg/types/ccipocr3/plugin_execute_types_test.go
```diff
@@ -38,6 +38,7 @@ func TestDecodeExecuteReportInfo(t *testing.T) {
 }
 
 func TestExecuteReportInfo_EncodeDecode(t *testing.T) {
+
```
makramkd, line 41: Can rm

execute/exectypes/outcome.go
```diff
@@ -104,8 +104,8 @@ func NewSortedOutcome(
 	sort.Slice(
 		pendingCommitsCP,
 		func(i, j int) bool {
-			if pendingCommitsCP[i].SourceChain != pendingCommitsCP[j].SourceChain {
-				return pendingCommitsCP[i].SourceChain < pendingCommitsCP[j].SourceChain
+			if !pendingCommitsCP[i].Timestamp.Equal(pendingCommitsCP[j].Timestamp) {
+				return pendingCommitsCP[i].Timestamp.Before(pendingCommitsCP[j].Timestamp)
 			}
 			return pendingCommitsCP[i].SequenceNumberRange.Start() < pendingCommitsCP[j].SequenceNumberRange.Start()
```
makramkd, line 110: I feel like there's a redundancy here, if (for the same source chain) report A has timestamp < report B, it's sequence numbers must necessarily be less than report B.

asoliman92: My test wasn't failing in both logics
It's a way for tests to keep tests without timestamps (zero timestamp) or to use same one without worrying.
Also just in the very rare case we have 2 identical timestamps to have a fallback.

Another test was failing because it can't sort them deterministically having same timestamp. I had a hack earlier to change timestamps for each report but then decided to go this way better.

makramkd (approved):

PR #710, opened 2025-03-13
[Solana][NONEVM-1403]: Fix Mint PDA derivation in tokenpools

PabloMansanet:
https://smartcontract-it.atlassian.net/browse/NONEVM-1403


agusaldasoro (commented):
Should we validate the `state.config.mint` value matches the seed and the mint when setting it up?

PabloMansanet:
> Should we validate the `state.config.mint` value matches the seed and the mint when setting it up?

In `base-token-pool/src/common.rs::BaseConfig::init`, the mint value of the config is already being set to equal the mint (which is tied to the state PDA through the seed) so we should be safe here.

agusaldasoro (approved):

PR #711, opened 2025-03-13
[Solana][NONEVM-1402] Token bucket config validation/setting

PabloMansanet:
https://smartcontract-it.atlassian.net/browse/NONEVM-1402


PabloMansanet:
Drafting as I work on understanding why this breaks some tests 👍

PR #713, opened 2025-03-13
CCIP-5449 Backgroung caching of config data

0xnogo:
Implementation of an asynchronous background polling system for chain configuration to reduce blocking operations. Key improvements:

* First call fetches data synchronously, all subsequent requests return cached data instantly
* Background thread refreshes configurations on a configurable interval
* Optimized network usage by batching chain and source chain config requests when possible
* Automatically tracks and refreshes newly discovered chains
* Added proper lifecycle management with Start/Stop methods and a Close method on `ccipChainReader`


pkg/reader/config_poller.go
```diff
@@ -57,27 +58,341 @@ type chainCache struct {
...
+	// Refresh each chain (and its source chains if applicable)
+	for _, destChain := range chainSelectors {
+		// Skip chains we don't know as dest chains
+		sourceChains, hasSourceChains := sourceChainsMap[destChain]
+
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+
+		// If this is a destination chain with source chains, batch refresh both
+		if hasSourceChains && len(sourceChains) > 0 {
+			err := c.batchRefreshChainAndSourceConfigs(ctx, destChain, sourceChains)
```
0xnogo, line 141: batchRefreshChainAndSourceConfigs will try merge the calls for the source chain config in the same batch in order to minimize the RPC calls.

pkg/reader/config_poller.go
```diff
@@ -57,27 +58,341 @@ type chainCache struct {
...
+func (c *configPoller) extractStandardChainConfigResults(
+	batchResult types.BatchGetLatestValuesResult,
+	standardOffRampRequestCount int,
+) types.BatchGetLatestValuesResult {
+	chainConfigResultsCopy := make(types.BatchGetLatestValuesResult)
+
+	for contract, results := range batchResult {
+		if contract.Name == consts.ContractNameOffRamp && len(results) > standardOffRampRequestCount {
+			// Only include the standard results (first N results)
+			chainConfigResultsCopy[contract] = results[:standardOffRampRequestCount]
```
0xnogo, line 184: We are getting the first N results:
* 0 -> N = config calls
* N+1 -> end = source chain calls

This is allowing us to reuse the processing methods

pkg/reader/config_poller.go
```diff
@@ -212,8 +528,56 @@ func (c *configPoller) GetOfframpSourceChainConfigs(
 	return result, nil
 }
 
-// RefreshSourceChainConfigs forces a refresh of source chain configurations
-func (c *configPoller) RefreshSourceChainConfigs(
+// refreshChainConfig forces a refresh of the chain configuration
+func (c *configPoller) refreshChainConfig(
```
0xnogo, line 532: Made the refresh methods private as they are not used.

pkg/reader/ccip.go
```diff
@@ -88,6 +88,10 @@ func newCCIPChainReaderInternal(
 		lggr.Errorw("failed to sync contracts", "err", err)
 	}
 
+	// After contracts are synced, start the background polling
+	reader.configPoller.Start()
+	lggr.Info("Started config background polling")
```
makramkd, line 93: nit: log belongs better probably in `configPoller.Start()` method

pkg/reader/ccip.go
```diff
@@ -98,6 +102,16 @@ func (r *ccipChainReader) WithExtendedContractReader(
 	return r
 }
 
+func (r *ccipChainReader) Close() error {
+	// Stop the background poller
+	r.configPoller.Stop()
+	r.lggr.Info("Stopped config background polling")
```
makramkd, line 108: nit: log belongs better probably in `configPoller.Stop()` method

makramkd: another nit: to be consistent maybe `Stop` can be called `Close` instead? Also makes it compatible with `io.Closer`

pkg/reader/ccip_test.go
```diff
@@ -90,6 +90,15 @@ func TestCCIPChainReader_getSourceChainsConfig(t *testing.T) {
 		}, nil, chainC, offrampAddress, mockAddrCodec,
 	)
 
+	// Add cleanup to ensure resources are released
+	t.Cleanup(func() {
+		if ccipReader.configPoller != nil {
+			ccipReader.configPoller.Stop()
+		}
+		// Allow some time for goroutines to stop
+		time.Sleep(50 * time.Millisecond)
```
makramkd, line 99: I think this is a recipe for flakes, can `Stop` be blocking in the configPoller instead?

pkg/reader/config_poller.go
```diff
@@ -65,11 +67,333 @@ func newConfigPoller(
...
+			}
+		}
+	}()
+}
+
+// Add Start method to configPoller
+func (c *configPoller) Start() {
+	c.startBackgroundPolling()
+
+	c.lggr.Info("Background poller started")
```
makramkd, line 102: Some field I would add to this log (unless they're already part of logger fields): `c.refreshPeriod`.

pkg/reader/config_poller.go
```diff
@@ -65,11 +67,333 @@ func newConfigPoller(
...
+		// If this chain has source chains, gather them
+		if sourceChains, exists := c.knownSourceChains[chainSel]; exists && len(sourceChains) > 0 {
+			sourceList := make([]cciptypes.ChainSelector, 0, len(sourceChains))
+			for sourceChain := range sourceChains {
+				sourceList = append(sourceList, sourceChain)
+			}
+			sourceChainsMap[chainSel] = sourceList
+		}
+	}
+	c.RUnlock()
```
makramkd, line 130: This seems like a good candidate for a helper method + unit test (if not already implemented). Can then make use of the `c.RLock()` -> `defer c.RUnlock()` paradigm for less error-prone locking

pkg/reader/config_poller.go
```diff
@@ -6,41 +6,43 @@ import (
...
 
-// configPoller handles caching of chain configuration data for multiple chains.
-// It is used by the ccipChainReader to store and retrieve configuration data,
-// avoiding unnecessary contract calls and improving performance.
 // configPoller handles caching of chain configuration data for multiple chains
 type configPoller struct {
 	sync.RWMutex
 	chainCaches   map[cciptypes.ChainSelector]*chainCache
 	refreshPeriod time.Duration
 	reader        *ccipChainReader // Reference to the reader for fetching configs
```
makramkd, line 37: qq: do we need a pointer to the impl here or is the interface type enough? Interface type would maybe make it easier for testing?

pkg/reader/config_poller.go
```diff
@@ -65,11 +67,333 @@ func newConfigPoller(
...
+				c.refreshAllKnownChains()
+			case <-c.stopChan:
+				return
+			}
+		}
+	}()
+}
+
+// Add Start method to configPoller
+func (c *configPoller) Start() {
```
makramkd, line 99: This should only ever be started / stopped once right? If thats the case we can make use of `services.StateMachine` and the `StartOnce` `StopOnce` helpers.

pkg/reader/config_poller.go
```diff
@@ -65,11 +67,333 @@ func newConfigPoller(
...
+			}
+			sourceChainsMap[chainSel] = sourceList
+		}
+	}
+	c.RUnlock()
+
+	// Refresh each chain (and its source chains if applicable)
+	for _, destChain := range chainSelectors {
+		// Skip chains we don't know as dest chains
+		sourceChains, hasSourceChains := sourceChainsMap[destChain]
```
makramkd, line 135: I don't think I'm fully understanding the configPoller context - we create one config poller per plugin, and it is responsible for refreshing the configs for a single dest chain + all associated source chains right? The comments here seem to suggest it may refresh configs for multiple destination chains, is that the case?

pkg/reader/config_poller.go
```diff
@@ -65,11 +67,333 @@ func newConfigPoller(
...
+
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
makramkd, line 137: nit: `30 * time.Second` can be a const `bgRefreshTimeout` at the top of this file

pkg/reader/config_poller.go
0xnogo: > refreshing the configs for a single dest chain + all associated source chains right

This is correct but we are also fetching config data from source chains

```
type ChainConfigSnapshot struct {
	Offramp   OfframpConfig <- dest chain
	RMNProxy  RMNProxyConfig <- dest chain
	RMNRemote RMNRemoteConfig <- dest chain
	FeeQuoter FeeQuoterConfig <- dest chain
	OnRamp    OnRampConfig <- src chain
	Router    RouterConfig <- src chain
	CurseInfo CurseInfo <- dest chain
}
```

So we can call `GetChainConfig(dstChain)` but also `GetChainConfig(any of the srcChain)`. Depending on whether the dest chain is being called or not we will route the batch request creation in `prepareBatchConfigRequests`. But ChainConfigSnapshot will be empty for a `srcChain` on OfframpConfig, RMNProxyConfig, ...

I think you raised a good point as the naming is not properly defined as this is not only `destChains`. I think we are also missing a check on destChains. It should be something like this

```
// Refresh each chain (and its source chains if applicable)
	for _, chainSel := range chainSelectors {
		// If this is the destination chain and has source chains, use batch refresh
		if chainSel == c.reader.destChain && len(sourceChainsMap[chainSel]) > 0 {
			sourceChains := sourceChainsMap[chainSel]
			err := c.batchRefreshChainAndSourceConfigs(ctx, chainSel, sourceChains)
		} else {
			// This is a source chain or a chain without source chains, just refresh its config
			_, err := c.refreshChainConfig(ctx, chainSel)
			
		}
		```

File: pkg/reader/config_poller.go (hunk 8 above)
Comment (0xnogo) [reply pkg/reader/config_poller.go:130]: Good point

Comment (0xnogo) [reply pkg/reader/config_poller.go:130]: Created the test `TestConfigPoller_GetChainsToRefresh`

File: pkg/reader/config_poller.go (hunk 9 above)
Comment (0xnogo) [reply pkg/reader/config_poller.go:37]: We actually the implementation here as we are getting some elements from the struct like the `destChain` or the `contractReaders` slice

Comment (makramkd) [reply pkg/reader/config_poller.go:37]: This doesn't seem like a great practice to follow - private members of `ccipReader` are implementation details and shouldn't be relied upon by users of it. I feel like defining an interface like `ccipReaderInternal` and some methods may be a better approach that is also more testable. e.g:

```golang
// private because only used by config poller
type ccipReaderInternal interface {
  getDestChain() cciptypes.ChainSelector
  getContractReader(chain cciptypes.ChainSelector) (types.ContractReader, error)
  // ....
}
```

pkg/reader/ccip_test.go
0xnogo: Indeed it could leads to flakes. I think this wait one was not needed at all. Removed it.

pkg/reader/config_poller.go
0xnogo: Good idea.

pkg/reader/config_poller.go
0xnogo: Let's do it! It's way cleaner to use the interface indeed

makramkd (dismissed):

mateusz-sekara (changes requested):

pkg/reader/config_poller.go
```diff
@@ -6,41 +6,50 @@ import (
...
 		ctx context.Context,
 		destChain cciptypes.ChainSelector,
 		sourceChains []cciptypes.ChainSelector) (map[cciptypes.ChainSelector]StaticSourceChainConfig, error)
-	// RefreshSourceChainConfigs forces a refresh of source chain configurations
-	RefreshSourceChainConfigs(
-		ctx context.Context,
-		destChain cciptypes.ChainSelector,
-		sourceChains []cciptypes.ChainSelector) (map[cciptypes.ChainSelector]StaticSourceChainConfig, error)
+	// Start starts the background polling
+	Start() error
```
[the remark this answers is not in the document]
0xnogo: For consistency, I guess we can switch to `Service`

pkg/reader/config_poller.go
```diff
@@ -61,15 +70,359 @@ type chainCache struct {
...
+	map[cciptypes.ChainSelector][]cciptypes.ChainSelector) {
+	c.RLock()
+	defer c.RUnlock()
+
+	// Get all chains to refresh
+	chainSelectors := make([]cciptypes.ChainSelector, 0, len(c.chainCaches))
+	sourceChainsMap := make(map[cciptypes.ChainSelector][]cciptypes.ChainSelector)
+
+	for chainSel := range c.chainCaches {
+		chainSelectors = append(chainSelectors, chainSel)
```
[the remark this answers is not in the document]
0xnogo: I did that to avoid holding the lock on the entire running of `refreshAllKnownChains`. `chainSelectors` is used there.

pkg/reader/config_poller.go
```diff
@@ -204,8 +558,56 @@ func (c *configPoller) GetOfframpSourceChainConfigs(
...
-// RefreshSourceChainConfigs forces a refresh of source chain configurations
-func (c *configPoller) RefreshSourceChainConfigs(
+// refreshChainConfig forces a refresh of the chain configuration
+func (c *configPoller) refreshChainConfig(
+	ctx context.Context,
+	chainSel cciptypes.ChainSelector,
+) (ChainConfigSnapshot, error) {
+	chainCache := c.getOrCreateChainCache(chainSel)
+
+	chainCache.chainConfigMu.Lock()
```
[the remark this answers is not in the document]
0xnogo: Makes sense, thanks

pkg/reader/ccip_test.go
```diff
@@ -90,6 +90,14 @@ func TestCCIPChainReader_getSourceChainsConfig(t *testing.T) {
 		}, nil, chainC, offrampAddress, mockAddrCodec,
 	)
 
+	// Add cleanup to ensure resources are released
+	t.Cleanup(func() {
+		err := ccipReader.Close()
```
mateusz-sekara, line 95: nit, but probably you can ignore that error and make it as simple as
```go
t.Cleanup(func() { ccipReader.Close() } )
```

mateusz-sekara (approved):
Looks good, I have two more questions:
* how do we want to test in more e2e way?
* do we need some sort of metrics to track the behaviour of the bg config poller?

0xnogo:
@mateusz-sekara 
* The config poller is integration tested throughout the ccip reader tests. I think we can introduce some more specific testing around it.
* I thought of some metrics we could have for the config poller. I'd like to discuss that with you. Indeed it could be a great add up for this component.

PR #716, opened 2025-03-14
CCIP-5449 NextSeqNum uses fresh source chain configs

0xnogo:
* Prevented use of stale `MinSeqNr` values by modifying `NextSeqNum` to bypass config cache
* Added direct fetch method for source chain configs to ensure fresh data retrieval
* Introduced `StaticSourceChainConfig` structure that excludes `MinSeqNr` to prevent misuse
* Changed the `ConfigPoller` interface to use `StaticSourceChainConfig` in its return types
* Ensured `MinSeqNr` is only used by `NextSeqNum`, which now always fetches fresh values


asoliman92 (approved):

pkg/reader/ccip.go
```diff
@@ -1322,6 +1305,89 @@ func (r *ccipChainReader) getOffRampSourceChainsConfig(
 	return configs, nil
 }
 
+// fetchDirectSourceChainConfigs always fetches fresh source chain configs directly from contracts
+// without using any cached values. Use this when up-to-date data is critical, especially
+// for sequence number accuracy.
+func (r *ccipChainReader) fetchDirectSourceChainConfigs(
```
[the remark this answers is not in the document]
0xnogo: sure

pkg/reader/ccip.go
```diff
@@ -1322,6 +1309,89 @@ func (r *ccipChainReader) getOffRampSourceChainsConfig(
...
+		return nil, fmt.Errorf("no contract reader for chain %d", destChain)
+	}
+
+	// Filter out destination chain
+	filteredSourceChains := filterOutChainSelector(sourceChains, destChain)
+	if len(filteredSourceChains) == 0 {
+		return make(map[cciptypes.ChainSelector]SourceChainConfig), nil
+	}
+
+	// Prepare batch requests for the sourceChains
```
winder, line 1333:
```suggestion
	// Prepare batch requests for the sourceChains to fetch the latest Unfinalized config values.
```

winder: You don't get to specify finalized/unfinalized for batch requests and I keep forgetting what the default is... worth calling out in comments I think.

pkg/reader/ccip_interface.go
```diff
@@ -78,6 +79,48 @@ func (ca ContractAddresses) Append(contract string, chain cciptypes.ChainSelecto
 	return resp
 }
 
+// CachedSourceChainConfig stores the static parts of SourceChainConfig
```
makramkd, line 82:
```suggestion
// StaticSourceChainConfig stores the static parts of SourceChainConfig
```

pkg/reader/config_poller.go
```diff
@@ -437,6 +356,17 @@ func filterOutChainSelector(
 	return filtered
 }
 
+// FromSourceChainConfig creates a StaticSourceChainConfig from a SourceChainConfig,
+// omitting the MinSeqNr field.
+func StaticSourceChainConfigFromSourceChainConfig(sc SourceChainConfig) StaticSourceChainConfig {
```
makramkd, line 361: Can be private

pkg/reader/config_poller.go
```diff
@@ -437,6 +356,17 @@ func filterOutChainSelector(
 	return filtered
 }
 
+// FromSourceChainConfig creates a StaticSourceChainConfig from a SourceChainConfig,
```
makramkd, line 359:
```suggestion
// StaticSourceChainConfigFromSourceChainConfig creates a StaticSourceChainConfig from a SourceChainConfig,
```

pkg/reader/config_poller.go
```diff
@@ -55,17 +54,10 @@ type chainCache struct {
 
 	// Source chain config specific lock and data
 	sourceChainMu      sync.RWMutex
-	sourceChainConfigs map[cciptypes.ChainSelector]SourceChainConfig
+	sourceChainConfigs map[cciptypes.ChainSelector]StaticSourceChainConfig
```
makramkd, line 57: nit:
```suggestion
	staticSourceChainConfigs map[cciptypes.ChainSelector]StaticSourceChainConfig
```

pkg/reader/ccip.go
```diff
@@ -1322,6 +1309,89 @@ func (r *ccipChainReader) getOffRampSourceChainsConfig(
...
+	)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to get source chain configs: %w", err)
+	}
+
+	// Process results
+	configs := make(map[cciptypes.ChainSelector]SourceChainConfig)
+
+	for _, readResult := range results {
```
makramkd, line 1364: Should we check that `len(readResult) == 1` and error otherwise? Seems like thats how the API is designed, one result per contract name https://github.com/smartcontractkit/chainlink-common/blob/6f1f48342e36fa00a3e89d73695922c47aa94987/pkg/types/contract_reader.go#L56

pkg/reader/ccip.go
```diff
@@ -1322,6 +1309,89 @@ func (r *ccipChainReader) getOffRampSourceChainsConfig(
...
+
+			cfg, ok := v.(*SourceChainConfig)
+			if !ok {
+				lggr.Errorw("Invalid result type from GetSourceChainConfig",
+					"chain", chain,
+					"type", fmt.Sprintf("%T", v))
+				return nil, fmt.Errorf("invalid result type from GetSourceChainConfig for chainSelector=%d", chain)
+			}
+
+			// Store the config - we don't filter here as that's done at the reader level
```
makramkd, line 1387: What does this comment mean `we don't filter here as that's done at the reader level`

pkg/reader/ccip.go
```diff
@@ -1322,6 +1309,89 @@ func (r *ccipChainReader) getOffRampSourceChainsConfig(
...
+					"error", err)
+				return nil, fmt.Errorf("GetSourceChainConfig for chainSelector=%d failed: %w", chain, err)
+			}
+
+			cfg, ok := v.(*SourceChainConfig)
+			if !ok {
+				lggr.Errorw("Invalid result type from GetSourceChainConfig",
+					"chain", chain,
+					"type", fmt.Sprintf("%T", v))
+				return nil, fmt.Errorf("invalid result type from GetSourceChainConfig for chainSelector=%d", chain)
```
makramkd, line 1384:
```suggestion
				return nil, fmt.Errorf("invalid result type (%T) from GetSourceChainConfig for chainSelector=%d, expected *SourceChainConfig", v, chain)
```

pkg/reader/ccip.go
0xnogo: `readResult` should encapsulate all the calls made for the Offramp contract, so I assume that we are referring to `results`? I can add a check on `len(readResult)`. Have no result should still return 1 entry.

pkg/reader/ccip.go
0xnogo: I was referring to the filtering logic based on `includeDisabled `, but I realize that it's not that clear, removing the comment

pkg/reader/ccip.go
makramkd: Ah yeah I meant `results`.

PR #717, opened 2025-03-14
Initial pass at converting access layer for CCIPReader.

winder:
This interface was created by going through each method in the [CCIPReader interface](https://github.com/smartcontractkit/chainlink-ccip/blob/263cd972f3e116c007003a949cbb79b97bcfa7ae/pkg/reader/ccip_interface.go#L122) and figuring out what chain access primitives are required.

Many of the changes are related to the access layer being tied 1:1 with a particular chain. The CCIPReader aggregates requests across many readers, which means additional parameters are needed to specify where data should be fetched from.

There are also some new specific design questions that need to be resolved:
* How does the access layer relate to ConfigPoller. The current CCIPReader API has several functions that pull data from ConfigPoller, or are direct wrappers around ConfigPoller. My current leaning is that ConfigPoller uses the accessor, and CCIPReader uses the ConfigPoller.
* How exactly does contract address discovery work? It seems that the existing DiscoverContracts/Sync function would still work, but those functions also use ConfigPoller.
* There are a handful of types that need to be redefined or moved.
* See inline `TODO` comments and the [design document notes](https://docs.google.com/document/d/13wHGMr6Odfxzqp9XLL4TxBWId1w17bMhF-S-YZaS18c/edit?tab=t.0#heading=h.pzedmeu7xpqk) for more.

### DIFF
--- a/pkg/types/ccipocr3/chain_accessor.go
+++ b/pkg/types/ccipocr3/chain_accessor.go
@@ -1,0 +1,313 @@
+package ccipocr3
+
+import (
+	"context"
+	"sort"
+	"time"
+
+	rmntypes "github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn/types"
+	iplugintypes "github.com/smartcontractkit/chainlink-ccip/internal/plugintypes"
+	"github.com/smartcontractkit/chainlink-ccip/plugintypes"
+	"github.com/smartcontractkit/chainlink-common/pkg/types"
+	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
+)
+
+// AccessorMetadata contains metadata about the chain accessor.
+type AccessorMetadata struct {
+	ChainSelector ChainSelector
+	Contracts     map[string]UnknownAddress
+}
+
+// ChainAccessor for all direct chain access. The accessor is responsible for
+// in addition to direct access to the chain, this interface also translates
+// onchain representations of data to the plugin representation.
+type ChainAccessor interface {
+	AllAccessors
+	SourceAccessor
+	DestinationAccessor
+	RMNAccessor
+}
+
+// AllAccessors contains functionality that is available to all types of accessors.
+type AllAccessors interface {
+	// Metadata associated with the chain accessor. Replaces GetContractAddress.
+	Metadata() AccessorMetadata
+
+	// TODO: Discovery/Binding functions here?
+
+	// GetContractAddress returns the contract address that is registered for the provided contract name and chain.
+	// WARNING: This function will fail if the oracle does not support the requested chain.
+	//
+	// Deprecated: use Metadata() instead.
+	GetContractAddress(contractName string) ([]byte, error)
+
+	// GetChainFeeComponents Returns all fee components for given chains if corresponding
+	// chain writer is available.
+	//
+	// Access Type: ChainWriter
+	// Contract: N/A
+	// Confidence: N/A
+	GetChainFeeComponents(
+		ctx context.Context,
+	) map[ChainSelector]ChainFeeComponents
+
+	// GetDestChainFeeComponents seems redundant. If the error is needed lets
+	// add it to GetChainFeeComponents.
+	//
+	// Deprecated: use GetChainFeeComponents instead.
+	GetDestChainFeeComponents(
+		ctx context.Context,
+	) (types.ChainFeeComponents, error)
+
+	// DiscoverContracts reads the destination chain for contract addresses. They are returned per
+	// contract and source chain selector.
+	// allChains is needed because there is no way to enumerate all chain selectors on Solana. We'll attempt to
+	// fetch the source config from the offramp for each of them.
+	DiscoverContracts(ctx context.Context, allChains []ChainSelector) (ContractAddresses, error)
+
+	// Sync can be used to perform frequent syncing operations inside the reader implementation.
+	// Returns a bool indicating whether something was updated.
+	Sync(ctx context.Context, contracts ContractAddresses) error
+}
+
+// DestinationAccessor contains all functions typically associated by the destination chain.
+type DestinationAccessor interface {
+
+	// CommitReportsGTETimestamp reads CommitReportAccepted events starting from a given timestamp.
+	// The number of results are limited according to limit.
+	//
+	// Access Type: Event(CommitReportAccepted)
+	// Contract: OffRamp
+	// Confidence: Unconfirmed
+	CommitReportsGTETimestamp(
+		ctx context.Context,
+		ts time.Time,
+		limit int,
+	) ([]plugintypes.CommitPluginReportWithMeta, error)
+
+	// ExecutedMessages looks for ExecutionStateChanged events for each sequence
+	// in a given range. The presence of these events indicates that an attempt to
+	// execute the message has been made, which the system considers "executed".
+	// A slice of all executed sequence numbers is returned.
+	//
+	// Access Type: Event(ExecutionStateChanged)
+	// Contract: OffRamp
+	// Confidence: Unconfirmed, Finalized
+	ExecutedMessages(
+		ctx context.Context,
+		source ChainSelector,
+		seqNumRange SeqNumRange,
+		confidence ConfidenceLevel,
+	) ([]SeqNum, error)
+
+	// NextSeqNum reads the source chain config to get the next expected
+	// sequence number for the given source chains.
+	//
+	// Access Type: Method(GetSourceChainConfig)
+	// Contract: OffRamp
+	// Confidence: Unconfirmed
+	//
+	// TODO: Have a single "GetSourceChainConfigs" method for the log poller to
+	//       use for cached values, and the plugin to use for non-cached values.
+	//       Rename to "GetSourceChainConfig".
+	//      See Design Doc (NextSeqNum) for notes.
+	NextSeqNum(ctx context.Context, sources []ChainSelector) (
+		seqNum map[ChainSelector]SeqNum, err error)
+
+	// Nonces for all provided selector/address pairs. Addresses must be encoded
+	// according to the source chain requirements by using the AddressCodec.
+	//
+	// Access Type: Method(GetInboundNonce)
+	// Contract: NonceManager
+	// Confidence: Unconfirmed
+	Nonces(
+		ctx context.Context,
+		addresses map[ChainSelector][]UnknownEncodedAddress,
+	) (map[ChainSelector]map[string]uint64, error)
+
+	// GetChainFeePriceUpdate Gets latest chain fee price update for the provided chains.
+	//
+	// Access Type: Method(GetChainFeePriceUpdate)
+	// Contract: FeeQuoter
+	// Confidence: Unconfirmed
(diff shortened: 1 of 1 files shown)


pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+
+// ChainAccessor for all direct chain access. The accessor is responsible for
+// in addition to direct access to the chain, this interface also translates
+// onchain representations of data to the plugin representation.
+type ChainAccessor interface {
+	AllAccessors
+	SourceAccessor
+	DestinationAccessor
+	RMNAccessor
+}
```
winder, line 29: Very rough organization of methods. These interfaces may become private or change before everything is finished. Some points I'm considering:

* Source/Destination accessors could be top level objects that the CCIPReader has access to rather than the Generic ChainAccessor. This would prevent misusing src/dest only function calls.
* Are the RMN functions globally accessible? It might belong in the All interface.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// DiscoverContracts reads the destination chain for contract addresses. They are returned per
+	// contract and source chain selector.
+	// allChains is needed because there is no way to enumerate all chain selectors on Solana. We'll attempt to
+	// fetch the source config from the offramp for each of them.
+	DiscoverContracts(ctx context.Context, allChains []ChainSelector) (ContractAddresses, error)
+
+	// Sync can be used to perform frequent syncing operations inside the reader implementation.
+	// Returns a bool indicating whether something was updated.
+	Sync(ctx context.Context, contracts ContractAddresses) error
```
winder, line 70: This interface seems fine as is. Except `DiscoverContracts` is a ConfigPoller wrapper so maybe it goes away and the interface only has a `Sync` function. Also, might rename `Sync` to `UpdateAddresses`.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+	// GetRmnCurseInfo returns rmn curse/pausing information about the provided chains
+	// from the destination chain RMN remote contract. Caller should be able to access destination.
+	GetRmnCurseInfo(ctx context.Context) (CurseInfo, error)
+}
+
+////////////////////////////////////////////////////////////////
+// TODO: Find a better location for the types below this line //
+//       For the purpose of designing these interfaces, the   //
+//       location is not critical.                            //
+////////////////////////////////////////////////////////////////
```
winder, line 243: Open to suggestions on types. I don't hate defining all of these in the ccipocr3 package.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+	// Access Type: Method(GetExpectedNextSequenceNumber)
+	// Contract: OnRamp
+	// Confidence: Unconfirmed
+	GetExpectedNextSequenceNumber(
+		ctx context.Context,
+		dest ChainSelector,
+	) (SeqNum, error)
+
+	// GetTokenPriceUSD looks up a token price in USD. The address value should
+	// be retrieved from a configuration cache (i.e. ConfigPoller).
```
makramkd, line 198: Haven't read the full PR yet, but question about this doc comment: are we saying that the CCIPReader (the object that manages / coordinates all of the Accessor objects) is supposed to access this via a cache or is this saying that the accessor should have this cached?

To keep things vanilla and simple I'd prefer the accessor doesn't have caching logic and just does the I/O necessary to give you the data you need. I think this is whats going on but wanted to confirm.

winder: I would prefer if the accessor does not cache anything. This was a point of discussion /w @mateusz-sekara and @0xnogo and we didn't have consensus.

After reviewing the code last week, I think it's clear that the accessor does NOT need a cache. In this case I added a new argument to the poller so that the CCIPReader can manage the cache and pass it into the accessor. It turns out there are many other functions like this.

All of that said, I don't think the CCIP Reader should be caching either. Having it abstracted that way makes it difficult for the plugin to know when IO is happening. I was really surprised by how many CCIPReader functions are implemented by the ConfigPoller. Among other things the `DiscoverContracts` function should be refactored to be a single simple function.

Moving the ConfigPoller into the plugins would be a much larger change, I'm not planning to touch it with this project.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+	// Deprecated: use GetChainFeeComponents instead.
+	GetDestChainFeeComponents(
+		ctx context.Context,
+	) (types.ChainFeeComponents, error)
+
+	// DiscoverContracts reads the destination chain for contract addresses. They are returned per
+	// contract and source chain selector.
+	// allChains is needed because there is no way to enumerate all chain selectors on Solana. We'll attempt to
+	// fetch the source config from the offramp for each of them.
+	DiscoverContracts(ctx context.Context, allChains []ChainSelector) (ContractAddresses, error)
```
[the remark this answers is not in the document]
winder: +1

I'll remove this from the interface.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// Nonces for all provided selector/address pairs. Addresses must be encoded
+	// according to the source chain requirements by using the AddressCodec.
+	//
+	// Access Type: Method(GetInboundNonce)
+	// Contract: NonceManager
+	// Confidence: Unconfirmed
+	Nonces(
+		ctx context.Context,
+		addresses map[ChainSelector][]UnknownEncodedAddress,
```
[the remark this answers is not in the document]
winder: Yes, I modified this from the current CCIPReader API.

The CCIPReader signature is
```
	Nonces(
		ctx context.Context,
		source ChainSelector,
		addresses []string,
	) (map[string]uint64, error)
```

The version here allows for improved batching.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+	// execute the message has been made, which the system considers "executed".
+	// A slice of all executed sequence numbers is returned.
+	//
+	// Access Type: Event(ExecutionStateChanged)
+	// Contract: OffRamp
+	// Confidence: Unconfirmed, Finalized
+	ExecutedMessages(
+		ctx context.Context,
+		source ChainSelector,
+		seqNumRange SeqNumRange,
```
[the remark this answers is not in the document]
winder: Thanks. I noticed this optimization for Nonces and missed it for ExecutedMessages. I'll update the signature.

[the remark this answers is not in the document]
0xnogo: This was already noticed, @0xAustinWang is working on the optimization: https://github.com/smartcontractkit/chainlink-ccip/pull/719

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+
+type ChainFeeComponents = types.ChainFeeComponents
+
+type TimestampedBig = iplugintypes.TimestampedBig
+
+// sourceChainConfig is used to parse the response from the offRamp contract's getSourceChainConfig method.
+// See: https://github.com/smartcontractkit/ccip/blob/a3f61f7458e4499c2c62eb38581c60b4942b1160/contracts/src/v0.8/ccip/offRamp/OffRamp.sol#L94
+//
+//nolint:lll // It's a URL.
+type SourceChainConfig struct {
```
0xnogo, line 267: We have introduced a `StaticSourceChainConfig` in order to keep caching it without the `MinSeqNr` in https://github.com/smartcontractkit/chainlink-ccip/pull/716 but I think the logic we remain the same here

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// GetOffRampConfigDigest returns the offramp config digest for the provided plugin type.
+	//
+	// TODO: this should go away, it's a call to the ConfigPoller.
+	GetOffRampConfigDigest(ctx context.Context, pluginType uint8) ([32]byte, error)
+
+	// GetOffRampSourceChainsConfig returns the sourceChains config for all the provided source chains.
+	// If a config was not found it will be missing from the returned map.
+	//
+	// TODO: this should go away, it's a call to the ConfigPoller.
```
0xnogo, line 154: Shouldn't we keep it in the Accessor and have the ConfigPoller call it instead? I was under the impression that the source of truth is going to be the Accessor and the ConfigPoller will simply have to deal with caching.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,313 @@
...
+	) (FeeQuoterDestChainConfig, error)
+}
+
+type RMNAccessor interface {
+	// GetRMNRemoteConfig
+	//
+	// TODO: This function should be removed and replaced with direct access to
+	//       the ConfigPoller. It also manages chain reader contract binding
+	//       directly instead of using DiscoverContracts.
+	GetRMNRemoteConfig(ctx context.Context) (rmntypes.RemoteConfig, error)
```
0xnogo, line 232: The ConfigPoller is also providing:
```
type ChainConfigSnapshot struct {
	Offramp   OfframpConfig
	RMNProxy  RMNProxyConfig
	RMNRemote RMNRemoteConfig
	FeeQuoter FeeQuoterConfig
	OnRamp    OnRampConfig
	Router    RouterConfig
	CurseInfo CurseInfo
}
```

We are batching all the calls (apart from the src chain calls which are batched separately) and merging everything into that struct. I have the feeling that we are loosing this aspect here as we have different methods per data? Example `RMNRemoteConfig` in the `RMNAccessor` or the `CurseInfo`

0xnogo (dismissed):

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,389 @@
...
+	) (FeeQuoterDestChainConfig, error)
+}
+
+type RMNAccessor interface {
+	// GetRMNRemoteConfig
+	//
+	// TODO: This function should be removed and replaced with direct access to
+	//       the ConfigPoller. It also manages chain reader contract binding
+	//       directly instead of using DiscoverContracts.
+	GetRMNRemoteConfig(ctx context.Context) (RemoteConfig, error)
```
winder, line 246: This should go to the dest accessor.

pkg/types/ccipocr3/chain_accessor.go
```diff
@@ -0,0 +1,389 @@
...
+	//
+	// Notes: This is a new general purpose function needed to implement
+	//        GetMedianDataAvailabilityGasConfig.
+	GetFeeQuoterDestChainConfig(
+		ctx context.Context,
+		dest ChainSelector,
+	) (FeeQuoterDestChainConfig, error)
+}
+
+type RMNAccessor interface {
```
winder, line 240: These are only on the destination chain, so this could be a subcomponent for the DestinationAccessor.

PR #719, opened 2025-03-15
ccip reader can check execution status of disjoint seqnr ranges

0xAustinWang:
core ref: 727156662432a6bf8d782304b17c1a9488c6ee60

https://smartcontract-it.atlassian.net/browse/CCIP-5300

Allow passing an array of seqnrranges to the ccipreader to optimize DB query


pkg/reader/ccip_interface.go
```diff
@@ -133,7 +133,7 @@ type CCIPReader interface {
 	ExecutedMessages(
 		ctx context.Context,
 		source cciptypes.ChainSelector,
```
makramkd, line 135: I think the func signature should be the following:

```suggestion
		ranges map[cciptypes.ChainSelector]SeqNumRange,
```

This allows us to query for ExecutionStateChanged messages emitted for potentially different chains in the same optimized batch query.

makramkd (changes requested):

pkg/reader/ccip_interface.go
```diff
@@ -133,7 +133,7 @@ type CCIPReader interface {
...
-		seqNumRange cciptypes.SeqNumRange,
+		seqNumRanges []cciptypes.SeqNumRange,
 		confidence primitives.ConfidenceLevel,
 	) ([]cciptypes.SeqNum, error)
```
makramkd, line 138:
```suggestion
	) (map[cciptypes.ChainSelector][]cciptypes.SeqNum, error)
```

execute/plugin.go
```diff
@@ -227,32 +227,32 @@ func getPendingReportsForExecution(
...
 			// Get finalized messages
-			finalizedMessages, err := ccipReader.ExecutedMessages(ctx, selector, seqRange, primitives.Finalized)
+			finalizedMessages, err := ccipReader.ExecutedMessages(ctx, selector, ranges, primitives.Finalized)
 			if err != nil {
-				return nil, nil, nil, fmt.Errorf("get finalized %d executed messages in range %v: %w", selector, seqRange, err)
+				return nil, nil, nil, fmt.Errorf("get finalized %d executed messages in range %v: %w", selector, ranges, err)
 			}
-			finalizedMsgSet = finalizedMsgSet.Union(mapset.NewSet(finalizedMessages...))
+			finalizedMsgSet = mapset.NewSet(finalizedMessages...)
 		}
```
makramkd, line 244: With the suggested API I think this would be:

```golang
executed, err := ccipReader.ExecutedMessages(ctx,map[cciptypes.ChainSelector][]cciptypes.SeqNumRange{selector: ranges}, primitives.Unconfirmed)
// ... rest of logic
```

execute/plugin.go
```diff
@@ -623,7 +623,11 @@ func (p *Plugin) checkAlreadyExecuted(
 	// TODO: batch these queries? these are all DB reads.
 	// maybe some alternative queries exist.
 	for sourceChainSelector, seqNrRange := range seqNrRangesBySource {
-		executed, err := p.ccipReader.ExecutedMessages(ctx, sourceChainSelector, seqNrRange, primitives.Unconfirmed)
+		executed, err := p.ccipReader.ExecutedMessages(
+			ctx,
+			sourceChainSelector,
+			[]cciptypes.SeqNumRange{seqNrRange},
+			primitives.Unconfirmed)
```
makramkd, line 630: With the suggested API this whole loop can be collapsed to a single call. We can change the type of `seqNrRangesBySource` to make the map value `[]SeqNumRange` instead of a single range:

```golang
executed, err := p.ccipReader.ExecutedMessages(ctx, seqNrRangesbySource, primitives.Unconfirmed)
if err != nil { ... }

// process results
```

makramkd (commented):
With the suggested API

execute/plugin.go
```diff
@@ -186,7 +186,8 @@ func getPendingReportsForExecution(
 	lggr.Debugw("grouped commits before removing fully executed reports",
 		"groupedCommits", groupedCommits, "count", len(groupedCommits))
 
-	// Remove fully executed reports.
+	rangesBySelector := make(map[cciptypes.ChainSelector][]cciptypes.SeqNumRange)
+	filteredReports := make(map[cciptypes.ChainSelector][]exectypes.CommitData)
```
makramkd, line 190:
```suggestion
	executableReportsPerChain := make(map[cciptypes.ChainSelector][]exectypes.CommitData)
```

Or just `executableReports` if we wanna be more compact.

execute/plugin.go
```diff
@@ -209,50 +210,55 @@ func getPendingReportsForExecution(
...
+		sort.Slice(filteredReports[selector], func(i, j int) bool {
+			return filteredReports[selector][i].SequenceNumberRange.Start() <
+				filteredReports[selector][j].SequenceNumberRange.Start()
 		})
 		// todo: remove this logs after investigating whether the sorting above can be safely removed
-		lggr.Debugw("sorted reports", "selector", selector, "reports", reports, "count", len(reports))
+		lggr.Debugw("sorted reports",
+			"selector", selector,
+			"reports", filteredReports[selector],
+			"count", len(filteredReports[selector]))
```
makramkd, line 227: I think we can simplify the logging here and just keep the sort, better to be safe than sorry. Also I think its cleaner to do the sort prior to the map assignment? And `filtered` as a name is not super expressive, maybe change to `executableReports`?

```golang
// ensure we are always executing from earliest to latest report (seq nr order).
sort.Slice(executableReports, func(i, j int) bool {
    return executableReports[i].SequenceNumberRange.Start() <
        executableReports[j].SequenceNumberRange.Start()
})
executableReportsPerChain[selector] = executableReports
lggr.Debugw("grouped reports and sorted",
    "selector", selector,
    "reports", executableReportsPerChain[selector],
    "count", len(executableReportsPerChain[selector]))
```

execute/plugin.go
```diff
@@ -209,50 +210,55 @@ func getPendingReportsForExecution(
...
+			fmt.Errorf("get executed messages in range %v: %w", rangesBySelector, err)
+	}
+	// Get finalized messages
+	finalizedMessages, err := ccipReader.ExecutedMessages(ctx, rangesBySelector, primitives.Finalized)
+	if err != nil {
+		return nil, nil, nil,
+			fmt.Errorf("get finalized executed messages in range %v: %w", rangesBySelector, err)
+	}
 
+	for selector, reports := range filteredReports {
```
makramkd, line 249: This function is actually really hard to follow despite it doing a supposedly simple thing. Here's my attempt at a refactoring into pure funcs that are separately testable and makes it easier to digest: https://github.com/smartcontractkit/chainlink-ccip/pull/759

execute/plugin.go
```diff
@@ -620,23 +626,36 @@ func (p *Plugin) checkAlreadyExecuted(
...
-			return fmt.Errorf("couldn't check if messages already executed: %w", err)
+	executed, err := p.ccipReader.ExecutedMessages(
+		ctx,
+		seqNrRangesBySource,
+		primitives.Unconfirmed)
+	if err != nil {
+		return fmt.Errorf("couldn't check if messages already executed: %w", err)
+	}
+
+	for sourceChainSelector, expectedSqNrRanges := range seqNrRangesBySource {
```
makramkd, line 637:
```suggestion
	for sourceChainSelector, reportRanges := range seqNrRangesBySource {
```

execute/plugin.go
```diff
@@ -620,23 +626,36 @@ func (p *Plugin) checkAlreadyExecuted(
...
+		snRangeSet := mapset.NewSet[cciptypes.SeqNum]()
+		totalSeqNums := 0
+		for _, snRange := range expectedSqNrRanges {
+			snRangeSet.Append(snRange.ToSlice()...)
+			totalSeqNums += snRange.Length()
 		}
```
makramkd, line 643: Can we move this below the `if _, exists` if statement which breaks the loop early?

execute/plugin.go
```diff
@@ -620,23 +626,36 @@ func (p *Plugin) checkAlreadyExecuted(
...
+	executed, err := p.ccipReader.ExecutedMessages(
+		ctx,
+		seqNrRangesBySource,
+		primitives.Unconfirmed)
+	if err != nil {
+		return fmt.Errorf("couldn't check if messages already executed: %w", err)
+	}
+
+	for sourceChainSelector, expectedSqNrRanges := range seqNrRangesBySource {
+		snRangeSet := mapset.NewSet[cciptypes.SeqNum]()
```
makramkd, line 638:
```suggestion
		snSet := mapset.NewSet[cciptypes.SeqNum]()
```

This is a set of seq nrs not ranges

execute/plugin.go
```diff
@@ -620,23 +626,36 @@ func (p *Plugin) checkAlreadyExecuted(
...
+		if _, exists := executed[sourceChainSelector]; !exists {
+			lggr.Debugw("no messages from source chain executed yet",
+				"sourceChain", sourceChainSelector)
+			return nil
+		}
+
+		executedSet := mapset.NewSet(executed[sourceChainSelector]...)
 		intersection := executedSet.Intersect(snRangeSet)
-		if intersection.Cardinality() != seqNrRange.Length() {
+		if intersection.Cardinality() != totalSeqNums {
```
makramkd, line 653: Can't you do `snSet.Length()` here and axe the extra var?

execute/plugin.go
0xAustinWang: renamed to `executableReports` so it's always accessed as `executableReports[selector]` rather than `executableReportsBySelector[selector]`

execute/plugin.go
```diff
@@ -154,6 +154,105 @@ func (p *Plugin) Query(ctx context.Context, outctx ocr3types.OutcomeContext) (ty
...
+	rangesBySelector := make(map[cciptypes.ChainSelector][]cciptypes.SeqNumRange)
+	executableReports := make(map[cciptypes.ChainSelector][]exectypes.CommitData)
+
+	for selector, reports := range groupedCommits {
+		if len(reports) == 0 {
+			continue
+		}
+
+		rangesBySelector[selector] = make([]cciptypes.SeqNumRange, 0)
+		executableReports[selector] = make([]exectypes.CommitData, 0)
```
makramkd, line 204: These lines shouldn't be necessary since we assign these keys right afterwards

execute/plugin.go
```diff
@@ -154,6 +154,105 @@ func (p *Plugin) Query(ctx context.Context, outctx ocr3types.OutcomeContext) (ty
...
+	for selector, reports := range executableReports {
+		unconfirmedMsgSet := mapset.NewSet(unconfirmedMessages[selector]...)
+		finalizedMsgSet := mapset.NewSet(finalizedMessages[selector]...)
+		// Get unfinalized messages by taking the difference
+		unfinalizedMessages := slicelib.ToSortedSlice(unconfirmedMsgSet.Difference(finalizedMsgSet))
+
+		sortedFinalizedMessages := slicelib.ToSortedSlice(finalizedMsgSet)
+
+		// Fully finalized roots are removed from the reports and set in groupedCommits
+		var executedCommitsFinalized []exectypes.CommitData
```
makramkd, line 242: Unnecessary declaration far as I can tell

makramkd (approved):

pkg/reader/ccip.go
```diff
@@ -397,25 +379,80 @@ func (r *ccipChainReader) ExecutedMessages(
...
 
-		executed = append(executed, stateChange.SequenceNumber)
+		executed[stateChange.SourceChainSelector] =
+			append(executed[stateChange.SourceChainSelector], stateChange.SequenceNumber)
 	}
 
 	return executed, nil
 }
 
+func createExecutedMessagesKeyFilter(
```
winder, line 402: I only wonder if this function will blow up CR at some point. Is there a limit to how many expressions there can be?

cc: @ilija42   @reductionista

winder (approved):

pkg/reader/ccip.go
ilija42: The filters are just parsed into sql, so I think as long as the sql query is not too long it should be fine? wdyt @reductionista

PR #721, opened 2025-03-17
[Solana][NONEVM-1296] RMN disabled placeholder

PabloMansanet:


chains/solana/contracts/programs/ccip-offramp/src/state.rs
```diff
@@ -91,6 +91,8 @@ pub struct ExternalExecutionConfig {}
 pub struct SourceChainConfig {
     pub is_enabled: bool, // Flag whether the source chain is enabled or not
 
+    pub is_rmn_verification_disabled: bool, // Currently a placeholder.
```
agusaldasoro, line 94: the default value of a not initialized field is true?

agusaldasoro (approved):

chains/solana/contracts/programs/ccip-offramp/src/state.rs
PabloMansanet: Do you mean in Rust in general?

No, the default would be false, but we don't implement or derive `Debug` on this struct so it shouldn't matter.

PabloMansanet: Ah sorry, I misunderstood your question I think. Yes, it will default to false, but it should be set at construction anyway so we should be OK.

PR #722, opened 2025-03-17
[Solana][NONEVM-1379] Rename SVM extra args to generic extra args

PabloMansanet:
https://smartcontract-it.atlassian.net/browse/NONEVM-1379


chains/solana/contracts/programs/fee-quoter/src/instructions/v1/messages.rs
```diff
@@ -116,7 +116,7 @@ pub fn process_extra_args(
             };
 
             let mut data = &extra_args[4..];
-            parse_and_validate_evm_extra_args(tag.try_into().unwrap(), &mut data)
+            parse_and_validate_generic_extra_args(tag.try_into().unwrap(), &mut data)
```
agusaldasoro, line 119: should this be under `CHAIN_FAMILY_SELECTOR_EVM`? Or should it be the default option and fail in case the parsing does?
If we need to add any other chain that's not EVM but it's compliant with the default extra args, then no change in code should be needed.

PabloMansanet: Good call, that does sound a bit more flexible.

chains/solana/contracts/programs/fee-quoter/src/instructions/v1/messages.rs
```diff
@@ -134,18 +123,30 @@ pub fn process_extra_args(
...
+            // Extra args are optional for non-SVM destinations. In case there
+            // are extra args, they must be prefixed by a four byte tag
+            // -> bytes4(keccak256("CCIP EVMExtraArgsV2"));
+            let Some(tag) = extra_args.get(..4) else {
+                return Ok(ProcessedExtraArgs::defaults(dest_config));
+            };
 
-        _ => Err(FeeQuoterError::InvalidChainFamilySelector.into()),
+            let mut data = &extra_args[4..];
+            parse_and_validate_generic_extra_args(tag.try_into().unwrap(), &mut data)
```
agusaldasoro, line 135: should we map this error and fail into a custom one?

PabloMansanet: I don't think we need to: we've already verified that the tag exists, so all that remains is expectations after having read that tag. The errors are still relevant to the caller (arguments couldn't be processed because the data after the tag was wrong).

agusaldasoro (approved):

PR #729, opened 2025-03-18
Improve vec capacity initialization

toblich:


PabloMansanet (approved):
Nice and simple, LGTM.

PR #730, opened 2025-03-18
New commit report option: MaxPricesPerReport

winder:
core ref: 6a481920beb027bdb083f182ac22df66747a3dc7

Refactor report building logic:
* Move to separate package.
* Delete "truncated report" option.
* Update initialization to support errors.

Implement new `ReportBuilderFunc` for splitting price updates into separate reports.


commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+
+	"github.com/smartcontractkit/chainlink-ccip/commit/committypes"
+	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot"
+	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+)
+
+// ReportBuilderFunc is used to inject different algorithms for building commit reports.
+type ReportBuilderFunc func(
```
winder, line 18: Now a public (internal) handle for exporting algorithms.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	ctx context.Context,
+	lggr logger.Logger,
+	reportCodec cciptypes.CommitPluginCodec,
+	transmissionSchedule *ocr3types.TransmissionSchedule,
+	outcome committypes.Outcome,
+	config pluginconfig.CommitOffchainConfig,
+) ([]ocr3types.ReportPlus[[]byte], error)
+
+// NewReportBuilder returns a ReportBuilderFunc based on the provided config.
+func NewReportBuilder(config pluginconfig.CommitOffchainConfig) (ReportBuilderFunc, error) {
```
winder, line 28: Builder selection algorithm moved from `NewPlugin` to here. Called by factory.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+		return nil, err
+	}
+	if report != nil {
+		return []ocr3types.ReportPlus[[]byte]{*report}, nil
+	}
+	return nil, nil
+}
+
+// buildMultiplePriceReports builds many reports of with at most maxMerkleRootsPerReport roots.
+func buildMultiplePriceAndMerkleRootReports(
```
winder, line 172: Entrypoint for new price splitting report builder.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+
+	// Default to the standard report.
+	return buildStandardReport, nil
+}
+
+// buildOneReport is the common logic for building a report. Different report building
+// algorithms can reassemble reports by selecting which blessed and unblessed merkle
+// roots to include in the report, and which price updates and rmn signatures to use.
+func buildOneReport(
```
winder, line 58: Unmodified, moved from commit package.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	return &ocr3types.ReportPlus[[]byte]{
+		ReportWithInfo: ocr3types.ReportWithInfo[[]byte]{
+			Report: encodedReport,
+			Info:   encodedInfo,
+		},
+		TransmissionScheduleOverride: transmissionSchedule,
+	}, nil
+}
+
+func buildStandardReport(
```
winder, line 126: Unmodified, moved from commit package.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	// Add merkle root reports to price reports.
+	reports = append(reports, rootReports...)
+
+	return reports, nil
+}
+
+// buildMultipleMerkleRootReports builds many reports of with at most maxMerkleRootsPerReport roots.
+// Any price reports in the outcome are included in the first merkle root.
+func buildMultipleMerkleRootReports(
```
winder, line 210: Unmodified, moved from commit package.

commit/internal/builder/report_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/smartcontractkit/chainlink-ccip/commit/committypes"
+	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot"
+	rmntypes "github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn/types"
+	"github.com/smartcontractkit/chainlink-ccip/commit/tokenprice"
+	"github.com/smartcontractkit/chainlink-ccip/internal/mocks"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
+)
+
+func TestReportBuilders(t *testing.T) {
```
winder, line 23: Moved from commit package. Updated with new testcase.

pluginconfig/commit.go
```diff
@@ -304,15 +310,22 @@ func (c *CommitOffchainConfig) Validate() error {
...
 		if c.MaxMerkleRootsPerReport != 0 {
-			errs = append(errs, fmt.Errorf("maxMerkleRootsPerReport is set with RMN enabled"))
-		}
-		if len(errs) > 0 {
-			return errors.Join(errs...)
+			errs = append(errs, fmt.Errorf("maxMerkleRootsPerReport does not support RMN, RMNEnabled cannot be true"))
 		}
 	}
+	if c.MaxMerkleRootsPerReport != 0 && !c.MultipleReportsEnabled {
+		errs = append(errs, fmt.Errorf("maxMerkleRootsPerReport cannot be used without MultipleReportsEnabled"))
```
winder, line 320: Previously this was the "truncated report" case.

pluginconfig/commit.go
```diff
@@ -157,6 +157,12 @@ type CommitOffchainConfig struct {
...
 
+	// MaxPricesPerReport is the maximum number of prices to include in a single report. When this option is set,
+	// the commit plugin will split the prices across multiple reports if there are more than this number of prices.
+	// It will also split prices and merkle roots into separate reports.
+	// NOTE: this can only be used if RMNEnabled == false.
+	MaxPricesPerReport uint64 `json:"maxPricesPerReport"`
+
 	// MultipleReportsEnabled is a flag to enable/disable multiple reports per round.
 	// This is typically set to true on chains that use 'MaxMerkleRootsPerReport'
 	// in order to avoid delays when there are reports from multiple sources.
```
makramkd, line 168: Is there no cap to the max # of reports generated in a round? e.g imagine 100 source chains, so we have like 100 merkle roots and 75 price reports, would we generate 100 reports for that round?

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+			updates.TokenPriceUpdates = append(updates.TokenPriceUpdates, update.TokenPrice)
+		}
+		if (update.GasPriceChain != cciptypes.GasPriceChain{}) {
+			updates.GasPriceUpdates = append(updates.GasPriceUpdates, update.GasPriceChain)
+		}
+		return updates
+	}
+
+	var updates []update
+	for _, tokenPriceUpdate := range outcome.TokenPriceOutcome.TokenPrices.ToSortedSlice() {
```
makramkd, line 326: Any significance here to the sort?

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+				transmissionSchedule,
+				outcome.MerkleRootOutcome.OutcomeType,
+				nil,
+				nil,
+				nil, // no RMN for partial reports.
+				0,   // no RMN for partial reports.
+				priceUpdates,
+			)
+			if err != nil {
+				return nil, err
```
makramkd, line 359: Would more ctx added to the error be useful?

```suggestion
				return nil, fmt.Errorf("building report with only price updates %+v: %w", priceUpdates, err)
```

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+			}
+			if report != nil {
+				reports = append(reports, *report)
+			}
+
```
makramkd, line 364: So at this point `report` could be `nil` is that something that should be logged? Feels like an invariant violation since `priceUpdates` must be non-empty?

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+			nil, // no RMN for partial reports.
+			0,   // no RMN for partial reports.
+			priceUpdates,
+		)
+		if err != nil {
+			return nil, err
+		}
+		if report != nil {
+			reports = append(reports, *report)
+		}
```
makramkd, line 392: Ditto here maybe a check that `report == nil` and return an error since its an invariant violation (postcondition of buildOneReport should be that report is non-nil if price reports are provided, which in this case they are)

pluginconfig/commit.go
```diff
@@ -157,6 +157,12 @@ type CommitOffchainConfig struct {
 	// NOTE: this can only be used if RMNEnabled == false.
 	MaxMerkleRootsPerReport uint64 `json:"maxRootsPerReport"`
 
+	// MaxPricesPerReport is the maximum number of prices to include in a single report. When this option is set,
+	// the commit plugin will split the prices across multiple reports if there are more than this number of prices.
+	// It will also split prices and merkle roots into separate reports.
+	// NOTE: this can only be used if RMNEnabled == false.
```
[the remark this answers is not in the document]
winder: It seemed unlikely that we would ever set this without also setting MaxMerkleRootsPerReport, so the RMN constraint is carried through.

pluginconfig/commit.go
winder: Yes, in that case up to 175 reports could be produced in 1 round.

commit/internal/builder/report.go
winder: No, this is just the existing helper function to simplify the code.

pluginconfig/commit.go
makramkd: Up to 175, thats if we don't combine MR's with price updates. What flag allows / disallows that?

makramkd: At least my understanding was if MaxMerkleRoots == 1 and MaxPriceReports == 1 then you could have a report w/ 1 MR and 1 price report

commit/internal/builder/report.go
winder: I probably added this check because of the dereference, maybe the linter complained. I can add a log.

commit/internal/builder/report.go
winder: I don't think we want to return an error. If other reports were generated they should still be submitted.

pluginconfig/commit.go
winder: No. If you set `MaxPriceReports > 0` then merkle root reports will be separated from the price reports. I'll update the MaxPriceReports comment to clarify the behavior.

makramkd: I see, yeah this info would be good to have.

commit/internal/builder/report.go
winder: It looks like this behavior is a carry-over from when this code was called from the OCR `Reports` function.

`buildOneReport` has special handling to check for `outcomeType == EmptyReport` and a check for `report.IsEmpty()`. If you want no-op reports to be an error we should change this behavior so that `buildOneReport` returns an error and we avoid calling it when there's no need.

Personally, I don't mind the condition-less logic:
1) pass all available data to the build function.
     a) if there's no data, there's no report.
     b) if there is data, there is one or more report.
2) Submit everything that was generated. If there was a problem with some of the data, `buildOneReport` should have returned an error.

makramkd: I don't think its no-op reports that should be an error, at this point in the code we have `priceUpdates` of nonzero length, so reports _must not_ be `nil`. All I'm saying is that if you submit a nonzero length price updates struct to `buildOneReport` and it returns `nil` this is a bug in `buildOneReport`. At this point it'd just be a sanity check that `report != nil` is the only case (i.e if the `if` condition on line 390 is not triggered, AND there is no error, there's something fatally wrong)

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+
+	// MerkleRoots and RMNSignatures will be empty arrays if there is nothing to report
+	rep = cciptypes.CommitPluginReport{
+		BlessedMerkleRoots:   blessedMerkleRoots,
+		UnblessedMerkleRoots: unblessedMerkleRoots,
+		PriceUpdates:         priceUpdates,
+		RMNSignatures:        rmnSignatures,
+	}
+
+	if outcomeType == merkleroot.ReportEmpty {
```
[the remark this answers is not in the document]
winder: Got 'em :+1:

commit/factory.go
```diff
@@ -216,6 +217,11 @@ func (p *PluginFactory) NewReportingPlugin(ctx context.Context, config ocr3types
 		return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to create metrics reporter: %w", err)
 	}
 
+	reportBuilder, err := builder.NewReportBuilder(offchainConfig)
```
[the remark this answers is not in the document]
winder: Updated signature and added tests.

Re: the variable name. The type of the variable includes `Func`, so adding it to the name as well seems redundant.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+
+	// Currently RMN is only supported for standard reports. Because the config is validated,
+	// we can assume that if RMN is enabled, we are building a standard report.
+	if config.RMNEnabled {
+		return buildStandardReport, nil
+	}
+
+	// MaxPricesPerReport includes MaxMerkleRootsPerReport, so check it first.
+	if config.MaxPricesPerReport != 0 {
+		return buildMultiplePriceAndMerkleRootReports, nil
```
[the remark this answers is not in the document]
winder: Correct. These values are also passed to the report building logic. When `MaxPricesPerReport` is non-zero the handler knows what to do if `MaxMerkleRootsPerReport` is zero.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("problem building price reports: %w", err)
+	}
+	outcome.TokenPriceOutcome.TokenPrices = make(cciptypes.TokenPriceMap)
+	outcome.ChainFeeOutcome.GasPrices = nil
+
+	// 2. Select algorithm for merkle root reports, and build merkle root report(s).
+	var rootReportBuilder ReportBuilderFunc
+	if config.MaxMerkleRootsPerReport == 0 {
+		rootReportBuilder = buildStandardReport
```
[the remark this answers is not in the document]
winder: Which part is confusing? Reports are based on data inside the "outcome" object, in this case I'm composing a few functions depending on how everything is configured.

It might be simpler if we had some sort of intermediate object rather than `Outcome` to make it clear that we're splitting things up based on `merkleroot.Outcome` and `tokenprice.Outcome`.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	maxPricesPerReport uint64, // passed in directly to avoid implementing ReportBuilderFunc
+) ([]ocr3types.ReportPlus[[]byte], error) {
+	// update joins together the different types of price updates so that they can be
+	// selected in one loop.
+	type update struct {
+		cciptypes.TokenPrice
+		cciptypes.GasPriceChain
+	}
+	// Helper to add an update to the PriceUpdates type.
+	add := func(updates cciptypes.PriceUpdates, update update) cciptypes.PriceUpdates {
```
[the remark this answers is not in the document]
winder: I made sure all of these helper types were private to the function.
Got rid of the add function.

commit/internal/builder/report.go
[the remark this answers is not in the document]
winder: Added tests.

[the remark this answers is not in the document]
winder: done.

commit/report.go
```diff
@@ -24,241 +23,44 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
 )
 
-// reportBuilderFunc is used to inject different algorithms for building commit reports.
-type reportBuilderFunc func(
+func encodeReports(
```
winder, line 26: Use split view for this function. `encodedReports` and `reportBuilderFunc` are totally different. The report encoding logic was split out from the builder and now lives in this function.

TODO: unit test for this function.

commit/report.go
```diff
@@ -290,17 +92,45 @@ func (p *Plugin) Reports(
...
+		dataPoints += len(outcome.ChainFeeOutcome.GasPrices)
+		dataPoints += len(outcome.MerkleRootOutcome.RMNReportSignatures)
+		if dataPoints > 0 {
+			lggr.Errorw("no reports generated despite having data",
+				"roots", outcome.MerkleRootOutcome.RootsToReport,
+				"tokenPriceUpdates", outcome.TokenPriceOutcome.TokenPrices,
+				"gasPriceUpdates", outcome.ChainFeeOutcome.GasPrices,
+				"rmnSignatures", outcome.MerkleRootOutcome.RMNReportSignatures,
+			)
+			return nil, fmt.Errorf("no reports generated despite having data")
```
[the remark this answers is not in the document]
winder: Good point. The builder behavior is now opaque to this function and updated the builder to return an error if there are unexpected empty reports.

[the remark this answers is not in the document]
winder: This is resolved now. I added stricter error checking to the prices report (the only place where an empty report is an error) and removed this catch-all check.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	outcome.TokenPriceOutcome.TokenPrices = make(cciptypes.TokenPriceMap)
+	outcome.ChainFeeOutcome.GasPrices = nil
+
+	// 2. Select algorithm for merkle root reports, and build merkle root report(s).
+	var rootReportBuilder ReportBuilderFunc
+	if config.MaxMerkleRootsPerReport == 0 {
+		rootReportBuilder = buildStandardReport
+	} else {
+		rootReportBuilder = buildMultipleMerkleRootReports
```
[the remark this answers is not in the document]
winder: Test added.

commit/internal/builder/report.go
```diff
@@ -0,0 +1,396 @@
...
+			unblessedMerkleRoots = make([]cciptypes.MerkleRootChain, 0)
+
+			// price updates are only included in the first report.
+			priceUpdates = cciptypes.PriceUpdates{}
+		}
+	}
+
+	// check for final partial report.
+	if numRoots > 0 {
+		report, err := buildOneReport(
```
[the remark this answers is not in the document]
winder: Test added.

PR #735, opened 2025-03-19
NONEVM-1415: create a new release on each push to `main`

chris-de-leon-cll:
On each push to `main`, a new release will automatically be created with the latest Solana program artifacts


archseer:
We should probably have some way of expiring old tags or this will quickly add up over time

chris-de-leon-cll:
Yup good call - I've created a ticket [here](https://smartcontract-it.atlassian.net/browse/NONEVM-1435) so we can follow up on it later

jlaveracll (approved):

PR #736, opened 2025-03-19
test keys for dry run

tt-cll:


tt-cll:
https://github.com/smartcontractkit/chainlink-ccip/actions/runs/14055240353/job/39353160110

did not rebuild artifacts when it should have, causing key mismatch

PR #739, opened 2025-03-19
[CCIP-5450] Deprecate FeeInfo in favor of ChainConfig in ChainFee processor

asoliman92:
So far in ChainFeeProcessor we've been using [FeeInfo](https://github.com/smartcontractkit/chainlink-ccip/blob/main/pluginconfig/commit.go#L35-L41).  It makes sense to use [ChainConfig](https://github.com/smartcontractkit/chainlink-ccip/blob/main/chainconfig/chainconfig.go#L15-L25) as it will be more dynamic. Checking the test setups in Chainlink, all FeeInfo related code seems to be removed and we're using [CCIPHome](https://github.com/smartcontractkit/chainlink/blob/3822907f158219b2b87cb3442d1360984c4c3b9c/deployment/ccip/changeset/testhelpers/test_environment.go#L711-L719) config. 

This PR fixes this by removing FeeInfo completely from the code an relying on ChainConfig

### DIFF
--- a/commit/chainfee/outcome_test.go
+++ b/commit/chainfee/outcome_test.go
@@ -5,6 +5,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/mock"
+
+	"github.com/smartcontractkit/chainlink-ccip/internal/libs/mathslib"
+
+	mapset "github.com/deckarep/golang-set/v2"
+	libocrtypes "github.com/smartcontractkit/libocr/ragep2p/types"
+
+	"github.com/smartcontractkit/chainlink-ccip/chainconfig"
+	mock_home_chain "github.com/smartcontractkit/chainlink-ccip/mocks/internal_/reader"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/reader"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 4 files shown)


pluginconfig/commit.go
```diff
@@ -22,7 +22,6 @@ const (
 	defaultNewMsgScanBatchSize                = merklemulti.MaxNumberTreeLeaves
 	defaultEvmDefaultMaxMerkleTreeSize        = merklemulti.MaxNumberTreeLeaves
 	defaultMaxReportTransmissionCheckAttempts = 5
-	defaultRMNEnabled                         = false
```
asoliman92, line 25: unused variable

commit/chainfee/outcome_test.go
```diff
@@ -70,6 +139,20 @@ func sameObs(n int, obs Observation) []plugincommon.AttributedObservation[Observ
 	return aos
 }
 
+type FeeInfo struct {
```
0xnogo, line 142: Why this is still needed? I thought we wanted to deprecated this struct?

asoliman92: Deprecated it, this is only for the test so that I don't change the test setup.

0xnogo (approved):

commit/chainfee/outcome_test.go
ogtownsend: @asoliman92 how involved is changing the test setup? wondering what the tradeoff is between leaving this `FeeInfo` struct in here and potentially confusing people who work on the test vs removing it

PR #744, opened 2025-03-19
Quicker test validator & test funding retries

toblich:
Main changes:
- Set `--ticks-per-slots=8` in local test validator, to speed up test execution
- Adds retries on funding steps, which were prone to failures in CI due to missing transactions

Related smaller changes:
- Fix curse tests in tokenpools suite so that they don't mess with curse state, which can cause problems with concurrent tests
- Add more specific test to a few mcm tests that were reusing the same title, just to have more specific messages in case of failures

You can see [here](https://github.com/smartcontractkit/chainlink-ccip/actions/runs/13980021849) that the `go tests` CI step has now worked at least 10 times in a row: 
image


jadepark-dev (approved):
🚀

PR #747, opened 2025-03-20
[Solana][NONEVM-1436] Remove unused version fields

PabloMansanet:
https://smartcontract-it.atlassian.net/browse/NONEVM-1436

I decided to remove the `version` fields rather than add checks because I don't think there's a need for versioning these, but would like a sanity check on that.


archseer:
There's no need until it becomes necessary :) I'd rather pre-emptively have these included and unchecked

toblich:
Agreed with Blaz, I'd rather keep the version field around, it makes future account schema migrations way easier. Even if we're not checking it now, we can begin checking it whenever we actually start working with different versions.

PabloMansanet:
Gotcha. Closing this, will let the auditors know it's a no-fix.

PR #748, opened 2025-03-20
[Solana][NONEVM-1437] Add missing attributes to token context struct

PabloMansanet:
https://smartcontract-it.atlassian.net/browse/NONEVM-1437


chains/solana/contracts/programs/ccip-router/src/token_context.rs
```diff
@@ -105,8 +105,10 @@ pub struct OverridePendingTokenAdminRegistryByOwner<'info> {
     )]
     pub config: Account<'info, Config>,
     #[account(
+        mut,
```
toblich, line 108: Wow, the fact that the `mut` was missing means that the method probably wasn't working at all, as Anchor won't persist the change. Could you please add a quick integration test that executes this method and checks the state after it?
It looks like we only have [this test](https://github.com/smartcontractkit/chainlink-ccip/blob/28994272f7ef0d58d70e723ab59c186355e05e6d/chains/solana/contracts/tests/ccip/ccip_router_test.go#L2675) using this method, but it is a test where the tx fails so no state checks are made.

toblich (approved):
LGTM! If you can please add a test, it would be much appreciated :)
Anyway, the change is fine to merge if you're in a hurry and can't get to the test

chains/solana/contracts/programs/ccip-router/src/token_context.rs
PabloMansanet: Added an integration test for this, give it a look :)

toblich (approved):

PR #724, opened 2025-03-17
solana: refactor token accounts validation in Anchor context

jadepark-dev:
This PR refactors the token account validation logic in CCIP Solana programs by leveraging Anchor's built-in account validation system.

## Primary Changes
- Created `TokenAccountsValidationContext` account struct in both offramp and router programs
- Replaced manual validation checks with Anchor's declarative constraints
- Consolidated PDA derivation and ownership validations in the account context struct
- Standardized error handling across programs

## Others
- Removed redundant error codes (`InvalidInputsConfigAccounts`)
- Created `AnchorError` enum in common utils to standardize error string handling in tests
- Updated tests to reference the new standardized error codes
- Simplified account iteration with `next_account_info` instead of manual splitting

- [ ] Dependency: https://github.com/smartcontractkit/chainlink-ccip/pull/750

PR #750, opened 2025-03-20
[Solana][NONEVM-1452] Multiple deduplications / refactor

PabloMansanet:
Addresses two main points:
* Duplication of token pool validation logic accross router and offramp.
* Duplication of seed strings (not covered directly by the audit, but a direct consequence of attempting to solve the above).


chains/solana/contracts/crates/common/src/v1.rs
```diff
@@ -0,0 +1,318 @@
...
+    pub pool_config: &'a AccountInfo<'a>,
+    pub pool_token_account: &'a AccountInfo<'a>,
+    pub pool_signer: &'a AccountInfo<'a>,
+    pub token_program: &'a AccountInfo<'a>,
+    pub mint: &'a AccountInfo<'a>,
+    pub fee_token_config: &'a AccountInfo<'a>,
+    pub remaining_accounts: &'a [AccountInfo<'a>],
+}
+
+pub fn validate_and_parse_token_accounts<'info>(
```
jadepark-dev, line 21: Leaving backlog - I'll probably rewrite the refactor PR https://github.com/smartcontractkit/chainlink-ccip/pull/724 once this PR is merged.

chains/solana/contracts/crates/common/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "common"
```
jadepark-dev, line 2: I would suggest `ccip-common` or more specific name per scope? since we have mcms programs together

PabloMansanet: Good shout. Will do that!

chains/solana/contracts/crates/common/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub mod seed;
+pub mod v1;
+
+#[error_code]
+pub enum CommonCcipError {
+    #[msg("The given sequence interval is invalid")]
+    // offset error code so that they don't clash with other programs
+    // (Anchor's base custom error code 6000 + offset 3000 = start at 9000)
+    InvalidSequenceInterval = 4000,
```
toblich, line 13: Please update the comment numbers

chains/solana/contracts/crates/common/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+    InvalidInputsTokenAdminRegistryAccounts,
+    #[msg("Invalid LookupTable account")]
+    InvalidInputsLookupTableAccounts,
+    #[msg("Invalid LookupTable account writable access")]
+    InvalidInputsLookupTableAccountWritable,
+}
+
+// Used only to enable the `account` attribute to enable account deserialization, for accounts
+// owned by the router (see TokenAdminRegistry below)
+declare_id!("Ccip842gzYHhvdDkSyi2YVCoAWPbYJoApMFzSxQroE9C");
```
toblich, line 30: I think this is a problem, unless what tooling does accounts for this. We will use different addresses in different environments, so tooling has to replace the id where needed when they compile for each env. Anchor may use the value set here for ownership checks, so it is important that each time this is compiled, the right value is specified there.

It might just be the case that the way tooling is doing the compilation already matches this and there's no further work required, but please do check with them.

chains/solana/contracts/crates/common/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub mod accounts_for_deserialization {
+    /// This mod holds structs that represent accounts owned by other programs,
+    /// but are used here just to deserialize data from them. Must be kept in sync
+    /// with their originals.
```
toblich, line 34: Wait, if they still must be kept in sync in multiple places, does it really make sense to centralize this here?

chains/solana/contracts/crates/common/src/v1.rs
toblich: I love that this gets deduplicated, even more if #724 gets merged as well (after rebasing/rewriting) 🫶

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -1,8 +1,10 @@
 use anchor_lang::prelude::*;
+use common::seed;
+use common::v1::{validate_and_parse_token_accounts, TokenAccounts};
```
toblich, line 3: I guess that if we ever release a new version of the common, it should be under `common:v2` leaving `common:v1` unchanged, and then we'll just have to create a new version of the offramp (and other programs) with the new import here, right? Sounds good to me, just double-checking.

chains/solana/contracts/crates/common/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0-dev"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib", "lib"]
+name = "common"
```
toblich, line 8: I like the approach of extracting some things into a common crate. Please check though how much this impacts final binary sizes, as we may have to notify tooling if they are growing considerably (as that affects the account size and thus the required rent for deploying the programs).

chains/solana/contracts/crates/common/src/lib.rs
PabloMansanet: Good shout, I asked that question [here](https://chainlink-core.slack.com/archives/C084CRQEQ5B/p1742547398330069). I do wonder if, aside from this, we can make their life easier with a little bit of `build.rs` magic (would be muc preferable to string substitution on the source if that's what's happening.)

chains/solana/contracts/crates/common/src/lib.rs
PabloMansanet: We're only centralizing this struct out of necessity, because it's referenced by TokenAccounts. However, since we only have this problem with a struct in `ccip_router`, we can get away with defining the "canonical" one here and losing the original. This will work fine as long as we don't have to centralize structs from other crates in here, because at that point we'd run into the need to have different top level `ID`.

chains/solana/contracts/crates/common/src/lib.rs
PabloMansanet: I had a sync conversation with tooling and my understanding after it is that this wouldn't be a problem: At worst there will be minor modifications needed on their end. 

Unrelated to this PR there may be opportunities for us to help alleviate some of the manual work when updating keys manually, but that's outside this scope.

chains/solana/contracts/crates/common/src/lib.rs
PabloMansanet: Alright, I was able to deduplicate this by moving some things around and centralizing the definition in `ccip_common`.Take a look at the new arrangement, I think it's better than before and reflects the approach we took with `base_token_pool` as Agustina suggested.

chains/solana/contracts/crates/common/Cargo.toml
PabloMansanet: I ran `du -h target/sbf-solana-solana/release/*.so` on both branches.

Main:

```
344K	target/sbf-solana-solana/release/access_controller.so
308K	target/sbf-solana-solana/release/base_token_pool.so
740K	target/sbf-solana-solana/release/burnmint_token_pool.so
1.1M	        target/sbf-solana-solana/release/ccip_offramp.so
1.1M	        target/sbf-solana-solana/release/ccip_router.so
492K	target/sbf-solana-solana/release/example_ccip_receiver.so
740K	target/sbf-solana-solana/release/example_ccip_sender.so
368K	target/sbf-solana-solana/release/external_program_cpi_stub.so
880K	target/sbf-solana-solana/release/fee_quoter.so
760K	target/sbf-solana-solana/release/lockrelease_token_pool.so
780K	target/sbf-solana-solana/release/mcm.so
536K	target/sbf-solana-solana/release/rmn_remote.so
480K	target/sbf-solana-solana/release/test_ccip_invalid_receiver.so
480K	target/sbf-solana-solana/release/test_ccip_receiver.so
772K	target/sbf-solana-solana/release/test_token_pool.so
812K	target/sbf-solana-solana/release/timelock.so
```

This PR branch:
```
344K	target/sbf-solana-solana/release/access_controller.so
308K	target/sbf-solana-solana/release/base_token_pool.so
740K	target/sbf-solana-solana/release/burnmint_token_pool.so
8.0K	        target/sbf-solana-solana/release/ccip_common.so
1.1M	        target/sbf-solana-solana/release/ccip_offramp.so
1.1M	        target/sbf-solana-solana/release/ccip_router.so
492K	target/sbf-solana-solana/release/example_ccip_receiver.so
740K	target/sbf-solana-solana/release/example_ccip_sender.so
368K	target/sbf-solana-solana/release/external_program_cpi_stub.so
880K	target/sbf-solana-solana/release/fee_quoter.so
760K	target/sbf-solana-solana/release/lockrelease_token_pool.so
780K	target/sbf-solana-solana/release/mcm.so
536K	target/sbf-solana-solana/release/rmn_remote.so
480K	target/sbf-solana-solana/release/test_ccip_invalid_receiver.so
480K	target/sbf-solana-solana/release/test_ccip_receiver.so
772K	target/sbf-solana-solana/release/test_token_pool.so
812K	target/sbf-solana-solana/release/timelock.so
```

No appreciable differences. There's the extra 8k of `ccip_common`, but that's not deployable, and the definitions in that crate are statically linked to the others, so there's no change in size at all.

toblich (approved):

PR #751, opened 2025-03-20
Introduce ConfigPollerSyncDuration

0xnogo:
* In both Exec and Commit configs
* Default is set to 30 sec


mateusz-sekara (commented):
Is there a real need to make it faster/slower on some chains? If there is no strong use case, I'm worried that we might add a new configuration param that will never be changed

0xnogo:
@mateusz-sekara Treating all chains equally for a caching seems off. I don't think we have explored much this frequency. 30s seemed to be like a safe choice but we could potentially be more aggressive on slow chains.

PR #752, opened 2025-03-20
Added SVM call args types

silaslenihan:


chains/solana/types.go
```diff
@@ -0,0 +1,27 @@
+package solana
+
```
agusaldasoro, line 2: Could you add a comment here on why we need this/who is using? Thanks!

agusaldasoro (dismissed):

chains/solana/types.go
silaslenihan: done!

agusaldasoro (approved):

PR #757, opened 2025-03-21
Make router::get_fee return fewer fields that FQ

toblich:
The router's `get_fee` method was added in #712 as just a proxy to fee quoter's method.
However, from the end-user PoV, not all fields returned there are valuable. So, make the router extract only the relevant fields.

Also, I've made the example sender use the router to get the fee instead of fee quoter, since we've added the method there and Product's take is that users should integrate with the router and not fee quoter (which I don't fully agree with, but I can live with it anyway)


PabloMansanet (approved):
LGTM with only one comment on the struct naming. Good job on the typo hunting too :)

PR #762, opened 2025-03-25
Commit report PDA refund

matYang:
Add a permissionless function that checks if a CommitReport has been fully executed, if so, it closes the CommitReport PDA, and sends the rent to OnRamp's fee token receiver

native Sol is being sent because Solana Runtime makes it difficult to withdraw sol from PDA, send it and call SyncNative all in 1 tx: https://github.com/solana-labs/solana/issues/9711

SyncNative will occur when someone calls ccipSend, it can also be called at withdrawal time.


chains/solana/contracts/programs/ccip-offramp/src/context.rs
```diff
@@ -595,6 +597,47 @@ pub struct ExecuteReportContext<'info> {
...
+    #[account(
+        seeds = [seed::FEE_BILLING_SIGNER],
+        bump,
+        seeds::program = reference_addresses.load()?.router,
+    )]
+    pub fee_billing_signer: UncheckedAccount<'info>,
+
+    pub token_program: Interface<'info, TokenInterface>,
+
+    pub system_program: Program<'info, System>,
```
PabloMansanet, line 638: Is this needed? It's not necessary for closing accounts IIRC.

PabloMansanet (commented):
Seems good on the Solana end. I can't comment on the logic for checking `execution_states` but I trust you on that one being OK. Just have a couple questions.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
```diff
@@ -230,6 +231,65 @@ impl Commit for Impl {
...
+    }
+}
+
+// Helper function to check if all messages have been executed
+fn all_messages_executed(report: &CommitReport) -> bool {
+    let num_messages = report.max_msg_nr.saturating_sub(report.min_msg_nr) + 1;
+
+    // execution_states follow geometric series 2^0 + 2^2 + 2^4 + ... + 2^(2 * (num_messages - 1))
+    // sum is calculated as (4^num_messages - 1) / 3
+    let fully_executed = (4u128.pow(num_messages as u32) - 1) / 3;
```
toblich, line 266: My understanding is that the `Success` execution state has a value of 2, so that e.g. a commit with 3 messages all successfully executed would be `0b00000000...0000101010`, right?

If the above holds, then geometric series would be `2^1 + 2^3 + 2^5 + 2^(2 * num_messages - 1)`, and the sum formula would need to be updated

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
```diff
@@ -230,6 +231,65 @@ impl Commit for Impl {
...
+    token_program: &Interface<'info, TokenInterface>,
+    commit_report: AccountInfo<'info>,
+    fee_token_receiver: &InterfaceAccount<'info, TokenAccount>,
+) -> Result<()> {
+    // Get lamports before closing
+    let lamports = commit_report.lamports();
+
+    // Close the account by setting its data length to 0
+    **commit_report.try_borrow_mut_lamports()? = 0;
+    commit_report.realloc(0, false)?;
```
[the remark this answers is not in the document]
toblich: +1, just removing the balance is should be enough, Solana will take care of the rest

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
```diff
@@ -230,6 +231,65 @@ impl Commit for Impl {
...
+
+    // Transfer wSOL to OnRamp fee_token_receiver
+    **fee_token_receiver
+        .to_account_info()
+        .try_borrow_mut_lamports()? += lamports;
```
toblich, line 286: Btw, a more readable way of doing this (which might use a bit more CUs though, not sure) is to invoke the `transfer` instruction, see https://github.com/smartcontractkit/chainlink-ccip/blob/dc0ed911405b8bf8523e767579dfd656841457b6/chains/solana/contracts/programs/ccip-router/src/instructions/v1/fees.rs#L70-L73 for an example.

The way you did it is also ok by me, it's not a required change, just a suggestion for you to choose which approach you prefer

toblich (commented):
Looking good! A few comments

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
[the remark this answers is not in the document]
matYang: Ack

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
matYang: this pattern seems canonical in examples of closing PDAs in function body, will stick with it for now

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
matYang: yeah good catch! sum formula *2 will do

chains/solana/contracts/programs/ccip-offramp/src/instructions/interfaces.rs
```diff
@@ -29,9 +29,16 @@ pub trait Commit {
         ss: Vec<[u8; 32]>,
         raw_vs: [u8; 32],
     ) -> Result<()>;
+
+    fn close_commit_report_account(
```
archseer, line 33: Can't this just be part of execute?

matYang: this was discussed, team raised the point that this would introduce a global write lock on the wsol receiver that collects the funds for all message executions, something we try to avoid to be in line with Solana's best practice for parallelism

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
```diff
@@ -315,3 +384,42 @@ mod helpers {
...
+            chain_selector: 0,
+            merkle_root: [0; 32],
+            timestamp: 0,
+            min_msg_nr: 42,
+            max_msg_nr: 42,
+            execution_states: 0,
+        };
+        assert!(
+            !all_messages_executed(&report),
+            "Message should still be paneding"
```
PabloMansanet, line 405:
```suggestion
            "Message should still be pending"
```

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
```diff
@@ -315,3 +384,42 @@ mod helpers {
...
+        );
+        report.execution_states = 3;
+        assert!(
+            !all_messages_executed(&report),
+            "Failed message does not acount as executed"
```
PabloMansanet, line 410:
```suggestion
            "Failed message does not count as executed"
```

PabloMansanet (approved):
LGTM. Just a couple typos.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/commit.rs
```diff
@@ -230,6 +231,74 @@ impl Commit for Impl {
...
+// Helper function to convert the SOL from the closed account to wrapped SOL
+fn transfer_and_wrap_native_sol<'info>(
+    token_program: &Program<'info, Token>,
+    commit_report: AccountInfo<'info>,
+    fee_token_receiver: &Account<'info, TokenAccount>,
+) -> Result<()> {
+    // Get lamports before closing
+    let lamports = commit_report.lamports();
+
+    // Close the account by setting its data length to 0
```
toblich, line 289: This comment seems outdated, you're closing it by just removing its rent-exempt balance

toblich (commented):
LGTM! Just one more thing before merging, could you please add 1 integration test that invokes your new method and then checks the following?
- The commit account is closed
- The WSOL receiver balance has increased by the right amount
- The event was emitted

toblich (approved):

PR #767, opened 2025-03-26
[Solana][NONEVM-1360] Restrict token pool instantiation to upgrade authority

PabloMansanet:
I've leveraged the upgrade authority to authorize the entity capable of instantiating token pools, in an approach similar to the one used in `rmn_remote`. This seems simpler and cleaner than adding another PDA with an explicit authority set on instantiation.

There's a minor drawback I suppose, which is that if ever a user wants to deploy their own immutable variant of our contract templates, they won't have an upgrade authority. But in such cases, they can just modify their code and add the PDA as necessary.

### DIFF
--- a/chains/solana/contracts/programs/burnmint-token-pool/src/context.rs
+++ b/chains/solana/contracts/programs/burnmint-token-pool/src/context.rs
@@ -6,7 +6,7 @@ use anchor_spl::{
 use base_token_pool::common::*;
 use ccip_common::seed;
 
-use crate::{ChainConfig, State};
+use crate::{program::BurnmintTokenPool, ChainConfig, State};
 
 #[derive(Accounts)]
 pub struct InitializeTokenPool<'info> {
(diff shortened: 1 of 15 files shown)


chains/solana/contracts/programs/burnmint-token-pool/src/context.rs
```diff
@@ -19,9 +19,18 @@ pub struct InitializeTokenPool<'info> {
...
+    pub authority: Signer<'info>,
     pub system_program: Program<'info, System>,
+
+    #[account(constraint = program.programdata_address()? == Some(program_data.key()))]
+    pub program: Program<'info, BurnmintTokenPool>,
+
+    // Token pool initialization only allowed by program upgrade authority. Initializing token pools managed
+    // by the CLL deployment of this program is limited to CLL. Users must deploy their own instance of this program.
+    #[account(constraint = program_data.upgrade_authority_address == Some(authority.key()) @ CcipTokenPoolError::Unauthorized)]
+    pub program_data: Account<'info, ProgramData>,
```
toblich, line 33: This change will likely affect tooling (who deploy and initialize pools) and perhaps e2e (assuming they might be setting up some other pool), please make sure to sync with them before merging so that they make whatever tiny changes they need to do and we don't break e2e

toblich (approved):
LGTM

PR #773, opened 2025-03-27
commit/merkleroot: fix report transmission checks

makramkd:
There seems to have been two issues with the report transmission checks:
* We didn't set RootsToReport correctly in the Outcome where we were
checking for the report transmission. This caused the next round to believe
that the report was successfully transmitted.
* We don't check for ReportInFlight in the Reports function. I believe
this would cause us to include the same merkle roots in the report again
even though they may have been transmitted successfully.

This commit fixes both of these issues.


commit/merkleroot/outcome.go
```diff
@@ -363,6 +363,9 @@ func checkForReportTransmission(
 		OutcomeType:                     ReportInFlight,
 		OffRampNextSeqNums:              previousOutcome.OffRampNextSeqNums,
 		ReportTransmissionCheckAttempts: previousOutcome.ReportTransmissionCheckAttempts + 1,
+		// Carry over the previous roots since they're still in-flight.
+		// We won't re-report since outcome type is ReportInFlight.
+		RootsToReport: previousOutcome.RootsToReport,
```
winder, line 368: may as well group this variable & comment /w `OffRampNextSeqNums` since they're both carried over for the same reason

makramkd: @dimkouv can you share your thoughts? I don't really see advantages / disadvantages either way.

PR #778, opened 2025-03-28
[CCIP-5576] pkg/reader: fix ExecutedMessages CR query

makramkd:
Core w/ regression test: https://github.com/smartcontractkit/chainlink/pull/17010

When chain selectors are passed in with empty seq num ranges,
the generated query is incorrect.


pkg/reader/ccip.go
```diff
@@ -234,9 +234,8 @@ func (r *ccipChainReader) queryCommitReports(
 // processCommitReports decodes the commit reports from the query results
 // and returns the ones that can be properly parsed and validated.
 func (r *ccipChainReader) processCommitReports(
-	iter []types.Sequence, ts time.Time, limit int,
+	lggr logger.Logger, iter []types.Sequence, ts time.Time, limit int,
```
makramkd, line 237: Use contextual logger, we weren't before.

pkg/reader/ccip.go
```diff
@@ -355,14 +354,25 @@ func (r *ccipChainReader) ExecutedMessages(
...
 ) (map[cciptypes.ChainSelector][]cciptypes.SeqNum, error) {
+	lggr := logutil.WithContextValues(ctx, r.lggr)
+
 	if err := validateExtendedReaderExistence(r.contractReaders, r.destChain); err != nil {
 		return nil, err
 	}
 
+	// trim empty ranges from rangesPerChain
+	// otherwise we may get SQL errors from the chainreader.
+	nonEmptyRangesPerChain := make(map[cciptypes.ChainSelector][]cciptypes.SeqNumRange)
```
makramkd, line 365: Found this cleaner than mutating the input, but open to other ideas.

pkg/reader/ccip.go
```diff
@@ -234,9 +234,8 @@ func (r *ccipChainReader) queryCommitReports(
 // processCommitReports decodes the commit reports from the query results
 // and returns the ones that can be properly parsed and validated.
 func (r *ccipChainReader) processCommitReports(
-	iter []types.Sequence, ts time.Time, limit int,
```
mateusz-sekara, line 237: `ccipChainReader` has a logger already, do we need to pass it via args?

mateusz-sekara (commented):
Is there any way to test that?

pkg/reader/ccip.go
makramkd: Yeah, the one passed in via args has the call context (sequence number, ocr phase)

mateusz-sekara (approved):

pkg/reader/ccip.go
makramkd: (makes debugging easier)

PR #781, opened 2025-03-31
Exec plugin: Optimize Outcome phase - Consensus observation

dimkouv:
# Optimizations to Exec Plugin Outcome Phase

This PR improves the logic used to reach consensus on observations in the exec plugin's outcome phase. It includes a mix of refactoring, bug fixes, and optimizations to ensure we aggregate multiple observations into a single, reliable result more effectively.

---

## Part 1 – Refactoring & Cleanup

Refactored and cleaned up stale code to improve readability and maintainability:
- Removed outdated comments and unnecessary logs
- Deduplicated function parameters
- General code hygiene

**Commit:** `f1c07e1f63b520f8b6afdfe948e09aa6748de54d`

---

### 1. **Commit Data Observations (Merkle Roots)**

Previously, we attempted consensus on full objects, which led to inconsistent results for nearly identical observations. Example:

```
f = 1
{root: 0x123, range: [10, 20], executedMsgs: [10, 11]}
{root: 0x123, range: [10, 20], executedMsgs: [10]}
{root: 0x123, range: [10, 20], executedMsgs: [10, 11, 12]}
```

These were treated as separate due to minor differences, despite having a shared merkle root and overlapping executed messages.

**New behavior:**
- `computeCommitObservationsConsensus` identifies consensus on merkle roots with at least `f+1` observations.
- We now tolerate minor data variation (e.g., executed messages).
- If the same root appears with differing critical fields (e.g., OnRamp address), consensus is considered invalid.
- Executed messages are agreed upon individually if observed by `f+1` nodes.

### 2. **Message Observations**

Improved handling of scenarios where more than one message meets the consensus threshold. Also switched to an `f+1` threshold for consistency across the system.

### 3. **Token Data**

Switched token data consensus to use the `f+1` threshold.

### 4. **Nonces**

Already optimized in a previous PR.

### 5. **Message Hashes**

Handled edge cases where different hashes are observed for the same message.  
`computeMessageHashesConsensus`:
- Aggregates hashes from all observations.
- Accepts a hash if it has at least `f+1` votes.
- If multiple hashes hit the threshold for a single message, logs an error and skips that message.

---

`core ref: 6a481920beb027bdb083f182ac22df66747a3dc7`

### DIFF
--- a/execute/plugin_functions.go
+++ b/execute/plugin_functions.go
@@ -399,7 +399,6 @@ func computeMessageObservationsConsensus(
 			case 0:
 				lggr.Debugw("no message reached consensus for sequence number, skipping it",
 					"chain", chain, "seqNum", seqNum)
-				continue
 			case 1:
 				if _, ok := results[chain]; !ok {
 					results[chain] = make(map[cciptypes.SeqNum]cciptypes.Message)


execute/plugin_functions.go
```diff
@@ -365,144 +367,278 @@ func decodeAttributedObservations(
...
-	results := make(exectypes.CommitObservations)
-	for selector, validator := range validators {
-		if values := validator.GetValid(); len(values) > 0 {
-			results[selector] = validator.GetValid()
+	return validators
+}
+
+// computeCommitObservationsConsensus come to consensus over observations of CommitData.
+// Selects a merkle root and the relevant data if it has at least f+1 observations.
+// If a merkle root is agreed twice but with different relevant data (e.g. OnRampAddress) then we don't have consensus.
```
mateusz-sekara, line 452: Aren't these problems already solved by just reaching a consensus when there are 2Fplus1 observations?

execute/plugin_functions.go
```diff
@@ -621,58 +757,57 @@ func computeNoncesConsensus(
...
+		computeTokenDataObservationsConsensus(lggr, observations, fChain),
+		computeNoncesConsensus(lggr, observations, destFChain),
+		dt.Observation{},
+		computeMessageHashesConsensus(lggr, observations, fChain),
+	)
 
-	lggr.Debugw("merged token data observations", "mergedTokenObservations", mergedTokenObservations)
+	lggr.Debugw("computeConsensusObservation has finished computing the consensus observation",
+		"fChain", fChain,
+		"observations", observations,
```
mateusz-sekara, line 786: Logging all the observations might lead to the heavy impact on the underlying Loki engine. Afaik we also trim logs longer than X bytes (not sure what's the thresholds) in Loki. Maybe instead of logging everything, we should only print chains and merkle roots here to limit the size?

execute/plugin_functions.go
```diff
@@ -365,144 +367,278 @@ func decodeAttributedObservations(
 	return decoded, nil
 }
 
-func mergeMessageObservations(
+func computeMessageObservationsConsensus(
```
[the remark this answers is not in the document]
winder: I think this is personal preference. I'd prefer putting everything into the plugin.go file rather than splitting things up. Everything is related and it's sometimes tedious to move between files because we have very similar functions & filenames for commit, execute, mocks, tests....

execute/plugin_functions.go
```diff
@@ -365,144 +367,278 @@ func decodeAttributedObservations(
...
+	observations []plugincommon.AttributedObservation[exectypes.Observation],
+	fChain map[cciptypes.ChainSelector]int,
+) map[cciptypes.ChainSelector]consensus.OracleMinObservation[cciptypes.Message] {
 	// Create a validator for each chain
-	validators := make(map[cciptypes.ChainSelector]consensus.OracleMinObservation[exectypes.CommitData])
+	validators := make(map[cciptypes.ChainSelector]consensus.OracleMinObservation[cciptypes.Message])
 	for selector, f := range fChain {
-		validators[selector] =
-			consensus.NewOracleMinObservation[exectypes.CommitData](consensus.TwoFPlus1(f), nil)
+		validators[selector] = consensus.NewOracleMinObservation[cciptypes.Message](consensus.FPlus1(f), nil)
```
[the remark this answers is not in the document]
winder: Did we ever create a test case that failed with FPlus1?

execute/plugin_functions.go
```diff
@@ -365,144 +367,278 @@ func decodeAttributedObservations(
...
+				lggr.Debugw("no message reached consensus for sequence number, skipping it",
+					"chain", chain, "seqNum", seqNum)
+				continue
+			case 1:
+				if _, ok := results[chain]; !ok {
+					results[chain] = make(map[cciptypes.SeqNum]cciptypes.Message)
+				}
+				results[chain][seqNum] = msgsWithConsensus[0]
+			default:
+				lggr.Warnw("more than one message reached consensus for a sequence number, skipping it",
```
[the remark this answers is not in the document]
winder: Could you change this to `Errorw`? It seems really bad. The content of a message is on-chain, so how would it ever self resolve?

execute/plugin_functions.go
```diff
@@ -365,144 +367,277 @@ func decodeAttributedObservations(
...
+	fChain map[cciptypes.ChainSelector]int,
+) exectypes.CommitObservations {
+	merkleRootsVotes, executedMsgVotes := aggregateMerkleRootObservations(observations)
+
+	validRoots := make([]merkleRootData, 0, len(merkleRootsVotes))
+	for mr, votes := range merkleRootsVotes {
+		if consensus.GteFPlusOne(fChain[mr.SourceChain], votes) {
+			validRoots = append(validRoots, mr)
+		} else {
+			lggr.Debugw("merkle root with less than f+1 votes was found, skipping it", "mr", mr, "votes", votes)
```
winder, line 465: nit: I'd consider replacing "mr" with "merkleRoot" or "root".

execute/plugin_functions.go
```diff
@@ -365,144 +367,277 @@ func decodeAttributedObservations(
...
+func computeMessageHashesConsensus(
 	lggr logger.Logger,
-	aos []plugincommon.AttributedObservation[exectypes.Observation],
+	observations []plugincommon.AttributedObservation[exectypes.Observation],
 	fChain map[cciptypes.ChainSelector]int,
 ) exectypes.MessageHashes {
-	// Single message can transfer multiple tokens, so we need to find consensus on the token level.
-	validators := make(map[cciptypes.ChainSelector]map[cciptypes.SeqNum]consensus.OracleMinObservation[cciptypes.Bytes32])
-	results := make(exectypes.MessageHashes)
+	lggr = logger.With(lggr, "function", "computeMessageHashesConsensus", "fChain", fChain)
```
winder, line 578: there's already a "caller" field in the logs that tells you the filename/line number

PR #786, opened 2025-04-01
optimize the nonces call to support map[chainselector][]address

0xAustinWang:
core ref: 5f988c244a5ff4d91b1c882dee7a0d9d1e1d25d0

https://smartcontract-it.atlassian.net/browse/CCIP-5454


execute/observation.go
```diff
@@ -411,21 +409,24 @@ func (p *Plugin) getFilterObservation(
 				continue
 			}
 
-			nonceRequestArgs[commitReport.SourceChain][sender] = struct{}{}
+			commitReportSenders[commitReport.SourceChain][sender] = struct{}{}
```
winder, line 412: It looks like you can skip the 2nd loop by changing the type of `commitReportSenders` and appending the sender addresses here.

winder (commented):
Looks good :+1:

execute/observation.go
```diff
@@ -397,35 +395,38 @@ func (p *Plugin) getFilterObservation(
...
-		if err != nil {
-			lggr.Errorw("unable to get nonces", "err", err)
-			continue
-		}
-		nonceObservations[srcChain] = nonces
+	// Get nonces of the addresses.
+	nonceObservations, err := p.ccipReader.Nonces(ctx, commitReportSenders)
+	if err != nil {
+		lggr.Errorw("unable to get nonces", "err", err)
+		return exectypes.Observation{}, fmt.Errorf("unable to get nonces: %w", err)
```
0xnogo, line 429: The previous version would log the error and continue. It seems like we have change the fault tolerance here. Is it intended?

execute/observation.go
```diff
@@ -397,35 +395,38 @@ func (p *Plugin) getFilterObservation(
...
+			sender, err := p.addrCodec.AddressBytesToString(msg.Sender[:], srcChain)
 			if err != nil {
-				lggr.Errorw("unable to convert sender address to string", "err", err, "sender address", msg.Sender)
+				lggr.Errorw("unable to convert sender address to string",
+					"err", err, "sender address", msg.Sender)
 				continue
 			}
 
-			nonceRequestArgs[commitReport.SourceChain][sender] = struct{}{}
+			if _, exists := uniqueSenders[srcChain][sender]; !exists {
```
0xnogo, line 418: Should we also have a check on uniqueSenders[srcChain] existence before?

```
if _, ok := uniqueSenders[srcChain]; !ok {
    uniqueSenders[srcChain] = make(map[string]struct{})
}
if _, exists := uniqueSenders[srcChain][sender]; !exists {
    commitReportSenders[report.SourceChain] = append(commitReportSenders[srcChain], sender)
    uniqueSenders[srcChain][sender] = struct{}{}
}
```

pkg/reader/ccip.go
```diff
@@ -761,42 +757,69 @@ func (r *ccipChainReader) Nonces(
...
-	res := make(map[string]uint64, len(addresses))
 	for _, results := range batchResult {
 		for i, readResult := range results {
-			address := getAddressByIndex(addressToIndex, i)
-
-			if address == "" {
-				lggr.Errorw("critical error, address not found for index", "index", i)
-				continue
-			}
+			key := responses[i]
```
0xnogo, line 762: Can `results` has more elements than `responses`? Maybe we could add a check here to avoid out of bounds panic?

pkg/reader/ccip.go
```diff
@@ -708,47 +708,43 @@ func (r *ccipChainReader) NextSeqNum(
...
-			return nil, fmt.Errorf("failed to convert address %s to bytes: %w", address, err)
+	// create the structure that will contain our result
+	res := make(map[cciptypes.ChainSelector]map[string]uint64)
+	var addressCount int
+	for chain, addresses := range addressesByChain {
+		if len(addresses) == 0 {
+			continue
 		}
+		res[chain] = make(map[string]uint64, len(addresses))
+		addressCount += len(addresses)
```
0xnogo, line 738: We could consider lazy initialization of the maps here


```
res := make(map[cciptypes.ChainSelector]map[string]uint64)
// Only count addresses for array sizing, defer map creation
var addressCount int
for _, addresses := range addressesByChain {
    addressCount += len(addresses)
}

// Later, when actually adding a value:
if _, ok := res[key.chain]; !ok {
    res[key.chain] = make(map[string]uint64)
}
res[key.chain][key.address] = *val
```

execute/observation.go
0xAustinWang: It was intended because we previously wanted to continue to check other srcchains. Now we only have one call. If it fails, there's no other calls to continue to. 

The behavior does change slightly though. Previously if all nonces calls failed, we would eventually return ([], nil). I can restore the behavior in that scenario

execute/observation.go
0xAustinWang: Is this possible? I create the `uniqueSenders[srcChain]` beforehand on line 406 

```
		if _, ok := uniqueSenders[srcChain]; !ok {
			uniqueSenders[srcChain] = make(map[string]struct{})
		}
```

pkg/reader/ccip.go
0xAustinWang: updated. does this provide performance improvement?

execute/observation.go
0xnogo: Makes sense

0xnogo (approved):

PR #794, opened 2025-04-03
solana: minor fixes

jadepark-dev:
- [x] incorrect comments
- [x] `UpdateSourceChain` param in account context
- [x] Code optimizations/improvements(partially covered)
- [x] Redundant `mut` constraint
- [x] Assertion improvement


jadepark-dev:
@agusaldasoro I think we can merge this as-is for now since these are all informational issues

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/ocr3base.rs
```diff
@@ -191,6 +191,18 @@ pub(super) fn ocr3_transmit<R: Ocr3Report>(
         );
 
         verify_signatures(ocr3_config, report.hash(&report_context), signatures)?;
+    } else {
+        // if signature verification is disabled, ensure that the signatures are empty
```
agusaldasoro, line 195: Do we need to do that? I remember in EVM when executing the signatures were sent to the contracts but weren't verified on chain. I'm wondering if this change is a breaking change for the offchain code.

agusaldasoro (approved):

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/ocr3base.rs
jadepark-dev: @agusaldasoro Thanks for checking in! I double checked evm and it does not explicitly verify that signatures are empty when verification is disabled. looks like adding this is redundant, I'll remove it

agusaldasoro (dismissed):

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -478,19 +478,16 @@ pub fn validate_execution_report<'info>(
     message_header: &RampMessageHeader,
     svm_chain_selector: u64,
 ) -> Result<()> {
-    require!(
-        execution_report.message.header.nonce == 0,
-        CcipOfframpError::InvalidNonce
-    );
+    require!(message_header.nonce == 0, CcipOfframpError::InvalidNonce);
```
agusaldasoro, line 481: nit:
```suggestion
    require_eq!(message_header.nonce, 0, CcipOfframpError::InvalidNonce);
```

agusaldasoro (approved):

PR #795, opened 2025-04-03
solana: missing version checks in context

jadepark-dev:
Fixes M4(+L1): Among the protocol’s programs, there are some context structs which do not validate the version of the data of the accounts used. In most cases, the helper valid_version function is used as a constraint, but it is missing from the following instances:


chains/solana/contracts/programs/fee-quoter/src/context.rs
```diff
@@ -26,6 +26,7 @@ pub fn uninitialized(v: u8) -> bool {
 const MAX_CONFIG_V: u8 = 1;
 const MAX_CHAINSTATE_V: u8 = 1;
 const MAX_TOKEN_AND_CHAIN_CONFIG_V: u8 = 1;
+const MAX_BILLING_TOKEN_CONFIG_V: u8 = 1;
```
jadepark-dev, line 29: @agusaldasoro @toblich I need eyes on this, I'm not sure `MAX_TOKEN_AND_CHAIN_CONFIG_V` can be shared for `billing_token_config`?

agusaldasoro (commented):
We need to add the same validation to the TokenAdminRegistry PDA as well, right?

chains/solana/contracts/programs/burnmint-token-pool/src/context.rs
```diff
@@ -8,6 +8,8 @@ use ccip_common::seed;
 
 use crate::{program::BurnmintTokenPool, ChainConfig, State};
 
+const MAX_POOL_STATE_V: u8 = 1;
```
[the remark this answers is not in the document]
toblich: This is fine, it's not imported by any user. It just identifies the schema version for that PDA, in case we ever need to migrate the state to another data shape.

chains/solana/contracts/programs/fee-quoter/src/context.rs
toblich: Yes, as it's the same PDA in different instructions, it's correct that there's a single constant for both usages :+1:

toblich (approved):

PR #797, opened 2025-04-04
fix: Don't check ContractAddress, since it'll be set if RMNRemote is bound

archseer:
We want to support RMNRemote on Solana, but not RMNHome.


commit/merkleroot/rmn/types/config.go
```diff
@@ -43,8 +43,8 @@ type RemoteConfig struct {
 }
 
 func (r RemoteConfig) IsEmpty() bool {
-	return len(r.ContractAddress) == 0 &&
-		r.ConfigDigest == (cciptypes.Bytes32{}) &&
+	// NOTE: contract address will always be present, since the code auto populates it
```
winder, line 46: If it's always present, why remove the length check?

archseer: Because we want this check to return `IsEmpty` in order to skip the rest of the config validation (or merkle root observations fail).

The `RMNRemote` contract is bound so that Solana can use `GetCursedSubjects`, but we don't actually want to enable the full config here, which would imply RMNHome support. We blank out the `Version` value etc to return an empty config, but `ContractAddress` is the one variable we can't control: https://github.com/smartcontractkit/chainlink-ccip/blob/721ad77875e556708f1ecb76a4ac4b02e19eaaa1/pkg/reader/ccip.go#L1007

ilija42 (approved):

PR #798, opened 2025-04-04
Remove msg.data bytes from exec logs

asoliman92:
`Message.Data` field is not important to log, this is user data and doesn't affect debugging while taking too much space unnecessarily.

### DIFF
--- a/execute/exectypes/commit_data.go
+++ b/execute/exectypes/commit_data.go
@@ -40,3 +40,23 @@ type CommitData struct {
 	// Length of this slice should equal to the length of Messages slice.
 	MessageTokenData []MessageTokenData `json:"messageTokenData"`
 }
+
+// CopyNoMsgData creates a copy of the CommitData without the messages.Data
+func (cd CommitData) CopyNoMsgData() CommitData {
+	msgsWitoutData := make([]cciptypes.Message, len(cd.Messages))
+	for i, msg := range cd.Messages {
+		msgsWitoutData[i] = msg.CopyWithoutData()
+	}
+	return CommitData{
+		SourceChain:         cd.SourceChain,
+		OnRampAddress:       cd.OnRampAddress,
+		Timestamp:           cd.Timestamp,
+		BlockNum:            cd.BlockNum,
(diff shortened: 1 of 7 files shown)


pkg/types/ccipocr3/generic_types.go
```diff
@@ -183,6 +183,20 @@ type Message struct {
 	TokenAmounts []RampTokenAmount `json:"tokenAmounts"`
 }
 
+func (m Message) CopyWithoutData() Message {
```
mateusz-sekara, line 186: Maybe 
```go
func (m Message) ToLogFormat() Message {}
```

mateusz-sekara: Or even further returning serialized value to prevent from accidental usage

```go
func (m Message) ToLogFormat() string {}
```

execute/exectypes/commit_data.go
```diff
@@ -40,3 +40,23 @@ type CommitData struct {
...
+
+// CopyNoMsgData creates a copy of the CommitData without the messages.Data
+func (cd CommitData) CopyNoMsgData() CommitData {
+	msgsWitoutData := make([]cciptypes.Message, len(cd.Messages))
+	for i, msg := range cd.Messages {
+		msgsWitoutData[i] = msg.CopyWithoutData()
+	}
+	return CommitData{
+		SourceChain:         cd.SourceChain,
+		OnRampAddress:       cd.OnRampAddress,
```
mateusz-sekara, line 52: Maybe a follow-up, but it would be great to understand which of these fields are really worth having in the logs. Blocknumber, timestamp, or onramp address takes some space in the logs but might not be very useful for 99% of the cases.

asoliman92: Agreed. I was gonna name functions `Clean` or `Sanitize` for the future but kept it for msgData for now.

pkg/types/ccipocr3/generic_types.go
asoliman92: I like it, I'll keep this one as is. It's used in other functions to get the logFormat. Will change the rest.

PR #803, opened 2025-04-04
Use unique CPI signers per called program [NONEVM-1500]

toblich:
This changes the signers used:
1) When the router invokes a token pool _--> Signer added to the token pool lookup table_
2) When the offramp invokes a token pool
3) When the offramp invokes a receiver

This way, no user can create a malicious pool/receiver that then just `invoke`s (thus forwarding the signature) another receiver/pool 

As there are different signers, they now have to be passed in per-token pool. As there are multiple offramps and multiple receivers, this also means that the signers for cases 2 & 3 won't be present in any lookup table, thus consuming more tx space on executions.


chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -8178,7 +8141,7 @@ func TestCCIPRouter(t *testing.T) {
 					cu := testutils.GetRequiredCU(ctx, t, solanaGoClient, []solana.Instruction{instruction}, transmitter, config.DefaultCommitment)
 
 					// This validation is like a snapshot for gas consumption
-					require.LessOrEqual(t, cu, uint32(100_000))
+					require.LessOrEqual(t, cu, uint32(110_000))
```
toblich, line 8144: @agusaldasoro FYI

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -8403,7 +8367,7 @@ func TestCCIPRouter(t *testing.T) {
 
 					// This validation is like a snapshot for gas consumption
 					// Execute: 1 Token Transfer + Message Execution
-					require.LessOrEqual(t, cu, uint32(230_000))
+					require.LessOrEqual(t, cu, uint32(250_000))
```
toblich, line 8370: @agusaldasoro FYI

chains/solana/contracts/tests/txsizing_test.go
```diff
@@ -45,49 +98,46 @@ func TestTransactionSizing(t *testing.T) {
...
+	tokenTable := TokenTable{
+		TokenAdminRegistryPDA: mustRandomPubkey(),
+		PoolLookupTable:       mustRandomPubkey(),
+		PoolProgram:           config.CcipTokenPoolProgram,
+		PoolConfig:            mustRandomPubkey(),
+		PoolTokenAccount:      mustRandomPubkey(),
+		PoolSigner:            mustRandomPubkey(),
+		TokenProgram:          config.Token2022Program,
+		Mint:                  mustRandomPubkey(),
+		RouterSigner:          mustRandomPubkey(),
```
toblich, line 140: @agusaldasoro please review that this list still makes sense for every program. I found inconsistencies with this test and the accounts that were actually static enough (meaning they didn't depend on the caller/chain/feetoken) to have in an ALT

chains/solana/contracts/programs/ccip-common/src/v1.rs
```diff
@@ -2,20 +2,32 @@ use anchor_lang::prelude::*;
...
 };
 
+pub const MIN_TOKEN_POOL_ACCOUNTS: usize = 13; // see TokenAccounts struct for all required accounts
+
 pub struct TokenAccounts<'a> {
     pub user_token_account: &'a AccountInfo<'a>,
     pub token_billing_config: &'a AccountInfo<'a>,
     pub pool_chain_config: &'a AccountInfo<'a>,
+    pub lookup_table: &'a AccountInfo<'a>,
+    pub token_admin_registry: &'a AccountInfo<'a>,
```
agusaldasoro, line 16: those two weren't part of the token accounts? 🤔

chains/solana/contracts/programs/ccip-common/src/v1.rs
```diff
@@ -2,20 +2,32 @@ use anchor_lang::prelude::*;
...
     pub pool_program: &'a AccountInfo<'a>,
     pub pool_config: &'a AccountInfo<'a>,
     pub pool_token_account: &'a AccountInfo<'a>,
     pub pool_signer: &'a AccountInfo<'a>,
     pub token_program: &'a AccountInfo<'a>,
     pub mint: &'a AccountInfo<'a>,
     pub fee_token_config: &'a AccountInfo<'a>,
+    pub ccip_router_pool_signer: &'a AccountInfo<'a>,
+    // as this one is optional, it doesn't count for the MIN_TOKEN_POOL_ACCOUNTS
```
agusaldasoro, line 25: Is it optional? Or only present when this is used by the offramp?

chains/solana/contracts/programs/ccip-common/src/v1.rs
```diff
@@ -33,9 +46,20 @@ pub fn validate_and_parse_token_accounts<'info>(
     // Changes in environment-specific program addresses will not affect the PDA derivation.
     let program_id = Pubkey::default();
 
+    let mut accounts = raw_acc_infos;
+
+    let ccip_offramp_pool_signer = if offramp.is_some() {
+        // When and only when the offramp program is calling this function, the first account is the offramp pool signer.
```
agusaldasoro, line 52: Reminder to update the docs properly with this change

chains/solana/contracts/programs/ccip-common/src/v1.rs
```diff
@@ -72,6 +96,29 @@ pub fn validate_and_parse_token_accounts<'info>(
     let token_program = next_account_info(&mut accounts_iter)?;
     let mint = next_account_info(&mut accounts_iter)?;
     let fee_token_config = next_account_info(&mut accounts_iter)?;
+    let ccip_router_pool_signer = next_account_info(&mut accounts_iter)?;
+
+    // As the offramp signer is optional (only used when the offramp program is calling this function),
+    // the context cannot be used to validate it, as it could try to apply those validations to the wrong account
+    let ccip_offramp_pool_signer_bump = if let Some(offramp) = offramp {
+        let (expected_key, bump) = Pubkey::find_program_address(
```
agusaldasoro, line 104: nit:
```suggestion
        let (expected_offramp_pool_signer, bump) = Pubkey::find_program_address(
```

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -378,22 +385,25 @@ fn should_execute_messaging<'a>(
     !remaining_accounts.is_empty() && !only_tokens
 }
 
+pub struct MessagingAccounts<'a> {
```
agusaldasoro, line 388: like this!

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -243,9 +244,16 @@ fn internal_execute<'info>(
             accs.user_token_account.to_account_info(),
         ];
         acc_infos.extend_from_slice(accs.remaining_accounts);
+
+        let seeds = &[
+            seed::EXTERNAL_TOKEN_POOL,
+            accs.pool_program.key.as_ref(),
+            &[accs.ccip_offramp_pool_signer_bump],
```
agusaldasoro, line 251: how does this work? Does the bump also include the actual offramp program id?

chains/solana/contracts/tests/ccip/ccip_router_test.go
agusaldasoro: 😢

chains/solana/contracts/programs/ccip-common/src/v1.rs
toblich: It was, but this method was "swallowing" that value and not returning it. So, for this method to be clearer, I made it return all the accounts it gets, and thus the `MIN_TOKEN_POOL_ACCOUNTS` constant is also more consistent (as that one relates to the input accounts, and the actual struct was only listing the accounts this method returns)

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
toblich: Given that this is executed in the offramp code, the offramp program id is implicit. It's not included in the bump, that's just a `u8`, but you can only sign with a PDA that belongs to the same program, so it uses always uses the offramp program id.

chains/solana/contracts/programs/ccip-common/src/context.rs
```diff
@@ -106,7 +106,6 @@ pub struct TokenAccountsValidationContext<'info> {
 
     /// CHECK: The signer to be used by the router program to invoke the pool program
     #[account(
-        mut,
```
agusaldasoro: why was this mutable? 🤔

chains/solana/contracts/tests/txsizing_test.go
agusaldasoro: This test could be outdated as it only builds the txs, and it doesnt' execute anything

agusaldasoro (dismissed):

PR #804, opened 2025-04-07
Add Solidity contracts

RensR:


RensR:
>Can we add gethwrappers for TokenPoolFactory and FactoryBurnMintERC20?

They don't exist right now as far as I can see, we can do it in a follow up to this. This PR should only make sure we have parity and can move development to this repo

go.mod
```diff
@@ -4,57 +4,76 @@ go 1.24
 
 require (
 	github.com/deckarep/golang-set/v2 v2.6.0
+	github.com/ethereum/go-ethereum v1.15.3
```
makramkd, line 7: Once upon a time we explicitly didn't want to import go-ethereum (guidance from foundations). I can't really recall the reasoning at the moment. I guess since libocr imports it directly, it was a transitive dep for us anyway.

makramkd (approved):

go.mod
RensR: Yeah it was transitive already and very outdated. Not sure if that means libocr uses that very old version 🤔 

In a distant past we had issues with go-ethereum versions but I think that has since been solved so I don't really see a reason not to import. Good callout though

RensR: Hmm yeah if it wasn't already imported through OCR it would have probably made sense to create a separate go.mod in the evm folder. Does that still have value given we already import it?

0xsuryansh (approved):

PR #812, opened 2025-04-08
Limit number of reports and messages in exec plugin GetMessages phase [CCIP-5647] 

asoliman92:
As Exec plugin has 3 states, the second state `GetMessages` was getting all possible messages (with 1K reports) without doing much filtering other than if the observation size exceeded max size using type_optimizer  - deprecated in this PR -. 

The new algorithm does filter based on the encoding size as well but it's more lenient at the `GetMessages` phase. Usually exec plugin is able to execute 10-20 msgs at a time so it doesn't make sense to keep on observing - in theory - thousands of messages every round to only come in the third round and remove most of them to send only 10 msgs. 

We're setting a couple of lenient maxMsgPerObs and maxObsSize that we're sure can keep up with without complicating the algorithm.

Also introduced logic to check for size limit which used iteratively for each message added to the observation (ditching the old way of reading all and then removing messages later)


execute/exectypes/commit_data.go
```diff
@@ -41,6 +41,17 @@ type CommitData struct {
 	MessageTokenData []MessageTokenData `json:"messageTokenData"`
 }
 
+func CompareCommitData(i, j CommitData) bool {
+	// Should be very rare to have the same timestamp but better to be deterministic
```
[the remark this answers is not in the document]
asoliman92: True 👍

execute/plugin.go
```diff
@@ -275,7 +275,7 @@ func getPendingReportsForExecution(
 	fullyExecutedUnfinalized []exectypes.CommitData,
 	err error,
 ) {
-	commitReports, err := ccipReader.CommitReportsGTETimestamp(ctx, ts, 1000) // todo: configurable limit
+	commitReports, err := ccipReader.CommitReportsGTETimestamp(ctx, ts, 20) // todo: configurable limit
```
asoliman92, line 278: 1000 is a lot to fetch. 20 is more sensible. They're already sorted by timestamps

pkg/types/ccipocr3/generic_types.go
```diff
@@ -205,7 +205,7 @@ func (m Message) String() string {
 // IsEmpty returns true if the message is empty. Can't use == Message{} without using reflect.DeepEqual.
 func (m Message) IsEmpty() bool {
 	return m.Header.MessageID == Bytes32{} && m.Header.SourceChainSelector == 0 &&
-		m.Header.DestChainSelector == 0 && m.Header.SequenceNumber == 0 && m.Header.Nonce == 0
```
asoliman92, line 208: We use the seqNum of the message during report building while working with proofs.

execute/plugin.go
winder: Woah. Yes, good call. But I thought these would be cached by log poller after the initial query.

execute/observation.go
```diff
@@ -259,68 +261,177 @@ func buildCombinedReports(
...
-				messageObs[srcChain][msg.Header.SequenceNumber] = msg
-				messageTimestamps[msg.Header.MessageID] = report.Timestamp
+			// Update the observation with current state
+			observation.CommitReports = availableReports
+			observation.Messages = messageObs
+			observation.Hashes = msgHashes
+
+			// Check if we've exceeded encoding size limits
+			if !stop {
+				stop = p.handleEncodingSizeLimits(observation, messageObs, srcChain, seqNum)
```
winder, line 326: Having `handleEncodingSizeLimits` be a receiver function makes me a little nervous that there could be a side effect. Could you make it a pure function?

execute/exectypes/commit_data.go
```diff
@@ -41,6 +41,16 @@ type CommitData struct {
 	MessageTokenData []MessageTokenData `json:"messageTokenData"`
 }
 
+func CompareCommitData(i, j CommitData) bool {
```
makramkd, line 44: Doc comment explaining the significance of the implementation would be good so that we don't modify this erroneously in the future.

execute/observation.go
```diff
@@ -259,68 +261,177 @@ func buildCombinedReports(
...
-	commitObs := regroup(commitData)
+	sort.Slice(commitData, func(i, j int) bool {
+		return exectypes.CompareCommitData(commitData[i], commitData[j])
+	})
 
-	for srcChain, reports := range commitObs {
-		if len(reports) == 0 {
-			continue
+	stop := false
+	gasSum := uint64(0)
```
makramkd, line 279: I don't think this is a viable approach, gas limit is an EVM concept, and even if it isn't, we should not constrain ourselves to a particular amount in compute/gas here since we could split into multiple reports in the 3rd phase (e.g, Solana/Aptos)

makramkd (changes requested):

execute/observation.go
winder: If this message causes the limit to be hit, should it be removed?
```suggestion
				stop = p.handleEncodingSizeLimits(observation, messageObs, srcChain, seqNum)
				if stop {
					messageObs[srcChain][seqNum] = createEmptyMessageWithSeqNum(seqNum)
				}
```

asoliman92: That's why I'm using lenient maxSize, The max size is set up to use 75% of actual max that we're looking for to allow for these things

execute/plugin.go
asoliman92: even with caching retrieving 1K reports sounds like a lot as we won't be able to process all of them anyways. something around 20-50 makes more sense.

execute/factory.go
```diff
@@ -187,7 +193,7 @@ func (p PluginFactory) NewReportingPlugin(
 			Limits: ocr3types.ReportingPluginLimits{
 				// No query for this execute implementation.
 				MaxQueryLength:       maxQueryLength,
-				MaxObservationLength: maxObservationLength,
+				MaxObservationLength: ocr3types.MaxMaxObservationLength,
```
asoliman92, line 196: Using this to allow the lenientMax to work as expected. Otherwise we hit max size and ocr3 gets stuck

execute/factory.go
```diff
@@ -34,14 +34,16 @@ const (
 	// maxQueryLength is set to disable queries because they are not used.
 	maxQueryLength = 0
 
-	// maxObservationLength is set to the maximum size of an observation
+	// lenientMaxObservationLength is set to  75% of ocr3 maximum recommended size
 	// check factory_test for the calculation.
+	// Using 75% to allow for some space while observing without hitting the max.
 	// this is being set to the max maximum observation length due to
 	// the observations being so large at the moment, especially when
 	// commit reports have many messages.
```
makramkd, line 42: This comment seems to contradict the above, so probably remove?

execute/factory.go
```diff
@@ -58,6 +60,10 @@ const (
 	// the actual exec report type (ExecutePluginReport) may contain multiple
 	// per-source-chain reports. These are not limited by this value.
 	maxReportCount = 1
+
+	// maxMsgsPerQuery is set to the maximum number of messages that can be observed in one observation, this is a bit
+	// linient and acts as an indicator other than a hard limit.
```
makramkd, line 65:
```suggestion
	// lenient and acts as an indicator other than a hard limit.
```

execute/factory.go
```diff
@@ -34,14 +34,16 @@ const (
 	// maxQueryLength is set to disable queries because they are not used.
 	maxQueryLength = 0
 
-	// maxObservationLength is set to the maximum size of an observation
+	// lenientMaxObservationLength is set to  75% of ocr3 maximum recommended size
 	// check factory_test for the calculation.
+	// Using 75% to allow for some space while observing without hitting the max.
```
makramkd, line 39: Is there a benefit to this 25% buffer? It's not really clear from this comment what we gain from only using 75% of the max allowable. Is there some kind of risk/probabilistic approach to the current approach of truncation?

execute/report/roots.go
```diff
@@ -36,7 +36,8 @@ func ConstructMerkleTree(
 				"malformed report, message %s sequence number %d outside of report range %s",
 				report.MerkleRoot.String(), msg.Header.SequenceNumber, report.SequenceNumberRange)
 		}
-		if report.SourceChain != msg.Header.SourceChainSelector {
+		// When msg is empty the chain selector will always be zero
+		if report.SourceChain != msg.Header.SourceChainSelector && !msg.IsEmpty() {
```
makramkd, line 40: I don't really understand this change or this comment, so if the message is not empty, we still try to use this message despite it having the wrong source chain selector?

execute/report/report.go
```diff
@@ -167,7 +167,7 @@ func Template() Check {
 func CheckIfPseudoDeleted() Check {
 	return func(lggr logger.Logger, msg ccipocr3.Message, idx int, report exectypes.CommitData) (messageStatus, error) {
 		if msg.IsEmpty() {
-			lggr.Errorw("message pseudo deleted", "index", idx)
+			lggr.Infow("message pseudo deleted", "index", idx)
```
makramkd, line 170: `Debugw` seems more appropriate, if this is not an error, or is this indicative of something else? I haven't seen this log personally.

execute/observation.go
```diff
@@ -259,68 +261,142 @@ func buildCombinedReports(
...
-	commitObs := regroup(commitData)
+	sort.Slice(commitData, func(i, j int) bool {
+		return exectypes.CompareCommitData(commitData[i], commitData[j])
+	})
 
-	for srcChain, reports := range commitObs {
-		if len(reports) == 0 {
-			continue
+	stop := false
+	//gasSum := uint64(0)
```
makramkd, line 279: IMO better to delete unused code to reduce confusion about its existence and create a followup on a potential further optimization.

execute/observation.go
```diff
@@ -259,68 +261,142 @@ func buildCombinedReports(
...
+			observation.CommitReports = availableReports
+			observation.Messages = messageObs
+			observation.Hashes = msgHashes
 
-			for _, msg := range msgs {
-				messageObs[srcChain][msg.Header.SequenceNumber] = msg
-				messageTimestamps[msg.Header.MessageID] = report.Timestamp
+			// Check if we've exceeded encoding size limits
+			if !stop {
+				stop = totalMsgs > maxMsgsPerObs ||
```
makramkd, line 326: Shouldn't this be `totalMsgs >= maxMsgsPerObs` if we want to honor the max?

execute/observation.go
```diff
@@ -259,68 +261,142 @@ func buildCombinedReports(
...
-		if _, ok := groupedCommits[report.SourceChain]; !ok {
-			groupedCommits[report.SourceChain] = []exectypes.CommitData{}
-		}
-		groupedCommits[report.SourceChain] = append(groupedCommits[report.SourceChain], report)
-	}
-	return groupedCommits
-}
-
-func readAllMessages(
+func (p *Plugin) getObsWithoutTokenData(
```
makramkd, line 264: Can this be made a pure function w/ some rigorous tests?

makramkd: Also whats the significance of `withoutTokenData`?

execute/observation.go
```diff
@@ -259,68 +261,142 @@ func buildCombinedReports(
...
+
+			// Handle message observation and gas calculation
+			if !stop && !p.inflightMessageCache.IsInflight(srcChain, msg.Header.MessageID) {
+				messageObs[srcChain][seqNum] = msg
+				totalMsgs++
+			} else {
+				// This is for when we calculate roots in reports later, we need the seqNum and
+				// the hash for this message
+				messageObs[srcChain][seqNum] = createEmptyMessageWithSeqNum(seqNum)
 			}
```
makramkd, line 308: I find this code confusing, why create an empty message and not just use `msg`?

execute/observation.go
```diff
@@ -259,68 +261,142 @@ func buildCombinedReports(
...
-			if !msgsConformToSeqRange(msgs, report.SequenceNumberRange) {
-				lggr.Errorw("missing messages in range",
-					"srcChain", srcChain, "seqRange", report.SequenceNumberRange)
-				continue
+			// Compute hash for all messages in report even if they are executed or not included in this round
+			hash, err := p.msgHasher.Hash(ctx, msg)
+			if err != nil {
+				return exectypes.Observation{}, fmt.Errorf("unable to hash message srcChain %d, seqNum %d: %w",
+					srcChain, seqNum,
+					err)
```
makramkd, line 315: Lets include msg id in the error

execute/plugin.go
```diff
@@ -275,7 +275,7 @@ func getPendingReportsForExecution(
...
 	if err != nil {
 		return nil, nil, nil, err
```
makramkd, line 280: 1000 -> 20 seems like a pretty big drop; can we separate this change to another PR so we can figure out the best way to address?

execute/report/roots.go
asoliman92: we discard it if it's empty, this is an extra check because empty messages still have default values which are all zeroes

execute/report/report.go
asoliman92: It only happens under heavy load, that's why you haven't seen it. I'd rather info in this case to be clearer. It indicates that some truncation happened on the original observation.

execute/observation.go
asoliman92: Already wrote a lot of tests to cover most cases. Making it pure will just make the code more complicated with more function calls. 

> Also whats the significance of withoutTokenData?

To get token data in batch afterwards. Will add a comment about it.

execute/plugin.go
asoliman92: What's the best way to address it in your opinion? We already did some calculations. If each report has 2-4 messages and our max is 100 then number of reports should be max 100 with 20-50 making more sense.

execute/observation.go
asoliman92: This is how we mark a message as not included. We need a way to delete a message but in the same time keep its seqNr and hash for the rest of the exec plugin to work properly in outcome and report phases.

execute/factory.go
asoliman92: It's simplifying the logic. Otherwise we need to delete messages once we hit the max, also as we add tokenData afterwards to gain batching optimization we need more space to cover for it.

asoliman92: Actually without it the error we've been talking about earlier today will happen unless we have more conditions and truncation logic

execute/observation.go
asoliman92: Yeah, I just missed it while removing other gas related code. thanks

execute/observation.go
asoliman92: it's lenient so it's okay, report can have already have more messages. otherwise we'll need to pseudo delete all messages above the size. changing it to have equal and changing the variable name as well.

execute/observation.go
winder: I think it's better to be consistent. Room is still needed for the remaining hashes. The worst case scenario is a commit report with 256 messages and we hit the limit on message 2. If message 2 is huge will there always be enough room for the remaining 254 hashes?

The builder also has logic to handle skipping the large message and continuing with smaller messages. So I wonder if it's possible that this optimization could cause problems with other optimizations in the builder.

execute/factory.go
winder: For me, this violates the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). The algorithm should be deterministic without depending on fuzzy buffers.

execute/report/roots.go
winder: The data structure is index based so gaps are not allowed. At the same time, we delete messages when capacity is reached. This check is a result of that approach.

execute/report/report.go
winder: Could the messageID be included here? Thats often the keyword we would be using to search for logs during an incident.

execute/plugin.go
winder: A constant in `factory.go` might be a good spot for this.

execute/factory.go
makramkd: So this is the "max" observation size w/out token data and without hashes? Seems like a pretty important detail to put in here. Actually, I feel like the truncation logic is pretty poorly documented in this PR and it would be great to have a TLDR somewhere on what is going on (PR description doesn't count, you won't have that when reading the code).

execute/observation.go
```diff
@@ -259,68 +261,158 @@ func buildCombinedReports(
...
-	for srcChain, reports := range commitObs {
-		if len(reports) == 0 {
-			continue
+	stop := false
+
+	for _, report := range commitData {
+		srcChain := report.SourceChain
+
+		// Read messages for this report's sequence number range
+		msgs, err := p.readMessagesForReport(ctx, lggr, srcChain, report)
```
makramkd, line 300: Just noticed that this ends up doing a db query.

execute/observation.go
```diff
@@ -259,68 +261,158 @@ func buildCombinedReports(
...
+) ([]cciptypes.Message, error) {
+	msgs, err := p.ccipReader.MsgsBetweenSeqNums(ctx, srcChain, report.SequenceNumberRange)
+	if err != nil {
+		lggr.Errorw("unable to read all messages for report",
+			"srcChain", srcChain,
+			"seqRange", report.SequenceNumberRange,
+			"merkleRoot", report.MerkleRoot,
+			"err", err,
+		)
+		return nil, err
```
makramkd, line 383: nit: seems cleaner to return a wrapped error and log that error at the call site.

execute/observation.go
```diff
@@ -259,68 +261,158 @@ func buildCombinedReports(
...
+			observation.Messages = messageObs
+			observation.Hashes = msgHashes
 
-			for _, msg := range msgs {
-				messageObs[srcChain][msg.Header.SequenceNumber] = msg
-				messageTimestamps[msg.Header.MessageID] = report.Timestamp
+			// Check if we've exceeded encoding size limits
+			if !stop {
+				stop = totalMsgs >= lenientMaxMsgsPerObs ||
+					exceedsMaxEncodingSize(observation, p.ocrTypeCodec, lenientMaxObservationLength)
```
makramkd, line 343: How far above the lenient threshold is fine? If lenient is 75% of max observation size from libocr and we exceed it by 15% for pathological edge cases, aren't we at risk of exceeding the libocr size limit? It seems safer to not include the message that put you over the limit no?

This function currently checks if its >= maxSize, and I feel like this detail is important and can be pretty catastrophic if we error out in this part since the plugin gets completely blocked.

makramkd (changes requested):
We seem to only be size-limiting the messages without token data and without hashes, so even with the "lenient" threshold we might be at risk of overflowing the max data size.

If i'm understanding it correctly, to be truly 100% sure that we fit within the limit, there needs to be a step after adding token data and hashes to make sure we're not above the max size.

asoliman92:
We discussed offline and agreed to add more checks and implementation after calling `getObsWithoutTokenData` function

execute/factory.go
asoliman92: Added another constant for it

pkg/types/ccipocr3/generic_types.go
asoliman92: We use messageID also for debugging purposes, so an empty message is one that has 0 source and est chains as well as no onRamp

execute/factory.go
```diff
@@ -58,6 +57,10 @@ const (
 	// the actual exec report type (ExecutePluginReport) may contain multiple
 	// per-source-chain reports. These are not limited by this value.
 	maxReportCount = 1
+
+	// lenientMaxMsgsPerObs is set to the maximum number of messages that can be observed in one observation, this is a bit
+	// lenient and acts as an indicator other than a hard limit.
+	lenientMaxMsgsPerObs = 100
```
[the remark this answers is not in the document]
asoliman92: we talked more about the lenient and the logic in a call. Do you still have any concerns?

execute/observation.go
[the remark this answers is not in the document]
asoliman92: No need, we read report by report with size check for every single message added. We can still do it but I think it's a micro optimization IMO

execute/observation.go
asoliman92: We discussed offline and the logic now after adding tokenData and all empty structs before this loop should fix this with no issue.

execute/observation.go
```diff
@@ -259,68 +261,158 @@ func buildCombinedReports(
...
-		messageObs[srcChain] = make(map[cciptypes.SeqNum]cciptypes.Message)
-		// Read messages for each range.
-		for _, report := range reports {
-			msgs, err := ccipReader.MsgsBetweenSeqNums(ctx, srcChain, report.SequenceNumberRange)
+		// Add the report to available reports
+		availableReports[srcChain] = append(availableReports[srcChain], report)
+
+		totalMsgs := 0
+		// Initialize data structures for this source chain if needed
+		initSourceChainMaps(srcChain, messageObs, msgHashes)
```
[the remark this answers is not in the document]
asoliman92: less cognitive load, already simple to get out like that.

execute/observation.go
```diff
@@ -259,70 +261,109 @@ func buildCombinedReports(
...
 	}
-	return groupedCommits
+	msgTkData, err := p.tokenDataObserver.Observe(ctx, msgObs)
+	if err != nil {
+		lggr.Errorw("failed to observe token data", "err", err)
+		// In case of failure, initialize the token data with an error, that will prevent this specific token
+		// from being processed and sent later but won't stop the rest of messages
+		msgTkData = make(exectypes.TokenDataObservations)
+		msgTkData[srcChain] = make(map[cciptypes.SeqNum]exectypes.MessageTokenData)
+		msgTkData[srcChain][seqNum] = exectypes.NewMessageTokenData(exectypes.NewErrorTokenData(err))
```
makramkd, line 289: This can be written more expressively as:

```golang
msgTokenData = exectypes.TokenDataObservations{
  srcChain: map[cciptypes.SeqNum]exectypes.MessageTokenData{
    seqNum: exectypes.NewMessageTokenData(exectypes.NewErrorTokenData(err))
  }
}
```

execute/observation.go
```diff
@@ -259,70 +261,109 @@ func buildCombinedReports(
...
+	messageObs exectypes.MessageObservations,
+	msgHashes exectypes.MessageHashes,
+	tkData exectypes.TokenDataObservations,
+) {
+	if _, ok := messageObs[srcChain]; !ok {
+		messageObs[srcChain] = make(map[cciptypes.SeqNum]cciptypes.Message)
+		msgHashes[srcChain] = make(map[cciptypes.SeqNum]cciptypes.Bytes32)
+		tkData[srcChain] = make(map[cciptypes.SeqNum]exectypes.MessageTokenData)
+	}
+}
```
makramkd, line 306: In terms of net value this function is 12 lines long but the body is just 5 lines and its only used in one place. Better to just get rid of it and put this code inline at the call site. Its also pretty clunky because it mutates inputs which is bad practice and leads to harder to understand code

execute/observation.go
```diff
@@ -333,49 +374,98 @@ func (p *Plugin) getMessagesObservation(
 	//          These messages will not be executed in the current round, but may be executed in future rounds
 	//          (e.g. if gas prices decrease).
 	if len(previousOutcome.CommitReports) == 0 {
-		lggr.Debug("TODO: No reports to execute. This is expected after a cold start.")
+		p.lggr.Debug("TODO: No reports to execute. This is expected after a cold start.")
```
makramkd, line 377: Please use the provided logger in the function and not `p.lggr`

execute/observation.go
```diff
@@ -333,49 +374,98 @@ func (p *Plugin) getMessagesObservation(
...
+				// remove the last message and token data
+				observation.Messages[srcChain][seqNum] = createEmptyMessageWithIDAndSeqNum(msg)
+				observation.TokenData[srcChain][seqNum] = exectypes.NewMessageTokenData()
+				stop = true
+				break
+			}
+		}
+
+		if stop {
+			lggr.Infow("Stop processing messages, observation is too large")
```
makramkd, line 466: Please include some more context here, like `totalMsgs` and if possible the last observation size in case we are debugging issues.

execute/report/roots.go
```diff
@@ -36,7 +36,9 @@ func ConstructMerkleTree(
 				"malformed report, message %s sequence number %d outside of report range %s",
 				report.MerkleRoot.String(), msg.Header.SequenceNumber, report.SequenceNumberRange)
 		}
-		if report.SourceChain != msg.Header.SourceChainSelector {
+		// When msg is empty we don't need to check the source chain selector
+		// empty messages means they've been pseudo deleted by earlier phases and the report is still valid
+		if !msg.IsEmpty() && report.SourceChain != msg.Header.SourceChainSelector {
```
makramkd, line 41: I'd still vote for just setting `SourceChainSelector` in the "empty" message instead, unless this introduces other problems

pkg/types/ccipocr3/generic_types.go
```diff
@@ -204,8 +204,8 @@ func (m Message) String() string {
 
 // IsEmpty returns true if the message is empty. Can't use == Message{} without using reflect.DeepEqual.
 func (m Message) IsEmpty() bool {
```
makramkd, line 206: Can we check where this is used so we don't accidentally introduce a bug due to change of semantics?

makramkd: Actually I'd rather have a new method called `IsPseudoDeleted` because `IsEmpty` currently doesn't actually check emptiness.

makramkd (changes requested):

execute/observation.go
asoliman92: I believe using the map is better and more expressive.

execute/observation.go
asoliman92: good catch

execute/report/roots.go
asoliman92: I'd rather not. I can change `IsEmpty` to smth like `IsPseudoDeleted` with some comments about the meaning of being empty like that.

asoliman92: I tried removing the `SourceChainSelector` from the `IsPseudoDeleted` check but that caused a lot of tests to fail. It's not a prio to do it and fix it for now IMO.

PR #814, opened 2025-04-08
[Solana][Dedaub L18] Missing checks in validate_dest_chain_config

PabloMansanet:


chains/solana/contracts/programs/fee-quoter/src/instructions/v1/admin.rs
```diff
@@ -229,9 +230,14 @@ fn validate_dest_chain_config(dest_chain_selector: u64, config: &DestChainConfig
         config.default_tx_gas_limit <= config.max_per_msg_gas_limit,
         FeeQuoterError::DefaultGasLimitExceedsMaximum
     );
+
     require!(
-        config.chain_family_selector != [0; 4],
+        matches!(
+            u32::from_be_bytes(config.chain_family_selector),
+            CHAIN_FAMILY_SELECTOR_EVM | CHAIN_FAMILY_SELECTOR_SVM
```
agusaldasoro, line 237: can we add a test to verify this check?

PabloMansanet: Hmm... I generally don't like tests that simply verify top level function guards, as the tests end up being pretty tautological (in other words, simply re-stating what the code actually does). Since the logic here is pretty straight forward, I don't see a lot of value in a test that simply declares an invalid config and asserts that an error is returned.

Beyond that, there was no preexisting test in this file checking any of the other checks, so it seems strange to add one specifically for the chain family selector.

Still, If you think it's worth it, I can add tests for the entire function (including the other checks).

jadepark-dev (dismissed):

jadepark-dev (approved):

PR #822, opened 2025-04-09
Makefile: fix `make test` to exclude gobindings

makramkd:
The gobindings were being included in coverage calculations, exclude them.


makramkd:
Dup of #821

PR #825, opened 2025-04-09
add simple foundry scripts for off-chain testing

jhweintraub:
To better perform live testing on deployed contracts this PR creates some simple foundry scripts instead of relying on transporter, etherscan, or go scripts.

1. Sending a message via CCIP
2. Manually executing a message on a destination chain
3.

### DIFF
--- a/chains/evm/.solhintignore
+++ b/chains/evm/.solhintignore
@@ -1,3 +1,4 @@
 # Test files run with a different solhint ruleset, ignore them here.
 ./**/*.t.sol
+./**/*.s.sol
 ./node_modules/
--- a/chains/evm/contracts/script/CCIPManualExecutionScript.s.sol
+++ b/chains/evm/contracts/script/CCIPManualExecutionScript.s.sol
@@ -1,0 +1,61 @@
+pragma solidity ^0.8.24;
+
+import {Internal} from "../libraries/Internal.sol";
+import {OffRamp} from "../offRamp/OffRamp.sol";
+
+import {IERC20} from "@vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import {Script} from "forge-std/Script.sol";
+import {Test} from "forge-std/Test.sol";
+
+// solhint-disable-next-line no-console
+import {console2 as console} from "forge-std/console2.sol";
+
+/* solhint-disable no-console */
+contract CCIPSendTestScript is Script, Test {
+  using SafeERC20 for IERC20;
+
+  // Ex: "ETHEREUM_RPC_URL" as defined in .env
+  string public RPC_IDENTIFIER;
+  
+  OffRamp public s_offRamp;
+
+  bytes32 public s_messageId;
+
+  uint64 public s_sourceChainSelector;
+  uint64 public s_sequenceNumber;
+  bytes public s_manualExecutionData;
+
+  function run() public {
+    vm.createSelectFork(RPC_IDENTIFIER);
+
+    uint256 privateKey = vm.envUint("PRIVATE_KEY");
+
+    address sender = vm.rememberKey(privateKey);
+
+    vm.startBroadcast(privateKey);
+
+    console.log("Sender: %s", sender);
+    console.log("Starting Script...");
+
+    // Check that the messageId is not empty
+    Internal.MessageExecutionState executionState = s_offRamp.getExecutionState(s_sourceChainSelector, s_sequenceNumber);
+    assertTrue(
+      executionState == Internal.MessageExecutionState.FAILURE
+        || executionState == Internal.MessageExecutionState.UNTOUCHED,
+      "Message is not ready for Manual Execution"
+    );
+
+    // Manual Execution data can be invoked from a different tool or front-end to avoid having to 
+    // gather execution report data manually
+    (bool success,) = address(s_offRamp).call(s_manualExecutionData);
+    assertTrue(success, "Manual execution call reverted");
+
+    executionState = s_offRamp.getExecutionState(s_sourceChainSelector, s_sequenceNumber);
+    assertTrue(executionState == Internal.MessageExecutionState.SUCCESS, "Message was not executed successfully");
+
+    console.log("Script completed...");
+  }
+}
+/* solhint-enable no-console */
--- a/chains/evm/contracts/script/CCIPSendTestScript.s.sol
+++ b/chains/evm/contracts/script/CCIPSendTestScript.s.sol
@@ -1,0 +1,102 @@
+pragma solidity ^0.8.24;
+
+import {Router} from "../Router.sol";
+import {Client} from "../libraries/Client.sol";
+
+import {IERC20} from "@vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import {Script} from "forge-std/Script.sol";
+
+/* solhint-disable-next-line no-console */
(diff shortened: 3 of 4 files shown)


0xsuryansh (approved):

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
```diff
@@ -0,0 +1,62 @@
+pragma solidity ^0.8.24;
+
+import {Internal} from "../libraries/Internal.sol";
+import {OffRamp} from "../offRamp/OffRamp.sol";
+
+import {IERC20} from "@chainlink/vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@chainlink/vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import {Script} from "forge-std/Script.sol";
+
```
RensR, line 10: nit: rm newline

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
```diff
@@ -0,0 +1,62 @@
...
+// solhint-disable-next-line no-console
+import {console2 as console} from "forge-std/console2.sol";
+
+/* solhint-disable no-console */
```
RensR, line 14: nit: comments should use // over /*

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
```diff
@@ -0,0 +1,62 @@
...
+    (bool success,) = address(s_offRamp).call(s_manualExecutionData);
+    if (!success) revert ManualExecutionFailed();
+
+    executionState = s_offRamp.getExecutionState(s_sourceChainSelector, s_sequenceNumber);
+    if (executionState != Internal.MessageExecutionState.SUCCESS) revert ManualExecutionFailed();
+
+    console.log("Script completed...");
+  }
+}
+/* solhint-enable no-console */
```
RensR, line 62: probably not required?

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
```diff
@@ -0,0 +1,62 @@
...
+  bytes public s_manualExecutionData;
+
+  function run() public {
+    vm.createSelectFork(RPC_IDENTIFIER);
+
+    uint256 privateKey = vm.envUint("PRIVATE_KEY");
+
+    address sender = vm.rememberKey(privateKey);
+
+    vm.startBroadcast(privateKey);
```
RensR, line 39: This only has to wrap the non-view calls right? So probably best to only have it cover the call

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
```diff
@@ -0,0 +1,62 @@
...
+  // Ex: "ETHEREUM_RPC_URL" as defined in .env
+  string public RPC_IDENTIFIER;
+
+  OffRamp public s_offRamp;
+
+  bytes32 public s_messageId;
+
+  uint64 public s_sourceChainSelector;
+  uint64 public s_sequenceNumber;
+  bytes public s_manualExecutionData;
```
RensR, line 30: I'm assuming a user is supposed to fill every param here? There's no docs on anything so not 100% sure. 

How useful is this script if you need to construct the tx manually?

chains/evm/contracts/scripts/CCIPSendTestScript.s.sol
```diff
@@ -0,0 +1,103 @@
+pragma solidity ^0.8.24;
+
+import {Router} from "../Router.sol";
+import {Client} from "../libraries/Client.sol";
+
+import {IERC20} from "@chainlink/vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@chainlink/vendor/openzeppelin-solidity/v5.0.2/contracts/token/ERC20/utils/SafeERC20.sol";
+
```
RensR, line 8: nit: rm newline

chains/evm/contracts/scripts/CCIPSendTestScript.s.sol
```diff
@@ -0,0 +1,103 @@
...
+  bytes public s_recipient;
+  bytes public s_data;
+
+  function run() public {
+    vm.createSelectFork(RPC_DESCRIPTOR);
+
+    uint256 privateKey = vm.envUint("PRIVATE_KEY");
+
+    address sender = vm.rememberKey(privateKey);
+    s_recipient = abi.encode(sender);
```
RensR, line 40: Why store this in storage? Why do we override the values that appeared to be the input params (storage vars)

chains/evm/contracts/scripts/CCIPSendTestScript.s.sol
```diff
@@ -0,0 +1,103 @@
...
+    // Create the EVMTokenAmount array and populate with the first token
+    Client.EVMTokenAmount[] memory tokens;
+    if (TOKEN0 != address(0)) {
+      tokens = new Client.EVMTokenAmount[](1);
+      tokens[0] = Client.EVMTokenAmount({token: TOKEN0, amount: TOKEN0_AMOUNT});
+    }
+
+    // Construct the EVM2AnyMessage
+    Client.EVM2AnyMessage memory message = Client.EVM2AnyMessage({
+      receiver: abi.encode(sender),
```
RensR, line 56: You define s_recipient but don't use it

chains/evm/contracts/scripts/CCIPSendTestScript.s.sol
```diff
@@ -0,0 +1,103 @@
...
+      data: s_data,
+      tokenAmounts: tokens,
+      feeToken: address(0),
```
RensR, line 59: You don't use FEE_TOKEN

RensR (changes requested):

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
0xsuryansh: Better to inject the values using env 
We can make a .env.example file which shows all the keys which will be used to inject the values to these variables

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
```diff
@@ -0,0 +1,62 @@
...
+
+  function run() public {
+    vm.createSelectFork(RPC_IDENTIFIER);
```
0xsuryansh, line 33: Use consistent key name across scripts `RPC_IDENTIFIER` or `RPC_DESCRIPTOR`

chains/evm/contracts/scripts/CCIPSendTestScript.s.sol
```diff
@@ -0,0 +1,103 @@
...
+
+  uint64 public DESTINATION_CHAIN_SELECTOR;
+  uint64 public SOURCE_CHAIN_SELECTOR;
+
+  // Ex: "ETHEREUM_RPC_URL" as defined in .env
+  string public RPC_DESCRIPTOR;
+
+  bytes public s_extraArgs;
+  bytes public s_recipient;
+  bytes public s_data;
```
0xsuryansh, line 32: Better to keep these as memory vars and read through env

chains/evm/contracts/scripts/CCIPManualExecutionScript.s.sol
jhweintraub: 1. There's no way to get the sequence number dynamically on chain from the script it needs to be provided manually. I  will include some comments for the user that they can get the sequencer number from the CCIP explorer front-end as well. I don't think using an ENV is any simpler if you still have to fill in the data it just adds more friction if you need to change things. 
2. The script is useful for simulating and debugging manual executions in foundry locally. During the Ronin situation recently we had to test manual executions. You can get manual execution data from a variety of other tools and I will include links to those on how to use them.

PR #832, opened 2025-04-10
Ensure that pool state version is initialized, and provide workaround for when it is not

toblich:
We were missing the initialization of the `state.version = 1` field when initializing a token pool. This PR thus makes the following changes:
- It ensures that _every field_ is set during initialization. By using `ctx.accounts.state.set_inner(State{...})` the compiler will force us to set every field explicitly
- It adds a method to initialize just that field in the case where the state account was initialized but that field was not. This happens. This is permissionless so that it can be easily invoked, and it performs the necessary checks to prevent a reinitialization attack (there's a new integration test explicitly checking this).
- It enables integration tests on both the burnmint & lockrelease pools. Previously we were just testing the burnmint one.


chains/solana/contracts/tests/examples/pools_test.go
```diff
@@ -101,13 +101,22 @@ func TestBaseTokenPoolHappyPath(t *testing.T) {
...
+			{poolName: "lockrelease", poolProgram: config.CcipBasePoolLockRelease},
+		} {
+			t.Run(p.poolName+" "+v.tokenName, func(t *testing.T) {
+				t.Parallel()
+
+				decimals := uint8(0)
+				amount := uint64(1000)
+				poolProgram := p.poolProgram
+
+				// for _, poolProgram := range []solana.PublicKey{config.CcipBasePoolBurnMint} {
```
PabloMansanet, line 119: This comment can be removed now.

PabloMansanet (approved):

PR #837, opened 2025-04-14
Optimize execute plugin observation sequential CR calls

dimkouv:
⚠️ The execute plugin was making sequential calls to get messages from the CR.
✅ In this PR we switch to making them in parallel.
💡 Ideally we would like to make a single DB query but CR does not allow us to do that.

### DIFF
--- a/execute/observation.go
+++ b/execute/observation.go
@@ -3,6 +3,8 @@ package execute
 import (
 	"context"
 	"fmt"
+	"sort"
+	"sync"
 	"time"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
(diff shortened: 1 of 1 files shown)


asoliman92 (changes requested):
This logic is being changed in this PR:
https://github.com/smartcontractkit/chainlink-ccip/pull/812

execute/observation.go
```diff
@@ -291,35 +295,58 @@ func readAllMessages(
...
-			}
-
-			for _, msg := range msgs {
-				messageObs[srcChain][msg.Header.SequenceNumber] = msg
-				messageTimestamps[msg.Header.MessageID] = report.Timestamp
-			}
-			availableReports[srcChain] = append(availableReports[srcChain], report)
+		for _, commitRoot := range reports {
+			wg.Add(1)
+			go func(commitRoot exectypes.CommitData) {
```
makramkd, line 300: Just some concern here: can't `reports` be pretty big? like lets say 100, this loop would spin up 100 goroutines, all would query DB in parallel?

makramkd: One idea would be to cap it to like 10 at a time using `sync.Pool` but I'm not sure of the performance characteristics

PR #843, opened 2025-04-15
Commit-RMN: Handle missing signed roots

dimkouv:
## About
✅ This PR prevents contract reverts caused by missing roots that are signed by RMN but not agreed upon by the oracles. It also ensures continued progress with roots from RMN-disabled chains.

## Context
An RMN signature may cover multiple lane updates (i.e., multiple roots).
⚠️ If any of the signed roots are not part of the commit plugin’s consensus, the contract tx will revert — because the RMN signature cannot be verified for only a subset of the roots.

To handle this safely, we now skip all roots associated with RMN-enabled chains when any root in the RMN signature is missing. This allows us to proceed with roots from RMN-disabled chains and avoids reverting reports.

### DIFF
--- a/commit/chainfee/outcome.go
+++ b/commit/chainfee/outcome.go
@@ -87,7 +87,10 @@ func (p *processor) Outcome(
 		return gasPrices[i].ChainSel < gasPrices[j].ChainSel
 	})
 
-	lggr.Infow("Gas Prices Outcome", "gasPrices", gasPrices)
+	lggr.Infow("Gas Prices Outcome",
+		"gasPrices", gasPrices,
+		"consensusTimestamp", consensusObs.TimestampNow,
+	)
 	out := Outcome{GasPrices: gasPrices}
 	return out, nil
 }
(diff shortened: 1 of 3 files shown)


commit/merkleroot/outcome.go
```diff
@@ -219,48 +219,99 @@ func buildMerkleRootsOutcome(
...
+		}
+		proposedRootsSet.Add(rootKey{
+			ChainSel:      root.ChainSel,
+			SeqNumsRange:  root.SeqNumsRange,
+			MerkleRoot:    root.MerkleRoot,
+			OnRampAddress: addrStr,
+		})
+	}
+	if !signedRoots.IsSubset(proposedRootsSet) {
+		lggr.Errorw("signed roots are not a subset of proposed roots, skipping RMN-enabled roots",
```
carte7000, line 256: Just for my understanding: Is this indicative of a bigger problem? i.e. RMN does not compute the same root as the commit plugin?

carte7000 (approved):

carte7000 (approved):

mateusz-sekara (approved):
I would love to see some e2e tests (either in Chainlink repository or in this (we have some pseudo e2e tests)) to cover these edge cases

PR #846, opened 2025-04-16
Disallow Aptos precompile range and cut release for FeeQuoter

RensR:


chains/evm/contracts/FeeQuoter.sol
```diff
@@ -169,7 +169,7 @@ contract FeeQuoter is AuthorizedCallers, IFeeQuoter, ITypeAndVersion, IReceiver,
   /// @dev The decimals that Keystone reports prices in.
   uint256 public constant KEYSTONE_PRICE_DECIMALS = 18;
 
-  string public constant override typeAndVersion = "FeeQuoter 1.6.1-dev";
```
RensR, line 172: This is intentional

makramkd: This and the below cause a bytecode change w/ what was deployed to mainnets, is that desirable?

PR #849, opened 2025-04-16
Make token transfer use the same fee_billing_signer to invoke the tokenprogram transfer instruction [NONEVM-1645]

toblich:
Before this change, we were using 2 different PDAs for invoking the TokenProgram to transfer tokens:
- `fee_billing_signer` to transfer fee-related tokens
- `ccip_router_pool_signer` to transfer tokens to the token pool (which would then lock/burn them)

But the TokenProgram actually only supports setting one delegate at a time, meaning that if a user wants to bridge a token (e.g. LINK) and pay fees with that same token, it was impossible for them to simultaneously approve both those PDAs to make the transfers work. 

With this change, now it's a single PDA (`fee_billing_signer`) making both calls to the token program to transfer the tokens, so the user now has to delegate tokens to that single PDA.

In the future we may rename the `fee_billing_signer`, as it now does more than just fee-related actions, but that is a backwards-incompatible change as it would change its seeds


agusaldasoro (dismissed):

PabloMansanet (approved):
Looks good!

agusaldasoro (approved):

PabloMansanet (approved):

PR #851, opened 2025-04-17
Modernize the codebase

archseer:
Autofixed via `gopls` `modernize`
$ go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...


chains/evm/gobindings/generation/abigen.go
```diff
@@ -387,32 +387,32 @@ func writeAdditionalMethods(contractName string, logNames []string, abi abi.ABI,
 `, contractName, logName, contractName, logName)
 		}
 
-		bs = append(bs, []byte(fmt.Sprintf(`
```
archseer, line 390: Need to revert this file

PR #860, opened 2025-04-18
Narrow exec report window ccip 5351

asoliman92:
core ref: 8d419377d36a1f2c7730dc5f3bb75ae26c98ffc8

Currently if we have # commit reports with no merkleRoots = query limit for getting reports the plugin will get stuck re reading same reports until the earliest empty report timestamp is before the starting window for querying.

To fix the above issue we've implemented the following: 

* Add an extra query that checks only finalized reports and keep latest timestamp for empty root reports
* Reuse [commitRootsCache](https://github.com/smartcontractkit/chainlink-ccip/blob/2551e92d1fda6748873b33e8a327237d6a7f0bf0/execute/internal/cache/commit_root_cache.go#L111-L111) to store the latestEmptyRoot timestamp
* In GetTimestampToQueryFrom choose the lowest timestamp between latestEmptyRoot and earliestUnexecutedRoot. This will ensure that the plugin is moving forward even if number of empty reports is large without missing earlier reports that still need executing.

Thanks @0xnogo for the diagram
Screenshot 2025-04-18 at 4 30 25 PM

### DIFF
--- a/cmd/carpenter/internal/format/summary/commit.go
+++ b/cmd/carpenter/internal/format/summary/commit.go
@@ -181,7 +181,7 @@ func commitReportSummary(logs []*parse.Data) string {
 	if len(reportParts) > 0 || numReports != "" {
 		var buf strings.Builder
 		buf.WriteString(padding)
-		buf.WriteString(section.Render("UnfinalizedReports"))
+		buf.WriteString(section.Render("Reports"))
 		buf.WriteString(": ")
 		buf.WriteString(number.Render(numReports))
 		if numReports != "" && len(reportParts) == 0 {
--- a/cmd/carpenter/internal/format/summary/commit.go
+++ b/cmd/carpenter/internal/format/summary/commit.go
@@ -181,7 +181,7 @@ func commitReportSummary(logs []*parse.Data) string {
 	if len(reportParts) > 0 || numReports != "" {
 		var buf strings.Builder
 		buf.WriteString(padding)
-		buf.WriteString(section.Render("Reports"))
+		buf.WriteString(section.Render("UnfinalizedReports"))
 		buf.WriteString(": ")
 		buf.WriteString(number.Render(numReports))
 		if numReports != "" && len(reportParts) == 0 {
--- a/commit/report.go
+++ b/commit/report.go
@@ -98,7 +98,7 @@ func (p *Plugin) Reports(
 		lggr.Errorw("failed to build reports",
 			"outcome", outcome,
 			"err", err)
-		return nil, fmt.Errorf("err in Reports(): %w", err)
+		return nil, fmt.Errorf("err in UnfinalizedReports(): %w", err)
 	}
 
 	encodedReports, err := encodeReports(ctx, lggr, reports, transmissionSchedule, p.reportCodec)
--- a/commit/report.go
+++ b/commit/report.go
@@ -98,7 +98,7 @@ func (p *Plugin) Reports(
 		lggr.Errorw("failed to build reports",
 			"outcome", outcome,
 			"err", err)
-		return nil, fmt.Errorf("err in UnfinalizedReports(): %w", err)
+		return nil, fmt.Errorf("err in Reports(): %w", err)
 	}
 
 	encodedReports, err := encodeReports(ctx, lggr, reports, transmissionSchedule, p.reportCodec)
--- a/execute/internal/cache/commit_root_cache.go
+++ b/execute/internal/cache/commit_root_cache.go
@@ -193,39 +193,40 @@ func (r *commitRootsCache) isExecuted(key string) bool {
 func (r *commitRootsCache) GetTimestampToQueryFrom() time.Time {
 	r.cacheMu.RLock()
 	defer r.cacheMu.RUnlock()
-
 	// Calculate current visibility window based on stored interval
 	messageVisibilityWindow := r.timeProvider.Now().Add(-r.messageVisibilityInterval).UTC()
-	minTimestamp := messageVisibilityWindow
-
+	// Start with message visibility window as the default (lower bound)
+	commitRootsFilterTimestamp := messageVisibilityWindow
+	// Determine the earliest timestamp between unexecuted root and latest empty root
+	var minTimestamp time.Time
 	// Check if earliestUnexecutedRoot is set
-	if r.earliestUnexecutedRoot.After(minTimestamp) {
+	if !r.earliestUnexecutedRoot.IsZero() {
 		minTimestamp = r.earliestUnexecutedRoot
 		r.lggr.Debugw("MinTimeStamp set to earliest unexecuted root")
 	}
-
 	// Check if latestEmptyRoot is set
-	if r.latestEmptyRoot.After(minTimestamp) {
-		minTimestamp = r.latestEmptyRoot
+	if !r.latestEmptyRoot.IsZero() {
+		// If minTimestamp is not set or latestEmptyRoot is earlier, use latestEmptyRoot
+		if minTimestamp.IsZero() || r.latestEmptyRoot.Before(minTimestamp) {
+			minTimestamp = r.latestEmptyRoot
+		}
 		r.lggr.Debugw("MinTimeStamp set to latest empty finalized root")
 	}
-
-	// If we know the earliest unexecuted root or latestEmptyTimestamp and it's AFTER the visibility window,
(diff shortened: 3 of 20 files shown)


cmd/carpenter/internal/format/summary/commit.go
```diff
@@ -181,7 +181,7 @@ func commitReportSummary(logs []*parse.Data) string {
 	if len(reportParts) > 0 || numReports != "" {
 		var buf strings.Builder
 		buf.WriteString(padding)
-		buf.WriteString(section.Render("Reports"))
+		buf.WriteString(section.Render("UnfinalizedReports"))
```
asoliman92, line 184:
```suggestion
		buf.WriteString(section.Render("Reports"))
```

commit/report.go
```diff
@@ -98,7 +98,7 @@ func (p *Plugin) Reports(
 		lggr.Errorw("failed to build reports",
 			"outcome", outcome,
 			"err", err)
-		return nil, fmt.Errorf("err in Reports(): %w", err)
+		return nil, fmt.Errorf("err in UnfinalizedReports(): %w", err)
```
asoliman92, line 101:
```suggestion
		return nil, fmt.Errorf("err in Reports(): %w", err)
```

execute/plugin.go
```diff
@@ -258,10 +257,10 @@ func removeUnconfirmedAndFinalizedMessages(
 // 1. Gets all executed messages (both finalized and unfinalized) via primitives.Unconfirmed
 // 2. Gets only finalized executed messages via primitives.Finalized
 //
-// Reports are then classified as:
+// UnfinalizedReports are then classified as:
```
asoliman92, line 260:
```suggestion
// Reports are then classified as:
```

execute/plugin_e2e_test.go
```diff
@@ -42,7 +42,82 @@ func TestPlugin(t *testing.T) {
 	outcome := runRoundAndGetOutcome(ctx, ocrTypeCodec, t, runner)
 	require.Equal(t, exectypes.Initialized, outcome.State)
 
-	// Round 1 - Get Commit Reports
+	// Round 1 - Get Commit UnfinalizedReports
```
asoliman92, line 45:
```suggestion
	// Round 1 - Get Commit Reports
```

execute/plugin_tokendata_test.go
```diff
@@ -89,7 +89,7 @@ func Test_USDC_Transfer(t *testing.T) {
 	outcome := runRoundAndGetOutcome(ctx, ocrTypeCodec, t, runner)
 	require.Equal(t, exectypes.Initialized, outcome.State)
 
-	// Round 1 - Get Commit Reports
+	// Round 1 - Get Commit UnfinalizedReports
```
asoliman92, line 92:
```suggestion
	// Round 1 - Get Commit Reports
```

execute/plugin_test.go
```diff
@@ -1013,7 +1015,7 @@ func TestPlugin_ValidateObservation_CallsDiscoveryValidateObservation(t *testing
 		ocrTypeCodec: ocrTypeCodec,
 	}
 
-	// Reports with duplicate roots.
+	// UnfinalizedReports with duplicate roots.
```
asoliman92, line 1018:
```suggestion
	// Reports with duplicate roots.
```

execute/plugin_test.go
```diff
@@ -974,7 +976,7 @@ func TestPlugin_ValidateObservation_ValidateObservedSeqNum_Error(t *testing.T) {
...
```
asoliman92, line 979:
```suggestion
	// Reports with duplicate roots.
```

execute/plugin_functions.go
```diff
@@ -318,7 +319,7 @@ func groupByChainSelectorWithFilter(
 }
 
 // combineReportsAndMessages returns a new reports slice with fully executed messages removed.
-// Reports that have all of their messages executed are not included in the result.
+// UnfinalizedReports that have all of their messages executed are not included in the result.
```
asoliman92, line 322:
```suggestion
// Reports that have all of their messages executed are not included in the result.
```

execute/plugin.go
```diff
@@ -261,7 +260,7 @@ func removeUnconfirmedAndFinalizedMessages(
 // Reports are then classified as:
 // - fullyExecutedFinalized: All messages executed with finality (mark as executed)
 // - fullyExecutedUnfinalized: All messages executed but not finalized (snooze)
-// - groupedCommits: Reports with unexecuted messages (available for execution)
+// - groupedCommits: UnfinalizedReports with unexecuted messages (available for execution)
```
asoliman92, line 263:
```suggestion
// - groupedCommits: Reports with unexecuted messages (available for execution)
```

execute/internal/cache/commit_root_cache.go
```diff
@@ -183,17 +198,36 @@ func (r *commitRootsCache) GetTimestampToQueryFrom() time.Time {
...
+	// Check if latestEmptyRoot is set
+	if !r.latestEmptyRoot.IsZero() {
+		// If minTimestamp is not set or latestEmptyRoot is earlier, use latestEmptyRoot
+		if minTimestamp.IsZero() || r.latestEmptyRoot.Before(minTimestamp) {
+			minTimestamp = r.latestEmptyRoot
+		}
+		r.lggr.Debugw("MinTimeStamp set to latest empty finalized root")
+	}
+
+	// If we know the earliest unexecuted root or latestEmptymTimestamp and it's AFTER the visibility window,
```
winder, line 219:
```suggestion
	// If we know the earliest unexecuted root or latestEmptyTimestamp and it's AFTER the visibility window,
```

execute/internal/cache/commit_root_cache.go
```diff
@@ -183,17 +198,36 @@ func (r *commitRootsCache) GetTimestampToQueryFrom() time.Time {
 	// Start with message visibility window as the default (lower bound)
 	commitRootsFilterTimestamp := messageVisibilityWindow
 
-	// If we know the earliest unexecuted root and it's AFTER the visibility window,
+	// Determine the earliest timestamp between unexecuted root and latest empty root
+	var minTimestamp time.Time
```
winder, line 202: I think you can simplify the logic a little. This psudocode probably messes up durations and times, but hopefully gets the idea across.

```
minTimestamp := time.Now().Minus(messageVisibilityWindow)
...
if r.earliestUnexecutedRoot.After(minTimestamp) {
    minTimestamp = r.earliestUnexecutedRoot
}
if r.latestEmptyRoot.After(minTimestamp) {
    minTimestamp = r.latestEmptyRoot
}
return minTimestamp
```

pkg/reader/ccip.go
```diff
@@ -173,40 +173,33 @@ type RMNCurseResponse struct {
 
 // ---------------------------------------------------
 
-func (r *ccipChainReader) CommitReportsGTETimestamp(
-	ctx context.Context, ts time.Time, limit int,
-) ([]cciptypes.CommitPluginReportWithMeta, error) {
-	lggr := logutil.WithContextValues(ctx, r.lggr)
+func (r *ccipChainReader) CommitReportsGTETimestamp(ctx context.Context, ts time.Time, limit int) (cciptypes.CommitReportsByConfidenceLevel, error) {
```
winder, line 176: I don't think you should change the return type like this. The pattern used for `ExecutedMessages` where the confidence level is passed in as a parameter is simpler.

mateusz-sekara (commented):
I think this solution is still vulnerable to the price spamming. Let's take the example from the PR description. The only difference is that Root 3 contains messages that cannot be executed now (because of slow USDC/LBTC attestation which is a valid case). Afther Root3, there are a lot of price updates. Until reaching messageVisibilityInterval you won't know about Root1501, because you will keep asking about roots starting from Root3's timestamp

image

execute/internal/cache/commit_root_cache.go
asoliman92: Thanks to your comment I figured we're missing an extra test 😁. We can't really do it like that because we need to pick up the earliest of them and make sure that this earliest is not actually zero. I reverted to the original implementation after finding the bug in the new code.

asoliman92:
We talked about this offline. This is an edge case on top of an edge case. Of course best course of action is to exclude empty reports from the query but that's not possible with ContractReader constraints. We agreed this solution is good enough for now.

0xnogo (approved):

PR #861, opened 2025-04-20
[CCIP-5718] pkg/reader: batch calls in GetChainFeePriceUpdate

makramkd:
Also added tests for various scenarios including happy path, empty selectors, batch call errors, partial successes, and missing results.

Integration tests enhanced here: https://github.com/smartcontractkit/chainlink/pull/17363

### DIFF
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -936,33 +936,119 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors 
 	lggr := logutil.WithContextValues(ctx, r.lggr)
 	if err := validateExtendedReaderExistence(r.contractReaders, r.destChain); err != nil {
 		lggr.Errorw("GetChainFeePriceUpdate dest chain extended reader not exist", "err", err)
+		// Return nil as per original logic if reader doesn't exist
 		return nil
 	}
 
+	// Initialize the result map once at the beginning
 	feeUpdates := make(map[cciptypes.ChainSelector]cciptypes.TimestampedBig, len(selectors))
+
+	if len(selectors) == 0 {
+		return feeUpdates // Return the initialized empty map
+	}
+
+	// 1. Build Batch Request
+	contractBatch := make([]types.BatchRead, 0, len(selectors))
 	for _, chain := range selectors {
-		update := cciptypes.TimestampedUnixBig{}
-		// Read from dest chain
-		err := r.contractReaders[r.destChain].ExtendedGetLatestValue(
-			ctx,
-			consts.ContractNameFeeQuoter,
-			consts.MethodNameGetFeePriceUpdate,
-			primitives.Unconfirmed,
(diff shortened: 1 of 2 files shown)


pkg/reader/ccip.go
```diff
@@ -936,33 +936,119 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
+}
+
+// processFeePriceUpdateResults iterates through batch results, validates them,
+// and populates the feeUpdates map.
+func (r *ccipChainReader) processFeePriceUpdateResults(
+	lggr logger.Logger,
+	selectors []cciptypes.ChainSelector,
+	results []types.BatchReadResult,
+	feeUpdates map[cciptypes.ChainSelector]cciptypes.TimestampedBig,
+) map[cciptypes.ChainSelector]cciptypes.TimestampedBig {
```
0xnogo, line 1012: I think the map is mutated in place. If so, we could drop the return.

pkg/reader/ccip.go
```diff
@@ -936,33 +936,119 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
+			lggr.Warnw("failed to get chain fee price update from batch result",
+				"chain", chain,
+				"err", err)
 			continue
 		}
+
+		// Type assert the result
+		update, ok := val.(*cciptypes.TimestampedUnixBig)
+		if !ok || update == nil {
+			lggr.Errorw("Invalid type or nil value received for chain fee price update",
```
0xnogo, line 1035: I'd rather go with a `Warnw` as the function is dealing gracefully with that.

pkg/reader/ccip.go
```diff
@@ -936,33 +936,119 @@ func (r *ccipChainReader) GetChainFeePriceUpdate(ctx context.Context, selectors
...
+	// 3. Find FeeQuoter Results
+	var feeQuoterResults []types.BatchReadResult
+	found := false
+	for contract, results := range batchResult {
+		if contract.Name == consts.ContractNameFeeQuoter {
+			feeQuoterResults = results
+			found = true
+			break // Found the results, exit loop
+		}
+	}
```
0xnogo, line 987: Can we do something like this instead:

```
feeQuoterResults, found := batchResult[types.BoundContract{
    Name: consts.ContractNameFeeQuoter,
}]
if !found {
    lggr.Errorw("FeeQuoter results missing from batch response")
    return feeUpdates
}
```

0xnogo (commented):
Minor comments

pkg/reader/ccip.go
makramkd: I don't think so cuz `BoundContract` has an address as well?

pkg/reader/ccip.go
makramkd: Ah nice find, don't like that actually, will just init the map and return it.

PR #867, opened 2025-04-24
allow zero usd per unit gas

dimkouv:
`CCIP-5748`


asoliman92 (changes requested):
That will allow zero gas for evms as well. We need a way to identify the family before allowing it like that.

asoliman92 (approved):

makramkd (approved):

PR #870, opened 2025-04-24
Add type_version method to contracts [NONEVM-1646]

toblich:
This will allow observability and tooling (and anyone who needs it) to know which version of the binary is deployed at any time.


chains/solana/contracts/crates/build-commit/src/lib.rs
```diff
@@ -0,0 +1,67 @@
+use std::process::Command;
+
+pub fn cargo_instructions(source_file: &str) {
+    let program_name = source_file
+        .split('/')
+        .nth_back(1)
+        .unwrap()
+        .trim()
+        .to_string();
```
PabloMansanet, line 9: I think using `Path` would be a bit cleaner here, but feel free to ignore.
```
Path::from(source_file).parent().file_name();
```

(paraphrasing from memory, you might need some unwraps in there.

PabloMansanet (dismissed):
Nice! Seems clean and straightforward.

PR #871, opened 2025-04-24
Catch overflow on sender

toblich:


archseer:
We do enable overflow checks here but maybe good to be explicit:

https://github.com/smartcontractkit/chainlink-ccip/blob/main/chains/solana/contracts/Cargo.toml#L12

PabloMansanet (approved):

PR #876, opened 2025-04-28
Set CCIP_BUILD_GIT_HASH when building release

archseer:


tt-cll (dismissed):

jadepark-dev (dismissed):

tt-cll (approved):

jadepark-dev (approved):

toblich (approved):
Thanks for the fix @archseer 👍 

Just for traceability: this fixes a build step missed in #870 (given that this is not invoking Anchor, I missed this one)

tt-cll:
@toblich unfortunately this doesn't actually fix the issue because the hermetic builds don't have access to the env var. Summary in the slack thread and I'm opening a ticket

PR #878, opened 2025-04-28
Add CalculateUsdPerUnitGas to EstimateProvider interface

asoliman92:
This allows the usd per gas per chain to be calculated differently based on the chain family.

core ref: 777ca421ae5e95e95bed4e7cb62cdaccb26c7092

### DIFF
--- a/commit/chainfee/outcome_test.go
+++ b/commit/chainfee/outcome_test.go
@@ -1,6 +1,7 @@
 package chainfee
 
 import (
+	mock_ccipocr3 "github.com/smartcontractkit/chainlink-ccip/mocks/pkg/types/ccipocr3"
 	"math/big"
 	"testing"
 	"time"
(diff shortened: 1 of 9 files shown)


commit/chainfee/outcome_test.go
```diff
@@ -398,3 +410,79 @@ func TestProcessor_Outcome(t *testing.T) {
...
+		func(sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) *big.Int {
+			// Using 1e15 division instead of 1e18
+			tmp := new(big.Int).Mul(sourceGasPrice, usdPerFeeCoin)
+			return tmp.Div(tmp, big.NewInt(1e15))
+		}).Maybe()
+
+	// Create processor with custom estimate provider
+	p := &processor{
+		lggr:      lggr,
+		destChain: 1,
```
makramkd, line 437: Can we create `destChain` and `sourceChain` vars here to make the test more canonical and representative of a real situation

pkg/types/ccipocr3/interfaces.go
```diff
@@ -50,4 +51,5 @@ type TokenDataEncoder interface {
 type EstimateProvider interface {
 	CalculateMerkleTreeGas(numRequests int) uint64
 	CalculateMessageMaxGas(msg Message) uint64
+	CalculateUsdPerUnitGas(sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) *big.Int
```
makramkd, line 54: Actually I don't think this interface would work, you need to have a source chain selector argument somewhere

commit/chainfee/outcome_test.go
```diff
@@ -398,3 +410,79 @@ func TestProcessor_Outcome(t *testing.T) {
...
+		estimateProvider: estimateProvider,
+	}
+
+	// Create observation with timestamp old enough to trigger update
+	observation := Observation{
+		FeeComponents: map[cciptypes.ChainSelector]types.ChainFeeComponents{
+			1: {ExecutionFee: big.NewInt(100), DataAvailabilityFee: big.NewInt(200)},
+		},
+		NativeTokenPrices: map[cciptypes.ChainSelector]cciptypes.BigInt{
+			1: cciptypes.NewBigInt(big.NewInt(1e18)),
```
makramkd, line 453: We should have a test w/ multiple source chains that have different calculations (e.g solana and evm sources) to replicate the real-world use case.

makramkd (changes requested):

matYang (approved):

PR #882, opened 2025-04-28
Add type_version method to contracts [NONEVM-1646] [NONEVM-1681]

toblich:
This adds back the functionality of #870 that was reverted in #880. It now also supports setting the commit hash via a custom file, given that builds using the `solana-verify build` command don't have neither git nor access to the environment variables, but they do mount the filesystem.

For reviewers: It is easiest if you look at the diff skipping the first commit, which is the big one that is equivalent to #870, and look mainly at the other commit(s) that just add the new changes


toblich:
Verifiable build (using the action code from this branch) has passed: https://github.com/smartcontractkit/chainlink-ccip/actions/runs/14718031803

PR #883, opened 2025-04-28
WIP: multiple exec reports

winder:


pkg/ocrtypecodec/v1/exec.go
```diff
@@ -79,14 +79,20 @@ func (e *ExecCodecProto) DecodeObservation(data []byte) (exectypes.Observation,
...
+		CommitReports:        e.tr.commitDataSliceToProto(outcome.CommitReports),
+		ExecutePluginReports: e.tr.execPluginReportsToProto(outcome.Reports),
+	}
+
+	// If there is only one report, use the legacy field. This way new clients can still
+	// form consensus with old ones.
+	// TODO: Remove backwards compatibility code after a few releases.
+	if len(pbObs.ExecutePluginReports) == 1 {
+		pbObs.ExecutePluginReport = pbObs.ExecutePluginReports[0]
+		pbObs.ExecutePluginReports = nil
```
winder, line 95: The encode function ensures that encoded exec outcomes leverage the legacy field unless multiple reports are detected. This should ensure new deployments produce outcomes that are identical to the previous version.

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -102,12 +108,19 @@ func (e *ExecCodecProto) DecodeOutcome(data []byte) (exectypes.Outcome, error) {
...
 	}
 
+	var reports []cciptypes.ExecutePluginReport
+	if pbOutc.ExecutePluginReport != nil {
+		// Migrate legacy field by wrapping it in a slice and assigning it to the new field.
+		// TODO: Remove temporary migration code after a few releases.
+		reports = e.tr.execPluginReportsFromProto([]*ocrtypecodecpb.ExecutePluginReport{pbOutc.ExecutePluginReport})
+	} else {
+		reports = e.tr.execPluginReportsFromProto(pbOutc.ExecutePluginReports)
+	}
```
winder, line 118: The decode function wraps the legacy field in a slice so that old deployments produce outcomes that look like those expected by the new version.

execute/exectypes/outcome.go
```diff
@@ -56,16 +54,16 @@ type Outcome struct {
...
 	// * GetCommitReports: All pending commit reports which were observed.
 	// * GetMessages: All pending commit reports with messages.
-	// * Filter: Commit reports associated with the final execution report.
+	// * Filter: Commit reports associated with the final execution report, concatenated together.
 	CommitReports []CommitData `json:"commitReports"`
 
-	// Report is built from the oldest pending commit reports.
-	Report cciptypes.ExecutePluginReport `json:"report"`
+	// Reports are built from the oldest pending commit reports.
+	Reports []cciptypes.ExecutePluginReport `json:"reports"`
```
winder, line 61: The protobuf layer is managing the type migration, allowing the plugin layer to remove the legacy field.

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -79,14 +79,20 @@ func (e *ExecCodecProto) DecodeObservation(data []byte) (exectypes.Observation,
...
 
 	pbObs := &ocrtypecodecpb.ExecOutcome{
-		PluginState:   string(outcome.State),
-		CommitReports: e.tr.commitDataSliceToProto(outcome.CommitReports),
-		ExecutePluginReport: &ocrtypecodecpb.ExecutePluginReport{
-			ChainReports: e.tr.chainReportsToProto(outcome.Report.ChainReports),
-		},
+		PluginState:          string(outcome.State),
+		CommitReports:        e.tr.commitDataSliceToProto(outcome.CommitReports),
+		ExecutePluginReports: e.tr.execPluginReportsToProto(outcome.Reports),
```
makramkd, line 87: Hmm actually not sure if you can do this, it'll break oracles on the previous version since they won't know how to decode this field, and they'll just get an empty `ExecutePluginReport`

execute/exectypes/outcome.go
```diff
@@ -89,51 +89,31 @@ func (o Outcome) ToLogFormat() Outcome {
...
-		})
-	sort.Slice(
-		reportCP,
-		func(i, j int) bool {
-			return reportCP[i].SourceChainSelector < reportCP[j].SourceChainSelector
-		})
 	return Outcome{
 		State:         state,
-		CommitReports: pendingCommitsCP,
-		Report:        cciptypes.ExecutePluginReport{ChainReports: reportCP},
```
makramkd, line 137: Are we sure about removing these sorts? IIRC they were there for a reason

execute/outcome.go
```diff
@@ -199,11 +201,12 @@ func (p *Plugin) getFilterOutcome(
 		return exectypes.Outcome{}, fmt.Errorf("unable to select report: %w", err)
 	}
 
-	execReport := cciptypes.ExecutePluginReport{
-		ChainReports: outcomeReports,
+	// Collapse the commit reports.
+	var mergedData []exectypes.CommitData
+	for _, commitData := range selectedCommitReports {
+		mergedData = append(mergedData, commitData...)
 	}
```
makramkd, line 208: This is kinda weird, so `selectedCommitReports` is not already flattened? How did the types even work out?

execute/report/builder.go
```diff
@@ -126,47 +135,102 @@ type execReportBuilder struct {
...
-	maxGas                uint64
-	maxMessages           uint64
-	maxSingleChainReports uint64
+	checks                 []Check
+	destChainSelector      cciptypes.ChainSelector
+	maxReportSizeBytes     uint64
+	maxGas                 uint64
+	maxMessages            uint64
+	maxSingleChainReports  uint64
+	multipleReportsEnabled bool
```
makramkd, line 144: Should we also have a config that limits the # of reports built in a round? ]

execute/report/builder.go
```diff
@@ -126,47 +135,102 @@ type execReportBuilder struct {
...
 func (b *execReportBuilder) Add(
 	ctx context.Context,
 	commitReport exectypes.CommitData,
 ) (exectypes.CommitData, error) {
-	execReport, updatedReport, err := b.buildSingleChainReport(ctx, commitReport)
+	b.checkInitialize()
+	index := len(b.execReports) - 1
+
+	// Check if max is reached.
+	if b.maxSingleChainReports != 0 && len(b.execReports[index].ChainReports) >= int(b.maxSingleChainReports) {
```
makramkd, line 188: This checks the num of single-chain reports within a `Report` right? Not total number of `Report` objects (not single chain)

execute/exectypes/outcome.go
winder: Moderately sure. At least in the report builder (for the filter phase), it should return data in a canonical format. Maybe the getMessages / getCommitReports functions could still use this, but I think business logic around the ordering belongs in those functions.

execute/exectypes/outcome.go
```diff
@@ -56,16 +54,19 @@ type Outcome struct {
 	// the state:
 	// * GetCommitReports: All pending commit reports which were observed.
 	// * GetMessages: All pending commit reports with messages.
-	// * Filter: Commit reports associated with the final execution report.
+	// * Filter: Commit reports associated with the final execution report, concatenated together.
 	CommitReports []CommitData `json:"commitReports"`
 
 	// Report is built from the oldest pending commit reports.
 	Report cciptypes.ExecutePluginReport `json:"report"`
```
asoliman92, line 61: Keeping Report for comprehensive backward compatibility testing

winder: I still think this should go away. Backwards compatibility is done in the protobuf layer, no need for it to leak into the protocol layer.

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -79,38 +79,63 @@ func (e *ExecCodecProto) DecodeObservation(data []byte) (exectypes.Observation,
...
+		pbOtcm.ExecutePluginReports = nil
+	}
+
+	// TODO: Remove after "Reports" is fully supported.
+	if len(outcome.Report.ChainReports) != 0 {
+		r := e.tr.execPluginReportsToProto([]ccipocr3.ExecutePluginReport{outcome.Report})
+		if len(r) > 0 {
+			pbOtcm.ExecutePluginReport = r[0]
+		}
 	}
```
winder, line 104: This isn't needed because `Reports` is fully supported by the rest of this PR.

0xnogo (approved):

PR #884, opened 2025-04-28
Make CalculateUsdPerUnitGas chain family aware

asoliman92:
This allows the usd per gas per chain to be calculated differently based on the chain family.


internal/libs/mathslib/calc.go
```diff
@@ -24,8 +29,26 @@ func Deviates(x1, x2 *big.Int, ppb int64) bool {
...
+	}
+
+	switch family {
+	case chainsel.FamilyEVM:
+		// (wei / gas) * (usd / eth) * (1 eth / 1e18 wei)  = usd/gas
+		tmp := new(big.Int).Mul(sourceGasPrice, usdPerFeeCoin)
+		return tmp.Div(tmp, big.NewInt(1e18)), nil
+
+	case chainsel.FamilySolana:
+		// (micro lamport / compute units) * (usd / 1 sol) * (1 sol / 1e15 micro lamport)  = usd/cu
```
matYang, line 46: isn't this missing the fact that usdPerFeeCoin actually represents price of 1e9 full Sol, and you'd need to further scale down by 1e9?

internal/libs/mathslib/calc_test.go
```diff
@@ -105,26 +114,38 @@ func TestCalculateUsdPerUnitGas(t *testing.T) {
...
-			exp:            big.NewInt(0), // What do we do in these cases? Charge the user 0?
+			chainSelector:  EvmChainSelector, // evm
+			exp:            big.NewInt(0),    // What do we do in these cases? Charge the user 0?
+		},
+		{
+			name:           "sol base case",
+			sourceGasPrice: big.NewInt(2000),
+			usdPerFeeCoin:  big.NewInt(3e18),
+			chainSelector:  SolChainSelector, // sol
+			exp:            big.NewInt(6e6),
```
matYang, line 139: we can use a lot more test cases, eg add a few cases with real life numbers, like gas price of 0.0001gwei, 1gwei, 100gwei, Eth price of $2000, Sol price of $150, cu price of 0, or 1000 microlamports, or 100,000 microlamports

matYang (changes requested):

internal/libs/mathslib/calc.go
makramkd: Are you sure it represents that?

makramkd: Isn't it usd18 per full token in the fee quoter? This calculation also made up the 6 orders of magnitude difference that was observed

makramkd: Ah nvm its usd18 per 1e18 of smallest denomination, which is 1e-9 (lamport). So you're right.

internal/libs/mathslib/calc.go
```diff
@@ -24,8 +29,26 @@ func Deviates(x1, x2 *big.Int, ppb int64) bool {
 }
 
 // CalculateUsdPerUnitGas returns: (sourceGasPrice * usdPerFeeCoin) / 1e18
-func CalculateUsdPerUnitGas(sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) *big.Int {
-	// (wei / gas) * (usd / eth) * (1 eth / 1e18 wei)  = usd/gas
-	tmp := new(big.Int).Mul(sourceGasPrice, usdPerFeeCoin)
-	return tmp.Div(tmp, big.NewInt(1e18))
+func CalculateUsdPerUnitGas(sourceChainSelector ccipocr3.ChainSelector, sourceGasPrice *big.Int, usdPerFeeCoin *big.Int) (*big.Int, error) {
```
makramkd, line 32: Is there a good reason we're breaking the existing pattern by putting this logic in chainlink-ccip?

internal/libs/mathslib/calc.go
```diff
@@ -24,8 +29,26 @@ func Deviates(x1, x2 *big.Int, ppb int64) bool {
...
+		tmp := new(big.Int).Mul(sourceGasPrice, usdPerFeeCoin)
+		return tmp.Div(tmp, big.NewInt(1e18)), nil
+
+	case chainsel.FamilySolana:
+		// (micro lamport / compute units) * (usd / 1 sol) * (1 sol / 1e15 micro lamport)  = usd/cu
+		tmp := new(big.Int).Mul(sourceGasPrice, usdPerFeeCoin)
+		return tmp.Div(tmp, big.NewInt(1e15)), nil
+
+	default:
+		return nil, fmt.Errorf("unsupported family for extra args type %s", family)
```
makramkd, line 51:
```suggestion
		return nil, fmt.Errorf("unsupported family %s", family)
```

internal/libs/mathslib/calc_test.go
```diff
@@ -4,9 +4,18 @@ import (
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/chainlink-ccip/pkg/types/ccipocr3"
+
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	SolChainSelector = ccipocr3.ChainSelector(16423721717087811551)
+	EvmChainSelector = ccipocr3.ChainSelector(16015286601757825753)
```
makramkd, line 16: Instead of these random looking numbers you can import the exact selectors from the chain-selectors lib

internal/libs/mathslib/calc_test.go
```diff
@@ -105,26 +114,38 @@ func TestCalculateUsdPerUnitGas(t *testing.T) {
...
 			sourceGasPrice: big.NewInt(1000),
 			usdPerFeeCoin:  big.NewInt(2000),
-			exp:            big.NewInt(0), // What do we do in these cases? Charge the user 0?
+			chainSelector:  EvmChainSelector, // evm
+			exp:            big.NewInt(0),    // What do we do in these cases? Charge the user 0?
+		},
+		{
+			name:           "sol base case",
+			sourceGasPrice: big.NewInt(2000),
+			usdPerFeeCoin:  big.NewInt(3e18),
```
makramkd, line 137: Lets use realistic values here

internal/libs/mathslib/calc.go
asoliman92: Whatever the solution we do from Chainlink per chain we'll need a wrapper on top of them to switch the logic per family based on the chain selector. The logic for the calculation is one liner per family and making it from outside felt like an overkill IMO.

internal/libs/mathslib/calc.go
matYang: it's never been usd18 per full token onchain

internal/libs/mathslib/calc.go
```diff
@@ -23,9 +28,43 @@ func Deviates(x1, x2 *big.Int, ppb int64) bool {
...
+	case chainsel.FamilySolana:
+		// In SVM, sourceGasPrice is denoted in microlamports/cu, or 1e-15 SOL/cu.
+		// We need to multiply by 1e3 to be in units of 1e-18 SOL/cu, equivalent to wei/gas.
+
+		// Sol has 9 decimals, usdPerFeeCoin represents 1e18 USD per 1e9 whole sol, or 1e18 USD/1e27wei.
+		// We need to divide by 1e9 to be in units of 1e18 USD/1e18 wei, equivalent to EVM model.
+
+		// To keep precision, we want to divide at the very end.
+		// The above cancels out, the net result is we need to further divide by 1e6,
+		// or 1e18 * 1e6 = 1e24 in total.
```
makramkd, line 61: I feel like the formula that was there before was pretty clear and mirrors the code almost exactly:

```
// (micro lamport / compute units) * (usd18 / 1e9 sol) * (1e9 sol / 1e24 micro lamport)  = usd18/cu
// or
// (microlamport / cu) * (usd18 / 1e18 lamport) * (1 lamport / 1e6 microlamport)
```

internal/libs/mathslib/calc.go
```diff
@@ -23,9 +28,43 @@ func Deviates(x1, x2 *big.Int, ppb int64) bool {
...
+	family, err := chainsel.GetSelectorFamily(uint64(sourceChainSelector))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get chain family for selector %d: %w", sourceChainSelector, err)
+	}
+
+	switch family {
+	case chainsel.FamilyEVM:
+		// In EVM, sourceGasPrice is denoted in wei/gas.
+		// Eth has 18 decimals, usdPerFeeCoin represents 1e18 USD/1e18 wei.
+		// To get 1e18 USD/gas, we multiply sourceGasPrice by usdPerFeeCoin and divide by 1e18.
```
makramkd, line 48: Formula plz, with units.

commit/plugin_e2e_test.go
```diff
@@ -50,8 +49,8 @@ import (
 
 const (
 	destChain         = ccipocr3.ChainSelector(1)
-	sourceChain1      = ccipocr3.ChainSelector(2)
-	sourceChain2      = ccipocr3.ChainSelector(3)
+	sourceEvmChain1   = ccipocr3.ChainSelector(16015286601757825753)
+	sourceSolChain    = ccipocr3.ChainSelector(16423721717087811551)
```
makramkd, line 53: Use imports from chain-selectors like the other test file

makramkd (approved):

PR #885, opened 2025-04-29
do not skip gas price updates of 0

dimkouv:


pkg/reader/ccip.go
```diff
@@ -1031,7 +1031,7 @@ func (r *ccipChainReader) processFeePriceUpdateResults(
 		}
 
 		// Check if the update is empty (timestamp or value is zero/nil)
```
makramkd, line 1033: Update comment - zero value is fine

makramkd (approved):

PR #886, opened 2025-04-29
Commit Plugin - Support MultiPriceReports

dimkouv:
This PR introduces several changes to properly support the `MultiPriceReports` feature.

In this PR we drop the `mainOutcome` logic for inflight checks and we start using feeQuoter price updates to determine if reports have been finalized on-chain.

The new state looks like this:

```mermaid
flowchart TD
    Outcome --> CheckInflight?
    CheckInflight? --> |No| OutcomeNewPrices
    CheckInflight? --> |Yes| StillSomeInflight?
    StillSomeInflight? --> |Yes| OutcomeRetries-1
    StillSomeInflight? --> |No| EmptyOutcome
```

We also conditionally check in `ShouldAccept / ShouldTransmit` ocr3 phases how to verify if report is valid by either comparing against `report <= onchainSeqNum` or `report < onchainSeqNum` based on MultiPriceReports being enabled or not.


commit/chainfee/types.go
```diff
@@ -23,6 +23,9 @@ type Query struct {
 type Outcome struct {
 	// Each Gas Price is the combination of Execution and DataAvailability Fees using bitwise operations
 	GasPrices []cciptypes.GasPriceChain `json:"gasPrices"`
+
+	InflightChainFeeUpdates map[cciptypes.ChainSelector]time.Time `json:"inflightChainFeeUpdates"`
+	InflightRemainingChecks int64                                 `json:"inflightRemainingChecks"`
```
winder, line 28: I don't see any feature flag to prevent using these new fields until all nodes have been updated.

commit/chainfee/outcome.go
```diff
@@ -20,22 +20,28 @@ import (
...
 ) (Outcome, error) {
 	lggr := logutil.WithContextValues(ctx, p.lggr)
 
+	inflightPricesOutcome := newInflightPricesOutcome(
+		prevOutcome.InflightChainFeeUpdates, prevOutcome.InflightRemainingChecks-1)
+
 	consensusObs, err := p.getConsensusObservation(lggr, aos)
 	if err != nil {
-		return Outcome{}, fmt.Errorf("get consensus observation: %w", err)
+		return inflightPricesOutcome, fmt.Errorf("get consensus observation: %w", err)
```
makramkd, line 34: Wouldn't this be better left as empty? It should not get used anyway, if an error is returned.

commit/chainfee/outcome.go
```diff
@@ -20,22 +20,28 @@ import (
...
 	}
 
 	// No need to update yet
 	if len(consensusObs.FeeComponents) == 0 {
-		lggr.Warn("no consensus on fee components, nothing to update",
-			"consensusObs", consensusObs)
-		return Outcome{}, nil
+		lggr.Warn("no consensus on fee components, nothing to update", "consensusObs", consensusObs)
```
makramkd, line 39:
```suggestion
		lggr.Warnw("no consensus on fee components, nothing to update", "consensusObs", consensusObs)
```

commit/chainfee/outcome.go
```diff
@@ -101,8 +107,53 @@ func (p *processor) Outcome(
...
+}
+
+// computeInflightPricesOutcome is called if in this round we wait for prices to appear OnChain.
+// If we still wait for some prices it will decrement the number of available retries.
+// If all prices appeared OnChain or no retries left it sends an empty outcome so we can transmit fresh prices in the
+// next round.
+func (p *processor) computeInflightPricesOutcome(
+	lggr logger.Logger, consensusObs Observation, prevOutcome Outcome,
+) Outcome {
+	lggr.Debugw("checking for previously transmitted chain fee price updates to appear on-chain",
```
makramkd, line 131: I think this can be Infow also, it won't get spammed, just once per round thing if at all.

commit/chainfee/types.go
makramkd: Ah actually I think we'll have a problem with these Outcome type updates even if the feat flag is there, if only a subset of nodes are updated, unless protobuf encodes an outcome `Outcome{GasPrice}` identically to an outcome `Outcome{GasPrices, rest not specified}`. 

The approach that would seem to work is to feat flag the usage of these fields via offchainConfig, upgrade all the nodes (eventually - via canary or whatever process) and then switch this flag on.

commit/plugin.go
```diff
@@ -511,34 +467,7 @@ func (p *Plugin) getMainOutcome(
...
-			InflightPriceOcrSequenceNumber: inflightOcrSeqNum,
-			RemainingPriceChecks:           remainingPriceChecks,
-		}
-	}
-
-	return committypes.MainOutcome{
-		InflightPriceOcrSequenceNumber: 0,
-		RemainingPriceChecks:           0,
-	}
+	return committypes.MainOutcome{}
```
makramkd, line 470: Is this function still needed if it just returns an empty MainOutcome?

commit/report.go
```diff
@@ -291,16 +293,22 @@ func (p *Plugin) decodeReport(
...
+		reportSeqNumStale = reportSeqNr < latestOnChainReportSeqNr
+	}
+
+	if reportSeqNumStale &&
 		len(decodedReport.BlessedMerkleRoots) == 0 &&
 		len(decodedReport.UnblessedMerkleRoots) == 0 {
 		lggr.Infow(
 			"skipping stale report due to stale price seq nr and no merkle roots",
-			"latestPriceSeqNr", latestPriceSeqNr)
+			"latestPriceSeqNr", latestOnChainReportSeqNr)
```
makramkd, line 311: lets include the flag `multiPriceReportsEnabled` here

commit/tokenprice/processor.go
```diff
@@ -89,27 +91,75 @@ func (p *processor) Outcome(
...
 	if len(tokenPriceOutcome) == 0 {
 		lggr.Debugw("No token prices to report")
-		return Outcome{}, nil
+		return inflightTokenPricesOutcome, nil
+	}
+
+	// Check if we have inflight token price updates.
+	if prevOutcome.HasInflightTokenPriceUpdates() {
+		return p.computeInflightPricesOutcome(lggr, consensusObservation, prevOutcome), nil
+	}
```
makramkd, line 116: Most of the logic here is identical to chainfee except that we have this check _after_ the tokens selected for update stuff is done. Maybe we should move this to prior to `selectTokensForUpdate` to be consistent? Or is there some other reason this is different than chainfee?

commit/chainfee/types.go
winder: Need to double check that protobuf has an implicit "omitEmpty" to ensure `Outcome{GasPrices, rest not specified}` is encoded the same as `Outcome{GasPrice}`.

I'm going to grab some sample "legacy" messages to test my Outcome changes to use as backwards/forward compatibility test cases to cover this type of thing.

commit/chainfee/outcome.go
makramkd: This is an error in building the consensus observation, could that be transient? seems like it shouldn't be

PR #888, opened 2025-04-30
execute: pseudo-delete already executed messages

makramkd:
The plugin was getting stuck on large commit reports because it kept
fully including already executed messages in the observation when it
should have pseudo-deleted those and included messages that are not
executed.


winder (approved):
Change looks great, nice find.

PR #890, opened 2025-05-01
Improve solana validation

matYang:
Validate combination of msg.data + msg.receiver accounts + token data against maxDataBytes to reduce the chance of user sending a message to SVM that exceeds max transaction size on Solana.


chains/evm/contracts/FeeQuoter.sol
```diff
@@ -1005,32 +1006,75 @@ contract FeeQuoter is AuthorizedCallers, IFeeQuoter, ITypeAndVersion, IReceiver,
...
+      uint256 svmExpandedDataLength = dataLength;
+
+      // This abi.decode is safe because the address is validated above.
+      if (abi.decode(message.receiver, (uint256)) == 0) {
+        // When message receiver is zero, CCIP receiver is not invoked on SVM.
+        // There should not be additional accounts specified for the receiver.
+        if (accountsLength > 0) {
+          revert TooManySVMExtraArgsAccounts(accountsLength, 0);
+        }
+      } else {
```
PabloMansanet, line 1040: Is this correct even for a token transfer only with no logic receiver?

chains/evm/contracts/libraries/Internal.sol
```diff
@@ -287,6 +287,8 @@ library Internal {
   // bytes4(keccak256("CCIP ChainFamilySelector APTOS"));
   bytes4 public constant CHAIN_FAMILY_SELECTOR_APTOS = 0xac77ffec;
 
+  bytes4 public constant CHAIN_FAMILY_SELECTOR_c4e05953 = 0xc4e05953;
```
jhweintraub, line 290: Please Better define what this chain family is supposed to be and what the pre-image is like above for Aptos

PR #891, opened 2025-05-02
[CCIP-5809] execute: handle pseudo-deleted msgs in getFilterObservation

makramkd:
The GetMessages observation could potentially include a pseudo-deleted
message which is either:
- a message that is inflight
- a message that has already been executed
- a message that was not included in the observation due to size limits

In all cases, since we don't need to execute these messages, we don't
need to fetch their sender's nonce.

### DIFF
--- a/execute/observation.go
+++ b/execute/observation.go
@@ -454,6 +454,10 @@ func (p *Plugin) getMessagesObservation(
 			}
 		}
 
+		// TODO: check if the commit report is all fully executed and inflight messages.
+		// Might be able to just not include it in the observation at all? Or is there a
+		// risk with that?
+
 		if stop {
 			lggr.Infow("Stop processing messages, observation is too large",
 				"totalMsgs", totalMsgs, "encodedObsSize", encodedObsSize)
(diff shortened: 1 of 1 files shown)


execute/observation.go
```diff
@@ -454,6 +454,10 @@ func (p *Plugin) getMessagesObservation(
 			}
 		}
 
+		// TODO: check if the commit report is all fully executed and inflight messages.
+		// Might be able to just not include it in the observation at all? Or is there a
+		// risk with that?
```
makramkd, line 459: @0xnogo @winder thoughts on this? Does it make sense?

winder: Yeah that makes sense. There's code in here somewhere that attempts to remove fully executed commit reports. Maybe the inflight cache should use seqNum so that we can include it in the "executed message" list to trigger that code.

0xAustinWang (approved):

execute/observation.go
makramkd: Follow up: https://smartcontract-it.atlassian.net/browse/CCIP-5816

JTCombs95-commits:
> The GetMessages observation could potentially include a pseudo-deleted
> message which is either:
> - a message that is inflight
> - a message that has already been executed
> - a message that was not included in the observation due to size limits
> 
> In all cases, since we don't need to execute these messages, we don't
> need to fetch their sender's nonce.

Yea

PR #897, opened 2025-05-06
[Solana] Executio buffer contract for large TX

PabloMansanet:


chains/solana/contracts/programs/execution-buffer/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+no-log-ix-name = []
+cpi = ["no-entrypoint"]
+default = []
+
+[dependencies]
+solana-program = "1.17.25" # pin solana to 1.17
+anchor-lang = { version = "0.29.0", features = ["init-if-needed"] }
+anchor-spl = "0.29.0"
+bytemuck = "1.7"
+ethnum = "1.5"
```
toblich, line 23: Do we really need ethnum and anchor-spl?

chains/solana/contracts/programs/execution-buffer/src/context.rs
```diff
@@ -0,0 +1,113 @@
...
+use ccip_common::seed;
+
+pub const ANCHOR_DISCRIMINATOR: usize = 8; // size in bytes
+
+#[derive(Accounts)]
+#[instruction(buffer_id: u64, data: Vec<u8>)]
+pub struct AppendExecutionReportData<'info> {
+    #[account(
+        mut,
+        seeds = [seed::EXECUTION_BUFFER, authority.key().as_ref(), &buffer_id.to_le_bytes()],
```
toblich, line 12: I'd still prefer to use a 32-byte value here instead of u64. That way, the caller may choose to use the msgId as a key for the buffer if they want to. By not naming it explicitly `msg_id`, you still leave it open to reuse in other cases, if you want this implementation to be usable later for commits as well (where the caller could use either the root or something like chainSelector&seqNum)

toblich (commented):
Looking good! Some small comments, but overall LGTM

chains/solana/contracts/programs/execution-buffer/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+        let cpi_context = CpiContext::new(ctx.accounts.offramp.to_account_info(), cpi_accounts)
+            .with_remaining_accounts(cpi_remaining_accounts);
+        ccip_offramp::cpi::manually_execute(
+            cpi_context,
+            ctx.accounts.buffered_report.raw_report_data.clone(),
+            token_indexes,
+        )
+    }
+
+    pub fn append_execution_report_data<'info>(
```
jadepark-dev, line 43: It looks good overall, but I'm curious if we need to implement a hard cap for the report data size on-chain? While the CPI limits would eventually act as a constraint, I'm just asking out of curiosity.

jadepark-dev (approved):

chains/solana/contracts/programs/execution-buffer/Cargo.toml
PabloMansanet: Nope, just a leftover. Removing it.

chains/solana/contracts/programs/execution-buffer/src/lib.rs
PabloMansanet: Hmm, not sure myself, what's the worst case scenario here? Even if we promise "infinite size TX", I would imagine that users experienced with Solana will expect something to break.

PabloMansanet:
@toblich addressed your comments, PTAL when you can 🙏

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -9424,6 +9426,176 @@ func TestCCIPRouter(t *testing.T) {
 				})
 			})
 
+			t.Run("When executing a report with oversized data, it fails, but it succeeds when buffered", func(t *testing.T) {
+				transmitter := getTransmitter()
+
+				msgAccounts := []solana.PublicKey{config.CcipLogicReceiver, config.ReceiverExternalExecutionConfigPDA, config.ReceiverTargetAccountPDA, solana.SystemProgramID}
+
+				sourceChainSelector := config.EvmChainSelector
+				message, _ := testutils.CreateNextMessage(ctx, solanaGoClient, t, msgAccounts)
```
toblich, line 9435: Out of curiosity, have you tested the program (either in integration tests or manually in devnet) with a PTT use-case?

PR #902, opened 2025-05-06
Change contract type_version to use pkg semver version defined in Cargo.toml instead

toblich:
This is changing the versioning definition of #870 & #882. We'll now expose the semver version of the package instead of the commit hash, so as to simplify the verified builds process.


tt-cll (approved):

toblich:
Verified build GHA tested & working - see [here](https://github.com/smartcontractkit/chainlink-ccip/actions/runs/14869824253)

PabloMansanet (approved):
Looks good and clean, we'll have to make sure we always bump semver when we make changes though. 👍

Could we automate that? Have a CI case that fails if the IDL has changed but there's no change to `cargo.toml`?

matYang (approved):

PR #907, opened 2025-05-07
Add setter method in FeeQuoter&Router for link_token_mint [NONEVM-1768]

toblich:


chains/solana/contracts/programs/fee-quoter/src/context.rs
```diff
@@ -75,6 +75,23 @@ pub struct UpdateConfig<'info> {
...
+pub struct UpdateConfigLinkMint<'info> {
+    #[account(
+        mut,
+        seeds = [seed::CONFIG],
+        bump,
+        constraint = valid_version(config.version, MAX_CONFIG_V) @ FeeQuoterError::InvalidVersion,
+    )]
+    pub config: Account<'info, Config>,
+
+    pub link_token_mint: InterfaceAccount<'info, Mint>,
```
toblich, line 88: I'm passing the new mint via an account (so using a new context struct) and not just instruction data so that we can look at the decimals configured in the mint

agusaldasoro (approved):

toblich:
Setting a `do-not-merge` label here, and we'll sync with @tt-cll to see if we really need this (and therefore should merge it) or not.

PabloMansanet (approved):
Looks good to me, though I have a question. Not too important, since I know it applies to more than this change in isolation, but still.

I don't think we'd ever want to have a mismatch between the `link_token_mint` configured in router and fee quoter. In cases like this, rather than have two endpoints, would it be slightly safer/cleaner to simply have the `router` endpoint, and make it perform a CPI to the fee quoter so that the config is updated in lockstep?

PR #913, opened 2025-05-09
bump npm refs

RensR:


chains/evm/.solhint-test.json
```diff
@@ -17,6 +17,7 @@
     "no-unused-import": "error",
     "gas-struct-packing": "warn",
     "interface-starts-with-i": "warn",
+    "import-path-check": "off",
```
RensR, line 20: New rule in 5.1, but it doesn't get forge imports somehow. Disabled for tests only

PR #917, opened 2025-05-12
fix flake on solana mainnet

tt-cll:


chains/solana/utils/common/transactions.go
```diff
@@ -185,7 +185,9 @@ func sendTransactionWithLookupTables(ctx context.Context, rpcClient *rpc.Client,
 		count++
 		statusRes, sigErr := rpcClient.GetSignatureStatuses(ctx, true, txsig)
 		if sigErr != nil {
-			return nil, sigErr
+			fmt.Println(sigErr) // debugging if tx errors; mainnet can be flakey
```
tt-cll, line 188: We're getting a lot of `not found` errors running in CLD, but the txns do eventually show up on chain. My best guess is we submit it, and the call above returns not found due to some race https://github.com/gagliardetto/solana-go/blob/ca3f5f643435c1614475b2e564272bb8e6e21c1e/rpc/getSignatureStatuses.go#L48

tt-cll: cc @toblich @agusaldasoro @PabloMansanet

agusaldasoro (approved):

PR #921, opened 2025-05-13
CCIP-5886 Cache pending commit reports

0xnogo:


execute/internal/cache/commit_report_cache.go
```diff
@@ -0,0 +1,190 @@
...
+		// First query, use the original fetchFrom
+		queryFrom = fetchFrom
+	} else {
+		// For subsequent queries, start from the last query timestamp
+		// This is the key optimization: we only query for new reports since our last query
+		queryFrom = c.lastQueryTimestamp
+	}
+
+	// Fetch finalized reports since last query and update the cache with them.
+	newFinalizedReports, err := ccipReader.CommitReportsGTETimestamp(ctx, queryFrom, primitives.Finalized, limit)
```
winder, line 77: Check if the cache already has enough pending reports to reach the limit?

Might also consider 2 limits:
1. The query limit when fetching the mix of roots & prices.
2. Number of pending commit reports.

execute/internal/cache/commit_report_cache.go
```diff
@@ -0,0 +1,190 @@
...
+
+// generateKey creates a unique key for a commit report
+func generateKey(report ccipocr3.CommitPluginReportWithMeta) string {
+	// We'll create a unique key for each report by combining relevant fields
+	// For each blessed merkle root
+	for _, mrc := range report.Report.BlessedMerkleRoots {
+		return fmt.Sprintf("%s_%s", mrc.ChainSel.String(), mrc.MerkleRoot.String())
+	}
+
+	// If no blessed roots (unlikely but possible), check unblessed roots
```
winder, line 45: We're shipping 1.6 without RMN, so I guess they'll all be un-blessed

execute/internal/cache/commit_report_cache.go
```diff
@@ -0,0 +1,190 @@
...
+	for _, mrc := range report.Report.UnblessedMerkleRoots {
+		return fmt.Sprintf("%s_%s", mrc.ChainSel.String(), mrc.MerkleRoot.String())
+	}
+
+	// Fallback if no merkle roots (should never happen)
+	return fmt.Sprintf("%d_%d", report.Timestamp.Unix(), report.BlockNum)
```
winder, line 51: May as well include the chain selector in here.
```suggestion
	// Fallback if no merkle roots (should never happen)
	return fmt.Sprintf("%s_%d_%d", mrc.ChainSel.String(), report.Timestamp.Unix(), report.BlockNum)
```

execute/internal/cache/commit_report_cache.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// GetCachedAndNewReports combines cached reports with newly fetched ones
+func (c *CommitReportsCache) GetCachedAndNewReports(
+	ctx context.Context,
+	ccipReader readerpkg.CCIPReader,
+	fetchFrom time.Time,
+	limit int,
```
winder, line 59: Should these be properties of the cache rather than parameters?

execute/internal/cache/commit_report_cache.go
makramkd, line 51: Why would this never happen? I don't see any code filtering out reports that don't have any roots (blessed or otherwise). IMO there's no benefit to wasting memory and storing these in the cache.

execute/internal/cache/commit_report_cache.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// Add cached reports
+	for _, report := range cachedReports {
+		if !report.Timestamp.Before(minTimestamp) {
+			key := generateKey(report)
+			reportMap[key] = report
+		}
+	}
+
+	// Add new reports (will override cached ones if duplicates)
```
makramkd, line 167: When would this happen? Re-orgs?

execute/internal/cache/commit_report_cache.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+}
+
+// generateKey creates a unique key for a commit report
+func generateKey(report ccipocr3.CommitPluginReportWithMeta) string {
+	// We'll create a unique key for each report by combining relevant fields
+	// For each blessed merkle root
+	for _, mrc := range report.Report.BlessedMerkleRoots {
+		return fmt.Sprintf("%s_%s", mrc.ChainSel.String(), mrc.MerkleRoot.String())
+	}
```
makramkd, line 43: IIUC, this key should be "re-org resistant"? i.e if a re-org occurs on the dest chain, they key generated from the newly mined commit report should be equal to the key prior to the re-org?

execute/internal/cache/commit_report_cache.go
0xnogo: We don't have access to the chain selector from the report. But as anyway we don't way to store reports with 0 roots, this final case won't have any impact.

execute/internal/cache/commit_report_cache.go
0xnogo: Alright. It's probably safer to leave this as the `Report` struct still has an `UnblessedMerkleRoots` slice.

PR #922, opened 2025-05-13
retry send

tt-cll:


chains/solana/utils/common/transactions.go
```diff
@@ -173,14 +173,20 @@ func sendTransactionWithLookupTables(ctx context.Context, rpcClient *rpc.Client,
...
 
-	txsig, err := rpcClient.SendTransactionWithOpts(ctx, tx, rpc.TransactionOpts{SkipPreflight: skipPreflight, PreflightCommitment: rpc.CommitmentProcessed})
-	if err != nil {
-		fmt.Println(tx) // debugging if tx errors
-		return nil, err
+	count := 0
+	var txsig solana.Signature
+	for count < 500 {
+		count++
+		txsig, err = rpcClient.SendTransactionWithOpts(ctx, tx, rpc.TransactionOpts{SkipPreflight: skipPreflight, PreflightCommitment: commitment})
```
gustavogama-cll, line 180: seems you're sending the transaction 500 times, no matter what. Is that what you desired?

tt-cll: 😂 no. good catch we need to bail

agusaldasoro (approved):

PR #923, opened 2025-05-13
[CCIP-5898] chain fee per-dest deviation

matYang:
core ref: bb2b9fe22de7636ed41d76cfb620384afdbba218

Gas price deviation are meant to be per-dest-chain, not per-soure-chain.

For example, when reporting onto Ethereum, the deviation threshold is 400% for all source chain gas prices.

This is because the deviation threshold is chosen largely based on how expensive writing to the dest chain is, hence it is a function of the dest chain.


0xnogo (approved):
lgtm

PR #929, opened 2025-05-15
Implement gas prices deviation curve

ogtownsend:
- Port gas price deviation curve over

### DIFF
--- a/chainconfig/chainconfig.go
+++ b/chainconfig/chainconfig.go
@@ -46,6 +46,16 @@ type ChainConfig struct {
 	// on this chain. If true, we will only report prices based on the heartbeat (configured
 	// in the commit plugin offchain config).
 	ChainFeeDeviationDisabled bool `json:"chainFeeDeviationDisabled"`
+
+	// ExecNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for exec gas.
+	// If the exec gas price is less than this value, we should never trigger a deviation since the
+	// value is so small any deviation is negligible. A value of 5e9 would correspond to 0.000000005 USD.
+	ExecNoDeviationThresholdUSDWei cciptypes.BigInt `json:"execNoDeviationThresholdUSDWei"`
+
+	// DataAvNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for DA gas.
+	// If the DA gas price is less than this value, we should never trigger a deviation since the
+	// value is so small any deviation is negligible. A value of 5e9 would correspond to 0.000000005 USD.
+	DataAvNoDeviationThresholdUSDWei cciptypes.BigInt `json:"dataAvNoDeviationThresholdUSDWei"`
 }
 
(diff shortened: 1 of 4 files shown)


chainconfig/chainconfig.go
```diff
@@ -65,6 +75,22 @@ func (cc ChainConfig) Validate() error {
...
+	}
+
+	if cc.DataAvNoDeviationThresholdUSDWei.Int == nil {
+		return errors.New("DataAvNoDeviationThresholdUSDWei not set. Default value is 20e9")
+	}
+
+	if cc.DataAvNoDeviationThresholdUSDWei.Int.Cmp(big.NewInt(0)) < 0 {
+		return errors.New("DataAvNoDeviationThresholdUSDWei not set or negative. Default value is 20e9")
+	}
+
```
ogtownsend, line 93: Is it worth setting the defaults here instead of returning a failure?

chainconfig/chainconfig.go
```diff
@@ -46,6 +46,16 @@ type ChainConfig struct {
...
+	// ExecNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for exec gas.
+	// If the exec gas price is less than this value, we should never trigger a deviation since the
+	// value is so small any deviation is negligible. A value of 5e9 would correspond to 0.000000005 USD.
+	ExecNoDeviationThresholdUSDWei cciptypes.BigInt `json:"execNoDeviationThresholdUSDWei"`
+
+	// DataAvNoDeviationThresholdUSDWei is the lower bound *no* deviation threshold for DA gas.
+	// If the DA gas price is less than this value, we should never trigger a deviation since the
+	// value is so small any deviation is negligible. A value of 5e9 would correspond to 0.000000005 USD.
+	DataAvNoDeviationThresholdUSDWei cciptypes.BigInt `json:"dataAvNoDeviationThresholdUSDWei"`
```
ogtownsend, line 58: Since these are USD-based values, we don't really need them to be chain-specific. Maybe a better place for them is in `CommitOffchainConfig`, or even just hardcoded as a `const` in code. Any preferences @matYang or @makramkd

commit/chainfee/outcome.go
```diff
@@ -289,33 +286,35 @@ func (p *processor) getGasPricesToUpdate(
...
+			destChainCfg.Config.ExecNoDeviationThresholdUSDWei.Int,
 		)
-
-		dataAvFeeDeviates := mathslib.Deviates(
+		dataAvFeeDeviates = mathslib.DeviatesOnCurve(
 			currentChainFee.DataAvFeePriceUSD,
 			lastUpdate.ChainFee.DataAvFeePriceUSD,
-			daGasPriceDeviation,
+			destChainCfg.Config.DataAvNoDeviationThresholdUSDWei.Int,
 		)
```
ogtownsend, line 300: Do we want to gate the deviation curve behind a feature flag, or just ship it?

PR #930, opened 2025-05-16
Chainfee observation improvements

dimkouv:
Parallelize and add custom timeout to observation individual calls.

### DIFF
--- a/commit/chainfee/observation.go
+++ b/commit/chainfee/observation.go
@@ -37,11 +37,19 @@ func (p *processor) Observation(
 		fChain            map[cciptypes.ChainSelector]int
 	)
 
-	operations := []func(ctx context.Context, l logger.Logger){
-		func(ctx context.Context, l logger.Logger) { feeComponents = p.obs.getChainsFeeComponents(ctx, l) },
-		func(ctx context.Context, l logger.Logger) { nativeTokenPrices = p.obs.getNativeTokenPrices(ctx, l) },
-		func(ctx context.Context, l logger.Logger) { chainFeeUpdates = p.obs.getChainFeePriceUpdates(ctx, l) },
-		func(ctx context.Context, l logger.Logger) { fChain = p.observeFChain(l) },
+	operations := map[string]func(ctx context.Context, l logger.Logger){
+		"getChainsFeeComponents": func(ctx context.Context, l logger.Logger) {
+			feeComponents = p.obs.getChainsFeeComponents(ctx, l)
+		},
+		"getNativeTokenPrices": func(ctx context.Context, l logger.Logger) {
+			nativeTokenPrices = p.obs.getNativeTokenPrices(ctx, l)
+		},
+		"getChainFeePriceUpdates": func(ctx context.Context, l logger.Logger) {
+			chainFeeUpdates = p.obs.getChainFeePriceUpdates(ctx, l)
+		},
+		"observeFChain": func(_ context.Context, l logger.Logger) {
+			fChain = p.observeFChain(l)
+		},
 	}
 
 	callTimeout := p.cfg.ChainFeeAsyncObserverSyncTimeout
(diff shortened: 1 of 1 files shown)


commit/chainfee/observation.go
```diff
@@ -28,10 +30,36 @@ func (p *processor) Observation(
...
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(operations))
+	for i, op := range operations {
+		go func(i int) {
+			defer wg.Done()
+			tStart := time.Now()
+			op(callCtx, logger.With(lggr, "opID", i))
+			lggr.Debugw("observing goroutine finished", "opID", i, "duration", time.Since(tStart))
+		}(i)
```
0xnogo, line 60: Is there a risk all goroutines end up seeing the last function in operations?

Should we do this instead?
```
for i, op := range operations {
    opCopy := op         
    go func() {
        opCopy(...)     
    }()
}
```

commit/chainfee/observation.go
```diff
@@ -28,10 +30,36 @@ func (p *processor) Observation(
...
+	}
+
+	callTimeout := p.cfg.ChainFeeAsyncObserverSyncTimeout
+	callCtx, cf := context.WithTimeout(ctx, callTimeout)
+	defer cf()
+	lggr.Debugw("spawning observing goroutines", "timeout", callTimeout)
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(operations))
+	for i, op := range operations {
```
mateusz-sekara, line 54: How would that behave in case of a timeout? Are we going to continue processing with partially initialized values (feeComponents, nativeTokenPrices etc).

Can we have a test for that logic with timeouts hitting / not hitting different processors?

commit/chainfee/observation.go
```diff
@@ -28,10 +30,36 @@ func (p *processor) Observation(
...
+		go func(i int, op func(ctx context.Context, l logger.Logger)) {
+			defer wg.Done()
+			tStart := time.Now()
+			op(callCtx, logger.With(lggr, "opID", i))
+			lggr.Debugw("observing goroutine finished", "opID", i, "duration", time.Since(tStart))
```
mateusz-sekara, line 59: This opID doesn't bring much value, I would consider passing some meaningful/human understandable value here via operations lambda or sth

for instance:
```go
operations := map[string]func(context, logger){
   "feecomponents": func(ctx context.Context, l logger.Logger) { feeComponents = p.obs.getChainsFeeComponents(ctx, l) }
   "nativeTokenPrices": ...
}
// and then


for opName, op := range operations {
			lggr.Debugw("observing goroutine finished", "operation", opName, "duration", time.Since(tStart))
}
```

pluginconfig/commit.go
```diff
@@ -136,7 +136,9 @@ type CommitOffchainConfig struct {
 	// ChainFeeAsyncObserverSyncFreq defines how frequently the async chain fee observer should sync.
 	ChainFeeAsyncObserverSyncFreq time.Duration `json:"chainFeeAsyncObserverSyncFreq"`
 
-	// ChainFeeAsyncObserverSyncTimeout defines the timeout for a single sync operation (e.g. fetch token prices).
+	// ChainFeeAsyncObserverSyncTimeout defines the timeout for a single
+	// chain fee observation operation (e.g. fetch token prices).
+	// NOTE: It is also used when the async observer is disabled, while making the sync calls.
```
makramkd, line 141: As part of this work lets make sure we set this value appropriately in the configs in chainlink (I think in ocr3.go)

commit/chainfee/observation.go
makramkd: Thats a good question, I think in the case of timeout for all these funcs that have ctx we will log the error (ctx deadline exceeded?) and just return an empty map / array

makramkd (approved):

commit/chainfee/observation.go
mateusz-sekara: > This concern though is not related to the changes of this PR, I'd rather not overcomplicat it with test scenarios not related to it.

Are you sure concern is not valid? Before we had a timeout applied on the OCR level, in case of a timeout, we break execution immediately and cancel the entire observation. Right now, we are terminating on a lower level and code keeps continuing to work, right? Therefore, after the timeout, we will have data that is partially initialized and might lead to some weird behaviour. This is new behaviour. I think tests are needed here

0xnogo (approved):

mateusz-sekara (approved):

PR #934, opened 2025-05-19
Async price observers - prevent price updates spam

dimkouv:
# Issue Description

1. Commit plugin detects a stale price `P1`
2. Submits a report with the new price `P2`
3. Detects the price was transmitted and moves to the next observing round
4. Still observes the old stale price `P1` even though `P2` is onchain because the cache still contains `P1`.

# Solution in this PR

At step 3 where price transmission is detected, commit plugin invalidates the price related async observer caches. That way old stale prices are cleared immediately and a 'sync' action is triggered to fetch fresh prices.

We prevent blocking/sync operations within the async processor, on average it will require one extra round to submit the new price reports, but since rounds are short `<5s` this is acceptable.

### DIFF
--- a/commit/plugin.go
+++ b/commit/plugin.go
@@ -316,6 +316,7 @@ func (p *Plugin) getPriceRelatedObservations(
 	prevOutcome committypes.Outcome,
 	decodedQ committypes.Query,
 ) (tokenprice.Observation, chainfee.Observation) {
+	invalidatePriceCache := false
 	waitingForPriceUpdatesToMakeItOnchain := prevOutcome.MainOutcome.InflightPriceOcrSequenceNumber > 0
 
 	// If we are waiting for price updates to make it onchain, but we have no more checks remaining, stop waiting.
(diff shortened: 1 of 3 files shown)


commit/plugin.go
```diff
@@ -343,6 +344,7 @@ func (p *Plugin) getPriceRelatedObservations(
 
 		if cciptypes.SeqNum(latestPriceOcrSeqNum) >= prevOutcome.MainOutcome.InflightPriceOcrSequenceNumber {
 			lggr.Infow("previous price report made it through", "ocrSeqNum", latestPriceOcrSeqNum)
+			ctx = context.WithValue(ctx, consts.InvalidateCacheKey, true)
```
makramkd, line 347: Can we move this closer to where the ctx is used, e.g have a `var invalidateCache bool` up top, set it to `invalidateCache = true` here and before the `go func()` call:

```golang
ctx = context.WithValue(ctx, consts.InvalidateCacheKey, invalidateCache)
```

commit/chainfee/observation.go
```diff
@@ -31,6 +32,10 @@ func (p *processor) Observation(
 ) (Observation, error) {
 	lggr := logutil.WithContextValues(ctx, p.lggr)
 
+	if invalidateCache, ok := ctx.Value(consts.InvalidateCacheKey).(bool); ok && invalidateCache {
+		p.obs.invalidateCaches(ctx, lggr)
```
makramkd, line 36: This invalidates the cache but we can't be sure we have a viable reading by the time we observe, right? Would it make more sense to bust + read in a blocking way?

makramkd (approved):

0xnogo (approved):

PR #939, opened 2025-05-20
Upgrade anchor

agusaldasoro:


chains/solana/contracts/programs/ccip-offramp/src/state.rs
```diff
@@ -34,20 +34,40 @@ pub struct ReferenceAddresses {
...
+#[derive(InitSpace, Default)]
 pub struct Ocr3ConfigInfo {
     pub config_digest: [u8; 32], // 32-byte hash of configuration
     pub f: u8,                   // f+1 = number of signatures per report
     pub n: u8,                   // number of signers
     pub is_signature_verification_enabled: u8, // bool -> bytemuck::Pod compliant required for zero_copy
 }
 
+#[derive(AnchorSerialize, AnchorDeserialize, Clone, Debug)]
+pub struct Ocr3ConfigInfoInput {
```
agusaldasoro, line 46: @jadepark-dev Does this change make sense? I copied it from your branch, I'm not sure if I should do the same to all the structs that I had to remove the AnchorSerialize and AnchorDeserialize?

jadepark-dev: No, iirc it was only for `Ocr3ConfigInfo`. This could be a bug from Anchor. They changed `AnchorSerialize`, `AnchorDeserialize` to be implemented automatically from 0.30.x. It's not ideal, but the compiler was complaining when it's used for input struct so I had to split the struct and give it those attributes. 

But please keep in mind that the work was experimental, so it might have been fixed in a more recent version of Anchor or there might be a better pattern now!

PR #942, opened 2025-05-22
Commit report query limit in exec offchain config

0xnogo:
* Introducing the commit report limit in the exec offchain config
* Adding test related to the price reports spam verifying that execution is still happening.
* Reduce the limit from 2000 (1000 * 2) to 500 as the previous limits seems to have triggered CPU util on DB servers. Also it is not needed to query that much as most of the reports are going to be filtered (fully executed for example). Setting the default to 500, but we might need to update it again depending on production environment.

core ref: 055288e08c957c12cc066d087ef95217ae698221

### DIFF
--- a/pluginconfig/execute.go
+++ b/pluginconfig/execute.go
@@ -45,6 +45,9 @@ type ExecuteOffchainConfig struct {
 	// MaxSingleChainReports is the maximum number of single chain reports that can be included in a report.
 	// When set to 0, this setting is ignored.
 	MaxSingleChainReports uint64 `json:"maxSingleChainReports"`
+
+	// MaxCommitReportsToFetch is the maximum number of commit reports that can be fetched in each round.
+	MaxCommitReportsToFetch uint64 `json:"maxCommitReportsToFetch"`
 }
 
 func (e *ExecuteOffchainConfig) ApplyDefaultsAndValidate() error {
@@ -76,6 +79,10 @@ func (e *ExecuteOffchainConfig) Validate() error {
 
 	if e.MessageVisibilityInterval.Duration() == 0 {
 		return errors.New("MessageVisibilityInterval not set")
+	}
+
+	if e.MaxCommitReportsToFetch == 0 {
+		return errors.New("MaxCommitReportsToFetch not set")
 	}
 
 	set := make(map[string]struct{})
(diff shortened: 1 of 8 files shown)


pluginconfig/execute.go
```diff
@@ -45,6 +45,9 @@ type ExecuteOffchainConfig struct {
 	// MaxSingleChainReports is the maximum number of single chain reports that can be included in a report.
 	// When set to 0, this setting is ignored.
 	MaxSingleChainReports uint64 `json:"maxSingleChainReports"`
+
+	// MaxCommitReportsToFetch is the maximum number of commit reports that can be fetched in each round.
+	MaxCommitReportsToFetch uint64 `json:"maxCommitReportsToFetch"`
```
makramkd, line 50: I'm not sure if we should add a new required field, things will start failing in production because I'm pretty sure we call `Validate` on the config in constructors.

In order to avoid failures in prod we can set config w/ this new field set explicitly, but it seems error-prone and something that can be easily lost in translation.

makramkd (changes requested):

pluginconfig/execute.go
0xnogo: Even if it's set in the `DefaultExecuteOffChainCfg`? Chain configs are deriving their config from it.

makramkd: There's already a config set in CCIPHome, so if you add this as a required field and deploy this code w/out updating that config, the plugin will just fail to start because of the validation error.

0xnogo: Makes sense. Done!

pluginconfig/execute.go
```diff
@@ -8,6 +8,10 @@ import (
 	commonconfig "github.com/smartcontractkit/chainlink-common/pkg/config"
 )
 
+const (
+	defaultMaxCommitReportsToFetch = 500
```
makramkd: Should the default be the same as before (1000)? Or do we actually want to drop in this PR?

0xnogo: I actually changed it to 250 + update the PR description

We are currently fetching 1000 * 2. I'm proposing 250 * 2.

Reason: ~750 fetched constantly was enough to trigger some CPU limit on DB side + do we need to fetch that much? Most reports are getting filtered out.

It might be too much changes in one go. So we can maybe set it to 1000.

makramkd (approved):

PR #948, opened 2025-05-26
Skip e2e on solidity-only changes

RensR:
After this is merged, there should be a significant drop in CI runs and therefore delays, flakiness and cost

- Solana onchain CI only runs on Solana onchain changes 
- E2E tests are skipped for Solidity changes, except for changes to released versions. 

TODOs: apply the E2E skip logic to the offchain Go tests as well. That means normal Solidity changes should only run Solidity CI. 

This PR also makes all CI use a consistent v4 version of `actions/checkout`

### DIFF
--- a/.github/workflows/solana.yml
+++ b/.github/workflows/solana.yml
@@ -19,7 +19,7 @@ jobs:
     name: Detect changes
     runs-on: ubuntu-latest
     outputs:
-      non_src_changes: ${{ steps.changes.outputs.non_src }}
+      solana_changes: ${{ steps.changes.outputs.solana_changes }}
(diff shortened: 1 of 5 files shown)


.github/workflows/solana.yml
```diff
@@ -15,7 +15,29 @@ defaults:
     working-directory: ./chains/solana
 
 jobs:
+  changes:
```
RensR, line 18: As far as I could see, Solana CI is not required so we can skip without causing issues

makramkd (approved):

matYang (approved):

PR #952, opened 2025-05-27
fix e2e test ignore filter

RensR:
Previous didn't work, this one is tested here https://github.com/smartcontractkit/chainlink-ccip/actions/runs/15271754388/job/42948794439?pr=950


makramkd (approved):
Think this will work, small comment on the approach that might make things simpler.

PR #955, opened 2025-05-28
only gen metadata when required

RensR:
Only run ZK wrappers and metadata generation for releases, not every PR. 

Fix bug in changes checker, should result in properly running CI for non-release PRs


kylesmartin (approved):

.github/workflows/solidity-build-and-publish.yml
```diff
@@ -64,17 +64,6 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Checkout diff-so-fancy
```
RensR, line 67: not needed complexity, actually made CI fail and adds no real value

PR #956, opened 2025-05-28
[SOLANA] Offramp contract buffering

PabloMansanet:


chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -31,9 +30,21 @@ impl Execute for Impl {
         report_context_byte_words: [[u8; 32]; 2],
         token_indexes: &[u8],
     ) -> Result<()> {
-        let execution_report =
-            ExecutionReportSingleChain::deserialize(&mut raw_execution_report.as_ref())
-                .map_err(|_| CcipOfframpError::FailedToDeserializeReport)?;
```
toblich, line 36: Qq, since I see this error mapping was removed, what is the new error that is thrown if the report input is bad and fails to deserialize correctly?

chains/solana/contracts/programs/ccip-offramp/src/buffering.rs
```diff
@@ -0,0 +1,275 @@
...
+        self.chunk_bitmap = 0;
+        self.total_chunks = 0;
+        self.chunk_length = 0;
+
+        // We reinsert the new chunk and terminator, which will be accepted as it's smaller. From now
+        // on, we won't accept bigger chunks than this again.
+        self.add_chunk(report_length, new_chunk, chunk_index)?;
+        self.add_chunk(report_length, &terminator, terminator_index)?;
+        Ok(())
+    }
```
toblich, line 133: Wdyt about asking the user to submit the chunks_size explicitly? Given that it's already up to them to decide how to chunk it, and that size correlates explicitly with the chunk_index that they already have to calculate (so they _must_ already have defined/calculated their desired chunks_size)? I completely get the goal for this function, but it seems like avoidable complexity to me. For example, I don't see any checks actually guaranteeing that it was the _last_ chunk that was undersized. Could it be the case that a user submits a smaller-than-desired first chunk? And then, this recover function would move the wrong chunk of bytes around. And I think we can come up with other wrong-input cases that could also lead to unexpected behaviour here.

So, for example, let's say that the report is 1500 bytes long. The user may split it in an 800 byte and a 700 byte. They would submit two concurrent txs:
- report_length=1500, chunks_size=800, chunk_index=0, chunk=[...800 bytes...]
- report_length=1500, chunks_size=800, chunk_index=1, chunk=[...700 bytes...]

This way, there's no ambiguity anywhere, regardless of the order in which those two txs arrive you always have all the necessary info to insert them in the right place

chains/solana/contracts/programs/ccip-offramp/src/buffering.rs
```diff
@@ -0,0 +1,275 @@
...
+            // in the buffer (terminator), which may be smaller than all others. It's okay,
+            // we can recover in place.
+            return self.recover_wrong_size(report_length, chunk, chunk_index);
+        }
+
+        require_gte!(
+            self.chunk_length,
+            chunk.len() as u32,
+            CcipOfframpError::ExecutionReportBufferInvalidChunkSize
+        );
```
toblich, line 87: Could we actually please check that only the last chunk could be undersized, and all other chunks should have exactly the same length as the buffer's constant chunk size?

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -368,104 +414,194 @@ fn internal_execute<'info>(
...
-fn should_execute_messaging<'a>(
-    remaining_accounts: &'a [AccountInfo<'a>],
-    token_indices: &[u8],
-) -> bool {
-    // The first entry in the accounts corresponds to a token, which means there is no logic receiver
-    let only_tokens = token_indices.first().map(|i| *i == 0).unwrap_or_default();
-    !remaining_accounts.is_empty() && !only_tokens
+// Encodes the structured layout of the Execute Report Context remaining accounts, so they can
+// be accurately used where needed instead of passed as a raw array.
+pub struct ExecuteReportContextRemainingAccountsLayout<'a> {
```
toblich, line 428: I love this concept, but I worry though about doing this refactor with an in-place upgrade and not creating an `instructions::v2::execute.rs`, mostly just because of the amount of lines changed and their sensitivity, not because of any particular code problem I've spotted (though I haven't gone through this with a fine comb yet :smile:). How confident are you on the changes and whether this is actually safe to deploy in the same `v1` now that the product has already GA'd?

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -522,6 +658,10 @@ pub fn validate_execution_report<'info>(
         commit_report.timestamp != 0,
         CcipOfframpError::RootNotCommitted
     );
+    require!(
+        commit_report.chain_selector == execution_report.source_chain_selector,
```
toblich, line 662: Please add a comment clarifying that this can't easily be done in the context, as the report may be buffered.
Also, please use `require_eq!` when possible instead (and I see a few other checks in this function that could also use that other macro, heh), as that gives more info if the error ever arises, making the troubleshooting a bit easier

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -9506,6 +9506,189 @@ func TestCCIPRouter(t *testing.T) {
...
 			})
 
+			t.Run("When executing a report with oversized data, it fails, but it succeeds when buffered", func(t *testing.T) {
+				transmitter := getTransmitter()
+
+				msgAccounts := []solana.PublicKey{config.CcipLogicReceiver, config.ReceiverExternalExecutionConfigPDA, config.ReceiverTargetAccountPDA, solana.SystemProgramID}
+
+				sourceChainSelector := config.EvmChainSelector
+				message, _ := testutils.CreateNextMessage(ctx, solanaGoClient, t, msgAccounts)
+				message.Data = bytes.Repeat([]byte{42}, 1000)
```
toblich, line 9516: Suggestion: run the test with a data that actually has different bytes here and there. That way, you'll actually check that the data isn't scrambled later on, making the check slightly more robust. I get that you're already achieving something similar with the rest of the fields of the execution report (as this only represents that data), but still :shrug:

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -9506,6 +9506,189 @@ func TestCCIPRouter(t *testing.T) {
...
+					// PDA of the buffer account
+					solana.NewAccountMeta(bufferPDA, true, false),
+				)
+				bufferedExecutionIx, err := raw2.ValidateAndBuild()
+				require.NoError(t, err)
+				tx = testutils.SendAndConfirm(ctx, t, solanaGoClient, []solana.Instruction{bufferedExecutionIx}, transmitter, config.DefaultCommitment, common.AddComputeUnitLimit(1000000))
+
+				executionEvents, err2 := common.ParseMultipleEvents[ccip.EventExecutionStateChanged](tx.Meta.LogMessages, "ExecutionStateChanged", config.PrintEvents)
+				require.NoError(t, err2)
+
```
toblich, line 9669: Could you please also add a check that the transmitter has recovered the SOL from the buffer rent after the fact? I get that anchor should handle it, but it's a relevant thing to guarantee so that NOPs don't leak funds when doing the buffered exec

toblich (commented):
I like the approach. Here are quite a few comments

chains/solana/contracts/programs/ccip-offramp/src/lib.rs
```diff
@@ -533,6 +535,64 @@ pub mod ccip_offramp {
...
+            chunk,
+            chunk_index,
+        )
+    }
+
+    /// Closes the execution report buffer to reclaim funds.
+    ///
+    /// Note this is only necessary when aborting a buffered transaction, or when a mistake
+    /// was made when buffering data. The buffer account will otherwise automatically close
+    /// and return funds to the caller whenever buffered manual execution succeeds.
```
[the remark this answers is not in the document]
PabloMansanet: Right, will remove the word "manual" from there.

chains/solana/contracts/programs/ccip-offramp/src/buffering.rs
PabloMansanet: Ah yes, that's a very good question. I planned to do exactly what you describe at first, but I was worried about a potentially confusing aspect of the resulting API.

If we ask the reporter to submit the chunk size alongside every chunk, it becomes very easy to make the mistake of sending the chunk size of *the current* chunk instead of the global "average case" one. Plus you pay the cost of an extra parameter.

As for not checking that the smaller chunk is last, that's a good point: It should be easy to add a check for that.

I'd personally prefer to keep this method to really minimize the complexity offchain (and remove the ambiguity of which size you're sending alongside your chunk) but I'll make it a bit more robust against edge cases with extra unit tests.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
PabloMansanet: Ah my bad, I forgot to replace this. It will bubble up some Borsh error, so I'll fix it to be as it used to.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/buffering.rs
```diff
@@ -0,0 +1,314 @@
...
+        require!(
+            !self.is_initialized(),
+            CcipOfframpError::ExecutionReportBufferAlreadyInitialized
+        );
+        require!(
+            chunk_length > 0 && report_length >= chunk_length,
+            CcipOfframpError::ExecutionReportBufferInvalidLength
+        );
+        self.data.resize(report_length as usize, 0);
+        self.total_chunks = (report_length + chunk_length - 1) / chunk_length;
```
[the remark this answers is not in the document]
PabloMansanet: This is just a rote trick to implement ceiling division. `(A + B - 1) / B` is the same as `A.div_ceil(B)`. I can leave a comment to that effect :)

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/buffering.rs
```diff
@@ -0,0 +1,328 @@
...
+            .map_err(|_| CcipOfframpError::FailedToDeserializeReport)?,
+        buffer.data.len(),
+    ))
+}
+
+fn div_ceil<T: Into<u32>>(a: T, b: T) -> u32 {
+    let (a, b) = (a.into(), b.into());
+    assert!(a + b - 1 > 0);
+    (a + b - 1) / b
+}
```
toblich, line 170: Qq, is there any reason to not use `a.div_ceil(b)` directly, given that `u32` implements that method?

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -368,104 +418,194 @@ fn internal_execute<'info>(
...
+        report_is_buffered: bool,
+        source_bitmap: &u64,
+    ) -> Result<Self> {
+        ////////////////////////
+        // Buffering subslice //
+        ////////////////////////
+
+        // First, if the report is buffered, it means the last account is the buffering account.
+        // We can simply remove it from consideration from now on.
+        let (remaining_accounts_without_buffering, buffering_accounts) = if report_is_buffered {
```
toblich, line 508: Wdyt about renaming `buffering_accounts` into `buffering_account`, given that it's at most 1?

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -368,104 +418,194 @@ fn internal_execute<'info>(
...
+                .unwrap_or(remaining_accounts_without_buffering.len());
+
+            require!(
+                1 <= end_index && end_index <= remaining_accounts_without_buffering.len(), // program id and message accounts need to fit in remaining accounts
+                CcipOfframpError::InvalidInputsTokenIndices
+            ); // there could be other remaining accounts used for tokens
+
+            let logic_receiver = &remaining_accounts_without_buffering[0];
+            let external_execution_signer = &remaining_accounts_without_buffering[1];
+            let msg_accounts = &remaining_accounts_without_buffering[2..end_index];
```
toblich, line 540: Unrelated to your changes, just a comment to think about and maybe do a follow-up PR on:

There's something slightly off here with regards to the `end_index`. We validate that it is `>=1`, and then do `let msg_accounts = &remaining_accounts_without_buffering[2..end_index];` which doesn't make much sense when the `end_index` is `=1`. It looks like the initial restriction might be slightly too permissive, and the actual lowest admissible `end_index` should be 2 or 3 here.

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -368,104 +418,194 @@ fn internal_execute<'info>(
...
+            .clone()
+            .zip(indices_with_sentinel.skip(1));
+
+        let token_accounts_per_token: Vec<ExecuteContextRemainingTokenAccounts> = pairwise_indices
+            .map(|(start, end)| {
+                assert!(
+                    start < end
+                        && end <= remaining_accounts_without_buffering.len()
+                        && end - start >= MIN_TOKEN_POOL_ACCOUNTS,
+                );
```
toblich, line 595: What will the error look like if this assert fails? Is it just a panic on the program? Could user-error get here on wrong input of a manual exec (in which case a nicer error would be preferable) or would previous checks already catch invalid inputs and this should not happen (so it's ok to have this safety check with an ugly error)?

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -9506,6 +9507,431 @@ func TestCCIPRouter(t *testing.T) {
...
+				finalLamportsAfterClose := getLamports(transmitter.PublicKey())
+				closeDiff := finalLamportsAfterClose - finalLamportsBeforeClose
+
+				within10PercentOf := func(a int, b int) bool {
+					diff := math.Abs(float64(a - b))
+					return diff <= 0.1*math.Max(float64(a), float64(b))
+				}
+
+				// TODO work out exact math: the rent is being recovered but there's some slight drift in the fees.
+				require.True(t, within10PercentOf(int(rent), int(closeDiff-result.Meta.Fee)))
```
toblich, line 9607: The math here should be 
```
closeDiff = rent - result.Meta.Fee
```
changing that might fix this TODO

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -9506,6 +9507,431 @@ func TestCCIPRouter(t *testing.T) {
...
+				var rootAccount ccip_offramp.CommitReport
+				err = common.GetAccountDataBorshInto(ctx, solanaGoClient, rootPDA, config.DefaultCommitment, &rootAccount)
+				require.NoError(t, err, "failed to get account info")
+				require.NotEqual(t, bin.Uint128{Lo: 0, Hi: 0}, rootAccount.Timestamp)
+				require.Equal(t, bin.Uint128{Lo: 2, Hi: 0}, rootAccount.ExecutionStates)
+				require.Equal(t, commitReport.MerkleRoot.MinSeqNr, rootAccount.MinMsgNr)
+				require.Equal(t, commitReport.MerkleRoot.MaxSeqNr, rootAccount.MaxMsgNr)
+			})
+
+			t.Run("Direct buffering execution, happy path", func(t *testing.T) {
```
toblich, line 9797: "Direct buffering" means no buffering, right? I find the title slightly confusing :smile:

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -9506,6 +9507,431 @@ func TestCCIPRouter(t *testing.T) {
...
+				ix = sendChunk(t, 3, terminatorSize, 4, data, bufferID, bufferPDA, transmitter)
+				result = testutils.SendAndConfirm(ctx, t, solanaGoClient, []solana.Instruction{ix}, transmitter, config.DefaultCommitment)
+				totalSolSpentInFees += result.Meta.Fee
+				ix = sendChunk(t, 3, terminatorSize, 4, data, bufferID, bufferPDA, transmitter)
+				result = testutils.SendAndFailWith(ctx, t, solanaGoClient, []solana.Instruction{ix}, transmitter, config.DefaultCommitment, []string{"Error Code: ExecutionReportBufferAlreadyContainsChunk"})
+				totalSolSpentInFees += result.Meta.Fee
+				ix = sendChunk(t, 0, chunkSize, 4, data, bufferID, bufferPDA, transmitter)
+				result = testutils.SendAndConfirm(ctx, t, solanaGoClient, []solana.Instruction{ix}, transmitter, config.DefaultCommitment)
+				totalSolSpentInFees += result.Meta.Fee
+				// Checking again after the terminator shuffling
```
toblich, line 9574: We no longer have terminator shuffling, right? Could the testcase be simplified?

toblich (commented):
Looks good to me, leaving here a few more comments, but all of them are pretty small anyway :+1:

chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/buffering.rs
PabloMansanet: It's not stable on the Rustc version we use on CI I believe.

chains/solana/contracts/tests/ccip/ccip_router_test.go
PabloMansanet: Sorry, "direct" here means "no manual exec", as in, we're emulating the offchain code detecting the TX was too large to begin with and directly attempting an `execute` with a buffer, as opposed to failing first.

chains/solana/contracts/tests/ccip/ccip_router_test.go
PabloMansanet: I think I'll remove the comment, but there's still value in keeping this somewhat convoluted order to catch edge cases :)

toblich (approved):

PR #959, opened 2025-05-28
Multiple exec reports - Unordered execution

asoliman92:
This PR adds functionality to return multiple exec reports instead of only one. It only works with OutOfOrder messages/chains. When an exec report reaches capacity we should generate multiple exec reports rather than discarding all of the remaining data.

Added multiple tests and code defensively against having a misconfiguration with multi report enabled for chains that are ordered or have ordered commits.

First real world scenario will be Solana with 1 message per 1 report.

https://github.com/smartcontractkit/chainlink/pull/17968/commits
core ref: 486ff774e32af69fd01543dc043e572dc16a1bc8


execute/report/builder.go
```diff
@@ -176,55 +176,108 @@ func (b *execReportBuilder) checkInitialize() {
 // The commit report with updated metadata is returned. It reflects which messages
 // were selected for the exec report.
 //
-// TODO: Finish this function to generate multiple reports..
+// TODO: Add support for multiple reports with ordered messages.
 func (b *execReportBuilder) Add(
```
asoliman92, line 180: Working logic. Want to simplify before merging though.

execute/report/report.go
```diff
@@ -555,20 +553,20 @@ func (b *execReportBuilder) buildSingleChainReport(
 
 		msgs[i] = struct{}{}
 
-		finalReport2, err := buildSingleChainReportHelper(b.lggr, commitData, msgs)
+		candidateReport, err := buildSingleChainReportHelper(b.lggr, commitData, msgs)
```
asoliman92, line 556: renaming for better understanding

execute/report/builder.go
```diff
@@ -176,55 +176,108 @@ func (b *execReportBuilder) checkInitialize() {
...
 
-	// No messages fit into the exec report
-	if errors.Is(err, ErrEmptyReport) {
-		if !b.multipleReportsEnabled {
-			return commitReport, nil
+			// Start a new exec report
+			b.execReports = append(b.execReports, cciptypes.ExecutePluginReport{})
+			b.commitReports = append(b.commitReports, []exectypes.CommitData{})
+			b.accumulated = append(b.accumulated, validationMetadata{})
+			index = len(b.execReports) - 1
```
winder, line 221:
```suggestion
			index++
```

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -89,38 +89,52 @@ func (e *ExecCodecProto) EncodeOutcome(outcome exectypes.Outcome) ([]byte, error
...
 	}
 
-	return proto.MarshalOptions{Deterministic: true}.Marshal(pbObs)
+	// TODO: Remove after "Reports" is fully supported.
+	if len(outcome.Report.ChainReports) != 0 {
+		r := e.tr.execPluginReportsToProto([]ccipocr3.ExecutePluginReport{outcome.Report})
+		if len(r) > 0 {
+			pbOtcm.ExecutePluginReport = r[0]
+		}
+	}
```
winder, line 103: Is `outcome.Report` used anywhere besides this file?

execute/plugin_test.go
```diff
@@ -1408,9 +1490,20 @@ func TestPlugin_Reports_UnableToEncode(t *testing.T) {
 	codec := codec_mocks.NewMockExecutePluginCodec(t)
 	codec.On("Encode", mock.Anything, mock.Anything).
 		Return(nil, fmt.Errorf("test error"))
-	p := &Plugin{reportCodec: codec, lggr: logger.Test(t), ocrTypeCodec: ocrTypeCodec}
+	// Mock chain support for transmission schedule
```
asoliman92, line 1493: Changed ordering in Reports function so needs to have transmitters before reaching encoding part

execute/plugin.go
```diff
@@ -584,18 +588,14 @@ func (p *Plugin) Reports(
 		return nil, nil
 	}
 
-	// TODO: Handle multiple reports.
-	encodedReport, err := p.reportCodec.Encode(ctx, decodedOutcome.Reports[0])
+	reportInfos, err := extractReportInfo(decodedOutcome)
+	lggr.Debugw("report info in UnfinalizedReports()", "reportInfos", reportInfos)
```
makramkd, line 592: This log should be added after the error check. Also what is "UnfinalizedReports()"?

execute/plugin.go
```diff
@@ -606,20 +606,34 @@ func (p *Plugin) Reports(
...
-		"transmissionSchedule", transmissionSchedule, "oracleIDToP2PID", p.oracleIDToP2pID)
 
-	r := []ocr3types.ReportPlus[[]byte]{
-		{
+	// Handle all reports in the outcome
+	var reports []ocr3types.ReportPlus[[]byte]
+	for i, execReport := range decodedOutcome.Reports {
+		encodedReport, err := p.reportCodec.Encode(ctx, execReport)
+		if err != nil {
+			return nil, fmt.Errorf("unable to encode report %d: %w", i, err)
```
makramkd, line 615: While this is unlikely to happen the index here isn't super useful, I think using `%+v` as the format specifier and passing in `execReport` will yield better error messages.

execute/plugin.go
```diff
@@ -606,20 +606,34 @@ func (p *Plugin) Reports(
...
+		}
+
+		encodedInfo, err := reportInfos[i].Encode()
+		if err != nil {
+			return nil, fmt.Errorf("unable to encode report info %d: %w", i, err)
```
makramkd, line 620: Ditto here re: format specifier

execute/report/builder.go
```diff
@@ -176,55 +176,165 @@ func (b *execReportBuilder) checkInitialize() {
...
-	if b.maxSingleChainReports != 0 && len(b.execReports[index].ChainReports) >= int(b.maxSingleChainReports) {
-		if !b.multipleReportsEnabled {
-			// TODO: this is the previous behavior. Should we return an empty report error?
-			return commitReport, nil
-		} /* else {
-			// TODO: start a new exec report.
-		}*/
+	// Validate nonces for multiple reports mode
+	if err := b.validateNoncesForMultipleReports(commitReport); err != nil {
+		return commitReport, err
```
makramkd, line 188: This is a sidenote but this function's signature doesn't seem to make much sense to me. We return a commit report, but far as I can tell, its not really used. We seem to update a slice in-place with it but that slice isn't used (see https://github.com/smartcontractkit/chainlink-ccip/blob/7d0894c3f66876664301f8d63e938422212c3906/execute/plugin.go#L493-L498)

The `return` statements in this function end up returning a commit report even when an error is found; Go practice dictates that for errors we shouldn't return a usable object, otherwise its not really an error. Anyway, I feel like it'll be simpler if we just don't return the commit report here.

execute/exectypes/outcome.go
```diff
@@ -117,3 +123,17 @@ func NewOutcome(
 		Reports:       reports,
 	}
 }
+
+func NewOutcomeWithSortedCommitReports(
```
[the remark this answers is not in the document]
asoliman92: Sorting in the third phase (Filter) is a bit trickier as we have Reports and CommitReports in slices that maps to each others. Will check it out after debugging more the smoke tests.

execute/report/builder.go
```diff
@@ -176,55 +176,165 @@ func (b *execReportBuilder) checkInitialize() {
...
+		if len(readyMessages) == 0 {
+			// No more messages to process
+			return currentCommitReport, nil
+		}
+
+		// Check if we need to create a new exec report due to limits
+		if b.shouldCreateNewExecReport() {
+			if !b.multipleReportsEnabled {
+				return currentCommitReport, nil
+			}
```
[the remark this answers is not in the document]
asoliman92: exactly 👍

execute/exectypes/outcome.go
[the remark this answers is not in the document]
asoliman92: I don't see the benefit of sorting outside of here. The function is called `NewOutcomeWithSorted...` which is pretty clear and straightforward IMO.

execute/plugin.go
```diff
@@ -537,14 +538,18 @@ func selectReport(
 	return execReports, selectedReports, err
 }
 
-func extractReportInfo(report exectypes.Outcome) []cciptypes.ExecuteReportInfo {
+func extractReportInfo(report exectypes.Outcome) ([]cciptypes.ExecuteReportInfo, error) {
```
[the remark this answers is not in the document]
asoliman92: It is the outcome of the previous phase. I agree in general that naming in exec plugin are not very easy to follow.

execute/report/builder.go
```diff
@@ -176,55 +176,165 @@ func (b *execReportBuilder) checkInitialize() {
...
+		}
+
+		if len(readyMessages) == 0 {
+			// No more messages to process
+			return currentCommitReport, nil
+		}
+
+		// Check if we need to create a new exec report due to limits
+		if b.shouldCreateNewExecReport() {
+			if !b.multipleReportsEnabled {
```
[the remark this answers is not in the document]
asoliman92: We can't do this. 
1. If we do it earlier we'll break out of the loop early for chains that has the flag disabled without even creating the first report.
2. `Add` method is called multiple times by nature, if it doesn't return here it will create more reports for even the ones that don't allow it.

pkg/ocrtypecodec/v1/compatability_test.go
```diff
@@ -0,0 +1,226 @@
...
+		t.Run(testcase.name, func(t *testing.T) {
+			enc := v1.NewExecCodecProto()
+			encoded, err := enc.EncodeOutcome(testcase.input)
+			require.NoError(t, err)
+			b64Encoded := base64.StdEncoding.EncodeToString(encoded)
+			require.Equal(t, testcase.encodedOutcome, b64Encoded)
+		})
+	}
+}
+
```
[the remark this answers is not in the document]
asoliman92: I think it's safer to keep and then clean it up once all merged and working on prod.

pkg/ocrtypecodec/v1/exec.go
```diff
@@ -146,8 +160,20 @@ func (*ExecCodecJSON) DecodeObservation(data []byte) (exectypes.Observation, err
 }
 
 func (*ExecCodecJSON) EncodeOutcome(outcome exectypes.Outcome) ([]byte, error) {
-	// We sort again here in case construction is not via the constructor.
-	return json.Marshal(exectypes.NewOutcome(outcome.State, outcome.CommitReports, outcome.Reports))
+	// Create a new outcome with the same data for consistent representation
+
+	// Backward compatibility handling (similar to Proto version)
+	// Ensure Report field is populated if Reports has data
+	if len(outcome.Reports) == 1 && len(outcome.Report.ChainReports) == 0 {
```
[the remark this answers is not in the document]
asoliman92: It's easier to do it here once and remove it later. This is temporary code that's going to be removed soon.

.github/workflows/ccip-integration-test.yml
```diff
@@ -49,18 +49,18 @@ jobs:
...
+          - cmd: cd integration-tests/smoke/ccip && go test ccip_messaging_test.go -timeout 18m -test.parallel=2 -count=1 -json
             name: "Messaging Test"
           - cmd: cd integration-tests/smoke/ccip && go test -run "Test_CCIPBatching_MaxBatchSizeEVM" ccip_batching_test.go -timeout 12m -test.parallel=1 -count=1 -json
             name: "Batching Test Test_CCIPBatching_MaxBatchSizeEVM"
           - cmd: cd integration-tests/smoke/ccip && go test -run "^Test_CCIPBatching_MultiSource$" ccip_batching_test.go -timeout 12m -test.parallel=1 -count=1 -json
             name: "Batching Test Test_CCIPBatching_MultiSource"
-          - cmd: cd integration-tests/smoke/ccip && go test -run "Test_CCIPBatching_MultiSource_MultiReports" ccip_batching_test.go -timeout 12m -test.parallel=1 -count=1 -json
-            name: "Batching Test Test_CCIPBatching_MultiSource_MultiReports"
+          - cmd: cd integration-tests/smoke/ccip && go test -run "Test_CCIPBatching_MultiSource_MultiRoot" ccip_batching_test.go -timeout 12m -test.parallel=1 -count=1 -json
```
asoliman92, line 58: was a typo and wasn't actually testing anything.

execute/plugin.go
```diff
@@ -525,6 +525,7 @@ func selectReport(
 		}
 	}
 
+	// TODO: Stop at MaxReportsPerRound in the builder
```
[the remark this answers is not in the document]
asoliman92: It's done, I forgot to remove the todo :))

PR #962, opened 2025-05-30
NONEVM-1891: add TVM support in Fee Quoter for barebone TON CCIP e2e

jadepark-dev:
This PR adds TVM(TON) support to the Fee Quoter contract for TON CCIP integration [NONEVM-1891]

## Primary Changes
* Added TVM chain family selector support to enable TON address validation in the Fee Quoter contract
* Implemented `_validateTVMAddress` function to validate 36-byte TON user-friendly addresses with basic format checks
* Extended fee calculation logic to include TVM chain family
* Added tests for the internal lib

[NONEVM-1891]: https://smartcontract-it.atlassian.net/browse/NONEVM-1891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@
 .idea
 
 vendor/
+
+# Ignore personal settings
+**/.claude/settings.local.json
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/smartcontractkit/chain-selectors v1.0.60
 	github.com/smartcontractkit/chainlink-common v0.4.2-0.20250121163309-3e179a73cb92
 	github.com/smartcontractkit/chainlink-protos/rmn/v1.6/go v0.0.0-20250131130834-15e0d4cde2a6
-	github.com/smartcontractkit/libocr v0.0.0-20241007185508-adbe57025f12
+	github.com/smartcontractkit/libocr v0.0.0-20250408131511-c90716988ee0
 	github.com/stretchr/testify v1.10.0
 	github.com/zksync-sdk/zksync2-go v1.0.0
 	go.uber.org/multierr v1.11.0
@@ -42,8 +42,8 @@ require (
 	github.com/crate-crypto/go-ipa v0.0.0-20240724233137-53bbb0ceb27a // indirect
 	github.com/crate-crypto/go-kzg-4844 v1.1.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
-	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.2.0 // indirect
-	github.com/ethereum/c-kzg-4844 v1.0.0 // indirect
+	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.3.0 // indirect
+	github.com/ethereum/c-kzg-4844 v1.0.3 // indirect
 	github.com/ethereum/go-verkle v0.2.2 // indirect
 	github.com/fsnotify/fsnotify v1.6.0 // indirect
 	github.com/go-logr/logr v1.4.2 // indirect
--- a/go.sum
+++ b/go.sum
@@ -78,11 +78,11 @@ github.com/decred/dcrd/crypto/blake256 v1.0.0/go.mod h1:sQl2p6Y26YV+ZOcSTP6thNdn
 github.com/decred/dcrd/crypto/blake256 v1.0.1 h1:7PltbUIQB7u/FfZ39+DGa/ShuMyJ5ilcvdfma9wOH6Y=
 github.com/decred/dcrd/crypto/blake256 v1.0.1/go.mod h1:2OfgNZ5wDpcsFmHmCK5gZTPcCXqlm2ArzUIkw9czNJo=
 github.com/decred/dcrd/dcrec/secp256k1/v4 v4.0.1/go.mod h1:hyedUtir6IdtD/7lIxGeCxkaw7y45JueMRL4DIyJDKs=
-github.com/decred/dcrd/dcrec/secp256k1/v4 v4.2.0 h1:8UrgZ3GkP4i/CLijOJx79Yu+etlyjdBU4sfcs2WYQMs=
-github.com/decred/dcrd/dcrec/secp256k1/v4 v4.2.0/go.mod h1:v57UDF4pDQJcEfFUCRop3lJL149eHGSe9Jvczhzjo/0=
+github.com/decred/dcrd/dcrec/secp256k1/v4 v4.3.0 h1:rpfIENRNNilwHwZeG5+P150SMrnNEcHYvcCuK6dPZSg=
+github.com/decred/dcrd/dcrec/secp256k1/v4 v4.3.0/go.mod h1:v57UDF4pDQJcEfFUCRop3lJL149eHGSe9Jvczhzjo/0=
 github.com/decred/dcrd/lru v1.0.0/go.mod h1:mxKOwFd7lFjN2GZYsiz/ecgqR6kkYAl+0pz0tEMk218=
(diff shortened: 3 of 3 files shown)


RensR:
Please change the version of the feeQuoter to 1.6.1-dev to indicate this is unaudited and not 1.6.0

chains/evm/contracts/libraries/Internal.sol
```diff
@@ -192,6 +193,36 @@ library Internal {
...
+  /// @param encodedAddress The 36-byte TON address.
+  function _validateTVMAddress(
+    bytes memory encodedAddress
+  ) internal pure {
+    if (encodedAddress.length != 36) {
+      revert InvalidTVMAddress(encodedAddress);
+    }
+
+    // check if the address is in the basechain
+    if (encodedAddress[1] != 0x00) {
```
RensR, line 211: Does this mean we ONLY ever support sending to the base chain? We will never support other TON networks?

jadepark-dev: I wouldn't say never, but currently there is only one workchain(BaseChain) for regular transactions and creating a workchain is not an easy task since they have to pay a high price and receive 2/3 votes from validators(according to doc). I think the scope of supporting additional TON workchain will be a clearly visible task. 

But on the other hand, limiting workchain in the chain address validation can be redundant. There are only Masterchain / BaseChain and transactions in Masterchain is expensive anyway so probably we don't need this check.

jadepark-dev: [`aca59f2` (#962)](https://github.com/smartcontractkit/chainlink-ccip/pull/962/commits/aca59f2e3655fdf042abe72932563a45653d8a64) Removed workchain validation. It seems that the scope is broader when workchain is added, compared to what can be gained through validation

chains/evm/contracts/libraries/Internal.sol
```diff
@@ -192,6 +193,24 @@ library Internal {
...
+  /// - 32 bytes: account_id
+  /// - 2 bytes: CRC16 checksum(computationally heavy, validation omitted for simplicity)
+  /// @param encodedAddress The 36-byte TON address.
+  function _validateTVMAddress(
+    bytes memory encodedAddress
+  ) internal pure {
+    if (encodedAddress.length != 36) revert InvalidTVMAddress(encodedAddress);
+    bytes32 accountId;
+    assembly {
+      accountId := mload(add(encodedAddress, 0x22)) // 0x22 = 0x20 (data start) + 2 (offset for account_id)
```
matYang, line 209: Realizing that we don't have an Internal.t.sol yet, we should, reference `RateLimiter.t.sol`, something like this should be covered in a foundry unit test

RensR: Agreed this needs test coverage.

chains/evm/contracts/test/helpers/InternalTestHelper.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {Internal} from "../../libraries/Internal.sol";
+
+/// @notice This helper contract exposes the internal functions of the Internal library
+contract InternalTestHelper {
```
RensR, line 7: Libs should not need helpers to expose functions

chains/evm/contracts/test/libraries/Internal.t.sol
```diff
@@ -0,0 +1,119 @@
...
+  InternalTestHelper internal s_helper;
+
+  function setUp() public virtual {
+    s_helper = new InternalTestHelper();
+  }
+}
+
+contract Internal_validateEVMAddress is InternalSetup {
+  function test_validateEVMAddress_succeeds_onValidAddress() public {
+    bytes memory validAddress = abi.encode(VALID_EVM_ADDRESS);
```
RensR, line 27: nit: can probably just do 

```
s_helper.validateEVMAddress(abi.encode(VALID_EVM_ADDRESS));
```

chains/evm/contracts/test/libraries/Internal.t.sol
```diff
@@ -0,0 +1,119 @@
...
+    s_helper.validateEVMAddress(validAddress);
+  }
+
+  function test_validateEVMAddress_reverts_onInvalidLength() public {
```
RensR, line 31: Please check the style/foundry guide for test naming

https://github.com/smartcontractkit/chainlink-evm/blob/develop/contracts/STYLE_GUIDE.md
https://github.com/smartcontractkit/chainlink-evm/blob/develop/contracts/FOUNDRY_GUIDE.md

chains/evm/contracts/test/libraries/Internal.t.sol
```diff
@@ -0,0 +1,119 @@
...
+  function test_validateEVMAddress_succeeds_onBoundaryAddresses() public {
+    bytes memory lowerBoundary = abi.encode(uint256(Internal.EVM_PRECOMPILE_SPACE));
+    s_helper.validateEVMAddress(lowerBoundary);
+
+    bytes memory upperBoundary = abi.encode(uint256(type(uint160).max));
+    s_helper.validateEVMAddress(upperBoundary);
+  }
+}
+
+contract Internal_validate32ByteAddress is InternalSetup {
```
RensR, line 58: We don't do multiple contracts per file, please split each contract into its own file

chains/evm/contracts/test/libraries/Internal.t.sol
```diff
@@ -0,0 +1,119 @@
...
+  function test_validate32ByteAddress_reverts_onInvalidLength() public {
+    bytes memory invalidAddress = new bytes(31);
+    vm.expectRevert(abi.encodeWithSelector(Internal.Invalid32ByteAddress.selector, invalidAddress));
+    s_helper.validate32ByteAddress(invalidAddress, 0);
+  }
+
+  function test_validate32ByteAddress_reverts_onAddressBelowMinValue() public {
+    bytes memory belowMinAddress = abi.encode(uint256(500));
+    vm.expectRevert(abi.encodeWithSelector(Internal.Invalid32ByteAddress.selector, belowMinAddress));
+    s_helper.validate32ByteAddress(belowMinAddress, 1000);
```
RensR, line 72: Can you reduce the magic values here? If you declare `minValue` and then use declare 

```
    bytes memory belowMinAddress = abi.encode(minValue - 1));
```

chains/evm/contracts/test/libraries/Internal.t.sol
```diff
@@ -0,0 +1,119 @@
...
+  }
+}
+
+contract Internal_validateTVMAddress is InternalSetup {
+  function test_validateTVMAddress_succeeds_onValidAddress() public {
+    s_helper.validateTVMAddress(VALID_TVM_ADDRESS);
+  }
+
+  function test_validateTVMAddress_reverts_onShortLength() public {
+    bytes memory shortAddress = new bytes(35);
```
RensR, line 94: It's not clear if this fails due to length checks or the zero address check if all bytes are 0

chains/evm/contracts/test/libraries/Internal.t.sol
```diff
@@ -0,0 +1,119 @@
...
+    vm.expectRevert(abi.encodeWithSelector(Internal.InvalidTVMAddress.selector, shortAddress));
+    s_helper.validateTVMAddress(shortAddress);
+  }
+
+  function test_validateTVMAddress_reverts_onLongLength() public {
+    bytes memory longAddress = new bytes(37);
+    vm.expectRevert(abi.encodeWithSelector(Internal.InvalidTVMAddress.selector, longAddress));
+    s_helper.validateTVMAddress(longAddress);
+  }
+
```
RensR, line 104: I'd like a test that covers the following line in a meaningful way 

```
      accountId := mload(add(encodedAddress, 0x22)) // 0x22 = 0x20 (data start) + 2 (offset for account_id)
```

PR #964, opened 2025-06-02
Role don fix - get contract address

dimkouv:
Make reader method GetContractAddress not depend on chain support.

https://github.com/smartcontractkit/chainlink/pull/17969
`core ref: 52df57adf27a386283b16d9c53698609e3712dac`


0xnogo (approved):
lgtm

PR #970, opened 2025-06-04
commit plugin - inflight prices with role don

dimkouv:
This PR resolves the final known Role DON inconsistency.

# Inconsistency

1. Within an outcome that contains prices, the ocrSeqNr of that round is included.
2. In the next observation round we read the offramp last price ocrSeqNr so we know if prices were transmitted.
3. If not we don't observe prices. If yes we invalidate the price cache and observe new prices.

The inconsistency is on step 2 since with role don in the near future not every oracle will be able to read the offramp.
This introduces several issues and is not a consistent approach.

# Solution

1. During the observation phase and if we have inflight prices we only observe ocr seqNr, by dest supporting oracles.
2. During outcome if we have inflight prices, every oracle knows if prices were transmitted (thanks to dest supporting oracles that observed the seqNrs) so everyone deterministically can invalidate their price cache.

This approach was not agreed upon initially due to the extra complexity of having to observe and come to consensus on ocrSeqNrs.

The consensus logic implemented on this PR is similar to other observed msg seq nums.
```
sort seqNrs asc and exclude zeroes
select seqNrs[f] // skipping the first f potentially faulty observations
```

# Backwards Compatibility

Having some oracles running the old release and some oracles running the new one is problematic, since we introduce different logic, different outcomes and observations.

For that reason we roll-out this feature with a new generic feature flag called `EnableDonBreakingChanges`.
If that is set to False even if an oracle has upgraded it still runs the old logic.
After every oracle has upgraded this flag can be set to True and then in the next release deprecated code can be removed.

NOTE

Due to the complexity of the diff, this PR does not introduce conditions within the existing Observation/Outcome logic.
For example it does NOT do:
```
observation() {
   foo()
   bar()
   if EnableDonBreakingChanges {
        foo2()
   } else {
       foo3()
   }

  ...

  if EnableDonBreakingChanges { .. }
 // etc..
}
```

Since that would make the code pretty chaotic. Instead what it does:
```
observation() {
   if !EnableDonBreakingChanges {
      return observationOld()
   }
   return observationNew()
}
```

With the cost of having temporarily duplicate code until the deprecated code is removed in the next release.

# Integration Tests
There is a chainlink PR that sets EnableDonBreakingChanges to True for all the tests so with the core ref override below this PR is running all the e2e tests using the new logic.

Developers should make changes in `observationNext` and `outcomeNext`.

Someone could argue that we don't run e2e tests for `EnableDonBreakingChanges=false` but implicitly this PR 'deletes' this code.

`core ref: aad88a584c81e0be68f68927d5ab041fc65ca646`

# Unit Tests

Are adjusted to run the new logic.

### DIFF
--- a/pluginconfig/commit.go
+++ b/pluginconfig/commit.go
@@ -233,18 +233,18 @@ func (c *CommitOffchainConfig) applyDefaults() {
 		if c.ChainFeeAsyncObserverSyncFreq == 0 {
 			c.ChainFeeAsyncObserverSyncFreq = defaultAsyncObserverSyncFreq
 		}
-		if c.ChainFeeAsyncObserverSyncTimeout == 0 {
-			c.ChainFeeAsyncObserverSyncTimeout = defaultAsyncObserverSyncTimeout
-		}
+	}
+	if c.ChainFeeAsyncObserverSyncTimeout == 0 {
+		c.ChainFeeAsyncObserverSyncTimeout = defaultAsyncObserverSyncTimeout
 	}
 
 	if !c.TokenPriceAsyncObserverDisabled {
 		if c.TokenPriceAsyncObserverSyncFreq.Duration() == 0 {
 			c.TokenPriceAsyncObserverSyncFreq = *commonconfig.MustNewDuration(defaultAsyncObserverSyncFreq)
 		}
-		if c.TokenPriceAsyncObserverSyncTimeout.Duration() == 0 {
-			c.TokenPriceAsyncObserverSyncTimeout = *commonconfig.MustNewDuration(defaultAsyncObserverSyncTimeout)
-		}
+	}
+	if c.TokenPriceAsyncObserverSyncTimeout.Duration() == 0 {
+		c.TokenPriceAsyncObserverSyncTimeout = *commonconfig.MustNewDuration(defaultAsyncObserverSyncTimeout)
 	}
 }
 
--- a/pluginconfig/commit_test.go
+++ b/pluginconfig/commit_test.go
@@ -420,7 +420,9 @@ func TestCommitOffchainConfig_ApplyDefaults(t *testing.T) {
 				InflightPriceCheckRetries:          defaultInflightPriceCheckRetries,
 				MerkleRootAsyncObserverDisabled:    true,
 				ChainFeeAsyncObserverDisabled:      true,
+				ChainFeeAsyncObserverSyncTimeout:   defaultAsyncObserverSyncTimeout,
 				TokenPriceAsyncObserverDisabled:    true,
+				TokenPriceAsyncObserverSyncTimeout: *commonconfig.MustNewDuration(defaultAsyncObserverSyncTimeout),
 			},
 		},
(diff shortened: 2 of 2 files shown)


commit/plugin.go
```diff
@@ -239,6 +241,14 @@ func (p *Plugin) ObservationQuorum(
 func (p *Plugin) Observation(
 	ctx context.Context, outCtx ocr3types.OutcomeContext, q types.Query,
 ) (types.Observation, error) {
+	// TODO: Replace Observation with ObservationNext
+	//       and remove the condition when every oracle upgraded.
+	// NOTE: If you are making changes to Observation make sure they are reflected in observationNext.
+	if p.offchainCfg.EnableDonBreakingChanges {
+		p.lggr.Info("running observation next")
+		return p.observationNext(ctx, outCtx, q)
```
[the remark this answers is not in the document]
mateusz-sekara: I think it would also make reviewing that PR simpler, because I don't see a diff between old and new logic. "new" logic is in a new place, looking like brand new code. However, some parts of it are copied and pasted from old impl

commit/plugin_next.go
```diff
@@ -0,0 +1,355 @@
...
+		FChain:                p.ObserveFChain(lggr),
+		OnChainPriceOcrSeqNum: 0,
+	}
+
+	inflightPricesExist := prevOutcome.MainOutcome.InflightPriceOcrSequenceNumber > 0
+	switch inflightPricesExist {
+	case true:
+		// If we have inflight prices destination chain supporting oracles only observe the onchain price ocr seq num.
+		// We use this observation to check if prices are still inflight within the Outcome.
+		obs.OnChainPriceOcrSeqNum, err = p.observeOnChainPriceOcrSeqNum(ctx)
```
mateusz-sekara, line 95: Is it safe to do it like this? I mean, it would be safer to not update `obs.OnChainPriceOcrSeqNum` when err is returned. That would prevent us from accidentally updating that field with some rubbish returned together with an error


```go
if priceOcrSeqNum, err := err = p.observeOnChainPriceOcrSeqNum(ctx); if err != nil {
     lggr.Errorw("failed to observe on-chain price seq number", "err", err)
} else {
     obs.OnChainPriceOcrSeqNum = priceOcrSeqNum
}
```

mateusz-sekara: tl;dr avoid assigning the result unless it's error-free

commit/tokenprice/processor.go
```diff
@@ -85,6 +86,10 @@ func (p *processor) Outcome(
 ) (Outcome, error) {
 	lggr := logutil.WithContextValues(ctx, p.lggr)
 
+	if invalidateCache, ok := ctx.Value(consts.InvalidateCacheKey).(bool); ok && invalidateCache {
```
mateusz-sekara, line 89: Some comment on that invalidation? It's not obvious why it landed here. How can we test that to avoid someone accidentally removing it?

mateusz-sekara (approved):

PR #979, opened 2025-06-04
Token Pool Solana Implementation V2

agusaldasoro:
This pull request introduces significant updates to the Solana token pool programs, focusing on adding global configuration management, improving mint authority handling, and enhancing token minting logic. The changes include new structs, methods, and validation logic to support these features.

_Note for the reviewer: I'd recommend to go through each commit separately, as they are divided per subject._

### Global Configuration Management:
* Added a `PoolConfig` struct with a `self_served_allowed` field to manage global configuration for token pools.
* Added `init_global_config` and `update_global_config` methods to manage global configuration settings programmatically.

### Mint Authority Handling:
* Added a `transfer_mint_authority` method to transfer mint authority securely, with support for multisig accounts. 
* Introduced a `MintAuthorityTransferred` event to track changes in mint authority. 
* Updated token pool initialization to validate mint authority using the new `allowed_to_initialize_token_pool` function. 

### Token Minting with Multisig:
* If the Token Mint Authority is a multisig, and it's correctly configured (the pool signer is one of the signers) then the Token Pool calls the mint message with the corresponding accounts.


toblich (dismissed):
LGTM! I left one relevant question to clarify still with regards to who is authorized to transfer mint authority. The way it is right now seems too restrictive IMHO, though it doesn't lead to any security concern (it may not possible to use in some desired scenarios though, but you have more context here than I do), so I wanted to double-check that it is intentional but it is not a blocker to merge this.

toblich (approved):

PR #1001, opened 2025-06-11
CCIP-6187 remove getAccumulatedPoolFees impl from abstraction

0xsuryansh:


chains/evm/scripts/compile_all
```diff
@@ -10,6 +10,7 @@ echo " └───────────────────────
 # as specified in the foundry.toml.
 OPTIMIZE_RUNS_OFFRAMP=800
 OPTIMIZE_RUNS_FEE_QUOTER=8000
+OPTIMIZE_RUNS_BURN_MINT_FAST_TRASNFER_TOKEN_POOL=8300
```
RensR, line 13: Please just make this 8k so we don't have custom params for every contract. FeeQuoter and the pool can share 8k

PR #1002, opened 2025-06-11
Switch usdc readers to Extended readers.

winder:
This PR swaps out the `ContractReaderFacade` with the `Extended` reader. This will enable using auto binding in future PRs rather than managing the bound contract manually.


pkg/reader/usdc_reader.go
```diff
@@ -111,8 +112,9 @@ func NewUSDCMessageReader(
 				lggr:           lggr,
 				contractReader: contractReaders[chainSelector],
 				cctpDestDomain: domains,
-				boundContract:  contract,
+				boundContract:  contract, // TODO: this is not needed if we switch to the Extended contract reader.
```
winder, line 115: This is one reason we might want to use `Extended`. The other is to ensure fault violation detection is enabled.

execute/factory.go
```diff
@@ -134,13 +134,16 @@ func (p PluginFactory) NewReportingPlugin(
...
 	for chain, cr := range p.contractReaders {
 		chainID, err1 := sel.GetChainIDFromSelector(uint64(chain))
 		if err1 != nil {
 			return nil, ocr3types.ReportingPluginInfo{}, fmt.Errorf("failed to get chain id from selector: %w", err1)
 		}
-		readers[chain] = contractreader.NewExtendedContractReader(
+		reader := contractreader.NewExtendedContractReader(
 			contractreader.NewObserverReader(cr, lggr, chainID))
+		readers[chain] = reader
+		extended[chain] = reader
```
winder, line 146: We were always using the `Extended` reader, but it was being downcast to `ContractReaderFacade`. Another followup improvement would be to switch `NewCCIPChainReader` to `Extended` to stop double wrapping the downcast `Extended` object inside another `NewExtendedContractReader`.

0xnogo (approved):

PR #1009, opened 2025-06-11
USDC Token Pool Upgrades to better support Solana

jhweintraub:
1. Add support for overriding mint recipients in CCTP Domains to better support USDC on Solana
2. Add migration support for previous versions of USDCTokenPool by parsing destinationCaller out of CCTP message and relaying to the appropriate previous pool.


chains/evm/contracts/pools/USDC/USDCTokenPool.sol
```diff
@@ -44,6 +44,7 @@ contract USDCTokenPool is TokenPool, ITypeAndVersion {
   // A domain is a USDC representation of a chain.
   struct DomainUpdate {
     bytes32 allowedCaller; //       Address allowed to mint on the domain
+    bytes32 mintRecipient; // Address to mint to on the destination chain
```
matYang, line 47: nit: alignment

chains/evm/contracts/pools/USDC/USDCTokenPool.sol
```diff
@@ -118,7 +130,13 @@ contract USDCTokenPool is TokenPool, ITypeAndVersion {
     if (lockOrBurnIn.receiver.length != 32) {
       revert InvalidReceiver(lockOrBurnIn.receiver);
     }
-    bytes32 decodedReceiver = abi.decode(lockOrBurnIn.receiver, (bytes32));
+
+    bytes32 decodedReceiver;
+    if (domain.mintRecipient != bytes32(0)) {
```
matYang, line 135: let's add more context here, i.e `mintRecipient` will be 0 across EVM, but we need it for Solana

chains/evm/contracts/pools/USDC/USDCTokenPool.sol
```diff
@@ -157,11 +177,26 @@ contract USDCTokenPool is TokenPool, ITypeAndVersion {
...
+
     MessageAndAttestation memory msgAndAttestation =
       abi.decode(releaseOrMintIn.offchainTokenData, (MessageAndAttestation));
 
     _validateMessage(msgAndAttestation.message, sourceTokenDataPayload);
 
+    // If the destinationCaller is the previous pool, indicating an inflight message during the migration, we need to
+    // route the message to the previous pool to satisfy the allowedCaller.
+    bytes32 destinationCallerBytes32;
+    bytes memory messageBytes = msgAndAttestation.message;
```
0xsuryansh, line 189: should we check that messageBytes length is at least 148 bytes long to avoid memory safety issues.

0xsuryansh: nvm, that's done in _validateMessage already

chains/evm/contracts/pools/USDC/USDCTokenPool.sol
```diff
@@ -99,6 +107,10 @@ contract USDCTokenPool is TokenPool, ITypeAndVersion {
     i_messageTransmitterProxy = cctpMessageTransmitterProxy;
     i_localDomainIdentifier = transmitter.localDomain();
     i_token.safeIncreaseAllowance(address(i_tokenMessenger), type(uint256).max);
+
+    // For new token pools, no previous pool exists, and so the previousPool is not needed, and thus
+    // the zero address is a valid value.
+    i_previousPool = previousPool;
```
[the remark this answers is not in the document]
jhweintraub: Good call. I'll add a check for address(this) and supportsInterface. We should not check though that it is equal to address(0) since for a newly deployed chain there might not be a previous pool to use, in which case i've also added another check in the actual call to previous pool that 
```
    if (i_previousPool != address(0) && destinationCaller == i_previousPool) {
      return USDCTokenPool(i_previousPool).releaseOrMint(releaseOrMintIn);
    }
```

chains/evm/contracts/pools/USDC/USDCTokenPool.sol
jhweintraub: I don't think it's necessary to check the full message length, since the data that is retrieved is checked against the sourceTokenData which is trusted since it comes from the source pool. We check source/destination domain and nonce against stored values/source token pool data, so if the message length was <148, those values would be zero when parsed, and thus the function would intentionally revert.

jhweintraub: good call.

0xsuryansh (approved):
LGTM!

chains/evm/contracts/pools/USDC/USDCTokenPool.sol
```diff
@@ -72,20 +78,27 @@ contract USDCTokenPool is TokenPool, ITypeAndVersion {
   /// For dest pool version 1.5.1, this is the destination chain's token pool.
   struct Domain {
     bytes32 allowedCaller; //      Address allowed to mint on the domain
+    bytes32 mintRecipient; // Address to mint to on the destination chain
```
matYang, line 81: nit: alignment

chains/evm/contracts/pools/USDC/USDCTokenPool.sol
```diff
@@ -157,11 +187,26 @@ contract USDCTokenPool is TokenPool, ITypeAndVersion {
...
       abi.decode(releaseOrMintIn.offchainTokenData, (MessageAndAttestation));
 
     _validateMessage(msgAndAttestation.message, sourceTokenDataPayload);
 
+    // If the destinationCaller is the previous pool, indicating an inflight message during the migration, we need to
+    // route the message to the previous pool to satisfy the allowedCaller.
+    bytes32 destinationCallerBytes32;
+    bytes memory messageBytes = msgAndAttestation.message;
+    assembly {
+      // Parse out the bytes32 destinationCaller from messageBytes
```
matYang, line 201: let's beef this up a bit, comments should explain why, the "what" should be obvious from reading self-documenting code, something like 
```
destinationCaller is a 32-byte word starting at position 84 in messageBytes body, add 32 to skip the 1st word representing bytes length
```

chains/evm/contracts/pools/USDC/USDCTokenPool.sol
```diff
@@ -192,11 +237,16 @@ contract USDCTokenPool is TokenPool, ITypeAndVersion {
   ///     * destinationCaller     32         bytes32    84
   ///     * messageBody           dynamic    bytes      116
   function _validateMessage(bytes memory usdcMessage, SourceTokenDataPayload memory sourceTokenData) internal view {
+    // 116 is the minimum length of a valid USDC message. Since destinationCaller needs to be checked for the previous
+    // pool, this ensures that it can be parsed correctly and that the message is not too short. Since messageBody is
+    // dynamic and not always used, it is not checked.
+    if (usdcMessage.length < 116) revert InvalidMessageLength(usdcMessage.length);
```
matYang, line 243: if messageBody is empty, doesn't it still result in a 32-byte word to represent length 0 at index 116, pushing the min  size to 148?

matYang (approved):

PR #1010, opened 2025-06-12
Add Role DON e2e test with previous outcome

dimkouv:
This test asserts expectations when we have inflight prices and seq num ranges previously selected (in this state we build roots and wait for previous prices to make it onchain).

Also updated the existing role don test by adding a chain without enough oracles supporting it and asserting that there are no results for this chain while we still get results for the others.

### DIFF
--- a/commit/plugin_roledon_e2e_test.go
+++ b/commit/plugin_roledon_e2e_test.go
@@ -298,7 +298,7 @@ func (s roleDonTestSetup) newRoleDonTestPlugin(oracleID commontypes.OracleID) *P
 			MerkleRootAsyncObserverDisabled: true,
 			ChainFeeAsyncObserverDisabled:   true,
 			TokenPriceAsyncObserverDisabled: true,
-			EnableDonBreakingChanges:        true,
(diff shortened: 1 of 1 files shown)


commit/plugin_roledon_e2e_test.go
```diff
@@ -182,6 +192,167 @@ func TestPlugin_RoleDonE2E_NoPrevOutcome(t *testing.T) {
...
+		}
+
+		// Dest Chain Expectations - Makes sure only oracles that support the destination chain are reading it.
+		{
+			if oracleChains.Contains(s.destChain) {
+				deps.ccipReader.EXPECT().GetLatestPriceSeqNr(mock.Anything).Return(53, nil) // <-- still inflight (less than 54)
+			}
+		}
+
+		p := NewPlugin(
```
mateusz-sekara, line 269: nit, but can we extract this to something like `createPluginForTests(deps Deps, oracleId int)`. 
It pollutes the test case with unnecessary details

mateusz-sekara (dismissed):

PR #1022, opened 2025-06-16
WIP - Solana USDC Reader

winder:


pkg/reader/usdc_reader.go
```diff
@@ -338,6 +360,148 @@ func (u evmUSDCMessageReader) recreateMessageTransmitterEvents(
...
+				},
+			),
+		))
+	}
+
+	// Parent expressions for the query.
+	keyFilter, err := query.Where(
+		// Using same as EVM for consistency. Only used by off-chain components so name does not have to align with on-chain
+		consts.EventNameCCTPMessageSent,
+		expressions...,
```
amit-momin, line 445: I think this logic could break if we get multiple CCTP data. The default operator in the LogPoller is `AND` so stringing expressions together like this would make a query like `nonce=1 && domain=5 && nonce=2 && domain=5`. I think we probably need to use the Or comparator here
```suggestion
		query.Or(expressions...),
```

amit-momin: Oh and we should probably move the `query.Confidence(primitives.Finalized)` from above here so it doesn't get mixed up the with the nonce/domain queries
```suggestion
		query.Confidence(primitives.Finalized),
		expressions...,
```

pkg/reader/usdc_reader.go
```diff
@@ -338,6 +360,148 @@ func (u evmUSDCMessageReader) recreateMessageTransmitterEvents(
...
+
+	// Parse the extra data field to get the CCTP nonces and source domains.
+	cctpData, err := u.getMessageTokenData(tokens)
+	if err != nil {
+		return nil, err
+	}
+
+	// Query the token pool contract for the MessageSent event data.
+	expressions := []query.Expression{query.Confidence(primitives.Finalized)}
+	for _, data := range cctpData {
```
mateusz-sekara, line 417: That would definitely require perf tests using database like the ones we have for evm usdc reader 
https://github.com/smartcontractkit/chainlink/blob/develop/core/capabilities/ccip/ccip_integration_tests/usdcreader/usdcreader_test.go

mateusz-sekara (commented):
It would be great to have some e2e tests to cover that, especially to cover the full flow with CR/LogPoller for Solana

pkg/reader/usdc_reader.go
```diff
@@ -338,6 +360,148 @@ func (u evmUSDCMessageReader) recreateMessageTransmitterEvents(
 	return messageTransmitterEvents, nil
 }
 
+type solanaUSDCMessageReader struct {
```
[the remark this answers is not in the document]
winder: done.

pkg/reader/usdc_reader.go
amit-momin: Unresolving this since I noticed the `query.Confidence(primitives.Finalized)` expression wasn't moved down into the where clause

pkg/reader/usdc_reader.go
```diff
@@ -20,6 +18,9 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/pluginconfig"
 )
 
+// USDCMessageReader retrieves each of the CCTPv1 MessageSent event created
+// during the Commit phase of the USDC token transfer. The events are created
```
mateusz-sekara, line 22: nit, but `MessageSent` events are created when a message is sent via Router - as a part of that transaction OnRamp interacts with USDCTokenPool, which interacts with Circle contracts. Technically, it's before Commit (assuming Sol works similar to EVM)

pkg/reader/usdc_reader.go
```diff
@@ -113,12 +111,32 @@ func NewUSDCMessageReader(
 				lggr:           lggr,
 				contractReader: contractReaders[chainSelector],
 				cctpDestDomain: domains,
-				boundContract:  contract, // TODO: this is not needed if we switch to the Extended contract reader.
+			}
+		case sel.FamilySolana:
+			// Bind the TokenPool contract, the contract re-emits the USDC MessageSent event along with other metadata.
+			bytesAddress, err := addrCodec.AddressStringToBytes(token.SourceMessageTransmitterAddr, chainSelector)
```
mateusz-sekara, line 117: Not sure if it's set for both families, but we will need to have time retention set for these events just to make sure we don't store them forever 

https://github.com/smartcontractkit/chainlink/blob/e6d7c15c29751ccd83e3ffaddce420471fd9bbe4/core/capabilities/ccip/configs/evm/contract_reader.go#L406

pkg/reader/usdc_reader_solana.go
```diff
@@ -0,0 +1,157 @@
...
+				},
+			),
+		))
+	}
+
+	// Parent expressions for the query.
+	keyFilter, err := query.Where(
+		// Using same as EVM for consistency. Only used by off-chain components so name does not have to align with on-chain
+		consts.EventNameCCTPMessageSent,
+		query.Confidence(primitives.Finalized),
```
mateusz-sekara, line 98: nit, not sure how adding finalization check impacts db query performance for Solana, but it might be redundant. By the time we reach this place message is already committed so it's already finalized - at least this is how it works for EVM

pkg/reader/usdc_reader.go
```diff
@@ -212,6 +230,12 @@ func AllAvailableDomains() map[uint64]uint32 {
 	return destDomains
 }
 
+type evmUSDCMessageReader struct {
```
mateusz-sekara, line 233: minor, but if you extracted Solana to a separate file then I would extract evm implementation to `usd_reader_evm.go` and keep only the factory/interfaces in `usdc_reader.go`

pkg/reader/usdc_reader_solana.go
```diff
@@ -0,0 +1,157 @@
+package reader
```
mateusz-sekara, line 1: Any tests for Solana? Or would that be covered/tested in Chainlink repository with e2e tests?

mateusz-sekara (approved):

pkg/reader/usdc_reader_solana.go
winder: It's technically a no-op on Solana. Everything is finalized.

pkg/reader/usdc_reader_solana.go
winder: TBD. I'm not really sure how to properly unit test this since everything would be mocked out. Even with an e2e test a lot of it is mocked out.

execute/observation.go
```diff
@@ -127,6 +127,13 @@ func (p *Plugin) Observation(
 		"numCommitReports", numCommitReports,
 		"numMessages", observation.Messages.Count())
 
+	lggr.Debugw("[verbose] execute plugin got observation",
```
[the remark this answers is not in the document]
winder: Apparently these observations are available in kafka somehow.

0xAustinWang (approved):

PR #1024, opened 2025-06-17
CCIP-6146 Add chain selector param to ftf hooks

0xsuryansh:
1. Added chain selector param to ftf hooks, this change enables the hybrid pools to utilise chain selector to perform lock/release OR burn/mint ops based on the chainselector
2. Removed `onlyRouter` from `_ccipReceive` since it is redundant with the same check in public function of `CCIPReceiver` abstract contract


chains/evm/contracts/pools/FastTransferTokenPoolAbstract.sol
```diff
@@ -309,7 +309,7 @@ abstract contract FastTransferTokenPoolAbstract is TokenPool, CCIPReceiver, ITyp
   // @inheritdoc CCIPReceiver
   function _ccipReceive(
     Client.Any2EVMMessage memory message
-  ) internal virtual override onlyRouter {
```
RensR, line 312: If you're making changes like these please provide context. This now looks like major security issue.

PR #1025, opened 2025-06-17
ccipcapability: move interface impls

makramkd:
Move over the interface implementations, from the core
smartcontractkit/chainlink repo, to a new go module within this repo
called ccipcapability.

This means that we can easily add an implementation for an interface
that is newly defined without breaking builds elsewhere. Better
co-location of interface and implementation also means less back and
forth across repos which has historically been a big time sink.

As of this commit, the ccipcapability module is importing
smartcontractkit/chainlink but the goal is to not do that. However, the
main sticking point is evmrelaytypes, the package that defines the EVM
chainreader configuration struct. Ideally this moves to chainlink-evm at
some point soon (see e.g.
https://github.com/smartcontractkit/chainlink/pull/16578) and we'd only
have to import chainlink-evm instead.

Some remaining things in chainlink would be the launcher and
oraclecreator packages, which currently heavily depend on chainlink
packages. They first need to be abstracted away from those so that we
can safely extract them.


ccipcapability/go.mod
```diff
@@ -0,0 +1,229 @@
...
+	github.com/gagliardetto/solana-go v1.12.0
+	github.com/google/uuid v1.6.0
+	github.com/pkg/errors v0.9.1
+	github.com/smartcontractkit/chain-selectors v1.0.60
+	github.com/smartcontractkit/chainlink-ccip v0.0.0-20250616193830-42c18cce582e
+	github.com/smartcontractkit/chainlink-ccip/chains/solana v0.0.0-20250616193830-42c18cce582e
+	github.com/smartcontractkit/chainlink-common v0.7.1-0.20250603190644-0dcaa6d2cc86
+	github.com/smartcontractkit/chainlink-evm v0.0.0-20250617093943-9df97249e67b
+	github.com/smartcontractkit/chainlink-solana v1.1.2-0.20250611193111-99542a0c5d58
+	github.com/smartcontractkit/chainlink/v2 v2.25.0
```
jmank88, line 20: We should be able to avoid importing all three of these

jmank88:
> the package that defines the EVM
chainreader configuration struct. 

This sounds wrong. The configs are just encoded bytes that don't actually require importing all that baggage. We can fix this basic types in chainlink-ckommon. Can you link me to the types?

ccipcapability/go.mod
winder: @jmank88 different teams are implementing CCIP for the different NonEVM chain, so we aren't really involved with how they decide to do it. The CCIP services could be in their own modules though, so there's a path to avoid the dependency in the CCIP protocol layer. They are already using a nice little driver registration service inspired by go's sql package.

PR #1036, opened 2025-06-19
Truncate data logged when fetching commit reports.

winder:
During a recent incident it was difficult to tell which reports were being returned by the get commit reports query. The token and fee price reports were too spammy. This PR improves the logging situation with two improvements:
1. Discard token and fee price reports, instead log a number for how many updates there were.
2. Log an error if there are so many reports returned that results were truncated.

### DIFF
--- a/pkg/chainaccessor/default_accessor.go
+++ b/pkg/chainaccessor/default_accessor.go
@@ -7,11 +7,12 @@ import (
 	"strconv"
 	"time"
 
+	"golang.org/x/exp/maps"
+
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	"github.com/smartcontractkit/chainlink-common/pkg/types"
 	"github.com/smartcontractkit/chainlink-common/pkg/types/query"
 	"github.com/smartcontractkit/chainlink-common/pkg/types/query/primitives"
-	"golang.org/x/exp/maps"
 
 	"github.com/smartcontractkit/chainlink-ccip/internal/libs/slicelib"
 	"github.com/smartcontractkit/chainlink-ccip/pkg/consts"
@@ -749,6 +750,7 @@ func (l *DefaultAccessor) processCommitReports(
 	if len(reports) < limit {
 		return reports
 	}
+	lggr.Errorw("too many commit reports received, commit report results are truncated", "numTruncatedReports", len(reports)-limit)
 	return reports[:limit]
 }
 
--- a/pkg/reader/ccip.go
+++ b/pkg/reader/ccip.go
@@ -205,6 +205,29 @@ type RMNCurseResponse struct {
 
 // ---------------------------------------------------
 
+// printTruncatedReports is used to reduce the amount of data logged. Especially on Solana where gas and token
+// price updates are split into separate reports there is an excessive amount of data logged.
+func printTruncatedReports(lggr logger.Logger, reports []cciptypes.CommitPluginReportWithMeta) {
+	tokenPriceUpdates := 0
+	gasPriceUpdates := 0
+	var truncatedReports []cciptypes.CommitPluginReportWithMeta
+	for _, report := range reports {
+		gasPriceUpdates += len(report.Report.PriceUpdates.GasPriceUpdates)
+		tokenPriceUpdates += len(report.Report.PriceUpdates.TokenPriceUpdates)
+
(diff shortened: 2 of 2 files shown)


pkg/chainaccessor/default_accessor.go
```diff
@@ -749,6 +750,8 @@ func (l *DefaultAccessor) processCommitReports(
 	if len(reports) < limit {
 		return reports
 	}
+	lggr.Errorw("too many commit reports received, commit report results are truncated",
+		"numTruncatedReports", len(reports)-limit)
 	return reports[:limit]
```
prashantkumar1982, line 755: As I understand, this was source where the events got dropped.
Shouldn't we be atleast logging the reports that were dropped, as error logs?
What about a version similar to below, but of error type calling out these as dropped:
printTruncatedReports(lgger, reports[limit:])

pkg/reader/ccip.go
```diff
@@ -205,6 +205,29 @@ type RMNCurseResponse struct {
 
 // ---------------------------------------------------
 
+// printTruncatedReports is used to reduce the amount of data logged. Especially on Solana where gas and token
+// price updates are split into separate reports there is an excessive amount of data logged.
+func printTruncatedReports(lggr logger.Logger, reports []cciptypes.CommitPluginReportWithMeta) {
```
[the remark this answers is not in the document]
winder: renamed

pkg/chainaccessor/default_accessor.go
winder: Good idea, adding additional logs for each report that doesn't make it to the result.

pkg/reader/ccip.go
```diff
@@ -207,6 +207,32 @@ type RMNCurseResponse struct {
...
 
+// printReports is used to trim the size of the printed report. There can be a
+// large number of reports, especially on Solana where gas and token
+// price updates are split into separate reports. This function removes price
+// only reports, and removes price data from combined reports in an attempt to
+// ensure merkle roots are easier to find if logs are truncated.
+func printReports(lggr logger.Logger, reports []cciptypes.CommitPluginReportWithMeta) {
+	tokenPriceUpdates := 0
+	gasPriceUpdates := 0
+	var truncatedReports []cciptypes.CommitPluginReportWithMeta
```
asoliman92, line 218: don't think this function cares whether they're truncated or not.
```suggestion
	var reportToPrint []cciptypes.CommitPluginReportWithMeta
```

pkg/reader/ccip.go
```diff
@@ -207,6 +207,32 @@ type RMNCurseResponse struct {
 
 // ---------------------------------------------------
 
+// printReports is used to trim the size of the printed report. There can be a
```
asoliman92, line 210: This will make it much harder to find issues with pricing though. I like the idea of separating them. So maybe instead of printing only the lengths of them print them normally but on 3 different logs.
Something like:
```
lggr.Infow("decoded price reports", tokenPriceUpdates) // where these are real updates
lggr.Infow("decoded chain fee reports", gasPriceUpdates)
lggr.Infow("decoded merkle reports", truncatedReports)
```

asoliman92 (approved):

pkg/chainaccessor/default_accessor.go
```diff
@@ -776,6 +777,11 @@ func (l *DefaultAccessor) processCommitReports(
 	if len(reports) < limit {
 		return reports
 	}
+	lggr.Errorw("too many commit reports received, commit report results are truncated",
```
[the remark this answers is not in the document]
makramkd: I thought we had some logic somewhere where we filter out price-only reports and paginate? Or is that not the case?

pkg/chainaccessor/default_accessor.go
```diff
@@ -788,6 +789,11 @@ func (l *DefaultAccessor) processCommitReports(
...
+		"numTruncatedReports", len(reports)-limit)
+	for l := limit; l < len(reports); l++ {
+		lggr.Warnw("dropping commit report which doesn't fit in limit", "report", reports[l])
```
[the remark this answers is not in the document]
makramkd: Also maybe consider grouping the relevant roots into one log instead?

0xAustinWang (approved):

asoliman92 (approved):

PR #1042, opened 2025-06-23
Execute plugin collects and transmits CCTPv2 attestations

rstout:
Implements CCTP v2 token data observation for Chainlink CCIP.

Cross-Chain Transfer Protocol (CCTP) is Circle's protocol for transferring USDC natively across
blockchains. 

Attestations are cryptographic proofs that validate cross-chain token transfers. When USDC is
burned on a source chain, Circle's attestation service provides signed attestations that can be
used to mint the equivalent amount on the destination chain. These attestations contain the
original message data and Circle's signature, proving the burn occurred. 

The CCTPv2 Token Data Observer implemented in this PR iterates through a batch of CCIP messages, determines which messages' token transfers are CCTPv2 transfers, calls the Circle CCTPv2 API to fetch attestations, and converts the attestations to `TokenData` that can be processed by on-chain contracts to mint the associated USDC.


execute/tokendata/cctp/v2/source_token_data_payload.go
```diff
@@ -0,0 +1,221 @@
...
+	vals, err := argTypes.Unpack(data)
+	if err != nil {
+		return nil, err
+	}
+	if len(vals) != 10 {
+		return nil, fmt.Errorf("unexpected number of unpacked values")
+	}
+
+	// Safely extract values with type checking
+	nonce, ok := vals[0].(uint64)
```
0xAustinWang, line 151: nit: we can combine these validations in a generic function like 
```
func extractAndValidate[T any](vals []interface{}, index int, fieldName string) (T, error) {
    var zero T
    if index >= len(vals) {
        return zero, fmt.Errorf("missing value for %s at index %d", fieldName, index)
    }
    
    value, ok := vals[index].(T)
    if !ok {
        return zero, fmt.Errorf("invalid %s type: expected %T, got %T", fieldName, zero, vals[index])
    }
    return value, nil
}

```

nit because it doesn't save many lines, but it does make it easier to test

execute/tokendata/cctp/v2/source_token_data_payload.go
```diff
@@ -0,0 +1,221 @@
...
+	if tokenData.CCTPVersion != CctpVersion2 {
+		return nil, fmt.Errorf("unsupported CCTP version: %d", tokenData.CCTPVersion)
+	}
+
+	return tokenData, nil
+}
+
+// DecodeSourceTokenDataPayload decodes a byte array into a SourceTokenDataPayload struct
+func DecodeSourceTokenDataPayload(data []byte) (*SourceTokenDataPayload, error) {
+	argTypes := abi.Arguments{
```
0xAustinWang, line 129: Is there a way for us to define this somewhere outside the function?

execute/tokendata/cctp/v2/observer.go
```diff
@@ -0,0 +1,522 @@
...
+// getSourceTokenDataPayloads iterates over ccipMessage.TokenAmounts and returns a map of token indexes to
+// SourceTokenDataPayload if the TokenAmount's ExtraData can decode to a valid (CCTP v2) SourceTokenDataPayload.
+func getSourceTokenDataPayloads(
+	ccipMessage cciptypes.Message,
+	cctpV2EnabledTokenPoolAddress string,
+) map[int]SourceTokenDataPayload {
+	sourceTokenDataPayloads := make(map[int]SourceTokenDataPayload)
+	for i, tokenAmount := range ccipMessage.TokenAmounts {
+		payload, err := getCCTPv2SourceTokenDataPayload(cctpV2EnabledTokenPoolAddress, tokenAmount)
+		if err == nil {
```
0xAustinWang, line 482: should this be `if err != nil`?

execute/tokendata/cctp/v2/observer.go
```diff
@@ -0,0 +1,522 @@
...
+	// Query the attestation service for all collected transaction hashes
+	// Each response contains the message data and attestation needed for on-chain minting
+	cctpV2Messages, err := getCCTPv2Messages(ctx, lggr, attestationClient, sourceChain, sourceDomainID, txHashes)
+	if err != nil {
+		// Return error tokens for all CCTP v2 tokens, others remain not supported
+		return errorMessageTokenData(err, ccipMessages, sourceTokenDataPayloads)
+	}
+
+	// Step 6: Match each SourceTokenDataPayload to its corresponding CCTP v2 Message
+	tokenIndexToCCTPv2Message := matchCCTPv2MessagesToSourceTokenDataPayloads(
```
jhweintraub, line 248: Is it possible to have this information on the cctp message details be retrieved from an event log rather than from the source token data payload? This would save us a LOT of space on the amount of bytes being sent between chains since most of the fields in the new struct are not used on destination chain?

execute/tokendata/cctp/v2/observer.go
```diff
@@ -0,0 +1,522 @@
...
+func matchCCTPv2MessagesToSourceTokenDataPayloads(
+	lggr logger.Logger,
+	cctpV2Messages map[string]Message,
+	sourceTokenDataPayloads map[cciptypes.SeqNum]map[int]SourceTokenDataPayload,
+	isMatch func(SourceTokenDataPayload, Message) bool,
+	sourceChain cciptypes.ChainSelector,
+	metricsReporter MetricsReporter,
+) map[cciptypes.SeqNum]map[int]CCTPv2MessageOrError {
+	matchedCCTPv2Messages := make(map[cciptypes.SeqNum]map[int]CCTPv2MessageOrError)
+	usedNonces := mapset.NewSet[string]()
```
matYang: `usedNonces` is only in function scope, aren't the knowledge of which attestations have been consumed persisted in different exec attempts? if not, and there are identical CCTP messages in same tx but executed at different times on dest, the Exec DON will be forever stuck on the 2nd one no?

rstout: > aren't the knowledge of which attestations have been consumed persisted in different exec attempts?

This info isn't persisted, in the review meeting the consensus was that it'd be too much effort to implement. 

> the Exec DON will be forever stuck on the 2nd one no?

Yes, unfortunately. This is a rare case that is hard to avoid, we agreed to add logic to manualExec in tools-ts to fix these issues if they ever do happen

PR #1053, opened 2025-06-26
Remove not necessary account

agusaldasoro:


PabloMansanet (approved):
Looks good. Out of curiosity, was this because we did a realloc before?

agusaldasoro:
> Looks good. Out of curiosity, was this because we did a realloc before?



> Looks good. Out of curiosity, was this because we did a realloc before?

This is a new message to modify the global config, I think that it was a copy/paste mistake as that account is needed when initializing.

PR #1058, opened 2025-07-02
[Solana] Add simple fund manager to CCTP token pool

PabloMansanet:


chains/solana/contracts/programs/cctp-token-pool/src/lib.rs
```diff
@@ -693,6 +699,9 @@ impl CctpMessage {
 pub struct State {
     pub version: u8,
     pub config: BaseConfig,
+    // Authority allowed to reclaim funds (i.e. from closing the CCTP event PDA, or de-funding an
+    // overfunded signer PDA.)
+    pub fund_manager: Pubkey,
```
toblich, line 704: We will need another method to extend the state PDA so that it fits this new field. Once migrated in staging, then we can remove the new method (no need to change the `version` field, this can be quick-and-dirty just for staging).
Also, if we are going to migrate, we could already add a allowed destinatary address for the withdrawed funds (similar to the router's `fee_aggregator` attribute), wdyt? And maybe also a `minimum_balance` field to ensure that the funds_manager never withdraws so much that the pools runs out of funds and leads to downtime. My thinking is maybe the owner sets the destinatary, and the funds manager is the one allowed to actually make the call to debit those funds while making (keeping the balance above a minimum).

toblich (commented):
Looks good to me! Just one comment regarding maybe adding a few more things on the same PR instead of a new one, just so that the PDA realloc that will need to happen happens only once.

chains/solana/contracts/programs/cctp-token-pool/src/lib.rs
```diff
@@ -70,6 +73,24 @@ pub mod cctp_token_pool {
...
+        Ok(())
+    }
+
+    pub fn set_fund_reclaim_destination(
+        ctx: Context<SetConfig>,
+        fund_reclaim_destination: Pubkey,
+    ) -> Result<()> {
+        ctx.accounts.state.fund_reclaim_destination = fund_reclaim_destination;
+        Ok(())
+    }
```
toblich, line 92: Please emit an event in each of these (it can be a specific event for each or the same event with all fields in all of them, as you prefer)

chains/solana/contracts/programs/cctp-token-pool/src/lib.rs
```diff
@@ -364,6 +385,39 @@ pub mod cctp_token_pool {
...
+    ///
+    /// The resulting funds on the PDA cannot drop below `minimum_signer_funds`.
+    pub fn reclaim_funds(ctx: Context<ReclaimFunds>, amount: u64) -> Result<()> {
+        require!(amount > 0, CctpTokenPoolError::InvalidSolAmount);
+
+        let pool_signer_lamports = ctx.accounts.pool_signer.lamports();
+        require!(
+            pool_signer_lamports >= amount + ctx.accounts.state.minimum_signer_funds,
+            CctpTokenPoolError::InsufficientFunds
+        );
```
toblich, line 413: Please use `require_gte!` just so that it is easier to troubleshoot on errors

chains/solana/contracts/programs/cctp-token-pool/src/lib.rs
```diff
@@ -364,6 +385,39 @@ pub mod cctp_token_pool {
...
+    /// Returns an amount of SOL from the pool signer account to the designated
+    /// fund reclaimer. There are three entities involved:
+    ///
+    /// * `owner`: can configure the reclaimer and fund manager.
+    /// * `fund_manager`: can execute this instruction.
+    /// * `fund_reclaim_destination`: receives the funds.
+    ///
+    /// The resulting funds on the PDA cannot drop below `minimum_signer_funds`.
+    pub fn reclaim_funds(ctx: Context<ReclaimFunds>, amount: u64) -> Result<()> {
+        require!(amount > 0, CctpTokenPoolError::InvalidSolAmount);
```
toblich, line 407: Please use `require_gt!` just so that logs have more info on errors

chains/solana/contracts/tests/ccip/tokenpool_test.go
```diff
@@ -1720,7 +1720,14 @@ func TestTokenPool(t *testing.T) {
...
 
-				testutils.SendAndConfirm(ctx, t, solanaGoClient, []solana.Instruction{ix}, admin, config.DefaultCommitment)
+				// There's no fund manager, so funds cannot be reclaimed
+				testutils.SendAndFailWith(ctx, t, solanaGoClient, []solana.Instruction{ix}, admin, config.DefaultCommitment, []string{"Fund Manager is invalid"})
+
+				fundManagerIxRaw := cctp_token_pool.NewSetFundManagerInstruction(admin.PublicKey(), cctpPool.State, usdcMint, admin.PublicKey())
+				fundManagerIx, err := fundManagerIxRaw.ValidateAndBuild()
+				require.NoError(t, err)
+
+				testutils.SendAndConfirm(ctx, t, solanaGoClient, []solana.Instruction{fundManagerIx, ix}, admin, config.DefaultCommitment)
```
toblich, line 1730: Please also add an integration test reclaiming funds, once hitting the minimum_balance constraint and once successfully

toblich (commented):
Looks good! Just a few tiny comments

toblich (approved):

PR #1060, opened 2025-07-02
Add CTTPv2AttestationClient

rstout:
Exposes an HTTP client to call the Circle Cross-Chain Transfer Protocol (CCTP) V2 HTTP API.

The goal is to fetch attestations for USDC transfers made via CCTPv2 by CCIP. The `/v2/messages/` endpoint
takes a source domain ID and transaction hash and returns a list of CCTPv2 messages executed on that source and
transaction. Each message contains an attestation and a decoded message body, which are needed to pass to
destination chain contracts where the attestation will be verified and the USDC transfer will be executed (e.g. the
USDC will be minted and transferred on the destination chain).

API docs are here: https://developers.circle.com/api-reference/stablecoins/common/get-messages-v-2

An example API calls is:
https://iris-api-sandbox.circle.com/v2/messages/0?transactionHash=0x8f1d58a2161b0bd662609b62537d57b76cfa6bb22a9afe8ec49b09f3810d44c2

### DIFF
--- a/execute/tokendata/usdc/cctpv2/http_client.go
+++ b/execute/tokendata/usdc/cctpv2/http_client.go
@@ -1,0 +1,186 @@
+// Exposes an HTTP client to call the Circle Cross-Chain Transfer Protocol (CCTP) V2 API.
+//
+// The goal is to fetch attestations for USDC transfers made via CCTPv2 by CCIP. The /v2/messages/ endpoint
+// takes a source domain ID and transaction hash and returns a list of CCTPv2 messages executed on that source and
+// transaction. Each message contains an attestation and a decoded message body, which are needed to pass to
+// destination chain contracts where the attestation will be verified and the USDC transfer will be executed (e.g. the
+// USDC will be minted and transferred on the destination chain).
+//
+// API docs are here: https://developers.circle.com/api-reference/stablecoins/common/get-messages-v-2
+//
+// An example API calls is:
+//nolint:lll
+// https://iris-api-sandbox.circle.com/v2/messages/0?transactionHash=0x8f1d58a2161b0bd662609b62537d57b76cfa6bb22a9afe8ec49b09f3810d44c2
+
+package cctpv2
+
+import (
+	"context"
+	"encoding/json"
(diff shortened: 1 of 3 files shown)


execute/tokendata/usdc/cctpv2/http_client.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func NewCTTPv2AttestationClient(
+	lggr logger.Logger,
+	config pluginconfig.USDCCCTPObserverConfig,
+) (*CTTPv2AttestationClient, error) {
+	client, err := http.GetHTTPClient(
+		lggr,
+		config.AttestationAPI,
+		config.AttestationAPIInterval.Duration(),
```
mateusz-sekara, line 50: How does the rate limiting work when both v1 and v2 are in use? I mean, is there a shared rate limit state between v1 and v2? Or is the rate limitting per api version? `http.GetHTTPClient` enforces a single instance (with shared rate limit) per `attestation api` with version exluded

rstout: I assume the rate limiting is shared across v1 and v2

execute/tokendata/usdc/cctpv2/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+	)
+)
+
+func TrackHTTPRequest(sourceDomainID uint32, status http.HTTPStatus, isErr bool, duration time.Duration) {
+	sourceDomainIDStr := fmt.Sprintf("%d", sourceDomainID)
+	statusStr := fmt.Sprintf("%d", status)
+	errorStr := fmt.Sprintf("%t", isErr)
+
+	PromHTTPRequest.
+		WithLabelValues(sourceDomainIDStr, statusStr, errorStr).
```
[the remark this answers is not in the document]
rstout: Good point, I've addressed this in the main PR I'm working on

rstout:
This PR was rolled into https://github.com/smartcontractkit/chainlink-ccip/pull/1042

PR #1061, opened 2025-07-02
Feat/limit router address

agusaldasoro:
- Added a default value for router address and rmn adress so token owners don't need to set that up
- Created a new message for token pool program upgrade authority to modify those addresses only for the token pools they owned
- Emit events for all those changes


chains/solana/contracts/programs/burnmint-token-pool/src/lib.rs
```diff
@@ -19,45 +19,78 @@ pub mod burnmint_token_pool {
...
+        emit!(GlobalConfigUpdated {
+            self_served_allowed: ctx.accounts.config.self_served_allowed,
+            router: ctx.accounts.config.router,
+            rmn_remote: ctx.accounts.config.rmn_remote,
+        });
+
+        Ok(())
+    }
+
+    pub fn initialize(ctx: Context<InitializeTokenPool>) -> Result<()> {
```
toblich, line 85: I think you're already working on it, but just in case: The removal of the two Pubkeys in the instruction data will require changes in tooling, probably e2e tests, and should be accounted for in the CCT docs (whenever those are available anyway)

toblich (approved):

PR #1068, opened 2025-07-07
Fix/metaplex extension

agusaldasoro:


chains/solana/contracts/programs/ccip-router/src/instructions/v1/pools.rs
```diff
@@ -47,7 +48,10 @@ pub fn transfer_token<'info>(
     signer: &AccountInfo<'info>,
     seeds: &[&[u8]],
 ) -> std::result::Result<(), ProgramError> {
-    let mint_data = Mint::unpack(*mint.try_borrow_data()?)?;
+    let data = mint.try_borrow_data()?;
+    let state = StateWithExtensions::<Mint>::unpack(&data)?;
```
agusaldasoro, line 52: This is the important change, this way of deserializing works for SPL Token, Token 22 without extensions and with extensions

toblich (approved):

PR #1073, opened 2025-07-08
Add TxHash to Message

rstout:
TxHash was recently added to the Sequence type in contract_reader. In this commit we bump the version of chainlink-common to access this new field and use it to populate the TxHash field of Message. The forthcoming CCTPv2 attestation logic needs TxHash to fetch attestations for a Message's USDC transfers.


pkg/chainaccessor/default_accessor.go
```diff
@@ -172,6 +174,7 @@ func (l *DefaultAccessor) MsgsBetweenSeqNums(
 		}
 
 		msg.Message.Header.OnRamp = onRampAddress
+		msg.Message.Header.TxHash = "0x" + hex.EncodeToString(item.TxHash)
```
mateusz-sekara, line 177: I suppose this pattern is already implemented as `Bytes.String()` function. It might be worth replacing it with `NewBytesFromString(item.TxHash).String()` or something similar

mateusz-sekara: Can we have a tiny unit test to cover that change?

PR #1079, opened 2025-07-11
Bugfix in chainfee validation

dimkouv:
While validating chain fee observations the wrong chains were used.

Which means that even when an oracle observed the actual chains it is supposed to observe it's observation might be blocked.

This is because chain fee update observations are only read from the destination chain and not from their individual source chains.

### DIFF
--- a/commit/chainfee/validate_observation.go
+++ b/commit/chainfee/validate_observation.go
@@ -31,7 +31,7 @@ func (p *processor) ValidateObservation(
 	if err != nil {
 		return fmt.Errorf("failed to get supported chains: %w", err)
 	}
-	if err := validateObservedChains(ao, observerSupportedChains); err != nil {
+	if err := validateObservedChains(ao, observerSupportedChains, p.destChain); err != nil {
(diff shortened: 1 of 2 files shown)


commit/chainfee/validate_observation.go
```diff
@@ -104,14 +104,20 @@ func validateFeeComponents(
...
-	observedChains = append(observedChains, maps.Keys(obs.ChainFeeUpdates)...)
+	if len(obs.ChainFeeUpdates) > 0 {
+		// chainFeeUpdates are read from the destination chain, if the oracle observed any chain fee updates
+		// it should support the destination chain.
+		observedChains = append(observedChains, destChain)
+	}
+
 	for _, chain := range observedChains {
 		if !observerSupportedChains.Contains(chain) {
 			return fmt.Errorf("chain %d is not supported by observer", chain)
```
winder, line 123: Is this the error you saw? I guess there is a fee update for a source chain that none of the nodes support?

PR #1089, opened 2025-07-18
Bump NPM deps (Solhint & @ changesets/cli)

RensR:
Solhint 6.0 enables many rules by default. See the changelog [here](https://github.com/protofire/solhint/releases/tag/v6.0.0).

Because of this, we turn some rules off that produce issues with our current codebase, and we can remove the explicit enablement of others, as they are now on by default. The intend is to re-enable some of the rules introduced in 6.0 in future PRs.


chains/evm/.solhint-test.json
```diff
@@ -4,36 +4,36 @@
   "rules": {
     "compiler-version": ["off", "^0.8.0"],
     "const-name-snakecase": "off",
-    "constructor-syntax": "error",
```
RensR, line 7: This file is re-ordered to first list all the items that are turned off, and then the rules that we turn on. A few items are added due to them being default in 6.0 and some are removed as they are either no longer required (reason-string) or default (interface-starts-with-i).

chains/evm/contracts/test/pools/FastTransferTokenPool/FastTransferTokenPool.ccipReceive.t.sol
```diff
@@ -87,7 +87,7 @@ contract FastTransferTokenPool_ccipReceive_Test is FastTransferTokenPoolSetup {
   function test_ccipReceive_WithDifferentDecimals() public {
     uint8 sourceDecimals = 6; // USDC-like decimals
     uint8 destDecimals = IERC20Metadata(address(s_token)).decimals();
-    require(sourceDecimals != destDecimals, "Test requires different source and destination decimals");
+    assertTrue(sourceDecimals != destDecimals, "Test requires different source and destination decimals");
```
RensR, line 90: The only change to code in this PR, as it was very minor and allows us to remove a rule override.

matYang (approved):

PR #1095, opened 2025-07-23
fix: ignore missing source CR in usdc

bukata-sa:


pkg/reader/usdc_reader.go
```diff
@@ -91,6 +91,11 @@ func NewUSDCMessageReader(
 		}
 		switch family {
 		case sel.FamilyEVM:
+			contractReader, ok := contractReaders[chainSelector]
+			if !ok {
+				lggr.Warnf("chain reader is missing for chain %d, skipping", chainSelector)
```
mateusz-sekara, line 96: Shouldn't that log on error level? It might be related to some underlying misconfiguration

mateusz-sekara (approved):

PR #1101, opened 2025-07-23
Set Rate Limit Admin

agusaldasoro:
Expose a message to modify the Rate Limit Admin in the Token Pool


chains/solana/contracts/programs/base-token-pool/src/common.rs
```diff
@@ -187,6 +187,25 @@ impl BaseConfig {
         self.can_accept_liquidity = allow;
         Ok(())
     }
+
+    pub fn set_rate_limit_admin(&mut self, new_rate_limit_admin: Pubkey) -> Result<()> {
+        require_keys_neq!(
```
agusaldasoro, line 192: Is this expected?

chains/solana/contracts/programs/lockrelease-token-pool/src/context.rs
```diff
@@ -377,6 +377,18 @@ pub struct SetChainRateLimit<'info> {
...
 }
 
+#[derive(Accounts)]
+#[instruction(mint: Pubkey)]
+pub struct SetRateLimitAdmin<'info> {
+    #[account(
+        seeds = [POOL_STATE_SEED, mint.key().as_ref()],
+        bump,
+    )]
+    pub state: Account<'info, State>,
```
toblich, line 387: Looks like this should be `mut`, otherwise changes just won't be persisted, right?

toblich (commented):
Could you please add a basic integration test for the new method?

chains/solana/contracts/tests/ccip/tokenpool_test.go
```diff
@@ -454,6 +454,54 @@ func TestTokenPool(t *testing.T) {
...
+									Capacity: amount,
+									Rate:     1,
+								}, test_token_pool.RateLimitConfig{
+									Enabled:  false,
+									Capacity: 0,
+									Rate:     0,
+								}, poolConfig, p.Chain[config.EvmChainSelector], user.PublicKey(), solana.SystemProgramID).ValidateAndBuild()
+							require.NoError(t, err)
+
+							testutils.SendAndFailWith(ctx, t, solanaGoClient, []solana.Instruction{ixRateAdmin, ixRatesValid, ixRateAdmin2, ixRates}, user, config.DefaultCommitment, []string{"SetChainRateLimit"}, common.AddSigners(anotherAdmin))
```
toblich, line 502: That `[]string{"SetChainRateLimit"}` only checks that that string is present in the logs, meaning that it will only check that the method was invoked. It won't enforce that that is the method that's failing, nor which of the two invocations you're making to is the one that fails.

I'll approve the PR anyway as I think this is minor compared to other priorities, but if you have the time to improve this check, please do so 🙂. You could check the events to make sure that the first call to that method did work for example.

chains/solana/contracts/programs/cctp-token-pool/src/context.rs
```diff
@@ -624,6 +630,35 @@ pub struct EditChainConfig<'info> {
...
+        seeds = [
+            POOL_CHAINCONFIG_SEED,
+            &remote_chain_selector.to_le_bytes(),
+            mint.key().as_ref(),
+        ],
+        bump,
+    )]
+    pub chain_config: Account<'info, ChainConfig>,
+
+    #[account(mut, constraint = authority.key() == state.config.owner @ CcipTokenPoolError::Unauthorized)]
```
toblich, line 648: Nitpick
```suggestion
    #[account(mut, address = state.config.owner @ CcipTokenPoolError::Unauthorized)]
```

chains/solana/contracts/programs/cctp-token-pool/src/context.rs
```diff
@@ -638,7 +673,21 @@ pub struct EditChainConfig<'info> {
     )]
     pub chain_config: Account<'info, ChainConfig>,
 
-    #[account(mut, constraint = authority.key() == state.config.owner || authority.key() == state.config.rate_limit_admin)]
```
toblich, line 641: Thanks a lot for finding this and fixing it 🫶

toblich (approved):
Approved, though if you have time to improve the test please do so (see comment)

PabloMansanet (approved):

PR #1102, opened 2025-07-23
Solana CCTP miscellanous fixes

toblich:
Suggestion for reviewers: Look at the diff commit-by-commit, as they are all fairly atomic

### DIFF
--- a/chains/solana/contracts/programs/ccip-common/src/v1.rs
+++ b/chains/solana/contracts/programs/ccip-common/src/v1.rs
@@ -266,7 +266,8 @@ fn load_v1_token_admin_registry(
 ) -> Result<TokenAdminRegistry> {
     const ANCHOR_DISCRIMINATOR_SIZE: usize = 8;
 
-    let data = token_admin_registry.try_borrow_data()?;
+    let borrowed_data = token_admin_registry.try_borrow_data()?;
+    let (discriminator, data) = borrowed_data.split_at(ANCHOR_DISCRIMINATOR_SIZE);
 
     require_keys_eq!(
         token_admin_registry.owner.key(),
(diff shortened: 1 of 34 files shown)


chains/solana/contracts/programs/ccip-offramp/src/instructions/v1/execute.rs
```diff
@@ -499,18 +499,15 @@ impl<'a> ExecuteReportContextRemainingAccountsLayout<'a> {
...
             .iter()
             .flat_map(|accs| {
                 Some(accs.logic_receiver.key())
                     .into_iter()
                     .chain(accs.remaining_messaging_accounts.iter().map(|a| a.key()))
             })
-            .collect();
-
-        keys.into_iter()
+            .collect()
```
PabloMansanet, line 510: The old function isn't quite right, but I'm not 100% on the new one either. Why not:

```rust
    pub fn receiver_and_message_account_keys(&self) -> impl Iterator<Item = Pubkey> + 'a {
        self.messaging_accounts.iter().flat_map(|accs| {
            Some(accs.logic_receiver.key())
                .into_iter()
                .chain(accs.remaining_messaging_accounts.iter().map(|a| a.key()))
        })
    }
```

Always best to return an existential iterator instead of a Vec in case the caller doesn't want to use heap memory.

chains/solana/contracts/programs/ccip-common/src/v1.rs
```diff
@@ -272,29 +275,15 @@ fn load_v1_token_admin_registry(
     );
 
     require!(
-        TokenAdminRegistry::DISCRIMINATOR == data[..8],
+        TokenAdminRegistry::DISCRIMINATOR == data[..ANCHOR_DISCRIMINATOR_SIZE],
```
PabloMansanet, line 278: This is just a nit, and one that there's no need to act on now as it's pervasive on the codebase, but for situations like this in the future I think it's a bit cleaner to use [split_at](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at) rather than using slice ranges.

chains/solana/contracts/programs/ccip-common/src/v1.rs
```diff
@@ -272,29 +275,15 @@ fn load_v1_token_admin_registry(
...
-        pending_administrator: Pubkey::new_from_array(data[41..73].try_into().unwrap()),
-        lookup_table: Pubkey::new_from_array(data[73..105].try_into().unwrap()),
-        writable_indexes: [
-            u128::from_le_bytes(data[105..121].try_into().unwrap()),
-            u128::from_le_bytes(data[121..137].try_into().unwrap()),
-        ],
-        mint: Pubkey::new_from_array(data[137..169].try_into().unwrap()),
-        supports_auto_derivation: false, // this is the field that is missing in v1, and it defaults to false
-    })
+    Ok(TokenAdminRegistry::try_from(token_admin_registry_v1)?)
```
PabloMansanet: Much much cleaner 👍

PabloMansanet (commented):
LGTM, just one error message change that I think needs to happen, but my other comments are minor thoughts.

agusaldasoro (approved):

PR #1104, opened 2025-07-24
Fix exec plugin observation validation

dimkouv:
Properly validate observer reading eligibility.
Depending on the state that we are currently in and depending on what the oracle has observed validate the chains it did access. So far the validation was wrong requiring the oracle to support all the source chains.

### DIFF
--- a/execute/plugin_test.go
+++ b/execute/plugin_test.go
@@ -953,7 +953,7 @@ func Test_getPendingReportsForExecution(t *testing.T) {
 				tt.fetchFrom,
 				tt.cursedSourceChains,
 				int(offchainConfigForTest.MaxCommitReportsToFetch), // limit int
-				currentTestLogger,                                  // lggr logger.Logger
+				currentTestLogger,
 			)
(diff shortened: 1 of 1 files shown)


bukata-sa (approved):

mateusz-sekara (dismissed):

execute/plugin.go
```diff
@@ -384,7 +384,9 @@ func (p *Plugin) ValidateObservation(
 		}
 	}
 
-	if err = validateCommonStateObservations(p, ao.Observer, decodedObservation, supportedChains); err != nil {
+	state := previousOutcome.State.Next()
+
+	if err = validateCommonStateObservations(p, ao.Observer, decodedObservation, supportedChains, state); err != nil {
```
asoliman92, line 389: I think it's better to have the switch case on the states in the main validation function instead for redladability and leave commonState for things that are really common.  You see already just before this call we check based on states. (using `state` as variable name is better than `nextState`) 

Otherwise logic looks correct.

asoliman92 (approved):

PR #1115, opened 2025-07-30
New ERC20 Lockbox Contract for easier migrations

jhweintraub:
This PR is the first of several which will be included in the changes to the USDC Token Pool PR. 

It includes a new contract, the ERC20 Lockbox which Lock-Release token pools can use to store tokens in long-term, preventing the need for a liquidity migration when token pools are upgraded. As of this PR only the SiloedLockReleaseTokenPool has been migrated to using the new LockBox. A future PR will be planned to migrate the Lock-Release token pool proper since that involves more heavy changes of tests. 

It also includes a small change to TokenPool.sol, renaming the error `ZeroAddressNotAllowed()` to `ZeroAddressInvalid()` to avoid a future naming collision with the Shared Library's AuthorizedCaller error, which is going to be used in other PR's after this.


chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+import {Ownable2StepMsgSender} from "@chainlink/contracts/src/v0.8/shared/access/Ownable2StepMsgSender.sol";
+
+import {IERC20} from
+  "@chainlink/contracts/src/v0.8/vendor/openzeppelin-solidity/v4.8.3/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from
+  "@chainlink/contracts/src/v0.8/vendor/openzeppelin-solidity/v4.8.3/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/// @title ERC20 Lock Box
+/// @notice A contract that holds ERC20 tokens for a token pool to simplify pool upgrades without requiring a liquidity migration.
+contract ERC20LockBox is Ownable2StepMsgSender {
```
matYang, line 15: nit: lets implement `ITypeAndVersion`

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+  struct AllowedCallerConfigArgs {
+    address token;
+    address caller;
+    bool allowed;
+  }
+
+  TokenAdminRegistry public immutable i_tokenAdminRegistry;
+
+  mapping(address => mapping(address => bool)) internal s_allowedCallers;
+  mapping(address => mapping(uint64 => uint256)) internal s_tokenBalances;
```
matYang, line 41: We should call out in the comments that this is using a model where many token & pools share the same lock box, this certainly adds a degree of risk, let's articulate the benefit in comments as well

chains/evm/contracts/pools/TokenPool.sol
```diff
@@ -42,7 +42,7 @@ abstract contract TokenPool is IPoolV1, Ownable2StepMsgSender {
   using RateLimiter for RateLimiter.TokenBucket;
 
   error CallerIsNotARampOnRouter(address caller);
-  error ZeroAddressNotAllowed();
+  error ZeroAddressInvalid();
```
matYang, line 45: since you ended up not using authorizedCaller, this is no longer needed right

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+      revert TokenAdminRegistryCannotBeZeroAddress();
+    }
+    i_tokenAdminRegistry = TokenAdminRegistry(tokenAdminRegistry);
+  }
+
+  /// @notice Deposits tokens for a specific remote chain selector
+  /// @param token The address of the ERC20 token to deposit
+  /// @param amount The amount of tokens to deposit
+  /// @param remoteChainSelector The chain selector for which to deposit tokens
+  function deposit(address token, uint256 amount, uint64 remoteChainSelector) external {
```
matYang, line 56: let's add a big note that remoteChainSelector = 0 is a valid case for a vanilla LockReleaseTokenPool where liquidity is to be shared across all dests

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+
+    // Transfer tokens from this contract to the recipient
+    IERC20(token).safeTransfer(recipient, amount);
+
+    // Emit the withdrawal event
+    emit Withdrawal(token, remoteChainSelector, recipient, amount);
+  }
+
+  /// @notice Configures the allowed callers for deposit and withdraw functions
+  /// @dev Only the owner can configure allowed callers
```
matYang, line 97: nit: Only the token admin returned by TokenAdminRegistry can configure allowed callers

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+    // Iterate through all configuration arguments
+    for (uint256 i = 0; i < configArgs.length; ++i) {
+      // Validate the token address
+      address token = configArgs[i].token;
+      if (token == address(0)) {
+        revert TokenAddressCannotBeZero();
+      }
+
+      // Only the CCIP administrator of the token can configure allowed callers
+      if (!i_tokenAdminRegistry.isAdministrator(token, msg.sender)) {
```
matYang, line 112: With the current setup, this contract does not have any owner-privileged functions (rightfully so), it doesn't need to inherit `Ownable2StepMsgSender` right?

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+
+  struct AllowedCallerConfigArgs {
+    address token;
+    address caller;
+    bool allowed;
+  }
+
+  TokenAdminRegistry public immutable i_tokenAdminRegistry;
+
+  mapping(address => mapping(address => bool)) internal s_allowedCallers;
```
RensR, line 40: Please always use named mappings

chains/evm/contracts/pools/ERC20LockBox.sol
RensR: +1, please add some comments on why all of this is happening

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+  constructor(
+    address tokenAdminRegistry
+  ) {
+    if (tokenAdminRegistry == address(0)) {
+      revert TokenAdminRegistryCannotBeZeroAddress();
+    }
+    i_tokenAdminRegistry = TokenAdminRegistry(tokenAdminRegistry);
+  }
+
+  /// @notice Deposits tokens for a specific remote chain selector
```
RensR, line 52: This comment explains *what* happens, not *why*. Why would anyone use this function?

chains/evm/contracts/pools/TokenPool.sol
RensR: Why rename this? We use ZeroAddressNotAllowed in more contracts and it would be best to keep the name consistent.

chains/evm/contracts/pools/SiloedLockReleaseTokenPool.sol
```diff
@@ -65,15 +76,21 @@ contract SiloedLockReleaseTokenPool is TokenPool, ITypeAndVersion {
     // super.lockOrBurn will validate the lockOrBurnIn and revert if invalid.
     out = super.lockOrBurn(lockOrBurnIn);
 
+    uint64 remoteChainSelector;
```
RensR, line 79: Please add a comment why the default value is used and why that's expected

chains/evm/contracts/pools/SiloedLockReleaseTokenPool.sol
```diff
@@ -101,15 +118,18 @@ contract SiloedLockReleaseTokenPool is TokenPool, ITypeAndVersion {
     uint256 availableLiquidity = chainIsSiloed ? remoteConfig.tokenBalance : s_unsiloedTokenBalance;
     if (localAmount > availableLiquidity) revert InsufficientLiquidity(availableLiquidity, localAmount);
 
+    uint64 remoteChainSelector;
```
RensR, line 121: again, comment on why default value is expected

chains/evm/contracts/pools/SiloedLockReleaseTokenPool.sol
```diff
@@ -267,11 +287,11 @@ contract SiloedLockReleaseTokenPool is TokenPool, ITypeAndVersion {
   /// @param amount The amount of liquidity to provide.
   function provideLiquidity(
```
RensR, line 288: You could argue these functions should no longer exist on this pool, and liquidity providers should use the lockbox directly

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+    if (!isAllowedCaller(token, msg.sender)) {
+      revert Unauthorized(msg.sender);
+    }
+
+    if (amount == 0) {
+      revert TokenAmountCannotBeZero();
+    }
+  }
+
+  /// @notice Checks if an address is allowed to call deposit and withdraw functions
```
RensR, line 154: For all comments: They end with a period `.`

chains/evm/contracts/pools/ERC20LockBox.sol
RensR: Maybe worth noting somewhere that this does give FULL control over all the funds to the token admin, something they don't (always) have with todays LR pools.

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+      revert InsufficientBalance(remoteChainSelector, amount, balance);
+    }
+
+    // Decrease the balance for the specified chain selector
+    s_tokenBalances[token][remoteChainSelector] -= amount;
+
+    // Transfer tokens from this contract to the recipient
+    IERC20(token).safeTransfer(recipient, amount);
+
+    // Emit the withdrawal event
```
RensR, line 92: nit: please remove comments like these

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /// @notice Configures the allowed callers for deposit and withdraw functions
+  /// @dev Only the owner can configure allowed callers
+  /// @dev Can add or remove multiple callers in a single transaction
+  /// @param configArgs Array of configuration arguments for allowed callers
+  function configureAllowedCallers(
+    AllowedCallerConfigArgs[] calldata configArgs
+  ) external {
+    // Iterate through all configuration arguments
```
RensR, line 103: nit: please remove comments like these

RensR (commented):
Do we need to support the case where a single tokens liquidity can be owned by multiple entities? e.g. USDC that doesn't use CCTP will have liquidity that belongs to many different destination chains on Eth. Or do we say "admin of the pool manages that, no need for the chain to have access/control over those funds" as they really are not property of the chain but rather property of the users that have bridged it to that chain?

chains/evm/contracts/pools/TokenPool.sol
jhweintraub: so it was changed to prevent a future issue with the Token Pools and Authorized caller but we can revisit that issue later and I will remove it from this PR.

chains/evm/contracts/pools/TokenPool.sol
```diff
@@ -297,7 +297,10 @@ abstract contract TokenPool is IPoolV1, Ownable2StepMsgSender {
   /// @param localAmount The local amount to be released or minted.
   /// @dev This function should always be called before executing a release or mint. Not doing so would allow
   /// for various exploits.
-  function _validateReleaseOrMint(Pool.ReleaseOrMintInV1 calldata releaseOrMintIn, uint256 localAmount) internal {
+  function _validateReleaseOrMint(
```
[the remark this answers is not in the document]
jhweintraub: _onlyOnRamp and _onlyOffRamp do need to be marked virtual since the next PR for USDC Token Pools is going to need to modify those functions. There's no way around it since we decided we did NOT want to make the USDC Pool proxy an allowed on/off Ramp. Pick your poison.

chains/evm/contracts/pools/ERC20LockBox.sol
jhweintraub: do we want to consider maybe adding rate limits then like we have with the pools?

chains/evm/contracts/pools/SiloedLockReleaseTokenPool.sol
matYang: no strong opinion but feels safer to keep interface as is

chains/evm/contracts/pools/ERC20LockBox.sol
matYang: doubt if rate limit solve the problem, because there will need to be a rate limit admin, which is probably the token admin, then we are back to square one;

I think that Rens is calling out is that token admin (or token issuer) and liquidity provider can differ, especially for more decentralized projects, there could be a DAO, but liquidity is managed by a market maker; in this scenario, the existing implementation will grant the DAO full control over the market maker's funds

RensR: Yeah you directly link token admin to access to tokens, which is new. Previously it would be up to the token team to decide access levels for setPool and internal pool management (including liquidity) separately. I'm not saying this is inheriently wrong, just that it should be very explicitly stated.

chains/evm/contracts/pools/TokenPool.sol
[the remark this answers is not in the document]
RensR: Probably best to not include "required in the next PR" items in a PR, I think there's two instances of review comments for stuff being useless because the change makes little sense in this context :)

chains/evm/contracts/pools/ERC20LockBox.sol
```diff
@@ -0,0 +1,166 @@
...
+      address token = configArgs[i].token;
+      if (token == address(0)) {
+        revert TokenAddressCannotBeZero();
+      }
+
+      // Note: Only the owner of the token pool itself, which MAY NOT be the administrator of the token in the token
+      // admin registry, can configure allowed callers. Currently, the token pool owner manages liquidity providers,
+      // who are allowed to withdraw liquidity at will. Since an allowed caller can do the same on this contract,
+      // limiting who can manage that configuration ensures there are no additional trust assumptions for managing liquidity.
+      address poolOwner = IOwner(i_tokenAdminRegistry.getPool(token)).owner();
```
matYang, line 120: this does restrict use of this lockbox to token pools that implement Ownable, won't work for pools that use e.g RBAC; should be ok tho given all reference pools use Ownable, if a pool uses RBAC or something else they can use a custom lock box

matYang (approved):

PR #1118, opened 2025-07-31
bump versions and wrappers for 1.6.2 release

jhweintraub:
Bump Versions for USDCTokenPool and CCTPMessageTransmitter Proxy to 1.6.2 official and generate new gobindings for release.


chains/evm/contracts/pools/USDC/CCTPMessageTransmitterProxy.sol
```diff
@@ -4,6 +4,8 @@ pragma solidity ^0.8.24;
 import {IMessageTransmitter} from "./interfaces/IMessageTransmitter.sol";
 import {ITokenMessenger} from "./interfaces/ITokenMessenger.sol";
 
+import {ITypeAndVersion} from "@chainlink/contracts/src/v0.8/shared/interfaces/ITypeAndVersion.sol";
+
```
matYang, line 8: nit: newline not required

matYang (commented):
when running wrapper release and zkSync gen, it does auto pollute with all these files, but we need to clean them up, the end result should be:

under latest, there is only net update of classic go wrappers for the 2 contracts, no metadata files or zksync wrappers
under 1_6_2, only copy the 2 contracts being released there, under each contract folder, there are classfic go wrapper, metadata file, and zksync wrapper

matYang (approved):

PR #1131, opened 2025-08-06
add modsec modules

makramkd:


chainlink-modsec/modsectypes/types.go
```diff
@@ -0,0 +1,181 @@
...
+	Transmit(ctx context.Context, encodedMessage []byte, proofs [][]byte, extraPayload []byte) error
+}
+
+// Signer is an interface that provides a way to sign a message.
+// This is expected to be implemented per chain family.
+type Signer interface {
+	// Sign signs the provided digest. For EVM the digest is 32 bytes long though for different
+	// hashing functions it may be longer, hence the []byte rather than [32]byte.
+	Sign(ctx context.Context, digest []byte) ([]byte, error)
+}
```
winder, line 172: I always wonder why we don't just use the crypto package for [this](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/crypto/crypto.go;l=180):
```
// Signer is an interface for an opaque private key that can be used for
// signing operations. For example, an RSA key kept in a hardware module.
type Signer interface {
	// Public returns the public key corresponding to the opaque,
	// private key.
	Public() PublicKey

	// Sign signs digest with the private key, possibly using entropy from
	// rand. For an RSA key, the resulting signature should be either a
	// PKCS #1 v1.5 or PSS signature (as indicated by opts). For an (EC)DSA
	// key, it should be a DER-serialised, ASN.1 signature structure.
	//
	// Hash implements the SignerOpts interface and, in most cases, one can
	// simply pass in the hash function used as opts. Sign may also attempt
	// to type assert opts to other types in order to obtain algorithm
	// specific values. See the documentation in each package for details.
	//
	// Note that when a signature of a hash of a larger message is needed,
	// the caller is responsible for hashing the larger message and passing
	// the hash (as digest) and the hash function (as opts) to Sign.
	Sign(rand io.Reader, digest []byte, opts SignerOpts) (signature []byte, err error)
}
```

PR #1135, opened 2025-08-09
Verifier main loop.

winder:
Sketch in a main loop for the verification logic.

image

### DIFF
--- a/chainlink-modsec/libmodsec/internal/verifier/main.go
+++ b/chainlink-modsec/libmodsec/internal/verifier/main.go
@@ -1,0 +1,100 @@
+package verifier
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"sync"
+)
+
+// run is the verifier main loop.
+//
+// Implementation of a pipeline:
+//   - Read events from blockchain.
+//   - Transform events to the verifier payload.
+//   - * Write raw message to storage.
+//   - Dispatch payloads to verifiers.
+//   - Collect results from verifiers as they are available.
+//   - Write verifier attestations to storage.
+func (s *Verifier) run() {
+	// Start reader
+	s.started = true
+	if s.poller == nil {
+		log.Println("No poller configured, verifier will not process any work.")
+		// TODO: surface an error.
+		return
+	}
+	if s.transformer == nil {
+		log.Println("No transformer configured, verifier will not process any work.")
+		// TODO: surface an error.
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	var wg sync.WaitGroup
+
+	workCh := make(chan Work)
+	messageCh := make(chan StoredMessage)
+	handlerPayloadCh := make(chan HandlerPayload)
+	resultCh := make(chan Result)
+
+	// produce data
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+
+		for true {
+			select {
+			case <-ctx.Done():
+				log.Println("Verifier shutting down poller...")
+				return
+			case data := <-s.poller.Watch(ctx):
+				fmt.Println(data)
+				workCh <- data
+			}
+		}
+	}()
+
+	// transform data
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+
+		for true {
+			select {
+			case <-ctx.Done():
+				log.Println("Verifier shutting down transformer...")
+				return
+			case work := <-workCh:
+				payload := s.transformer.Transform(work)
+				fmt.Println("Payload transformed:", payload)
+				handlerPayloadCh <- payload
+				// TODO:: send StoredMessage to 'messageCh', it can be written ahead of verification.
+			}
+		}
+	}()
+
+	// run handlers
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+
+		for true {
+			select {
+			case <-ctx.Done():
+				log.Println("Verifier shutting down handlers...")
+				return
+			case payload := <-handlerPayloadCh:
+				log.Printf("Running handlers for payload: %v\n", payload)
+				// Dispatch payload to each handler.
+				for _, handler := range s.handlers {
+					go handler(ctx, payload, resultCh)
+				}
+			}
+		}
+	}()
+
+	// write results
+}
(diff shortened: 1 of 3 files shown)


chainlink-modsec/libmodsec/internal/verifier/interfaces.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// Reader is an interface for reading the blockchain for new work.
+type Reader interface {
+	// Next returns a channel that will yield the next piece of work to be processed.
+	Next(ctx context.Context) <-chan Work
+
+	// Watch returns a channel that will yield all work to be processed.
+	Watch(ctx context.Context) <-chan Work
+}
```
makramkd, line 14: This needed with the `SourceReader` interface already defined?

```golang
// SourceReader is an interface that provides a way to read messages from the source chain onramp.
// This is expected to be implemented per chain family.
type SourceReader interface {
	Service

	// Messages returns a channel of messages that are read from the source chain onramp.
	Messages() <-chan Message

	// GetMessages returns a slice of messages from the source chain onramp starting
	// from the given sequence number.
	GetMessages(ctx context.Context, query MessageQueryArgs) ([]Message, error)
}
```

chainlink-modsec/libmodsec/internal/verifier/interfaces.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Transformer is an interface that defines how to transform a Work item into
+// the generic modsec.Message format and encoded into a destination specific
+// payload. This interface encapsulates most of the chain agnostic services
+// required by the 1.6 implementation.
+type Transformer interface {
+	Transform(work Work) HandlerPayload
+}
```
makramkd, line 22: Encapsulated in the interface above - could make things more generic if we think its useful though.

chainlink-modsec/libmodsec/internal/verifier/interfaces.go
```diff
@@ -0,0 +1,51 @@
...
+	// The writer implementation is responsible for deciding how to store the message.
+	// It should consider things like how it will be retrieved later, and how to avoid duplicates.
+	// For example, it may decide to store messages in a hierarchy based on the time, block number, sequence number, etc.
+	WriteMessage(ctx context.Context, msg HandlerPayload) error
+
+	// WriteAttestation stores the attestation for a message.
+	// The writer implementation is responsible for deciding how to store the attestation.
+	// It should consider things like how it will be retrieved later, and how to avoid duplicates.
+	WriteAttestation(ctx context.Context, msg Attestation) error
+}
```
makramkd, line 41: What I landed on:

```golang
// AttestationWriter defines the interface for storing attestations.
type AttestationWriter interface {
    Store(ctx context.Context, att Attestation) error
}
```

The `Attestation` is the object that can be used by the implementation to come up with the key path (if backend is S3) and contains the full CCIP message as well as the message proof, so I don't think we need two separate methods.

chainlink-modsec/libmodsec/internal/verifier/interfaces.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Handler is a function that processes incoming work from the work channel and sends results to the result channel.
+// There may be no results for a particular piece of work, in which case the result channel should not be written to.
+// For example, the verifier may have a rule to skip processing until a certain block depth is reached. In that case,
+// the work may be received, but no result will be produced until later.
+//
+// The verifier should respect the context for cancellation. If the context is cancelled, the verifier should stop.
+//
+// TODO: should verifier be a full service with start/stop/cancel, a cache, maybe a db connection, etc?
+type Handler func(ctx context.Context, payload HandlerPayload, result chan<- Attestation)
```
makramkd, line 51: I have the verifier logic encapsulated in an interface:

```golang
// Verifier is responsible for performing the specific verification logic for a message.
type Verifier interface {
    // Verify tries to verify the given message and returns nil if the message is valid.
    // For example, the commit verifier can check that the message is sufficiently deeply buried in the blockchain,
    // or is finalized/safe.
    Verify(ctx context.Context, message CCIPMessage) error
}
```

Method on it like that is synchronous, though we could have an async version or make it fully async with the channel input.

chainlink-modsec/libmodsec/internal/verifier/main.go
```diff
@@ -0,0 +1,143 @@
...
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+
+		for true {
+			select {
+			case <-ctx.Done():
+				fmt.Println("Verifier shutting down handlers...")
+				return
+			case payload := <-writePayloadCh:
```
makramkd, line 112: Seems kinda overkill to make message encoding an async thing

chainlink-modsec/libmodsec/internal/verifier/main.go
```diff
@@ -0,0 +1,143 @@
...
+				for _, w := range s.writer {
+					if err := w.WriteMessage(ctx, payload); err != nil {
+						// TODO: surface an error.
+						fmt.Println("Error writing payload.")
+					}
+				}
+			case attestation := <-writeAttestationCh:
+				for _, w := range s.writer {
+					if err := w.WriteAttestation(ctx, attestation); err != nil {
```
makramkd, line 121: Message encoding is required for attestation; if you'd want to make it a full pipeline you'd have to encode first then attest second

chainlink-modsec/libmodsec/internal/verifier/interfaces.go
winder: I'd leave out `GetMessages` and especially `query`. If we need to rollback the reader for some reason it should be as simple as possible, boil it down to which round to start on and simply stream everything again from that point.

chainlink-modsec/libmodsec/internal/verifier/interfaces.go
winder: I think its important to split reading and re-encoding. This layer will have chain family specific re-encoding details. So may as well split it up from the blockchain reading so that each has a single responsibility.

chainlink-modsec/libmodsec/internal/verifier/interfaces.go
winder: I think this assumes that `Verify` deals with creating the attestation, encoding a dest specific message, signing and writing.

IMO this is way too complex for a component we hope for 3rd parties to implement.

chainlink-modsec/libmodsec/internal/verifier/main.go
asoliman92: Why full message encoding needed? Shouldn't just the message hash be enough here?

PR #1139, opened 2025-08-14
[WIP] add initial executor

0xAustinWang:
Initial Executor based on [discussion from 8/13](https://chainlink-core.slack.com/archives/C090DUHHJU9/p1755106115654429). This models a lot of the code off [Will's initial PR for the verifier](https://github.com/smartcontractkit/chainlink-ccip/pull/1135/files)

## Highlights:
* In-memory attestation indexer
  * Uses a syncmap to model messageID: {verifierID1: Proof, verifierID2: Proof}
* In-memory attestation reader
  * An executor-level object for interacting with the in-memory attestation indexer
* `executor/interfaces.go`
  * SubscribeMessages
      * This subscribes to the messages coming from the CL verifier and uses any messages there as a trigger for work to process
    * GetAttestations
      * Fetches attestations for a message. AttestationReader that wants to subscribe to verifier storages directly should handle that in the implementation. the simple in-memory attestation reader interacts with an indexer
 * `executor/main.go`
    *  Spawns a bunch of workers and the piping between them. 
    * Step 1: subscribe to messages
       * Checks if it's my turn, how do other executes receive the message if it's not my turn?
    * Step 2: verify and get attestations for messages
    * Step 3: Transmit attestation and message to contract
  * `executor/types.go`
    * Creates some internal types for communicating within the executor. 
    
## Open questions 
* Do proofs ([]byte) need to be associated with the verifier ID when submitting to chain? 
* What order to the proofs need to be in?
* Do we assume that all verifiers specified in the message must submit a proof for it to be valid? What does requesting `m of n` verifiers look like in a message?

### DIFF
--- a/chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
+++ b/chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
@@ -3,16 +3,33 @@ package attestation_indexer
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	"golang.org/x/sync/syncmap"
 
-	"github.com/smartcontractkit/chainlink-modsec/libmodsec/internal/executor"
 	"github.com/smartcontractkit/chainlink-modsec/libmodsec/pkg/modsectypes"
 )
 
 // InMemoryAttestationIndexer is an in-memory implementation of an attestation indexer
-// It stores all proofs in an in-memory map, indexed by messageID([32]byte) and then by verifierID([32]byte)
+// It stores all proofs in an in-memory map, indexed by messageID([32]byte) and then by veriiferType(enum) and then verifierID([32]byte)
 // This implementation is useful when assuming that attestations will be requested via messageID
+// we only organize by verifierType here for the sake of "VerifierTypeMeetsQuorum" function
+// Normally, we can just get verifierType from verifierID, but that would require loading all verifiers and iterating over them
+// There are significantly more optimizations we can make in a real indexer solution
+/**
+at.attestations: {
+	[messageID1]: {
+		[verifierType1]: {
+			[verifierID1]: proof1,
+			[verifierID2]: proof2,
+		},
+		[verifierType2]: {
+			[verifierID3]: proof3,
+			[verifierID4]: proof4,
+		},
+	},
+}
+*/
 type InMemoryAttestationIndexer struct {
 	attestations syncmap.Map // concurrency-safe map of messageID to map of verifierID to attestation(proof)
 }
@@ -26,18 +43,24 @@ func NewInMemoryAttestationIndexer() *InMemoryAttestationIndexer {
 	return a
 }
 
-// GetAttestationsForMessage retrieves attestations for a given message and converts it into the executor.Attestation type
-func (at *InMemoryAttestationIndexer) GetAttestationsForMessage(_ context.Context, msg modsectypes.Message) ([]executor.Attestation, error) {
+// GetAttestationsForMessage retrieves attestations for a given message and converts it into the modsectypes.Attestation type
+func (at *InMemoryAttestationIndexer) GetAttestationsForMessage(_ context.Context, msg modsectypes.Message) ([]modsectypes.Attestation, error) {
 	// Retrieve the map of attestations for the given messageID
-	verifierProofs, ok := at.attestations.Load(msg.Header.MessageID)
+	verifierTypesToIDs, ok := at.attestations.Load(msg.Header.MessageID)
 	if !ok {
 		return nil, nil // No attestations found for this messageID
 	}
-	attestations := make([]executor.Attestation, 0)
-	verifierProofs.(*syncmap.Map).Range(func(key, value interface{}) bool {
-		attestations = append(attestations, executor.Attestation{
-			Proof:      value.([]byte),
-			VerifierId: key.([32]byte),
+	attestations := make([]modsectypes.Attestation, 0)
+
+	// Iterate over the verifier types for this messageID
+	verifierTypesToIDs.(*syncmap.Map).Range(func(verifierType, verifierIdsMap interface{}) bool {
+		verifierIdsMap.(*syncmap.Map).Range(func(verifierId, proof interface{}) bool {
+			// Append the attestation to the list
+			attestations = append(attestations, modsectypes.Attestation{
+				Proof:      proof.([]byte),
+				VerifierId: verifierId.([32]byte),
+			})
+			return true // Continue iterating over verifierIdsMap
 		})
 		return true
 	})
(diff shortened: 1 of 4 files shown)


chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
```diff
@@ -0,0 +1,87 @@
...
+	verifierProofs, ok := at.attestations.Load(msg.Header.MessageID)
+	if !ok {
+		return nil, nil // No attestations found for this messageID
+	}
+	attestations := make([]executor.Attestation, 0)
+	verifierProofs.(*syncmap.Map).Range(func(key, value interface{}) bool {
+		attestations = append(attestations, executor.Attestation{
+			Proof:      value.([]byte),
+			VerifierId: key.([32]byte),
+		})
```
KodeyThomas, line 41: i don't know how we'd do the m of n and if it would be one verifier id with 16 proofs for example

chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
```diff
@@ -0,0 +1,87 @@
...
+		return true
+	})
+
+	return attestations, nil
+}
+
+// MessageHasQuorum checks if the message has a quorum of attestations
+// Although this is less performant in an in-memory indexer, it will be much more performant in production with a real indexer
+// TODO: do we even need this? Not currently used
+func (at *InMemoryAttestationIndexer) MessageHasQuorum(_ context.Context, msg modsectypes.Message, required int) bool {
```
KodeyThomas, line 51: Think this may need to be a `VerifierHasQuorum` as imo we should split up the individual verifier checking of quorum to the message as a whole.

This could then be a thing wrapper around that and checks if `MessageIsReadyToExecute`

chainlink-modsec/libmodsec/internal/executor/executor.go
```diff
@@ -0,0 +1,110 @@
...
+	// custom executors want to subscribe directly to a verifier storage
+	// Example could be a kafka topic, CL commit verifier, etc.
+	MessageSources []MessageReader
+
+	// List of sources which can potentially provide attestations for messages
+	// This could be changed to a map of verifierID to attestation reader
+	AttestationReader AttestationReader
+
+	// Map of destination chains to writers. A writer is some abstraction on the chain access layer
+	Transmitters map[uint64]modsectypes.ContractTransmitter
```
KodeyThomas, line 21: Think this brings up a really interesting point, do we want one off-chain executor per destination chain or one off-chain executor per NOP

chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
0xAustinWang: It wouldn't make sense to have multiple nodes represented with a single verifier id as that would make billing a nightmare. 

We would instead have some mapping of verifier id to verifier type somewhere

```
verifier_types: {
   v_id1: COMMIT
   v_id2: COMMIT
   v_id3: COMMIT
   v_id4: USDC
   v_id5: LBTC
}

CLNodes: {
  node1: [v_id1, v_id4, v_id5],
  node2: [v_id2],
  node3: [v_id3]
}
```

Then checking for "m of n" is looking for 11 of 16 verifiers of type COMMIT

chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
0xAustinWang: You're right, we need some way to check quorum of verifier types, not just quorum across the entire message. This implementation was just to get something on the board assuming that all verifiers specified are needed. 

We would probably have something like 
```
const verifierType int
const (
  commit verifierType = iota
  usdc
  lbtc
)

func verifierTypeMeetsQuorum(ctx context.Context, msgId [32]byte, type verifierType, required int) bool

func messageIsReady(ctx context.Context, msg modsectypes.Message) bool {
  verifierTypesReady := make([]bool, 0)
  for _, verifierType := range msg.ExtraArgs.Verifiers{ 
     verifierTypesReady = append(verifierTypesReady, verifierTypeMeetsQuorum(ctx, msg.Header.Id, verifierType, numRequired)
  }

  // perform an && on all results
  return slices.Contains(verifierTypesReady, false)
}
```

chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
```diff
@@ -0,0 +1,125 @@
...
+func NewInMemoryAttestationIndexer() *InMemoryAttestationIndexer {
+	// Create a new InMemoryAttestationIndexer with an initialized attestations map
+	a := &InMemoryAttestationIndexer{
+		attestations: syncmap.Map{},
+	}
+	return a
+}
+
+// GetAttestationsForMessage retrieves attestations for a given message and converts it into the modsectypes.Attestation type
+func (at *InMemoryAttestationIndexer) GetAttestationsForMessage(_ context.Context, msg modsectypes.Message) ([]modsectypes.Attestation, error) {
```
makramkd, line 47: Does this method only need message ID? Would make it more explicit if we just take `messageID [32]byte` as an argument rather than the full message.

chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
```diff
@@ -0,0 +1,125 @@
...
+	msgToVerifierTypes, _ := at.attestations.LoadOrStore(msgId, &syncmap.Map{})
+
+	// if this is the first attestation for this verifiertype, we need to create a new map for verifierId
+	verifierTypesToVerifierIds, _ := msgToVerifierTypes.(*syncmap.Map).LoadOrStore(verifierType, &syncmap.Map{})
+
+	verifierTypesToVerifierIds.(*syncmap.Map).Store(verifierId, proof)
+	return nil
+}
+
+func (at *InMemoryAttestationIndexer) VerifierTypeMeetsQuorum(_ context.Context, msgId [32]byte, verifierType modsectypes.VerifierType, required int) bool {
```
makramkd, line 99: I think we need to revise `verifierType` - in my mind it should not be used in the API and only verifier ID should be used.

chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
```diff
@@ -0,0 +1,125 @@
...
+// InMemoryAttestationIndexer is an in-memory implementation of an attestation indexer
+// It stores all proofs in an in-memory map, indexed by messageID([32]byte) and then by veriiferType(enum) and then verifierID([32]byte)
+// This implementation is useful when assuming that attestations will be requested via messageID
+// we only organize by verifierType here for the sake of "VerifierTypeMeetsQuorum" function
+// Normally, we can just get verifierType from verifierID, but that would require loading all verifiers and iterating over them
+// There are significantly more optimizations we can make in a real indexer solution
+/**
+at.attestations: {
+	[messageID1]: {
+		[verifierType1]: {
```
makramkd, line 22: Shouldn't this be `verifierID` instead of `verifierType`?

chainlink-modsec/libmodsec/internal/attestation-indexer/in_memory_indexer.go
```diff
@@ -0,0 +1,125 @@
...
+			[verifierID1]: proof1,
+			[verifierID2]: proof2,
+		},
+		[verifierType2]: {
+			[verifierID3]: proof3,
+			[verifierID4]: proof4,
+		},
+	},
+}
+*/
```
makramkd, line 32: How are users supposed to use this API? Do they poll `VerifierTypeMeetsQuorum` until they get `true` and then call `GetAttestationsForMessage`? It seems kinda clunky cuz the former specifies the verifier whereas the latter doesn't.

chainlink-modsec/libmodsec/internal/executor/executor.go
makramkd, line 21: Feels like its more complicated to make the executor multi-chain, I was thinking of the object as scoped to a single chain, could be wrong though.

chainlink-modsec/libmodsec/pkg/modsectypes/types.go
```diff
@@ -179,3 +179,17 @@ type Hasher interface {
...
 	Hash(ctx context.Context, data []byte) ([]byte, error)
 }
+
+type VerifierType int
+
+const (
+	VerifierTypeCLCommit VerifierType = iota // CL Commit verifier
+	VerifierTypeUSDC                         // USDC verifier
+	VerifierTypeLBTC                         // LBTC verifier
+)
```
makramkd, line 189: I feel like this shouldn't be a thing, reasoning is that in a permissionless setup you won't be able to keep up with new verifier "types" being created. Also you could have multiple instances of the same verifier type, e.g. a finalized commit, ftf commit, etc.

chainlink-modsec/libmodsec/internal/executor/executor.go
```diff
@@ -0,0 +1,110 @@
+package executor
+
+import (
+	"fmt"
+	"github.com/smartcontractkit/chainlink-modsec/libmodsec/pkg/modsectypes"
+)
+
+type Executor struct {
+	stopCh    chan struct{}
+	messageCh chan modsectypes.Message
```
makramkd, line 10: These channels are not initialized

chainlink-modsec/libmodsec/internal/executor/main.go
```diff
@@ -0,0 +1,117 @@
...
+			wg.Add(1)
+			defer wg.Done()
+
+			for {
+				select {
+				case <-ctx.Done():
+					fmt.Println("Executor shutting down message reader...")
+					return
+				case msg := <-reader.SubscribeMessages(ctx):
+					if e.isMyTurn(msg) {
```
makramkd, line 28:
```suggestion
					if !e.isMyTurn(msg) {
```

chainlink-modsec/libmodsec/internal/executor/main.go
```diff
@@ -0,0 +1,117 @@
...
+				fmt.Println("Executor shutting down attestation poller...")
+				return
+			case msg := <-e.messageCh:
+
+				// TODO: We need some abstraction here to loop and handle retries when attestations are not yet ready
+				attestations, err := e.AttestationReader.GetAttestations(ctx, msg)
+				if err != nil {
+					// handle error, log it, etc.
+					continue
+				}
```
makramkd, line 61: See "Verifier Watcher", "Per Verifier Launcher", and "Attestation Aggregator" in the design diagram I shared, I believe they play this role.

chainlink-modsec/libmodsec/internal/executor/main.go
```diff
@@ -0,0 +1,117 @@
...
+	done := false
+	for !done {
+		select {
+		case <-e.stopCh:
+			cancel()
+			done = true
+		case <-ctx.Done():
+			done = true
+		}
+	}
```
makramkd, line 114: nit: the idiomatic way to do this is with an `outer` label:

```golang
outer:
  for {
    select {
      case <-e.stopCh:
        break outer
      case <-ctx.Done():
        break outer
    }
  }
```

Though there's room for simplification here, you don't need both `ctx.Done()` and `stopCh`, one would suffice.

chainlink-modsec/libmodsec/internal/executor/main.go
```diff
@@ -0,0 +1,117 @@
...
+		go func(r MessageReader) {
+			wg.Add(1)
+			defer wg.Done()
+
+			for {
+				select {
+				case <-ctx.Done():
+					fmt.Println("Executor shutting down message reader...")
+					return
+				case msg := <-reader.SubscribeMessages(ctx):
```
makramkd, line 27: This calls `SubscribeMessages` every time the branch is exercised? I'm not really sure how this function will be implemented but presumably a new channel will be created for each subscriber, perhaps better to call once and listen to the chan?

makramkd (commented):
Good start - one thing to figure out is the client behavior when reading from the offchain storage endpoint, we have a bunch of methods but its not clear how they can be properly composed together.

In addition I think the `verifierType` approach doesn't work in general, better to just stay on `verifierID`

PR #1146, opened 2025-08-21
CommitOffRamp & CCVAggregator improvements

RensR:
- Ability to revoke config
- Allow using of older signatures
- Versioned ICCVOffRamp
- Use message type in ICCVOffRamp


chains/evm/contracts/applications/CCIPClientExample.sol
```diff
@@ -19,7 +19,7 @@ import {IERC20} from
 // @dev If the client is upgradeable you have significantly more flexibility and
 // can avoid storage based options like the below contract uses. However it's
 // worth carefully considering how the trust assumptions of your client dapp will
-// change if you introduce upgradeability. An immutable dapp building on top of CCIP
+// change if you introduce upgradability. An immutable dapp building on top of CCIP
```
RensR, line 22: Typo

chains/evm/contracts/FeeQuoter.sol
```diff
@@ -508,7 +508,7 @@ contract FeeQuoter is AuthorizedCallers, IFeeQuoter, ITypeAndVersion, IReceiver,
     bytes4 interfaceId
   ) public pure override returns (bool) {
     return interfaceId == type(IReceiver).interfaceId || interfaceId == type(IFeeQuoter).interfaceId
-      || interfaceId == type(ITypeAndVersion).interfaceId || interfaceId == type(IERC165).interfaceId;
```
RensR, line 511: Not needed

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -249,20 +254,25 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
...
-      report.message.header.sourceChainSelector,
-      report.message.receiver,
-      report.message.tokenAmounts.destTokenAddress, // TODO fix
-      report.ccvs
-    );
-
     {
-      bytes memory encodedMessage = abi.encode(report.message);
+      address[] memory requiredPoolCCVs = new address[](0);
+      if (report.message.tokenAmounts.length > 0) {
```
0xsuryansh, line 259: We assumes single token transfers but the code doesn't enforce this assumption anywhere. 

If multiple tokens are included, only the first one's pool `tokenAmounts[0]` CCVs are validated..

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -249,20 +254,25 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
...
+      );
+
       // TODO real hash
-      bytes32 messageHash = keccak256(encodedMessage);
-      // TODO iterate over receiver CCVs not report.
-      for (uint256 i = 0; i < report.ccvs.length; ++i) {
-        ICCVOffRamp(report.ccvs[i]).validateReport(encodedMessage, messageHash, report.ccvData[i], originalState);
+      bytes32 messageHash = keccak256(abi.encode(report.message));
+      for (uint256 i = 0; i < ccvsToQuery.length; ++i) {
+        ICCVOffRampV1(ccvsToQuery[i]).validateReport(report.message, messageHash, report.ccvData[i], originalState);
```
0xsuryansh, line 275: are we expecting `ccvToQuery[I]` to align with `ccvData[I]` ? If not then this might pass the wrong proof to a wrong verifier.

`ccvToQuery` is subset of `AggregatedReport.ccvs`

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -3,32 +3,34 @@ pragma solidity ^0.8.24;
 
 import {IAny2EVMMessageReceiver} from "../interfaces/IAny2EVMMessageReceiver.sol";
 import {IAny2EVMMessageReceiverV2} from "../interfaces/IAny2EVMMessageReceiverV2.sol";
-import {ICCVOffRamp} from "../interfaces/ICCVOffRamp.sol";
+import {ICCVOffRampV1} from "../interfaces/ICCVOffRampV1.sol";
+import {IPoolV2} from "../interfaces/IPoolV2.sol";
+import {Client} from "../libraries/Client.sol";
```
0xsuryansh, line 8: nit : newline

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -3,32 +3,34 @@ pragma solidity ^0.8.24;
...
-  "@chainlink/contracts/src/v0.8/vendor/openzeppelin-solidity/v5.0.2/contracts/utils/structs/EnumerableSet.sol";
 
 contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
   using ERC165CheckerReverting for address;
   using EnumerableSet for EnumerableSet.UintSet;
 
   error ZeroChainSelectorNotAllowed();
   error ExecutionError(bytes32 messageId, bytes err);
-  error NoCCVQuorumReached();
+  error OptionalCCVQuorumNotReached(uint256 wanted, uint256 got);
```
0xsuryansh, line 33: How about `error OptionalCCVQuorumNotReached(uint256 wanted, uint256 got, address[] foundCCVs, address[] missingCCVs);` ?

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -249,20 +254,25 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
...
-    _ensureCCVQuorumIsReached(
-      report.message.header.sourceChainSelector,
-      report.message.receiver,
-      report.message.tokenAmounts.destTokenAddress, // TODO fix
-      report.ccvs
-    );
-
     {
-      bytes memory encodedMessage = abi.encode(report.message);
+      address[] memory requiredPoolCCVs = new address[](0);
```
0xsuryansh, line 258: We should validate `report.ccvData.length != report.ccvs.length`

Maybe check for duplicates here ?

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -289,22 +299,42 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
...
-      _getCCVsFromReceiverAndPool(sourceChainSelector, receiver, tokenPool);
+      _getCCVsFromReceiver(sourceChainSelector, receiver);
+
+    ccvsToQuery = new address[](ccvs.length);
+    uint256 numCCVsToQuery = 0;
 
     for (uint256 i = 0; i < requiredCCV.length; ++i) {
       bool found = false;
       for (uint256 j = 0; j < ccvs.length; ++j) {
         if (ccvs[j] == requiredCCV[i]) {
```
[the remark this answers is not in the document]
RensR: Nah, it's not optional. We also don't check that our FeeQuoter supports the FeeQuoter interfaces. 165 is only for when support is optional

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -313,42 +393,80 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
       for (uint256 j = 0; j < ccvs.length && optionalCCVsToFind > 0; ++j) {
         if (ccvs[j] == optionalCCVs[i]) {
           optionalCCVsToFind--;
```
0xsuryansh, line 395: This means an optional CCV that’s already included as required still counts toward the optional quorum right ? 
If we want optionals beyond required behaviour, then this should below `if (ccvAlreadyIncluded[j]) break;`

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -313,42 +393,80 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
...
-      if (sourceConfig.requiredCCV == address(0)) {
-        requiredCCV = new address[](1);
-        requiredCCV[0] = sourceConfig.defaultCCV;
-
-        return (requiredCCV, new address[](0), 0);
+    SourceChainConfig memory sourceConfig = s_sourceChainConfigs[sourceChainSelector];
+
+    // If the receiver is a contract
+    if (receiver.code.length != 0) {
+      // And the contract implements the IAny2EVMMessageReceiverV2 interface
```
0xsuryansh, line 430:
```suggestion
      // And the contract implements the IAny2EVMMessageReceiverV2 interface.
```

chains/evm/contracts/offRamp/CCVAggregator.sol
RensR: They should. If a pool makes one required, it should not break the optionals. This is intended

0xsuryansh (approved):

0xsuryansh:
🚀

chains/evm/contracts/ocr/SignatureQuorumVerifier.sol
```diff
@@ -2,150 +2,169 @@
...
+
   struct SignatureProof {
     bytes32[] rs; // R components of the signatures.
     bytes32[] ss; // S components of the signatures.
   }
 
-  SignatureConfigConfig internal s_ocrConfig;
-
-  EnumerableSet.AddressSet internal s_signers;
+  mapping(bytes32 configDigest => SignatureConfigConfig config) private s_signatureConfig;
```
makramkd, line 46: Do we have a TODO to transition to no config digest? Think that was a discussion recently cc @connorwstein

chains/evm/contracts/ocr/SignatureQuorumVerifier.sol
```diff
@@ -2,150 +2,169 @@
...
-      revert ConfigDigestMismatch(s_ocrConfig.configDigest, configDigest);
+  /// @notice Validates the signatures of a given report hash.
+  /// @dev The configDigest should be included in the report hash!
+  /// @param reportHash The hash of the report to validate signatures for.
+  /// @param configDigest The config digest determines the signature set to use for validation.
+  /// @param signatureProof The signatures to validate, encoded as a SignatureProof struct.
+  function _validateSignatures(bytes32 reportHash, bytes32 configDigest, bytes memory signatureProof) internal view {
+    // We allow proving of older messages that might have been signed by an older set of signers. This means we need to
+    // get the set of signers for the given configDigest.
+    SignatureConfigConfig storage config = s_signatureConfig[configDigest];
```
makramkd, line 63: Ah no config digest means doing this old signer check is kinda hard. I guess we haven't really landed on the end state of this part of the commit verifier?

In the DD we're basically saying we have to resign in certain cases and we just don't support old signers. I'm not sure the full implications of that, in the steady state, it seems like it should be OK if we have more lenient signature checks as described in [the DD](https://docs.google.com/document/d/1hD0s6YnpWr17sqpxZZmgLnRwt7XdL6Q5Z9I5iCrgJB8/edit?tab=t.0#heading=h.axfiogfv4b3u)

chains/evm/contracts/ocr/SignatureQuorumVerifier.sol
```diff
@@ -2,150 +2,169 @@
...
+  function setSignatureConfigs(
+    SignatureConfigArgs[] calldata signatureConfigs
+  ) external onlyOwner {
+    for (uint256 i = 0; i < signatureConfigs.length; ++i) {
+      SignatureConfigArgs memory signatureConfig = signatureConfigs[i];
 
-    _clearAllSigners();
+      if (
+        signatureConfig.F == 0 || signatureConfig.F > signatureConfig.signers.length
+          || signatureConfig.configDigest == ""
```
kylesmartin, line 124: Just want to call out that we should be able to reliably generate the configDigest offchain so we can avoid multi-proposal operational workflows (i.e. we can set a config on the home chain and set the digest here in the same proposal without any issues)

kylesmartin (approved):

chains/evm/contracts/ocr/SignatureQuorumVerifier.sol
```diff
@@ -2,150 +2,164 @@
...
     // from chain A and so OCR reports will be valid on both forks.
     if (i_chainID != block.chainid) revert ForkedChain(i_chainID, block.chainid);
 
-    if (report.rs.length != s_ocrConfig.F + 1) revert WrongNumberOfSignatures();
-    if (report.rs.length != report.ss.length) revert SignaturesOutOfRegistration();
+    uint256 numberOfSignatures = proofs.rs.length;
 
-    _verifySignatures(keccak256(abi.encode(messageHash, blobHash, configDigest)), report.rs, report.ss);
-  }
+    if (numberOfSignatures != config.F + 1) revert WrongNumberOfSignatures();
```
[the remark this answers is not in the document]
makramkd: Possibility mentioned in the DD:
> If the CCVOfframp logic is such that >=threshold is permitted, then we can raise the threshold without downtime by doing it first on the homechain (so the aggregator just starts producing more signatures in CCVData), then on the remote chain so that they become required. Similarly lowering the threshold would be the reverse, first on the remote chains, then on the homechain.

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -65,22 +68,25 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
...
-  // TODO actually set this config
   /// @dev Per-chain source config (defining a lane from a Source Chain -> Dest OffRamp).
   struct SourceChainConfig {
     IRouter router; // ─╮ Local router to use for messages coming from this source chain.
     bool isEnabled; // ─╯ Flag whether the source chain is enabled or not.
-    address defaultCCV; // Default CCV to use for messages from this source chain.
-    address requiredCCV; // Required CCV to use for all messages from this source chain.
+    bytes onRamp; // OnRamp address on the source chain.
+    address[] defaultCCV; // Default CCV to use for messages from this source chain.
+    address[] laneMandatedCCVs; // Required CCV to use for all messages from this source chain.
```
makramkd, line 77: Outlining the relationship between these two in a comment would be useful when reading the code because I'm already mixing up between default and lane mandated (formerly required). My understanding is that `laneMandatedCCV` is a subset of `defaultCCV`, i.e. lane mandated must always be specified and therefore the default set must include them if a user doesn't specify any?

chains/evm/contracts/offRamp/CCVAggregator.sol
```diff
@@ -289,66 +316,159 @@ contract CCVAggregator is ITypeAndVersion, Ownable2StepMsgSender {
...
+    address[] calldata ccvs,
+    address[] memory tokenRequiredCCVs
+  ) internal view returns (address[] memory ccvsToQuery, uint256[] memory dataIndexes) {
     (address[] memory requiredCCV, address[] memory optionalCCVs, uint8 optionalThreshold) =
-      _getCCVsFromReceiverAndPool(sourceChainSelector, receiver, tokenPool);
+      _getCCVsFromReceiver(sourceChainSelector, receiver);
+    address[] memory laneMandatedCCVs = s_sourceChainConfigs[sourceChainSelector].laneMandatedCCVs;
+
+    ccvsToQuery = new address[](ccvs.length);
+    dataIndexes = new uint256[](ccvs.length);
```
makramkd, line 327: Haven't read this code thoroughly yet but is `ccvs.length` sufficient for these arrays? Cuz you have `ccvs` and `tokenRequiredCCVs`, couldn't they be entirely different and you'd end up having to append more to this array than `ccvs.length` items?

makramkd (commented):
still leafing through

chains/evm/contracts/offRamp/CCVAggregator.sol
RensR: ccvs is the passed in ccvs, so everything we have data for. You can never end up with more than that

PR #1156, opened 2025-08-27
fix ccip reader to not skip chain accessor

huangzhen1997:
Support TON2EVM e2e test, TON doesn't have contract reader and uses accessor instead


pkg/reader/ccip.go
```diff
@@ -680,8 +680,10 @@ func (r *ccipChainReader) DiscoverContracts(ctx context.Context,
 			continue
 		}
 		// Sanity check that we have a reader for this chain
-		if _, exists := r.contractReaders[chain]; !exists {
-			lggr.Errorw("Contract reader not found for a supported chain", "chain", chain)
+		_, crExists := r.contractReaders[chain]
+		_, caExists := r.accessors[chain]
```
ogtownsend, line 684: there's a `getChainAccessor()` helper function to check this that returns a custom error if the CA is not found

archseer (approved):

PR #1172, opened 2025-09-04
CCIP-7093 CCVProxy config validation

0xsuryansh:


chains/evm/contracts/libraries/CCVConfigValidation.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.24;
```
RensR, line 2: libs should always have the lowest possible version, so 0.8.4 in this case

RensR (dismissed):

kylesmartin (dismissed):

kylesmartin (approved):

PR #1199, opened 2025-09-16
Remove FeeQuoter calls from 1.7.0 contracts

kylesmartin:
- Return 0 instead of FeeQuoter fee for now. The FeeQuoter needs changes to be able to support 1.7.0 and billing is out of scope for M1.
- Do not processMessage against the FeeQuoter for similar reasons.
- Update tools to set CCIPMessageDest (i.e. CCVAggregator)


0xsuryansh (approved):

chains/evm/contracts/libraries/Client.sol
```diff
@@ -169,6 +169,12 @@ library Client {
     bytes tokenArgs;
   }
 
+  function _argsToBytes(
+    EVMExtraArgsV3 memory extraArgs
+  ) internal pure returns (bytes memory bts) {
+    return abi.encodeWithSelector(GENERIC_EXTRA_ARGS_V3_TAG, extraArgs);
+  }
+
   // TODO milestone 2
```
jhweintraub, line 178: nit: todo

chains/evm/contracts/onRamp/CommitOnRamp.sol
```diff
@@ -45,25 +43,22 @@ contract CommitOnRamp is Ownable2StepMsgSender, BaseOnRamp {
...
     bytes32, // messageId, unused
-    address feeToken,
-    uint256 feeTokenAmount,
-    bytes calldata verifierArgs
+    address, // feeToken, unused
+    uint256, // feeTokenAmount, unused
+    bytes calldata // verifierArgs, unused
   ) external returns (bytes memory verifierReturnData) {
     // For EVM, sender is expected to be 20 bytes.
     address senderAddress = address(bytes20(message.sender));
```
jhweintraub, line 58: do we need an explicit check for length here?

chains/evm/contracts/onRamp/CommitOnRamp.sol
```diff
@@ -45,25 +43,22 @@ contract CommitOnRamp is Ownable2StepMsgSender, BaseOnRamp {
...
     _assertSenderIsAllowed(message.destChainSelector, senderAddress);
 
-    // TODO
```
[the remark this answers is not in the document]
kylesmartin: Expected, out of scope for PR.

chains/evm/contracts/onRamp/CommitOnRamp.sol
kylesmartin: I think it is safe to leave as is, already handled by CCVProxy & was like this before.

chains/evm/contracts/onRamp/CCVProxy.sol
```diff
@@ -177,8 +177,7 @@ contract CCVProxy is IEVM2AnyOnRampClient, ITypeAndVersion, Ownable2StepMsgSende
 
     Client.EVMExtraArgsV3 memory resolvedExtraArgs = _parseExtraArgsWithDefaults(destChainConfig, message.extraArgs);
     // TODO where does the TokenReceiver go? Exec args feels strange but don't have a better place.
-    bytes memory tokenReceiver =
-      IFeeQuoterV2(s_dynamicConfig.feeQuoter).resolveTokenReceiver(resolvedExtraArgs.executorArgs);
+    bytes memory tokenReceiver = IFeeQuoterV2(s_dynamicConfig.feeQuoter).resolveTokenReceiver(message.extraArgs);
```
0xsuryansh, line 180: Thinking about this, `resolveTokenReceiver` should take in `resolvedExtraArgs.executorArgs` as we set `executorArgs` same as `evm2AnyMessage.extraArgs`

We should add a condition in `FeeQuoterV2.resolveTokenReceiver`

```
    if (extraArgs.length >= 4 && bytes4(extraArgs[0:4])
```

RensR: This entire function was a bit of a hack to get things to work. I agree that exec extra args is probably the right place (and yes, legacy args get copied to it). Then token receiver should go into the destBlob to be included in the msg id

0xsuryansh (approved):

chains/evm/deployment/v1_7_0/sequences/configure_chain_for_lanes_test.go
```diff
@@ -248,54 +251,37 @@ func TestConfigureChainForLanes(t *testing.T) {
...
-							"name": "args",
-							"type": "tuple"
-						}
-					],
-					"outputs": [],
-					"stateMutability": "pure"
-				}
-			]
-			`
+			_, tx, msgHasher, err := message_hasher.DeployMessageHasher(evmChain.DeployerKey, evmChain.Client)
```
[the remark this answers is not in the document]
0xsuryansh: Yeah, rhis should technically be like how ExtraDataCodec is used in core, can fix this later..

PR #1203, opened 2025-09-17
Ccip-7204 mcms input processing for changesets

AnieeG:
Assumption - The usual timelock need not have any qualifier in datastore. Newly added timelock for RMNRemote will have some qualifier to denote the timelock is only for RMNRemote.

Implementation Details - 

- Introduced MCMSConfig which is to specify user input for MCMS ops.

```
type MCMSConfig struct {
	// Description is a human-readable description of the proposal.
	Description string
	// OverridePreviousRoot indicates whether to override the root of the MCMS contract.
	OverridePreviousRoot bool
	// ValidUntil is a unix timestamp indicating when the proposal expires.
	// Root can't be set or executed after this time.
	ValidUntil uint32
	// TimelockDelay is the amount of time each operation in the proposal must wait before it can be executed.
	TimelockDelay mcms_types.Duration
	// TimelockAction is the action to perform on the timelock contract (schedule, bypass, or cancel).
	TimelockAction mcms_types.TimelockAction
}
```

- Every CS input config should implement interface `WithMCMS`

```
type WithMCMS interface {
	MCMSConfig() *MCMSConfig
	TimelockAddressQualifier() string
	MCMSAddressQualifier() string
}
```
MCMSConfig() should return user specified values for MCMS ops. `MCMSConfig` can be made a part of input config struct.
If there is no qualifier for Timelock and MCMS addresses, the method implementation for TimelockAddressQualifier and MCMSAddressQualifier should just return empty string, otherwise it will return the qualifier with which the timelock and mcms address set can be uniquely identified. For example, RMNRemote related timelock may return "RMNRemoteMCMS" and rest of the timelock contract might default to empty string.

-  Added a method on `ResolveMCMS` for `NewFromOnChainSequenceParams` similar to ResolveDep.
- Added `chains/evm/deployment/utils/changesets/mcms.go` which can be used in changesets to derive complete MCMS params.
- Sample implementation is in `ConfigureChainForLanes` changeset.


chains/evm/deployment/utils/changesets/mcms.go
```diff
@@ -0,0 +1,164 @@
...
+
+var (
+	BypasserManyChainMultisig  datastore.ContractType = "BypasserManyChainMultiSig"
+	CancellerManyChainMultisig datastore.ContractType = "CancellerManyChainMultiSig"
+	ProposerManyChainMultisig  datastore.ContractType = "ProposerManyChainMultiSig"
+	ManyChainMultisig          datastore.ContractType = "ManyChainMultiSig"
+	RBACTimelock               datastore.ContractType = "RBACTimelock"
+	CallProxy                  datastore.ContractType = "CallProxy"
+
+	// roles
```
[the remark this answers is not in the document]
kylesmartin: ManyChainMultiSig + labels is legacy & I wouldn't want to migrate it here if possible.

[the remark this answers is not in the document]
kylesmartin: Also, do Solana & TON use the same nomenclature for these? Just curious, as I would like to try and keep this package EVM-agnostic.

cc @tt-cll @archseer

chains/evm/deployment/utils/changesets/mcms.go
```diff
@@ -0,0 +1,164 @@
...
+	if cfg.MCMSConfig() == nil {
+		return MCMSParams{}, nil
+	}
+	// map of chain selector to timelock address
+	timelockAddresses := make(map[types.ChainSelector]string)
+	chainMetadata := make(map[types.ChainSelector]mcms_types.ChainMetadata)
+	for sel := range e.BlockChains.EVMChains() {
+		// find and format the MCMS related addresses
+		timelockAddrs, err := datastore_utils.FindAndFormatEachRef(e.DataStore, []datastore.AddressRef{
+			{
```
kylesmartin, line 53: We should just take the full ref from user input (i.e. Timelock Ref), same as the MCM contract. We then replace the selector in the ref with the target chain selector, as we should override the correct value there despite whether or not the user set anything for it.

Maybe we can add a common func for this in the datastore pkg: updateRefWithChainSelector(ref datastore.AddressRef, chainSelector uint64)

chains/evm/deployment/utils/changesets/mcms.go
```diff
@@ -0,0 +1,164 @@
...
+		MCMSConfig:        cfg.MCMSConfig(),
+		TimelockAddresses: timelockAddresses,
+		ChainMetadata:     chainMetadata,
+	}, nil
+}
+
+// findMCMS finds the MCMS address for the given chain selector, role, and contract type.
+// It first looks for addresses stored with type ManyChainMultisig , label PROPOSER/BYPASSER/CANCELLER and qualifier
+// If not found, it falls back to type ProposerManyChainMultiSig/BypasserManyChainMultiSig/CancellerManyChainMultiSig and qualifier.
+func findMCMS(e cldf_deployment.Environment, sel uint64, role MCMSRole, contractType datastore.ContractType, qualifier string) (common.Address, error) {
```
kylesmartin, line 107: Can be removed with suggested changes

chains/evm/deployment/utils/changesets/mcms.go
```diff
@@ -0,0 +1,164 @@
...
+			e.Logger.Errorf("failed to find timelock addresses for chain selector %d: %v", sel, err)
+			return MCMSParams{}, err
+		}
+		if len(timelockAddrs) == 0 {
+			e.Logger.Errorf("no timelock address found for chain selector %d with qualifier %s", sel, cfg.TimelockAddressQualifier())
+			return MCMSParams{}, fmt.Errorf("no timelock address found for chain selector %d with qualifier %s", sel, cfg.TimelockAddressQualifier())
+		}
+		timelockAddresses[types.ChainSelector(sel)] = timelockAddrs[0].String()
+		var mcmsAddr common.Address
+		switch cfg.MCMSConfig().TimelockAction {
```
kylesmartin, line 70: Switching logic can be removed with suggested changes

chains/evm/deployment/utils/operations/contract/function.go
```diff
@@ -11,4 +11,14 @@ type FunctionInput[ARGS any] struct {
 	ChainSelector uint64 `json:"chainSelector"`
 	// Args are the parameters passed to the contract call.
 	Args ARGS `json:"args"`
+	// TimelockAddress is the address of the timelock contract, if applicable for MCMS proposals
```
kylesmartin, line 14: Operations shouldn't be aware of this, MCM proposal construction should happen at the changeset level & only the changeset should be aware of these addresses.

chains/evm/deployment/utils/operations/contract/write.go
```diff
@@ -19,6 +19,10 @@ type WriteOutput struct {
 	ChainSelector uint64 `json:"chainSelector"`
 	// Tx is the prepared transaction (in MCMS format)
 	Tx mcms_types.Transaction `json:"tx"`
+	// TimelockAddress is the address of the timelock contract, if applicable for MCMS proposals
+	TimelockAddress common.Address `json:"timelockAddress,omitempty"`
```
kylesmartin, line 23: Same as above

chains/evm/deployment/utils/changesets/changesets.go
```diff
@@ -19,33 +18,40 @@ type NewFromOnChainSequenceParams[IN any, DEP any, CFG any] struct {
...
 	ResolveDep func(e deployment.Environment, cfg CFG) (DEP, error)
+	// ResolveMCMS resolves the MCMS configuration based on the environment and changeset config.
+	ResolveMCMS func(e deployment.Environment, cfg CFG) (MCMSParams, error)
 	// Describe returns a human-readable description of the changeset.
 	Describe func(in IN, dep DEP) string
 }
 
 // NewFromOnChainSequence creates a Changeset from an operations.Sequence that deploys contracts on-chain and performs write operations.
 // It wraps sequence execution with DataStore and MCMS integration.
 func NewFromOnChainSequence[IN any, DEP any, CFG any](params NewFromOnChainSequenceParams[IN, DEP, CFG]) deployment.ChangeSetV2[CFG] {
```
kylesmartin, line 29:
```suggestion
func NewFromOnChainSequence[IN any, DEP any, ChangesetCfg[CFG] any](params NewFromOnChainSequenceParams[IN, DEP, CFG]) deployment.ChangeSetV2[CFG] {
```
Instead of having sequences define `WithMCMS interface`, we just create a ChangesetCfg type that defines the standard MCMS input (currently represented as the MCMSCfg struct) alongside a generic changeset config.

chains/evm/deployment/utils/types.go
```diff
@@ -0,0 +1,9 @@
+package utils
```
kylesmartin, line 1: Would like to avoid "types" files as they get bloated in scope.

With the suggested changes around inputting the full refs for MCM and timelock contract we shouldn't need this.

chains/evm/deployment/v1_7_0/changesets/configure_chain_for_lanes.go
```diff
@@ -34,6 +35,21 @@ type ConfigureChainForLanesCfg struct {
 	RemoteChains map[uint64]RemoteChainConfig
 }
 
+func (c ConfigureChainForLanesCfg) MCMSConfig() *changesets.MCMSConfig {
```
kylesmartin, line 38: Rm interface implementation

chains/evm/deployment/utils/changesets/changesets.go
```diff
@@ -19,33 +18,40 @@ type NewFromOnChainSequenceParams[IN any, DEP any, CFG any] struct {
 	ResolveInput func(e deployment.Environment, cfg CFG) (IN, error)
 	// ResolveDeps resolves the dependencies for the sequence based on the environment and changeset config.
 	ResolveDep func(e deployment.Environment, cfg CFG) (DEP, error)
+	// ResolveMCMS resolves the MCMS configuration based on the environment and changeset config.
+	ResolveMCMS func(e deployment.Environment, cfg CFG) (MCMSParams, error)
```
kylesmartin, line 22: Not needed here, as much of the ResolveMCMS logic can just be baked into NewFromOnChainSequence.

What we should have here is `ResolveMCMSMetadata(e cldf_deployment.Environment, mcmsRef datastore.AddressRef) (mcms_types.ChainMetadata, error)`, which would get called for each MCM contract used in the proposal. The ref loaded from datastore gets passed in.

Then we would have a resolveMCMSMetadataForEVM that we can pass in

PR #1204, opened 2025-09-18
TokenPoolV2

0xsuryansh:


chains/evm/contracts/pools/TokenPoolV2.sol
```diff
@@ -0,0 +1,90 @@
...
+import {Pool} from "../libraries/Pool.sol";
+import {TokenPool} from "./TokenPool.sol";
+
+import {IERC165} from
+  "@chainlink/contracts/src/v0.8/vendor/openzeppelin-solidity/v5.0.2/contracts/utils/introspection/IERC165.sol";
+
+abstract contract TokenPoolV2 is IPoolV2, TokenPool {
+  event DynamicVerifiersSet(uint64 indexed remoteChainSelector, address[] outbound, address[] inbound);
+
+  struct DynamicVerifierConfig {
```
0xsuryansh, line 14: can be a part of `struct DestChainConfig` which has more fields like gas and bytes overhead for the dest chain

0xsuryansh: However customer token pool implementation might need a different storage layout based on some other params like FTF or other params which come in from token extraArgs 


So I wondering if we should define a storage layout for  CCVs 🤔

chains/evm/contracts/pools/TokenPoolV2.sol
```diff
@@ -0,0 +1,142 @@
...
+  // ================================================================
+  // │                          CCV                                  │
+  // ================================================================
+
+  function applyCCVConfigUpdates(
+    CCVConfigArg[] calldata verifierConfigArgs
+  ) external virtual onlyOwner {
+    for (uint256 i = 0; i < verifierConfigArgs.length; ++i) {
+      uint64 remoteChainSelector = verifierConfigArgs[i].remoteChainSelector;
+      if (!isSupportedChain(remoteChainSelector)) {
```
RensR, line 73: This expects a lane to be enabled before you can set CCVs, which hurts security. Can probably remove this check

chains/evm/contracts/pools/TokenPoolV2.sol
```diff
@@ -0,0 +1,142 @@
...
+
+  /// @notice Returns the set of required CCVs for incoming messages from a source chain.
+  /// @param sourceChainSelector The source chain selector for incoming messages.
+  /// @dev The second param is the amount and the third param is the tokenArgs.
+  /// This implementation assumes the same set of CCVs are used for all transfers on a lane.
+  /// Implementers can override this function to define custom logic based on these params.
+  /// @return requiredCCVs Set of required CCV addresses.
+  function getRequiredInboundCCVs(
+    uint64 sourceChainSelector,
+    uint256,
```
RensR, line 100: Missing `// amount` and `// tokenArgs`

chains/evm/contracts/pools/TokenPoolV2.sol
```diff
@@ -0,0 +1,142 @@
...
+  /// @param ccvs The array of CCV addresses to validate.
+  function _validateCCVArray(
+    address[] calldata ccvs
+  ) private pure {
+    if (ccvs.length == 0) {
+      revert CCVArrayCannotBeEmpty();
+    }
+
+    for (uint256 i = 0; i < ccvs.length; ++i) {
+      if (ccvs[i] == address(0)) {
```
RensR, line 130: I thought we decided to have address(0) mean the lane default. That would make it a valid input and it would need to be handled as such in the onramp/offramps

chains/evm/contracts/pools/TokenPoolV2.sol
```diff
@@ -0,0 +1,142 @@
...
+      || interfaceId == type(IERC165).interfaceId;
+  }
+
+  // ================================================================
+  // │                        Lock or Burn                          │
+  // ================================================================
+
+  function lockOrBurn(
+    Pool.LockOrBurnInV1 calldata lockOrBurnIn,
+    bytes calldata /* tokenArgs */
```
RensR, line 46: nit: prefer `// tokenArgs` as we don't use `/* */` syntax

0xsuryansh: Hmm, saw this at some other places in the codebase, will change

chains/evm/contracts/pools/TokenPoolV2.sol
```diff
@@ -0,0 +1,142 @@
...
+  ) external view virtual returns (address[] memory requiredCCVs) {
+    return s_verifierConfig[destChainSelector].outboundCCVs;
+  }
+
+  /// @notice Validates a CCV array for zero addresses and duplicates.
+  /// @param ccvs The array of CCV addresses to validate.
+  function _validateCCVArray(
+    address[] calldata ccvs
+  ) private pure {
+    if (ccvs.length == 0) {
```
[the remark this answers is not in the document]
0xsuryansh: yeah, this makes sense..

chains/evm/contracts/pools/TokenPoolV2.sol
0xsuryansh: Good point, we can configure first and then enable the lane.. I was thinking the other way

chains/evm/contracts/pools/TokenPoolV2.sol
kylesmartin: Okay, this makes sense as long as we make it clear.

chains/evm/contracts/poolsV2/TokenPool.sol
```diff
@@ -0,0 +1,140 @@
...
+  function applyCCVConfigUpdates(
+    CCVConfigArg[] calldata ccvConfigArgs
+  ) external virtual onlyOwner {
+    for (uint256 i = 0; i < ccvConfigArgs.length; ++i) {
+      uint64 remoteChainSelector = ccvConfigArgs[i].remoteChainSelector;
+      address[] calldata outboundCCVs = ccvConfigArgs[i].outboundCCVs;
+      address[] calldata inboundCCVs = ccvConfigArgs[i].inboundCCVs;
+
+      // Validate and check for duplicates in outbound CCVs.
+      _validateCCVArray(outboundCCVs);
```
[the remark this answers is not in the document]
0xsuryansh: That function not only checks for duplicates in an array but also contains additional logic for validating lane and default CCV. But I get your point.. we can revisit this later as well, I'll create a ticket

chains/evm/contracts/interfaces/IPoolV2.sol
```diff
@@ -1,50 +1,58 @@
...
-  function isSupportedToken(
-    address token
-  ) external view returns (bool);
+  /// @notice Returns the set of required CCVs for outgoing messages to a destination chain.
+  /// @param localToken The address of the local token.
+  /// @param destChainSelector The chain selector of the destination chain.
+  /// @param amount The amount of tokens to be transferred.
+  /// @param tokenArgs Additional token arguments.
+  /// @return requiredCCVs A set of addresses representing the required outbound CCVs.
+  function getRequiredOutboundCCVs(
```
RensR, line 30: Has this been validated with product? If not, let's not block on it but please get in writing somewhere that this is the full set of args

chains/evm/contracts/interfaces/IPoolV2.sol
```diff
@@ -1,50 +1,58 @@
...
+    bytes calldata sourcePoolData
   ) external view returns (address[] memory requiredCCVs);
+
+  /// @notice Returns a fee quote for transferring tokens to a destination chain.
+  /// @param destChainSelector The chain selector of the destination chain.
+  /// @param message The message to be sent to the destination chain.
+  /// @return feeTokenAmount The amount of fee token needed for the fee.
+  function getFee(
+    uint64 destChainSelector,
+    Client.EVM2AnyMessage calldata message
```
RensR, line 56: Probably don't want pools to decode finality themselves?

chains/evm/contracts/poolsV2/TokenPool.sol
```diff
@@ -0,0 +1,140 @@
...
+    uint64 destChainSelector,
+    uint256, // amount
+    bytes calldata // tokenArgs
+  ) external view virtual returns (address[] memory requiredCCVs) {
+    return s_verifierConfig[destChainSelector].outboundCCVs;
+  }
+
+  /// @notice Checks a CCV address array for duplicate entries.
+  /// @param ccvs The array of CCV addresses to validate.
+  function _validateCCVArray(
```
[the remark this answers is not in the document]
RensR: Or just deduplicateAddressArray

chains/evm/contracts/test/helpers/TokenPoolV2Helper.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.24;
+
+import {TokenPool} from "../../poolsV2/TokenPool.sol";
+
+import {IERC20} from "@openzeppelin/contracts@4.8.3/token/ERC20/IERC20.sol";
+
+/// @notice Helper contract for testing TokenPool V2 functionality.
+contract TokenPoolV2Helper is TokenPool {
```
RensR, line 9: This doesn't do anything as it doesn't expose any internal functions

chains/evm/contracts/poolsV2/TokenPool.sol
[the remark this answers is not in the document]
0xsuryansh: it doesn't dedups it reverts... so renaming to `_checkNoDuplicateAddresses`

PR #1230, opened 2025-10-02
[CCIP-7351] Resolve gas bug

kylesmartin:
https://smartcontract-it.atlassian.net/browse/CCIP-7351

Issue was the gasLimit passed to routeMessage: `gasleft() - i_gasForCallExactCheck - 10000`

Led to revert based on this condition: `if iszero(gt(sub(g, div(g, 64)), gasLimit))`, where `g` is based on the gas left

Problem: `(~gasleft) - (~gasleft)/64` (amount passed into call per EIP-150) is not greater than `~gasleft`

The gas limit passed into routeMessage must account for EIP-150: `gasleft() - 2 * (gasleft() / 64) - i_gasForCallExactCheck - 10000`

Added test to ensure that CCVAggregator usage of routeMessage isn't broken

### DIFF
--- a/.github/workflows/ccip-integration-test.yml
+++ b/.github/workflows/ccip-integration-test.yml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'develop'
 
 jobs:
   changes:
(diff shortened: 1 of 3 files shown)


makramkd:
> Problem: (~gasleft) - (~gasleft)/64 (amount passed into call per EIP-150) is not greater than ~gasleft

Do you mean its not greater than gasLimit? It'll definitely be less than gasleft() cuz you're subtracting from gasleft()

RensR (approved):

0xAustinWang (approved):

PR #1237, opened 2025-10-03
[CCIP-7090] Move chain family agnostic utilities to deployment module

kylesmartin:
https://smartcontract-it.atlassian.net/browse/CCIP-7090

Move common datastore, changeset, sequence and operation types to `chainlink-ccip/deployment` to align practices.

Also, add testing of `deployment` and `chains/evm/deployment` to CI. Want to ensure we aren't breaking things without realizing it.


deployment/utils/operations/contract/write.go
```diff
@@ -0,0 +1,66 @@
+package contract
```
kylesmartin, line 1: TODO: Could probably just move this one back to EVM, since sequence outputs are aligned with mcms_types.BatchOperation

PR #1246, opened 2025-10-06
Remove replace directive from `chains/evm/deployment` on develop

kylesmartin:
https://smartcontract-it.atlassian.net/browse/CCIP-7464

Also moves module definition to `chains/evm`, since gobindings and operations change together. Safe to make this change on develop because `chainlink-ccv` is the only consumer.


kylesmartin:
Will actually need to move the module definition to `chains/evm` to ensure that gobindings and ops can be changed together

PR #1251, opened 2025-10-07
Validate dest pool in FTF pool

0xsuryansh:


chains/evm/contracts/pools/FastTransferTokenPoolAbstract.sol
```diff
@@ -38,7 +38,8 @@ abstract contract FastTransferTokenPoolAbstract is TokenPool, CCIPReceiver, ITyp
   error TransferAmountExceedsMaxFillAmount(uint64 remoteChainSelector, uint256 amount);
   error InsufficientPoolFees(uint256 requested, uint256 available);
   error QuoteFeeExceedsUserMaxLimit(uint256 quoteFee, uint256 maxFastTransferFee);
-  error InvalidReceiver(bytes receiver);
+  error InvalidEncodedAddress(bytes encodedAddress);
+  error InvalidDestinationPool(uint64 destChainSelector, bytes destinationPool);
```
matYang, line 42: I think you are missing a test case for this error

chains/evm/contracts/pools/FastTransferTokenPoolAbstract.sol
```diff
@@ -263,6 +264,12 @@ abstract contract FastTransferTokenPoolAbstract is TokenPool, CCIPReceiver, ITyp
       );
     }
 
+    // Only check for empty destination pool here, as proper validation is done during config update.
+    // An empty value means the config was never set.
+    if (destChainConfig.destinationPool.length == 0) {
+      revert InvalidDestinationPool(destinationChainSelector, destChainConfig.destinationPool);
```
matYang, line 270: while this is valid in most other places, this error `InvalidDestinationPool` accepts bytes but will only be thrown with empty bytes, maybe just call it `DestinationPoolCannotBeEmpty` and lose the bytes field?

RensR (approved):

PR #1594, opened 2026-01-23
Check for valid lane combinations before calling LatestMsgSeqNum

ogtownsend:
Some misc log cleanup:
- Only query LatestMsgSeqNum for actually valid source<>dest pairs
- Adds more buckets to CCIPReader latency histogram
- Correct log label in default accessor
- Don't log offramp error log if it's a no bindings error


ogtownsend:
replaced with: https://github.com/smartcontractkit/chainlink-ccip/pull/2134

PR #1969, opened 2026-04-09
Remove versioned folder ccv

carte7000:


ma33r (dismissed):

tt-cll (dismissed):

tt-cll:
I think there will be an issue because we're removing a required check. I think we'll have to open a ticket with security to remove the check before this can be approved and merged 😢 
```
test-ccv-evm-deployments Expected — Waiting for status to be reported
```

RensR (approved):

tt-cll (approved):

PR #1971, opened 2026-04-10
solana: Remove smalloc requirement by boxing large accounts

archseer:


PabloMansanet (dismissed):
Code LGTM, and we definitely should not be relying on a custom allocator, but I'm somewhat surprised that we're not hitting heap issues after boxing so many accounts, when we certainly had hit them before. Maybe it's due to anchor improvements upstream?

Either way, welcome change.

agusaldasoro (dismissed):

agusaldasoro (approved):

PabloMansanet (approved):

PR #1973, opened 2026-04-11
feat: simplify token pool rate limit handling

chris-de-leon-cll:
### Summary

Simplifies rate limiter configurations from four fields to two. This is possible because for v2.0.0 token pools, exactly one rate limiter bucket is only ever used at a time - the pool uses either the default or custom finality bucket based on its `allowedFinalityConfig`, but never both. Therefore, it's possible to improve the user experience by having the tooling read the finality config from the pool and automatically route rate limits to the correct bucket.

### Motivation

Previously, users had to configure four rate limiter fields:
- `DefaultFinalityInboundRateLimiterConfig`
- `DefaultFinalityOutboundRateLimiterConfig`
- `CustomFinalityInboundRateLimiterConfig`
- `CustomFinalityOutboundRateLimiterConfig`

This is not only unnecessarily complex, but leads to a higher risk of misconfiguring things on chain. Users are essentially forced to provide values for both the active and inactive rate limit buckets when realistically they'll almost always want to modify the currently active one. In fact, leaving out the inactive bucket from the input will result in it being needlessly reset on-chain since Go would provide 0-values for `rate`, `capacity`, and `isEnabled`. Thus, in order to use the tooling properly and avoid accidental overwrites, the user would need to do extra work to determine (1) which bucket is currently active and (2) what the existing values are on chain (both of which make this op more difficult to run).

### Change

Users now configure just two fields:
- `InboundRateLimiterConfig`
- `OutboundRateLimiterConfig`

For the 2.0 tooling, the code now reads `allowedFinalityConfig` from the pool and automatically chooses the appropriate rate limit bucket to update:
- `allowedFinalityConfig == 0` → default finality bucket
- `allowedFinalityConfig != 0` → custom finality bucket

When `AllowedFinalityConfig` is specified in the input, it is set on the pool **before** determining which bucket to update, which ensures the rate limits are applied to the intended bucket. This simplification covers all practical use cases. The only capability "lost" is setting values on an inactive bucket without activating it, but this was an anti-pattern anyway. This change also reduces the mental gymnastics needed to understand rate limits between v1.5.x, v1.6.x, and v2.0.0 - folks familiar with v1.5.x / v1.6.x token pool tooling can now pick up v2.0.0 more easily.

### References

`_validateLockOrBurn`:

https://github.com/smartcontractkit/chainlink-ccip/blob/2499ef05fcdebe2e25172d3a3bd20990d7ba11a7/chains/evm/contracts/pools/TokenPool.sol#L428-L436

`_validateReleaseOrMint`:

https://github.com/smartcontractkit/chainlink-ccip/blob/2499ef05fcdebe2e25172d3a3bd20990d7ba11a7/chains/evm/contracts/pools/TokenPool.sol#L486-L490


chains/evm/deployment/v2_0_0/adapters/tokens_test.go
chris-de-leon-cll, line 228: The original code in this test is a prime example of why this change is needed - notice that it configures both the default and custom rate limits, then later the test completely ignores the custom rate limit and only checks the default one

So if the intention was to only activate the default one, then it doesn't really make sense to force the user to set both the custom and default rate limits up front

It may be tempting to just omit the rate limit pair you don't want, but this actually leads to Go using 0-values for the omitted rate limit struct causing the bucket to be needlessly reset on-chain (which is not only wasteful but could even lead to misconfiguration issues later as well).

To avoid this, you'd need to look up the existing rate limit values on chain and pass them to the operation, which is not only error-prone but just adds additional complexity in running this op correctly. The new code in this PR prevents the user from running into these hazards entirely.

agusaldasoro (approved):

tt-cll (approved):

PR #1976, opened 2026-04-14
add cursing test

makramkd:
Add a test that curses a source chain on the destination and asserts that the commit plugin doesn't commit those messages.


deployment/testadapters/adapters.go
```diff
@@ -152,6 +152,21 @@ type TestAdapter interface {
...
+// TestAdapter implementations that don't implement this won't be able to be used
+// in cursing related tests.
+type Curser interface {
+	// IsCursed returns true if the given subject is cursed on the chain.
+	IsCursed(ctx context.Context, subject [16]byte) (bool, error)
+	// Curse curses the given subject on the chain.
+	Curse(ctx context.Context, subject [16]byte) error
+	// Uncurse uncurses the given subject on the chain.
+	Uncurse(ctx context.Context, subject [16]byte) error
+}
```
makramkd: Optional interface that is checked for in the cursing test.

devenv/chainimpl/ccip-evm/impl.go
```diff
@@ -73,7 +74,13 @@ func (m *CCIP16EVM) SetCLDF(e *deployment.Environment) {
...
 		panic(fmt.Sprintf("failed to find testadapter factory for %s", chainsel.FamilyEVM))
 	}
-	m.TestAdapter = factory(e, m.chainDetails.ChainSelector)
+	adapter := factory(e, m.chainDetails.ChainSelector)
+	m.TestAdapter = adapter
+	curser, ok := adapter.(testadapters.Curser)
+	if !ok {
+		panic(fmt.Sprintf("expected concrete testadapter type for %s to be a curser, got %T", chainsel.FamilyEVM, adapter))
+	}
+	m.Curser = curser
```
makramkd: Adopted this pattern here that can probably be replicated for other families.

devenv/tests/loki.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return fmt.Errorf("failed to connect to Loki WebSocket (status %d): %w", resp.StatusCode, err)
+}
+
+// WaitForLokiLogs waits for logs with the given query to appear in Loki
+// and returns the latest log line.
+//
+// logQuery is the same log query that is used in the Grafana UI, e.g.
+// `{container="don-node1"} | json | logger="SomeLoggerName" | msg="SomeMessage"`.
+func WaitForLokiLogs(ctx context.Context, lokiWSURL, logQuery string) (string, error) {
```
makramkd, line 66: This is basically a stripped down Loki log streamer compared to the logasserter package in chainlink-ccv. Eventually I think we should have a common library but this'll do for now.

devenv/go.mod
```diff
@@ -58,6 +58,8 @@ require (
 	golang.org/x/sync v0.20.0
 )
 
+require github.com/gorilla/websocket v1.5.3
```
makramkd, line 61: Used for loki log streaming

winder (approved):

devenv/tests/loki.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// DefaultLokiURL is the default Loki WebSocket URL for log streaming in tests.
+// TODO: ideally we'd get this from a config or an env var.
+const DefaultLokiURL = "ws://localhost:3030"
+
+// lokiPingQuery is a minimal valid LogQL selector for /tail. Loki rejects an empty `{}`
+// selector (HTTP 400 before the WebSocket upgrade, seen as "bad handshake").
+// Devenv pushes CCIP streams with a `job` label (see chainimpl monitoring).
+const lokiPingQuery = `{job=~".+"}`
```
winder, line 24: AFAIK we're moving away from loki, so this test has an expiration date

devenv/tests/loki.go
winder: I like this. In the past the options for assertions were really limited.

devenv/tests/curse.go
```diff
@@ -0,0 +1,122 @@
...
+	// the commit plugin that the source chain is being ignored due to being cursed on the destination.
+	// In this particular case, since we only have 2 chains and one is cursed, we should see the log
+	// "nothing to observe from the offRamp, no active source chains exist".
+	family, err := chainsel.GetSelectorFamily(toImpl.ChainSelector())
+	require.NoError(t, err)
+
+	chainID, err := chainsel.GetChainIDFromSelector(toImpl.ChainSelector())
+	require.NoError(t, err)
+
+	// Get offramp from the datastore
```
tt-cll: theoretically, this should be behind the test adapter so each chain family can serve their own version and we can eliminate the evm dep

devenv/tests/curse.go
```diff
@@ -0,0 +1,122 @@
...
+
+	curser, ok := toImpl.(testadapters.Curser)
+	if !ok {
+		t.Skipf("destination chain %d does not implement Curser, chain family: %s", toImpl.ChainSelector(), toImpl.Family())
+	}
+
+	// Ping the loki URL to ensure its available, else skip the test.
+	err := PingLoki(t.Context(), DefaultLokiURL)
+	if err != nil {
+		t.Skipf("failed to ping Loki WebSocket URL, is loki running?: %v", err)
```
winder, line 51: How robust are these checks? Could we get into a situation where they're accidentally skipped and nobody knows about it?

devenv/tests/curse.go
```diff
@@ -0,0 +1,122 @@
...
+	offRampRef, err := e.DataStore.Addresses().Get(datastore.NewAddressRefKey(
+		toImpl.ChainSelector(),
+		datastore.ContractType(offramp.ContractType), // TODO: evm only?
+		offramp.Version,
+		"",
+	))
+	require.NoError(t, err)
+
+	var (
+		expectedLogger = fmt.Sprintf("CCIPCommitPlugin.%s.%s.%d.%s", family, chainID, toImpl.ChainSelector(), offRampRef.Address)
```
winder: Nit: the structured logging let's you be much more precise than this.

deployment/testadapters/adapters.go
```diff
@@ -152,6 +152,21 @@ type TestAdapter interface {
 	TestAdapterForFamily
 }
 
+// Curser is an interface that allows tests to curse subjects on a chain.
+// It is used to test the curse and uncurse operations on a chain.
+//
+// This is optionally implemented by TestAdapter implementations.
+// TestAdapter implementations that don't implement this won't be able to be used
+// in cursing related tests.
+type Curser interface {
```
winder: I'd consider different parameters to the functions here to make it easier for someone to use if they aren't familiar with RMN curse subjects.

winder (dismissed):
Amazing.

deployment/testadapters/adapters.go
tt-cll: I am wondering if it makes more sense for the tests to call the tooling API curse adapter directly. This is what is called in live environments and any chain would need that to be implemented to actually be able to curse. I don't want to hold this PR up though
https://github.com/smartcontractkit/chainlink-ccip/blob/03ca8d0f23803609b89fe9c655b529dede1cd04f/deployment/fastcurse/product.go#L27

makramkd: Oh interesting, let me have a look at it!

devenv/tests/curse.go
makramkd: Yeah good note, I was thinking the same but now your other comment got me thinking - would this be more appropriate on some deployment chain-agnostic interface instead?

devenv/tests/loki.go
makramkd: Really? Even for development?

devenv/tests/curse.go
makramkd: The ping part specifically? It seems to "work" based on the local testing I did (the ping fails if the loki API server is not up) but it may be different for real Loki deployments

devenv/tests/curse.go
makramkd: Which extra fields were you thinking of adding?

deployment/testadapters/adapters.go
makramkd: True, we could do `chainSelector` instead of `subject`...

0xAustinWang (dismissed):

devenv/tests/curse.go
winder: I don't remember them all, but we at least have stuff like `plugin="Commit"`, you can also filter to specific componts like `component="merkleroot"`

devenv/tests/curse.go
tt-cll: That is actually a great point. We do have one of these on the lane adapter today but that might not be the right place for it. https://github.com/smartcontractkit/chainlink-ccip/blob/03ca8d0f23803609b89fe9c655b529dede1cd04f/deployment/lanes/product.go#L29

devenv/tests/curse.go
```diff
@@ -0,0 +1,137 @@
...
+	require.NotEqual(t, fromImpl, toImpl)
+
+	destFamily, err := chainsel.GetSelectorFamily(toImpl.ChainSelector())
+	require.NoError(t, err)
+
+	curseAdapter, ok := fastcurse.GetCurseRegistry().GetCurseAdapter(destFamily, semver.MustParse("1.6.0"))
+	if !ok {
+		t.Skipf("no curse adapter registered for chain family: %s and version: %s", destFamily, semver.MustParse("1.6.0"))
+	}
+	require.NotNil(t, curseAdapter, "registered curse adapter is nil")
```
makramkd, line 44: No new interface needed, using the deployment API directly here.

devenv/tests/curse.go
```diff
@@ -0,0 +1,137 @@
...
+				{
+					ChainSelector:        toImpl.ChainSelector(),
+					SubjectChainSelector: fromImpl.ChainSelector(),
+					Version:              semver.MustParse("1.6.0"),
+					IsGlobalCurse:        false,
+				},
+			},
+		})
+		require.NoError(t, err)
+		require.Greater(t, len(output.Reports), 0)
```
makramkd, line 79: Applying the changeset to curse.

devenv/tests/curse.go
makramkd: I refactored the test to not need the LaneAdapter, so all good here :-)

devenv/tests/curse.go
makramkd: Made the log check much more precise.

deployment/testadapters/adapters.go
makramkd: Interface no longer needed, using the deployment API directly :-)

deployment/testadapters/adapters.go
makramkd: @tt-cll thanks for the suggestion, switched to using the CurseAdapter instead! Don't need to duplicate stuff :-)

tt-cll (approved):

winder (approved):

PR #1977, opened 2026-04-14
Simplify fq config, address length and selector family

carte7000:


deployment/v2_0_0/changesets/configure_chains_for_lanes_from_topology.go
tt-cll, line 70: I think the next simplification would be to have a default set of CCVs that the user can override. So if these are unset, we auto-populate

tt-cll (approved):

ma33r (approved):

PR #1999, opened 2026-04-23
fix FQ 1.6 op

RensR:
Fixed the manually edited FQ 1.6 operation and adds additional CI steps that will be set as required to avoid this in the future


.github/workflows/solidity-build-and-publish.yml
```diff
@@ -93,6 +93,34 @@ jobs:
           git add --all
           git diff --minimal --color --cached --exit-code
 
+  native-compile-check:
```
RensR, line 96: These will always run and be required CI. They allow skips or passes, but not failures

KodeyThomas (dismissed):

tt-cll (dismissed):

KodeyThomas (approved):

0xsuryansh (approved):

PR #2007, opened 2026-04-24
Ccip-11140 ownership check hook

AnieeG:
There is no way to access the environment's blockchains as part of prehook, therefore we just use all evm chains from datastore


chains/evm/deployment/hooks/ownership_verification.go
```diff
@@ -0,0 +1,183 @@
...
+	datastore_utils "github.com/smartcontractkit/chainlink-ccip/deployment/utils/datastore"
+)
+
+var _ cciphooks.ContractOwnership = (*EVMContractOwnership)(nil)
+
+func init() {
+	cciphooks.GetContractOwnershipRegistry().Register(chain_selectors.FamilyEVM, &EVMContractOwnership{})
+}
+
+var laneMigratorContractTypesForOwnershipCheck = map[datastore.ContractType]struct{}{
```
AnieeG: @carte7000 do you think these should be sufficient?

chains/evm/deployment/hooks/ownership_verification.go
```diff
@@ -0,0 +1,183 @@
...
+	datastore.ContractType(committee_verifier.ContractType):   {},
```
carte7000: Maybe we can add the resolvers as well. The executor also has a proxy we can add

carte7000 (changes requested):

ma33r (approved):

carte7000 (approved):

tt-cll (approved):

matYang (approved):

PR #2010, opened 2026-04-27
fix(devenv): ton plugin config

jadepark-dev:
This PR adds overrides plugin configs for TON in devenv


devenv/common/implcommon.go
```diff
@@ -444,10 +445,34 @@ func AddNodesToContracts(
...
+					ocrParams.ExecuteOffChainConfig.MaxCommitReportsToFetch = 250
+					ocrParams.ExecuteOffChainConfig.InflightCacheExpiry = *config.MustNewDuration(1 * time.Minute)
+					ocrParams.ExecuteOffChainConfig.MessageVisibilityInterval = *config.MustNewDuration(8 * time.Hour)
+					ocrParams.ExecuteOffChainConfig.RootSnoozeTime = *config.MustNewDuration(5 * time.Minute)
+					ocrParams.ExecuteOffChainConfig.MultipleReportsEnabled = true
+				}
+			}
 			return ocrParams
 		}
 		commitOCRConfigs[chain] = DeriveOCRParamsForCommit(SimulationTest, ccipHomeSelector, nil, ocrOverride)
```
huangzhen1997, line 457: Instead of hardcode the default value here in the override function above, maybe we can follow the existing pattern ? like instead of using `SimulationTest` we use the actual family you got here, and define the default values config.go.

jadepark-dev: [`2fb7bb2` (this PR)](https://github.com/smartcontractkit/chainlink-ccip/pull/2010/commits/2fb7bb2253c99e74b4c25cd18aca7b8d4f0ac708)

PR #2014, opened 2026-04-28
Auto detect onramp version when setting fee configs

matYang:


chains/evm/deployment/v1_0_0/adapters/fee_contract_resolver.go
```diff
@@ -0,0 +1,127 @@
...
+				return datastore.AddressRef{}, fmt.Errorf("failed to bind v2.0 OnRamp at %s on src %d: %w", onRampAddr.Hex(), src, err)
+			}
+			cfg, err := c.GetDynamicConfig(&bind.CallOpts{Context: e.GetContext()})
+			if err != nil {
+				return datastore.AddressRef{}, fmt.Errorf("failed to call v2.0 OnRamp.getDynamicConfig at %s on src %d: %w", onRampAddr.Hex(), src, err)
+			}
+			fqAddr = cfg.FeeQuoter
+		default:
+			return datastore.AddressRef{}, fmt.Errorf("unsupported OnRamp major version %d (%s) at %s on src %d", onRampRef.Version.Major(), onRampRef.Version.String(), onRampAddr.Hex(), src)
+		}
```
chris-de-leon-cll: We should ideally move this logic to the adapter itself for that specific version

Ideally, this shared helper would only import the router, then delegate the call to `getDynamicConfig` to a helper function that is supplied via EVMFeeContractResolver

There's an existing pattern here that you can reference: https://github.com/smartcontractkit/chainlink-ccip/blob/fac5ac247a021f8b8f66dd5499eefdc76968c0bb/chains/evm/deployment/v1_0_0/adapters/pool_adapter.go#L33-L55

deployment/fees/set_token_transfer_fee.go
```diff
@@ -92,9 +97,9 @@ func makeApply(feeRegistry *FeeAdapterRegistry, mcmsRegistry *changesets.MCMSRea
 				return cldf.ChangesetOutput{}, fmt.Errorf("failed to get chain family for selector %d: %w", src.Selector, err)
 			}
 
-			adapter, exists := feeRegistry.GetFeeAdapter(srcFamily, cfg.Version)
+			resolver, exists := resolverRegistry.GetFeeContractResolver(srcFamily)
```
chris-de-leon-cll: We should also be mindful of Solana as well (this new call forces the operation to only work for EVM) - for Solana the contracts/architecture is a bit different but the idea of inferring the relevant fee contract should still apply - I believe this info is stored in the router as well

chris-de-leon-cll:
This PR is a nice starting point - I continued to iterate on it using codex 5.5 and managed to get it to a state that's ready for merge here:

https://github.com/smartcontractkit/chainlink-ccip/pull/2028

I also jotted down some AI learnings/notes while developing, so I'll be able to turn this into a skill for future coding

PR #2021, opened 2026-04-30
Register USDCTokenPoolProxy on legacy v1.6.2 pool for all remotes

0xsuryansh:
+ Fixes the following 

- ERC20 Lockbox address were not persisted to datastore
- Skip Solana related config for CCTP v2 flows
- We were not checking if SiloedUSDCTokenPool is already deployed

### DIFF
--- a/chains/evm/deployment/v2_0_0/sequences/cctp/configure_cctp_chain_for_lanes.go
+++ b/chains/evm/deployment/v2_0_0/sequences/cctp/configure_cctp_chain_for_lanes.go
@@ -1,7 +1,9 @@
 package cctp
 
 import (
+	"bytes"
 	"fmt"
+	"slices"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


chains/evm/deployment/v2_0_0/sequences/cctp/configure_cctp_chain_for_lanes.go
```diff
@@ -349,6 +362,100 @@ func resolveConfigureCCTPChainRefs(
...
+	}
+	supportedChains := supportedChainsReport.Output
+
+	var writes []contract_utils.WriteOutput
+	for remoteSel := range input.RemoteChains {
+		poolRef := input.RemoteRegisteredPoolRefs[remoteSel]
+		if poolRef.Type != datastore.ContractType(usdc_token_pool_proxy.ContractType) {
+			continue
+		}
+		cfg, ok := remoteChainConfigs[remoteSel]
```
crispymangoes, line 412: Shouldn't this always result in `ok == true`?
Since remoteChainConfigs is built from input.RemoteChains? Might be worth it to actually throw an error here instead of just silently continuing if there isn't a scenario where we expect ok to be false.

0xsuryansh: https://github.com/smartcontractkit/chainlink-ccip/pull/2021/commits/c18f6c75fc6307a1a3ab92b3acc2730610bad5f9

carte7000 (approved):

crispymangoes (approved):

PR #2038, opened 2026-05-06
feat: add chain-agnostic AuthorizedCallers adapter with EVM implementation

asoliman92:
Introduce a registry-backed AuthorizedCallers system mirroring the existing fastcurse pattern:

- deployment/authorizedcallers: chain-agnostic interface, registry singleton, and ConfigureAuthorizedCallersChangeset
- chains/evm/deployment/v2_1_0/adapters: EVMAuthorizedCallersAdapter backed by injected per-contract generated ops so MCMS metadata (ContractType, ABI) is accurate per target contract
- init.go registers (EVM, RMN, v2.1.0); additional contracts follow the same NewEVMAuthorizedCallersAdapter pattern


chains/evm/deployment/v2_1_0/changesets/configure_rmn_curse_admins_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func TestConfigureRMNCurseAdmins_Apply(t *testing.T) {
```
carte7000: Do we still need this changeset now it would be in the deployment package?

carte7000 (approved):

PR #2051, opened 2026-05-09
fix: Solana batch ops logic in ConfigureTokensForTransfers

chris-de-leon-cll:
## Summary

This PR fixes a bug in the Solana **`ConfigureTokenForTransfers`** sequence related to the `RegisterTokenAdminRegistry` and `AcceptTokenAdminRegistry` steps. We illustrate the bug using a **real-world** example. Suppose:

- For Solana, we want to register an existing mint under the CLL LnR self-service pool
- For EVM, we want to deploy a fresh token and pool
- At the end, we want to connect these two pools

Assume the state of the token admin registry (TAR) on Solana looks like this:

| Field                    | State                 |
| ------------------------ | ----------------------------------------------------------------------- |
| **Mint**                 | SPL mint pubkey                               |
| **Administrator**        | `11111111111111111111111111111111`                                      |
| **PendingAdministrator** | **Third-party** pubkey (**≠** mcms timelock signer PDA, **≠** deployer) |

With the current code on `main`, running token expansion for this scenario leads to the following error:

```text
Error: failed to process token configs for transfers: failed to configure token pool on chain
with selector 124615329519749607: failed to register token metadata: pending admin
GcqdKBdgcJNdBeC1TnZvJTaWuRXXg8WotC5qw1BNBSEp does not match timelock signer
GX38bfJaViHyypF7BTykt2tRVZfx5vXBr21Y6JMnyyq4 or deployer FqrKUJWwei7r8BTwXEJRprf16WvgzmUn7QarDqf8keyB
```

In this case, `RegisterTokenAdminRegistry` needs a proposal and adds an `NewCcipAdminOverridePendingAdministratorInstruction` to the batch [here](https://github.com/smartcontractkit/chainlink-ccip/blob/3e4cbbada64754dad71525d4d1b976c530fca2c1/chains/solana/deployment/v1_6_0/operations/router/router.go#L450-L469).

Then, `AcceptTokenAdminRegistry` runs next and queries RPC before the proposal was applied on-chain, so it sees the stale pending admin and rejects it [here](https://github.com/smartcontractkit/chainlink-ccip/blob/3e4cbbada64754dad71525d4d1b976c530fca2c1/chains/solana/deployment/v1_6_0/operations/router/router.go#L598-L601).

To fix this, we need to conditionally handle the output of `RegisterTokenAdminRegistry`:

1. If `RegisterTokenAdminRegistry` requires a proposal and we're overriding the pending admin with timelock, then we should avoid inspecting the current pending admin in the TAR and schedule an Accept instruction in the same batch.
2. If `RegisterTokenAdminRegistry` does not require a proposal, then we can keep the existing flow where `AcceptTokenAdminRegistry` runs as a separate operation and inspects the live onchain state of TAR.
3. In any other case, the acceptance of the token admin role is deferred to the external user and not performed inside the sequence

## What changed

1. **`TokenAdminRegistryOut`**: add **`ProposalInstructions []solana.Instruction`** (paired with **`BatchOps`** whenever register/transfer wrapped a proposal **`ix`**).
2. **`BuildAcceptTokenAdminRegistrySolanaInstruction`**: builds **`AcceptAdminRole`** for a given **`pendingAdministrator`** pubkey; **`Accept`** reuses this after its RPC/policy checks.
4. **`ConfigureTokenForTransfers`** (Solana adapter):
   - If **`ProposalInstructions`** and **`BatchOps`** present → **`append`(proposal ix(s), built accept ix)** → **`BuildMCMSBatchOperation`** once (**no** interim **`Accept`** that rereads stale TAR for that fork).
   - Else → unchanged path: **`Accept`** as a separate **`ExecuteOperation`**, **`BatchOps`** from register then accept appended.

**Integration (“Scenario 6”):** two **`TokenExpansion`** passes with explicit third-party **`PendingAdministrator`**, then cross-family pools; asserts success and **final** TAR: **`Mint`** unchanged, **`Administrator`** **=** timelock signer PDA, **`PendingAdministrator`** **=** zero, customer key **≠** **`Administrator`**.


chains/solana/deployment/v1_6_0/sequences/tokens.go
```diff
@@ -47,39 +48,82 @@ func (a *SolanaAdapter) ConfigureTokenForTransfersSequence() *cldf_ops.Sequence[
...
-				ExistingAddresses: input.ExistingDataStore.Addresses().Filter(),
-			})
-			if err != nil {
-				return sequences.OnChainOutput{}, fmt.Errorf("failed to register token metadata: %w", err)
+			timelockSigner := utils.GetTimelockSignerPDA(addrs, chain.Selector, common_utils.CLLQualifier)
+			tokenMintPK := solana.MustPublicKeyFromBase58(tokenAddr.Address)
+			deployerPK := chain.DeployerKey.PublicKey()
+			routerPK := solana.PublicKeyFromBytes(routerAddr)
+
+			// If the proposed token admin is timelock or the deployer, then we can run Register + Accept
```
agusaldasoro, line 76: Can it happen that maybe you need to config the token pool first (for example by running `set_pool`) and then propose a new administrator with everything already configured?

tt-cll (dismissed):

agusaldasoro (dismissed):

chains/solana/deployment/v1_6_0/sequences/tokens.go
tt-cll: SetPool should be called after this during the sequence. I guess the only scenario where we couldn't call SetPool after admin registration is if we were registering to an address CLL didn't control? So that would be the manual registration flow and the customer can set it afterwards?

tt-cll (approved):

sowgandhi11 (approved):

PR #2064, opened 2026-05-12
Chain agnostic token agnostic admin removal changeset

matYang:


chains/evm/deployment/tokens/tokenimpl/impl.go
```diff
@@ -76,3 +77,8 @@ type Token interface {
 	// outer token deployment sequence.
 	Deploy(b cldf_ops.Bundle, chain evm.Chain, in tokensapi.DeployTokenInput) (datastore.AddressRef, []contract.WriteOutput, error)
 }
+
+// Separate interface for tokens that support admin role checks.
+type AdminRoleToken interface {
+	HasAdminRole(ctx context.Context, chain evm.Chain, token, user common.Address) (bool, error)
+}
```
chris-de-leon-cll: Ah I'm assuming the agent put this in a separate interface to avoid forcing this on all existing token implementations? In this case, I think it's okay to put this in the Token interface above - we're already modifying a good portion of the tokens anyways and for the ones that don't support RBAC we can return an error.

deployment/tokens/revoke_admin_role.go
```diff
@@ -0,0 +1,174 @@
...
+			return errors.New("at least one token admin role revocation is required")
+		}
+
+		for i, revocation := range cfg.Revocations {
+			if revocation.ChainSelector == 0 {
+				return fmt.Errorf("revocation[%d]: chain selector is required", i)
+			}
+			if !e.BlockChains.Exists(revocation.ChainSelector) {
+				return fmt.Errorf("revocation[%d]: chain selector %d not found in environment", i, revocation.ChainSelector)
+			}
```
chris-de-leon-cll: [style guide violation]: checking `ChainSelector == 0` is most likely redundant. We can drop that and leave the other check:

```suggestion
			if !e.BlockChains.Exists(revocation.ChainSelector) {
				return fmt.Errorf("revocation[%d]: chain selector %d not found in environment", i, revocation.ChainSelector)
			}
```

deployment/tokens/revoke_admin_role.go
```diff
@@ -0,0 +1,174 @@
...
+			if datastore_utils.IsAddressRefEmpty(revocation.TokenRef) {
+				return fmt.Errorf("revocation[%d]: token ref is required", i)
+			}
+			if revocation.TokenRef.ChainSelector != 0 && revocation.TokenRef.ChainSelector != revocation.ChainSelector {
+				return fmt.Errorf("revocation[%d]: token ref chain selector mismatch: expected %d, got %d", i, revocation.ChainSelector, revocation.TokenRef.ChainSelector)
+			}
```
chris-de-leon-cll, line 93: [style guide violation]: this is using an unresolved token ref. Instead, prefer looking up the token in the datastore using `datastore_utils.FindAndFormatRef()` or `resolveTokenAdminRoleRef` as it already encapsulates these checks

deployment/tokens/revoke_admin_role.go
```diff
@@ -0,0 +1,174 @@
...
+func revokeTokenAdminRoleApply(tokenRegistry *TokenAdapterRegistry, mcmsRegistry *changesets.MCMSReaderRegistry) func(cldf.Environment, RevokeTokenAdminRoleInput) (cldf.ChangesetOutput, error) {
+	return func(e cldf.Environment, cfg RevokeTokenAdminRoleInput) (cldf.ChangesetOutput, error) {
+		batchOps := make([]mcms_types.BatchOperation, 0)
+		reports := make([]cldf_ops.Report[any, any], 0)
+		opsBundle := cldf_ops.NewBundle(
+			e.GetContext,
+			e.Logger,
+			cldf_ops.NewMemoryReporter(),
+			cldf_ops.WithOperationRegistry(e.OperationsBundle.OperationRegistry),
+		)
```
chris-de-leon-cll: [style guide violation]: there's no need to create a fresh bundle - prefer using `e.OperationsBundle` which is already pre-configured

deployment/tokens/revoke_admin_role.go
```diff
@@ -0,0 +1,174 @@
...
+					return cldf.ChangesetOutput{}, fmt.Errorf("revocation[%d]: timelock ref is empty for chain selector %d", i, revocation.ChainSelector)
+				}
+			} else if revocation.AdminAddress == "" {
+				return cldf.ChangesetOutput{}, fmt.Errorf("revocation[%d]: no MCMS reader registered for chain family '%s'", i, family)
+			}
+
+			adminAddress := revocation.AdminAddress
+			if adminAddress == "" {
+				adminAddress = timelockAddress
+			}
```
chris-de-leon-cll, line 159: Logic seems a bit odd here - I think we want something like this right?

```go
addressToRevokeAccessFrom := ""
if revocation.AdminAddress != "" {
	addressToRevokeAccessFrom = revocation.AdminAddress
} else {
	// fetch timelock ref
	// if it's not empty -> addressToRevokeAccessFrom = timelock.Address
	// if it's empty -> keep addressToRevokeAccessFrom empty; adapter should fall back to DeployerKey
}


```

deployment/tokens/revoke_admin_role_test.go
chris-de-leon-cll, line 1: Love the table driven tests :) although seems like there's a lot of boilerplate in mocking the MCMS reader and creating helper variables to call the verify function

To help cut down on that, I think it'd be better to:
1. consolidate this into one table driven test
2. move it to `integration-tests/deployment/revoke_token_admin_role_test.go`
3. refactor the test so that it calls `VerifyPreconditions`
4. remove this file

The integration tests already have a harness for a lot of things, so creating a bunch of mocks and helpers in this file seems a bit overkill for some simple validation checks

chris-de-leon-cll (commented):
This is actually pretty solid 😄  only noticed a few areas that seemed a bit inconsistent, but overall nice work!

deployment/tokens/revoke_admin_role.go
matYang: ExecuteSequence checks the bundle reporter for a prior successful run with the same sequence definition and input. If it finds one, it returns the old output instead of running the sequence again.

For this changeset, it can introduce caveats because the sequence output depends on current on-chain state:

1. First Apply: admin still has role, so it emits a revoke op.
2. Proposal executes and role is revoked.
3. If changeset is invoked again in the sequence, apply with the same input should re-read chain state, see the role is already gone, and emit no proposal.

chains/evm/deployment/tokens/tokenimpl/impl.go
matYang: Addressed by moving HasAdminRole onto the main Token interface and adding explicit unsupported implementations for token types without admin-role support.

deployment/tokens/revoke_admin_role.go
matYang: Addressed by dropping the explicit ChainSelector == 0 check and relying on BlockChains.Exists for the environment lookup.

deployment/tokens/revoke_admin_role.go
matYang: Addressed by moving token-ref empty and chain-mismatch validation into resolveTokenAdminRoleRef, so verification now goes through the same resolver path.

deployment/tokens/revoke_admin_role.go
matYang: Addressed by making explicit AdminAddress win, otherwise using the timelock when it resolves, and letting the EVM adapter default to the deployer when no admin is provided.

deployment/tokens/revoke_admin_role_test.go
matYang: Addressed by removing the standalone unit test file and moving the validation cases into the integration-test harness via VerifyPreconditions.

chains/evm/deployment/v1_0_0/sequences/revoke_token_admin_role.go
```diff
@@ -0,0 +1,165 @@
...
+		deployerHasAdminRole := false
+		if chain.DeployerKey.From != (common.Address{}) {
+			if hasRole, err := tokenImpl.HasAdminRole(b.GetContext(), chain, tokenAddress, chain.DeployerKey.From); err != nil {
+				return sequences.OnChainOutput{}, fmt.Errorf("failed to check admin role for deployer %s on token %s: %w", chain.DeployerKey.From.Hex(), tokenAddress.Hex(), err)
+			} else {
+				deployerHasAdminRole = hasRole
+			}
+		}
+		if !timelockHasAdminRole && !deployerHasAdminRole {
+			b.Logger.Warnf("neither timelock %s nor deployer %s has an admin role on token %s on chain %d; skipping revoke since there is no account with sufficient permissions to perform the operation", timelockAddress.Hex(), chain.DeployerKey.From.Hex(), tokenAddress.Hex(), input.ChainSelector)
```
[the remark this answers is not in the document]
chris-de-leon-cll: Hallucination: the `IsAllowedCaller` field already handles this

deployment/tokens/revoke_admin_role.go
```diff
@@ -0,0 +1,180 @@
...
+		if version == nil {
+			return errors.New("chain adapter version is required")
+		}
+
+		for i, revocation := range cfg.Revocations {
+			selector := revocation.ChainSelector
+			if revocation.TokenRef.ChainSelector != 0 && revocation.TokenRef.ChainSelector != revocation.ChainSelector {
+				return fmt.Errorf("revocation[%d]: chain selector mismatch in TokenRef: expected %d, got %d", i, revocation.ChainSelector, revocation.TokenRef.ChainSelector)
+			}
+			if datastore_utils.IsAddressRefEmpty(revocation.TokenRef) {
```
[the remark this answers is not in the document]
chris-de-leon-cll: This is unnecessary - a partial ref is intentional here

All we need to do in Verify is check that at least one field is given

Later in Apply, we will use whatever filters the user gave us to get a fully populated ref

deployment/tokens/revoke_admin_role.go
```diff
@@ -0,0 +1,180 @@
...
+				return fmt.Errorf("revocation[%d]: token ref is required", i)
+			}
+			if !e.BlockChains.Exists(selector) {
```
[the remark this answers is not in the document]
chris-de-leon-cll: This is a micro-optimization that's not worth the squeeze - regardless of this change the correct error is still raised

deployment/tokens/revoke_admin_role.go
```diff
@@ -0,0 +1,180 @@
...
+	FallbackAddress string
+	TimelockAddress string
+}
+
+func RevokeTokenAdminRole() cldf.ChangeSetV2[RevokeTokenAdminRoleInput] {
+	return cldf.CreateChangeSet(
+		revokeTokenAdminRoleApply(GetTokenAdapterRegistry(), changesets.GetRegistry()),
+		revokeTokenAdminRoleVerify(GetTokenAdapterRegistry()),
+	)
+}
```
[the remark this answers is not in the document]
chris-de-leon-cll: Unit and integration tests have been added in ./integration-tests/deployment

chains/evm/deployment/v1_0_0/sequences/revoke_token_admin_role.go
```diff
@@ -0,0 +1,165 @@
...
+		// Validate the chain
+		chain, ok := chains.EVMChains()[input.ChainSelector]
+		if !ok {
+			return sequences.OnChainOutput{}, fmt.Errorf("chain with selector %d not found among provided EVM chains", input.ChainSelector)
+		}
+		if !datastore_utils.IsAddressRefFullyPopulated(input.TokenRef) {
+			return sequences.OnChainOutput{}, fmt.Errorf("token ref is incomplete: %v", input.TokenRef)
+		}
+
+		// Validate the token address
```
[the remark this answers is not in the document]
chris-de-leon-cll: Unit and integration tests have been added in ./integration-tests/deployment

chains/evm/deployment/v1_0_0/sequences/revoke_token_admin_role.go
```diff
@@ -0,0 +1,165 @@
...
+	"Revoke an admin role from an EVM token",
+	func(b cldf_ops.Bundle, chains cldf_chain.BlockChains, input tokensapi.RevokeTokenAdminRoleSequenceInput) (sequences.OnChainOutput, error) {
+		// Validate the chain
+		chain, ok := chains.EVMChains()[input.ChainSelector]
+		if !ok {
+			return sequences.OnChainOutput{}, fmt.Errorf("chain with selector %d not found among provided EVM chains", input.ChainSelector)
+		}
+		if !datastore_utils.IsAddressRefFullyPopulated(input.TokenRef) {
+			return sequences.OnChainOutput{}, fmt.Errorf("token ref is incomplete: %v", input.TokenRef)
+		}
```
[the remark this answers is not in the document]
chris-de-leon-cll: Requiring the full ref is intentional and defensive here - we should ensure that a fully populated ref from the datastore is being passed in via a top level changeset

chris-de-leon-cll (approved):

0xsuryansh (approved):

PR #2068, opened 2026-05-15
Add monitoring for config digest missmatch

emate:
Add per-round config digest mismatch beholder metric for exec and commit plugins

Start emmiting `ccip_{exec|commit}_config_digest_mismatch` gauge (1=mismatch, 0=match) every round by comparing the home chain config digest against the offramp's on-chain config digest. Both reads are cached, so no additional RPC load.

Changes:
* Extract shared `ConfigDigestsMatch`helper 
* Add `TrackConfigDigestMismatch` to exec and commit metrics reporters (
* Refactor existing `checkConfigDigest` (exec) and `validateReport` (commit) to use the shared helper


commit/metrics/prom.go
```diff
@@ -161,8 +172,7 @@ func NewPromReporter(
 
 		sequenceNumbers:        promSequenceNumbers,
 		commitLatestRound:      promCommitLatestRoundID,
-		looppProviderSupported: promLooppCCIPProviderSupported,
-
+		looppProviderSupported: promLooppCCIPProviderSupported, configDigestMismatch: promCommitConfigDigestMismatch,
```
makramkd: **nit**: I think adding `configDigestMismatch` on its own line (nearer to the bottom) would be more in line with expected formatting.

commit/metrics/prom.go
```diff
@@ -340,3 +351,15 @@ func (p *PromReporter) TrackLooppProviderSupported(looppCCIPProviderSupported ma
...
+func (p *PromReporter) TrackConfigDigestMismatch(mismatch bool) {
+	var value float64
+	if mismatch {
+		value = 1
+	}
+	p.configDigestMismatch.WithLabelValues(p.chainFamily, p.chainID).Set(value)
+	p.bhConfigDigestMismatch.Record(context.Background(), int64(value), metric.WithAttributes(
+		attribute.String("chainFamily", p.chainFamily),
+		attribute.String("chainID", p.chainID),
+	))
```
makramkd, line 366: **question**: I see the use of `context.Background()` a lot in these `Track*` methods. I'm wondering what these methods actually do: is it blocking network I/O or is it some kind of IPC (to a `statsd` daemon or similar)?

commit/plugin.go
```diff
@@ -412,6 +415,17 @@ func (p *Plugin) ObserveFChain(lggr logger.Logger) map[cciptypes.ChainSelector]i
 	return fChain
 }
 
+func (p *Plugin) trackConfigDigestMismatch(ctx context.Context, lggr logger.Logger) {
+	configMatch, _, err := plugincommon.ConfigDigestsMatch(
+		ctx, p.ccipReader, consts.PluginTypeCommit, p.reportingCfg.ConfigDigest,
+	)
+	if err != nil {
+		lggr.Errorw("failed to check config digest", "err", err)
```
makramkd: **nit**: more descriptive error message.


```suggestion
		lggr.Errorw("failed to check for config digest mismatch", "err", err)
```

commit/metrics/prom.go
```diff
@@ -92,6 +92,10 @@ var (
 		Name: "ccip_commit_loopp_ccip_provider_supported",
 		Help: "Tracks whether LOOPP CCIP provider is supported for each chain family (1 = supported, 0 = not supported)",
 	}, []string{"chain_family"})
+	promCommitConfigDigestMismatch = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ccip_commit_config_digest_mismatch",
+		Help: "Reports whether the home chain config digest differs from the offramp config digest (1 = mismatch, 0 = match)",
+	}, []string{"chainFamily", "chainID"})
```
makramkd: **question**: this metric has `"chainFamily"` in the string slice, whereas the previous metric has `"chain_family"` (promLooppCCIPProviderSupported) and one pretty far up this list also has `"chainFamily"`. Should we settle on one instead and preferably use a `const` to refer to this label?

commit/metrics/prom.go
```diff
@@ -92,6 +92,10 @@ var (
 		Name: "ccip_commit_loopp_ccip_provider_supported",
 		Help: "Tracks whether LOOPP CCIP provider is supported for each chain family (1 = supported, 0 = not supported)",
 	}, []string{"chain_family"})
+	promCommitConfigDigestMismatch = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ccip_commit_config_digest_mismatch",
```
makramkd, line 96: **question**: should these metric names be `const`? They are referred to in multiple places by the string literal, which as we see in this file can quickly go out of sync.

execute/metrics/prom.go
```diff
@@ -101,6 +101,10 @@ var (
 		Name: "ccip_exec_loopp_ccip_provider_supported",
 		Help: "Tracks whether LOOPP CCIP provider is supported for each chain family (1 = supported, 0 = not supported)",
 	}, []string{"chain_family"})
+	promExecConfigDigestMismatch = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ccip_exec_config_digest_mismatch",
+		Help: "Reports whether the home chain config digest differs from the offramp config digest (1 = mismatch, 0 = match)",
+	}, []string{"chainFamily", "chainID"})
```
makramkd: **question**: same question here with `"chainFamily"` vs. `"chain_family"`.

commit/metrics/prom.go
emate: fixed

commit/plugin.go
emate: fixed, added additional labels
```
			"homeChainConfigDigest", p.reportingCfg.ConfigDigest,
			"pluginType", consts.PluginTypeExecute,
		)
```

commit/metrics/prom.go
emate: fair, adjusted the label names to what we currently have

execute/metrics/prom.go
emate: fixed

commit/metrics/prom.go
emate: The beholder Record/Add calls are non-blocking ops - they buffer metrics for async export

commit/metrics/prom.go
emate: yeah, it can be the way to go. I'm just wondering if we should start with having one metric as const leaving the rest for the other PR is. WDYT @makramkd ?

makramkd: I think its fine if we do the cleanup in a follow-up.

makramkd (approved):

0xsuryansh (approved):

PR #2071, opened 2026-05-15
Don't spam config poller cache miss retries

ogtownsend:
NOPs reported excessive error log spam from `processSourceChainConfigResults()` when a source chain's `GetSourceChainConfig` RPC call fails persistently (e.g., due to an RPC misconfiguration like gas limit below intrinsic gas).

Root cause: 
- When `GetAllConfigsLegacy` fetches source chain configs and receives some failure, `processSourceChainConfigResults` logs the error and skips that chain and it's never added to the config poller cache. 
- In `configPollerV2.GetOfframpSourceChainConfigs()`, any chain not in the cache is treated as a "cache miss," which triggers an inline call to `batchRefreshChainAndSourceConfigs`. 
- Since the failing chain never gets cached, every subsequent call triggers another full batch fetch, producing the same error logs every time. 
- Two concurrent callers in the merkle root `asyncObserver.sync()` (`ObserveOffRampNextSeqNums` and `ObserveLatestOnRampSeqNums`) both hit this path on every tick, multiplying the log volume.

Fix:
`pkg/reader/config_poller_v2.go`:
- Added `attemptedSourceChains` map to `chainCache` to track which source chains have been included in at least one fetch attempt.
- `GetOfframpSourceChainConfigs`: A chain missing from the cache is only treated as a "cache miss" (triggering an inline fetch) if it has never been attempted. Chains that were previously attempted but failed are not re-fetched inline 
- The background poller (`refreshAllKnownChains`) will still retry them on its regular schedule.
- `batchRefreshChainAndSourceConfigs`: After each fetch, all requested source chain selectors are recorded in `attemptedSourceChains` and sourceChainRefresh is always updated, regardless of whether individual chains succeeded.

`pkg/chainaccessor/config_processors.go`:
- Downgraded the "Failed to get source chain config from result" log from Errorw to Warnw. 
- This is a per-chain partial failure that is retried by the background poller, `Error` level creates excessive noise

### DIFF
--- a/pkg/reader/config_poller_v2.go
+++ b/pkg/reader/config_poller_v2.go
@@ -71,6 +71,11 @@ type chainCache struct {
 	sourceChainMu            sync.RWMutex
 	staticSourceChainConfigs map[cciptypes.ChainSelector]StaticSourceChainConfig
 	sourceChainRefresh       time.Time // Single timestamp for all source chain configs
+	// attemptedSourceChains tracks which source chains have been included in at least one
+	// fetch attempt. Chains in this set that are NOT in staticSourceChainConfigs were either
+	// unconfigured on-chain or had an RPC error, either way, re-fetching inline on every
+	// call creates a doom-loop of redundant RPC calls and error log spam.
+	attemptedSourceChains map[cciptypes.ChainSelector]struct{}
 }
 
 // newConfigPollerV2 creates a new instance of configPollerV2 with improved batch fetching capabilities.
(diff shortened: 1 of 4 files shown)


makramkd (dismissed):

pkg/reader/config_poller_v2.go
```diff
@@ -297,16 +302,19 @@ func (c *configPollerV2) GetOfframpSourceChainConfigs(
...
 		if exists {
 			cachedSourceConfigs[chain] = staticSourceChainConfig
-		} else {
-			// This chain isn't in cache yet
+		} else if _, attempted := destChainCache.attemptedSourceChains[chain]; !attempted {
+			// Only treat as missing if we haven't attempted to fetch this chain before.
+			// Chains that were attempted but not returned are either unconfigured on-chain
+			// or had an RPC error. Re-fetching inline on every call creates a doom-loop
+			// of redundant RPC calls and error log spam. The background poller will retry.
 			missingChains = append(missingChains, chain)
```
winder: **question**: Would it be possible for a bad config to cause the config to disappear? In that case this cache would need to have some sort of expiry to make sure we detect it's now missing.

winder (dismissed):

PR #2074, opened 2026-05-19
add outbound only tprl

tt-cll:


chains/solana/deployment/v1_6_0/sequences/tokens.go
```diff
@@ -498,13 +499,35 @@ func (a *SolanaAdapter) SetTokenPoolRateLimits() *cldf_ops.Sequence[tokenapi.TPR
...
+			if rl.OutboundOnly {
+				remoteChainConfigPDA, _, pdaErr := tokens.TokenPoolChainConfigPDA(input.RemoteChainSelector, tokenMint, tokenPool)
+				if pdaErr != nil {
+					return sequences.OnChainOutput{}, fmt.Errorf("failed to derive remote chain config PDA for pass-through inbound: %w", pdaErr)
+				}
+				var remoteChainConfigAccount base_token_pool.BaseChain
+				if readErr := chain.GetAccountDataBorshInto(b.GetContext(), remoteChainConfigPDA, &remoteChainConfigAccount); readErr != nil {
+					return sequences.OnChainOutput{}, fmt.Errorf("failed to read remote chain config account for pass-through inbound: %w", readErr)
+				}
+				inboundRL = tokenapi.RateLimiterConfig{
```
agusaldasoro: shouldn't we call `adapter.GetOnchainInboundRateLimit()` here?

chains/solana/deployment/v1_6_0/sequences/tokens.go
```diff
@@ -498,13 +499,35 @@ func (a *SolanaAdapter) SetTokenPoolRateLimits() *cldf_ops.Sequence[tokenapi.TPR
 
 			tokenMint := solana.MustPublicKeyFromBase58(tokenAddr.Address)
 			tokenPool := solana.MustPublicKeyFromBase58(input.TokenPoolRef.Address)
+
+			inboundRL := rl.InboundRateLimiterConfig
+			// In OutboundOnly mode the inbound side of the lane is not authored from user input;
+			// read the current on-chain inbound from the token pool's remote-chain config PDA and
+			// pass it through unchanged. setChainRateLimit on Solana takes both inbound and outbound
+			// atomically.
+			if rl.OutboundOnly {
```
agusaldasoro: Is it okay to have this here, or should be done before calling `SetTokenPoolRateLimits`?

deployment/tokens/rate_limits.go
```diff
@@ -400,6 +486,82 @@ func buildTPRLRemotesForSetRateLimitsLane(
 	return defaultOutboundRL, defaultInboundRL, buckets, nil
 }
 
+// validateOutboundOnlyAgainstCounterpartInbound reads the counterpart chain's on-chain
+// inbound rate limit (one per FastFinality bucket present in localOutbounds) and checks
+// that each is at least 110% of the new outbound being set on the local chain (both
+// Capacity and Rate). Per the design, IsEnabled is ignored when checking — adapters are
+// expected to return a zero RateLimiterConfig for unconfigured lanes, which will cause
+// the check to reject any positive outbound.
```
agusaldasoro, line 494: Couldn't it happen that we need to set an outbound only rate limit to a token pool that didn't have any configured? 🤔

deployment/tokens/rate_limits.go
```diff
@@ -318,6 +351,42 @@ func setTokenPoolRateLimitsApply() func(cldf.Environment, TPRLInput) (cldf.Chang
...
+					// on-chain inbound for each finality bucket and pass it through unchanged.
+					localReader, ok := tokenPoolAdapter.(RateLimitReaderAdapter)
+					if !ok {
+						return cldf.ChangesetOutput{}, fmt.Errorf(
+							"adapter for local chain selector %d (family %s, version %v) does not implement RateLimitReaderAdapter; outbound-only mode unsupported for this lane",
+							selector, family, tokenPool.Version,
+						)
+					}
+					for i, bucket := range tprlRemote.RateLimitBuckets {
+						current, err := localReader.GetOnchainInboundRateLimit(
```
agusaldasoro, line 365: nit: you are calling `GetOnchainInboundRateLimit` also inside `validateOutboundOnlyAgainstCounterpartInbound`, you could send current as param to avoid that repeated call.

tt-cll: these should actually be different inbounds (one for each direction)

deployment/tokens/rate_limits.go
tt-cll: in that case we should be going with the normal flow I think of outbound + inbound

deployment/tokens/rate_limits.go
```diff
@@ -44,6 +44,14 @@ type RemoteOutbounds struct {
 	// for all token pool versions. The slice should only contain up to two entries (one per finality
 	// flag). Pre-v2 adapters ignore fast-finality rows and only apply default buckets if they exist.
 	Outbounds []RateLimitConfig `yaml:"outbounds" json:"outbounds"`
+
+	// OutboundOnly, when true, configures only the outbound side of this lane on the local chain.
+	// The counterpart chain's TPRLConfig need not provide a matching RemoteOutbounds entry; the
+	// changeset will read the counterpart's on-chain inbound for the same lane and reject the
+	// update if it is not at least 110% of the new outbound (capacity AND rate). The counterpart
+	// chain's TPRLConfig must still exist with TokenPoolRef + TokenRef so the changeset can
+	// resolve the counterpart pool and its decimals for the validation read.
```
chris-de-leon-cll, line 53: Wanna make sure I understand the expected behavior for mixed pre-V2 and V2+ pools in this PR - do the tables below accurately depict what happens in each case?

#### Case 1: you are setting outbound only on pre-v2:

| What you set for the lane | Pre‑V2 pool | V2 pool |
| --- | --- | --- |
| **Default only** | Applies **default** RL bucket. | Unchanged (read only). |
| **FF only** | Errors (FF not supported). | Unchanged (read only). |
| **Default + FF** | Applies **default** RL only. | Unchanged (read only). |
| **Empty** (no default, no FF) | Unchanged (no-op). | Unchanged (no-op). |

#### Case 2: you are setting outbound only on v2:

| What you set for the lane | Pre‑V2 pool | V2 pool |
| --- | --- | --- |
| **Default only** | Unchanged (read only). | Applies **default** RL bucket. |
| **FF only** | Unchanged (read only). | Applies **FF** RL bucket. |
| **Default + FF** | Unchanged (read only). | Applies **default** and **FF** buckets in one batch. |
| **Empty** (no default, no FF) | Unchanged (no-op). | Unchanged (no-op). |

#### Case 3: you are setting outbound only on both pre-v2 AND v2:

| What you set for the lane | Pre‑V2 pool | V2 pool |
| --- | --- | --- |
| **Default only** | Applies **default** RL bucket. | Applies **default** RL bucket. |
| **FF only** | Ignores **FF** RL bucket. | Applies **FF** RL bucket. |
| **Default + FF** | Applies **default** RL only. | Applies **default** and **FF** buckets in one batch. |
| **Empty** (no default, no FF) | Rate limits remain as-is. | Rate-limits remain as-is. |

chris-de-leon-cll: In particular, if we have both Default + FF buckets involved for a pre-v2 pool, then would this PR make the tooling hard error because pre-v2 pools don't support FF rate limits? Or would it log a warning and only apply the default bucket for the pre-v2 pool?

tt-cll: It would currently hard error, which is different from the old behavior. But if a user is specifying FF for a pre-v2 pool, that seems like something we'd want to raise to the user.

chris-de-leon-cll (approved):

agusaldasoro (approved):

PR #2077, opened 2026-05-19
Improve override logic for FeeQuoterDestChainConfig

AnieeG:
Existing override logic did not allow updating to zero values if the existing on-chain values are non-zero. Updated the logic to use pointer fields so that values if specified in inputs are accepted even if it's zero


chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
```diff
@@ -492,18 +494,58 @@ func maybeAddOnRampDestChainConfigArgOnLocalChain(
...
+	if desired.MaxPerMsgGasLimit == nil {
+		desired.MaxPerMsgGasLimit = ptr.To(cur.MaxPerMsgGasLimit)
+	}
+	if desired.DestGasOverhead == nil {
+		desired.DestGasOverhead = ptr.To(cur.DestGasOverhead)
+	}
+	if desired.DestGasPerPayloadByteBase == nil {
+		desired.DestGasPerPayloadByteBase = ptr.To(cur.DestGasPerPayloadByteBase)
+	}
+	if desired.ChainFamilySelector == [4]byte{} {
```
tt-cll: nit: imo this shouldn't be here since it's static for a given chain family. Safer to remove it

deployment/v2_0_0/changesets/configure_chains_for_lanes_from_topology.go
```diff
@@ -627,41 +607,45 @@ func mergeCommitteeVerifierRemoteChainConfig(
...
 	}
 	if overrides.MaxPerMsgGasLimit != nil {
-		defaults.MaxPerMsgGasLimit = *overrides.MaxPerMsgGasLimit
+		out.MaxPerMsgGasLimit = overrides.MaxPerMsgGasLimit
 	}
 	if overrides.DestGasPerPayloadByteBase != nil {
-		defaults.DestGasPerPayloadByteBase = *overrides.DestGasPerPayloadByteBase
+		out.DestGasPerPayloadByteBase = overrides.DestGasPerPayloadByteBase
+	}
+	if overrides.ChainFamilySelector != [4]byte{} {
```
tt-cll: again, don't think we should ever expose this to a user to override via config

tt-cll (approved):

chris-de-leon-cll (approved):

0xsuryansh (approved):

PR #2109, opened 2026-06-04
Attempt to fix expandLanesToPartialChainConfigs to support all qualifiers

archseer:
…iers

### DIFF
--- a/deployment/v2_0_0/changesets/build_lanes_cross_family.go
+++ b/deployment/v2_0_0/changesets/build_lanes_cross_family.go
@@ -58,17 +58,20 @@ type ConfigureChainsForLanesFromTopologyConfig struct {
 
 // expandLanesToPartialChainConfigs converts bidirectional lane pairs into the internal
 // chain-centric representation used by the changeset apply path.
-func expandLanesToPartialChainConfigs(lanes []CrossFamilyLanePair, topology *offchain.NOPTopology) ([]partialChainConfig, error) {
+func expandLanesToPartialChainConfigs(lanes []CrossFamilyLanePair, committees map[string]offchain.CommitteeConfig) ([]partialChainConfig, error) {
 	if len(lanes) == 0 {
 		return nil, fmt.Errorf("at least one lane must be specified")
 	}
 
-	qualifiers := []string{defaultQualifier}
-	if topology != nil && len(topology.Committees) > 0 {
-		qualifiers = make([]string, 0, len(topology.Committees))
-		for qualifier := range topology.Committees {
+	var qualifiers []string
+	if len(committees) > 0 {
+		qualifiers = make([]string, 0, len(committees))
+		for qualifier := range committees {
 			qualifiers = append(qualifiers, qualifier)
 		}
+	}
+	if len(qualifiers) == 0 {
+		qualifiers = []string{defaultQualifier}
 	}
 
 	byChain := make(map[uint64]*partialChainConfig)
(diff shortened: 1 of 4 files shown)


deployment/v2_0_0/changesets/build_lanes_cross_family.go
```diff
@@ -58,13 +58,17 @@ type ConfigureChainsForLanesFromTopologyConfig struct {
 
 // expandLanesToPartialChainConfigs converts bidirectional lane pairs into the internal
 // chain-centric representation used by the changeset apply path.
-func expandLanesToPartialChainConfigs(lanes []CrossFamilyLanePair) ([]partialChainConfig, error) {
+func expandLanesToPartialChainConfigs(lanes []CrossFamilyLanePair, topology *offchain.NOPTopology) ([]partialChainConfig, error) {
```
agusaldasoro: do you need the topology here or just the committees?

agusaldasoro (dismissed):

deployment/v2_0_0/changesets/build_lanes_cross_family.go
```diff
@@ -108,15 +112,21 @@ func sortedKeys(m map[uint64]*partialChainConfig) []uint64 {
...
-		cfg = &partialChainConfig{
-			ChainSelector: local,
-			CommitteeVerifiers: []committeeVerifierInputConfig{{
+		cvConfigs := make([]committeeVerifierInputConfig, 0, len(qualifiers))
+		for _, qualifier := range qualifiers {
+			cvConfigs = append(cvConfigs, committeeVerifierInputConfig{
 				CommitteeQualifier: defaultQualifier,
 				RemoteChains:       make(map[uint64]committeeVerifierRemoteChainInput),
-			}},
+				// TODO: AllowedFinalityConfig: local.AllowedFinalityConfig, sourced from where?
```
AnieeG: There is a method on the adapter GetDefaultFinalityConfig and it gets populated later in the changeset

deployment/v2_0_0/changesets/build_lanes_cross_family.go
```diff
@@ -58,13 +58,23 @@ type ConfigureChainsForLanesFromTopologyConfig struct {
...
+	var qualifiers []string
+	if len(committees) > 0 {
+		qualifiers = make([]string, 0, len(committees))
+		for qualifier := range committees {
+			qualifiers = append(qualifiers, qualifier)
+		}
+	}
+	if len(qualifiers) == 0 {
+		qualifiers = []string{defaultQualifier}
+	}
```
AnieeG, line 76: **Bug: non-deterministic qualifier ordering**

Map iteration in Go is randomized, so the `qualifiers` slice has a different order on every run. This means `CommitteeVerifiers` in the resulting `partialChainConfig` will have qualifiers at different indices across calls — including between `validate` and `apply`. Any downstream code that processes verifiers by position will silently get a different qualifier than expected.

The same file already uses `sortedKeys` for `map[uint64]*partialChainConfig` to ensure determinism. Apply the same pattern here:

```go
qualifiers = make([]string, 0, len(committees))
for qualifier := range committees {
    qualifiers = append(qualifiers, qualifier)
}
sort.Strings(qualifiers)
```

deployment/v2_0_0/changesets/build_lanes_cross_family_internal_test.go
```diff
@@ -13,7 +13,7 @@ func TestExpandLanesToPartialChainConfigs_MinimalLane(t *testing.T) {
 
 	chains, err := expandLanesToPartialChainConfigs([]CrossFamilyLanePair{
```
AnieeG, line 14: **Missing test coverage: multiple qualifiers (len > 1) path is untested**

All internal unit tests pass `nil` for `committees`, which falls back to the single-`"default"`-qualifier path. The integration tests in `configure_chains_for_lanes_from_topology_test.go` cover the single-committee non-nil case, but no test exercises `len(committees) > 1`.

The new code paths that are untested:
- `qualifiers` slice built from multiple map keys (and the ordering issue noted above)
- `mergeLaneLeg` creating multiple `committeeVerifierInputConfig` entries per chain
- `CommitteeVerifiers` slice containing more than one entry in downstream enrichment

Consider adding a unit test like `TestExpandLanesToPartialChainConfigs_MultipleCommittees` with a 2-entry committees map to cover this path.

archseer:
Replaced by #2112

PR #2118, opened 2026-06-08
changeset to remove extra fee tokens from feequoter 2.0

AnieeG:
This changes existing changeset to only import prices for feetokens from previous versions
Adds a changeset to remove extra fee tokens from feequoter 2.0.

### DIFF
--- a/chains/evm/deployment/v2_0_0/changesets/remove_fee_tokens.go
+++ b/chains/evm/deployment/v2_0_0/changesets/remove_fee_tokens.go
@@ -229,6 +229,10 @@ func extraFeeTokens(fq20Tokens, legacyFeeTokens []common.Address) []common.Addre
 		legacySet[token] = struct{}{}
 	}
 
+	if len(legacySet) == 0 {
+		return nil
+	}
+
 	extra := make([]common.Address, 0)
 	for _, token := range fq20Tokens {
 		if _, exists := legacySet[token]; !exists {
--- a/chains/evm/deployment/v2_0_0/changesets/remove_fee_tokens.go
+++ b/chains/evm/deployment/v2_0_0/changesets/remove_fee_tokens.go
@@ -1,0 +1,235 @@
+package changesets
+
+import (
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/ethereum/go-ethereum/common"
+	cldf_chain "github.com/smartcontractkit/chainlink-deployments-framework/chain"
+	"github.com/smartcontractkit/chainlink-deployments-framework/chain/evm"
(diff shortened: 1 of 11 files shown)


chris-de-leon-cll (dismissed):

chris-de-leon-cll (dismissed):

simsonraj (dismissed):

chains/evm/deployment/v2_0_0/sequences/fee_quoter.go
```diff
@@ -692,6 +693,72 @@ var (
 		})
 )
 
+// RemoveFeeTokensPerChainInput holds per-chain parameters for removing extra fee tokens from FeeQuoter 2.0.
+type RemoveFeeTokensPerChainInput struct {
+	ChainSelector     uint64
+	FeeQuoter20Ref    datastore.AddressRef
```
RensR, line 699: Why do we need to assume there's multiple FQv2s on a chain? If we assume there's only one we don't have to worry about this input, right?

chains/evm/deployment/v2_0_0/changesets/remove_fee_tokens.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	return sequences.RemoveFeeTokensPerChainInput{
+		ChainSelector:     chainSel,
+		FeeQuoter20Ref:    fq20Ref,
+		FeeTokensToRemove: extraFeeTokens(fq20Tokens, legacyFeeTokens),
+	}, nil
+}
+
+func collectLegacyFeeTokens(
```
RensR, line 97: Can you not shortcut this entire process inc fetching routers by just looking for 1.6 FQ in address refs, if found query fee tokens. If not found try PriceReg and do the same?

Fee tokens are not lane config, they're chain config. And if you find the correct tokens on 1.6 you don't even have to query 1.5 right?

chains/evm/deployment/v2_0_0/sequences/fee_quoter.go
AnieeG: I have set this as an input to `sequence` to avoid duplication of finding the FQ20 address twice( first in changeset and then in sequence) , We are already resolving it in changeset ResolveInput,  just convenient to pass the same in the sequence.

chains/evm/deployment/v2_0_0/changesets/remove_fee_tokens.go
AnieeG: done

chris-de-leon-cll (approved):

crispymangoes (approved):

PR #2122, opened 2026-06-09
fix: prevent hung source chain RPC from blocking merkleroot observation

nvsriram:
## Description

A hung or slow RPC on one source chain can block commit observations for all chains when work fans out per chain on a shared OCR `ctx`.

This PR adds:
- Bounded per-chain contexts via `internal/libs/rpctimeout`:
   - Timeout: `min(rpctimeout.Default (10s), 80% of remaining OCR deadline)`
   - On timeout or error: log, skip that chain, continue with others
 - Updated `ObserveLatestOnRampSeqNums`, `ObserveMerkleRoots`, `GetChainsFeeComponents`, `GetWrappedNativeTokenPriceUSD` to use the per-chain context instead

## Testing

- Added unit tests for the timeout formula
- Added hung-RPC isolation subtests to ensure per-chain context is applied

### DIFF
--- a/commit/merkleroot/observation.go
+++ b/commit/merkleroot/observation.go
@@ -649,9 +649,18 @@ func (o observerImpl) ObserveMerkleRoots(
 	for _, chainRange := range ranges {
 		if supportedChains.Contains(chainRange.ChainSel) {
 			wg.Go(func() {
-				msgs, err := o.ccipReader.MsgsBetweenSeqNums(ctx, chainRange.ChainSel, chainRange.SeqNumRange)
+				chainCtx, chainCancel := rpctimeout.Context(ctx)
+				defer chainCancel()
+				msgs, err := o.ccipReader.MsgsBetweenSeqNums(chainCtx, chainRange.ChainSel, chainRange.SeqNumRange)
 				if err != nil {
-					lggr.Warnw("call to MsgsBetweenSeqNums failed", "err", err)
+					if errors.Is(err, context.DeadlineExceeded) {
+						lggr.Warnw("timed out getting messages for source chain",
+							"sourceChain", chainRange.ChainSel,
+							"range", chainRange.SeqNumRange,
+						)
+					} else {
+						lggr.Warnw("call to MsgsBetweenSeqNums failed", "err", err)
+					}
 					return
 				}
 
--- a/commit/merkleroot/observation.go
+++ b/commit/merkleroot/observation.go
@@ -26,6 +26,7 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 
 	"github.com/smartcontractkit/chainlink-ccip/commit/merkleroot/rmn"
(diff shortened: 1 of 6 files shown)


KodeyThomas (dismissed):

commit/merkleroot/observation.go
```diff
@@ -590,12 +620,16 @@ func (o observerImpl) ObserveLatestOnRampSeqNums(ctx context.Context) []pluginty
 				lggr.Warnw("rmn enablement is misconfigured on this lane, skipping observations", "source", sourceChain)
 				return
 			}
-			latestOnRampSeqNum, err := o.ccipReader.LatestMsgSeqNum(ctx, sourceChain)
+			chainCtx, chainCancel := chainRPCContext(ctx)
+			latestOnRampSeqNum, err := o.ccipReader.LatestMsgSeqNum(chainCtx, sourceChain)
+			chainCancel()
```
RensR: defer cancel instead?

commit/merkleroot/observation.go
```diff
@@ -53,6 +53,30 @@ func (p *Processor) ObservationQuorum(
...
 const SendingObservation = "sending merkle root processor observation"
 
+const (
+	defaultChainRPCTimeout           = 10 * time.Second
+	chainRPCTimeoutOCRBudgetFraction = 0.8
+)
+
+// chainRPCTimeoutDuration returns the per-chain RPC timeout as
+// min(80% of remaining OCR budget, defaultChainRPCTimeout).
+func chainRPCTimeoutDuration(ctx context.Context) time.Duration {
```
RensR: Could move this to a more generic file and then use it for the other calls as well, like ObserveMerkleRoots

EDIT: maybe fine to leave it here unchanged if we only need it in this file

commit/merkleroot/observation.go
```diff
@@ -53,6 +53,30 @@ func (p *Processor) ObservationQuorum(
 
 const SendingObservation = "sending merkle root processor observation"
 
+const (
+	defaultChainRPCTimeout           = 10 * time.Second
```
RensR: Could make this config like MerkleRootAsyncObserverSyncTimeout

commit/merkleroot/observation.go
nvsriram: Done! Refactored the helper to its own internal lib and I've extended RPC timeout logic to `ObserveMerkleRoots`, `GetWrappedNativeTokenPriceUSD`, and `GetChainsFeeComponents` as well.

commit/merkleroot/observation.go
nvsriram: Adding it as a config would require more checks to be in place to ensure the value is valid etc. and increase operator overhead. Since this value is also capped by `min(Default, 80% of remaining OCR deadline)`, the OCR parent deadline would be the binding limit in OCR rounds and not the constant.

commit/merkleroot/observation.go
nvsriram: Done!

RensR (approved):

KodeyThomas (approved):

PR #2131, opened 2026-06-12
Change Minimum NOPs in Topology to be Validated for Each Chain Family

ma33r:
### What
- Changes minimum number of NOPs to be verified per-chain-family 
- Changes tests to match this new pattern

### Why
- Some chain families such as EVM will need a minimum of 15 NOPs while others such as Canton will only need 4


deployment/v2_0_0/adapters/chain_family.go
```diff
@@ -118,6 +118,7 @@ type ChainFamily interface {
 	GetDefaultRemoteChainConfig(sourceChainSelector, remoteChainSelector uint64) RemoteChainDefaults
 	GetDefaultCommitteeVerifierRemoteChainConfig() CommitteeVerifierRemoteChainDefaults
 	GetDefaultFinalityConfig() finality.Config
+	ValidateMinimumNOPsTopology(chainSelector string, nopCount int) error
```
AnieeG: instead of MinimumNOPs let's have it as `ValidateNOPsTopology` in future we can add more validation here whenever required

AnieeG (approved):

chris-de-leon-cll (approved):

PR #2134, opened 2026-06-13
Check for valid lanes before issuing chain queries, log cleanup

ogtownsend:
I queried live testnet and mainnet logs for the highest firing logs and found some are extremely spammy, like `rmn enablement is misconfigured on this lane, skipping observations` and`sibling source chain config not found, skipping observations` which logs over 260k times per hour

From what I can tell, the log noise stems from two main sources:
1. deprecated chains still configured (already flagged to team for cleanup)
2. the fact that the plugin queries CCIPHome for ALL supported source chains when issuing queries even if there is no valid lane from that source chain to the current plugin's dest chain

This PR aims to reduce log volume and reduce unnecessary RPC calls by:
1. Checking for valid lane configurations before calling `ObserveLatestOnRampSeqNums()` by checking `IsEnabled`, has an onRamp, and `IsRMNVerificationDisabled`
3. No longer calling `LatestMsgSeqNum()` for disabled/invalid lanes
4. Grouping legitimate errors into a single error log instead of spamming them, but not dropping any information about errored chains 
5. Adding a log sampler to warning logs for inactive lanes


commit/merkleroot/observation.go
```diff
@@ -40,6 +41,53 @@ import (
...
+	// skippedDisabled have an offramp config but are disabled.
+	skippedDisabled []cciptypes.ChainSelector
+	// rmnMisconfigured are live lanes that unexpectedly have RMN verification enabled.
+	rmnMisconfigured []cciptypes.ChainSelector
+}
+
+// classifyOffRampSourceLanes buckets the supported source chains based on their offramp source chain
+// config. Only "live" lanes should be queried for onRamp seq nums, the other buckets are skipped and
+// exist for logging.
+func classifyOffRampSourceLanes(
```
ogtownsend, line 68: this mirrors the previous per-chain skip behavior in the current state of the codebase, so observation output should be unchanged

commit/merkleroot/observation.go
```diff
@@ -581,22 +629,28 @@ func (o observerImpl) ObserveLatestOnRampSeqNums(ctx context.Context) []pluginty
 		return nil
 	}
 
+	classification := classifyOffRampSourceLanes(supportedSourceChains, sourceChainsCfg)
+
+	if len(classification.skippedNotALane) > 0 || len(classification.skippedDisabled) > 0 {
+		logutil.LogWhenExceedFrequency(&lastSkippedLanesLog, skippedLanesLogFrequency, func() {
+			lggr.Debugw("skipping onRamp seq num observations for non-live lanes",
+				"noConfigOrOnRamp", classification.skippedNotALane,
+				"disabled", classification.skippedDisabled,
```
ogtownsend, line 638: sampled every 30 minutes since it doesn't seem valuable to log this every time for non-live lanes

pkg/reader/ccip.go
```diff
@@ -341,36 +341,57 @@ func (r *ccipChainReader) NextSeqNum(
...
 
 		res[chain] = cciptypes.SeqNum(cfg.MinSeqNr)
 	}
 
+	if len(configNotFound) > 0 {
+		lggr.Debugw("source chain config not found, chains skipped", "chains", configNotFound)
+	}
+	if len(disabledChains) > 0 {
+		lggr.Debugw("source chains disabled, chains skipped", "chains", disabledChains)
+	}
```
makramkd: **question**: Should these also use `logutil.LogWhenExceedFrequency`?

pkg/reader/ccip.go
```diff
@@ -411,7 +432,7 @@ func (r *ccipChainReader) GetChainsFeeComponents(
 		func(chain cciptypes.ChainSelector) {
 			chainAccessor, err := getChainAccessor(r.accessors, chain)
 			if err != nil {
-				lggr.Errorw("failed to get chain accessor", "chain", chain, "err", err)
+				lggr.Debugw("failed to get chain accessor", "chain", chain, "err", err)
```
makramkd: **question**: Ditto here, this is pretty spammy in mainnet logs IIRC.

pkg/reader/ccip.go
```diff
@@ -469,7 +490,7 @@ func (r *ccipChainReader) GetWrappedNativeTokenPriceUSD(
 
 			chainAccessor, err := getChainAccessor(r.accessors, chain)
 			if err != nil {
-				lggr.Errorw("chain accessor not found, chain native price skipped", "chain", chain, "err", err)
+				lggr.Debugw("chain accessor not found, chain native price skipped", "chain", chain, "err", err)
```
makramkd: **question**: Ditto here

pkg/reader/ccip.go
```diff
@@ -478,7 +499,7 @@ func (r *ccipChainReader) GetWrappedNativeTokenPriceUSD(
 				if errors.Is(err, context.DeadlineExceeded) {
 					lggr.Warnw("timed out getting chain config for native token address", "chain", chain)
 				} else {
-					lggr.Warnw("failed to get chain config for native token address", "chain", chain, "err", err)
+					lggr.Debugw("failed to get chain config for native token address", "chain", chain, "err", err)
```
makramkd: **question**: Ditto here

pkg/reader/config_poller_v2.go
```diff
@@ -274,11 +281,7 @@ func (c *configPollerV2) GetOfframpSourceChainConfigs(
 
 	// Add any new source chains to list of tracked source chains for background refreshing
 	for _, chain := range filteredSourceChains {
-		if !c.trackSourceChainForDest(chain) {
-			c.lggr.Warnw("Could not track source chain for background refreshing",
-				"destChain", c.destChainSelector,
-				"sourceChain", chain)
-		}
+		c.trackSourceChainForDest(chain)
```
makramkd, line 284: **question**: Will the return value of `trackSourceChainForDest` ever be used? If not, can we get rid of it?

pkg/reader/ccip.go
ogtownsend: sure might as well, grouping them together here instead of logging them in each loop iteration should already help a bunch, but we can add sampling too

makramkd (dismissed):

winder (approved):

makramkd (approved):

PR #2151, opened 2026-06-22
Various minor CLD fixes & code dedup

RensR:
- Set min nops to 9 (threshold for v2)
- Transfer MCM ownership on deploy
- Extract shared logic to reduce code dup
- Allow passing in extra RMN cursers on deployment

### DIFF
--- a/deployment/deploy/mcms.go
+++ b/deployment/deploy/mcms.go
@@ -167,7 +167,7 @@ func updateMCMSConfigApply(d *DeployerRegistry, mcmsRegistry *changesets.MCMSRea
 			reports = append(reports, report.ExecutionReports...)
 		}
 
-		return changesets.NewOutputBuilder(e, mcmsRegistry).
+		return changesets.NewOutputBuilder(e, mcmsRegistryForDeploy(mcmsRegistry)).
 			WithReports(reports).
 			WithBatchOps(batchOps).
 			Build(cfg.MCMS)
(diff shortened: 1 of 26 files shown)


deployment/fastcurse/product.go
```diff
@@ -127,6 +127,9 @@ func (cr *CurseRegistry) groupRMNSubjectBySelector(e cldf.Environment, rmnSubjec
 		if err != nil {
 			return nil, err
 		}
+		if s.Version == nil {
```
RensR, line 130: Version is required

KodeyThomas (dismissed):

deployment/deploy/mcms.go
```diff
@@ -336,10 +336,49 @@ func deployMCMSApply(
...
+	if registry != nil {
+		return registry
+	}
+	return changesets.GetRegistry()
+}
+
+// mcmsInputForDeployment returns MCMS proposal input for deploy output. When ownership
+// accept batch operations are produced and the caller did not supply MCMS config, a
+// default bypass proposal is used so timelock can accept MCM contract ownership.
+func mcmsInputForDeployment(cfg MCMSDeploymentConfig, batchOps []mcmstypes.BatchOperation) mcms.Input {
```
simsonraj: wondering in what case we would need a bypass proposal? if the only caller is the deployment route.
Bypass would need a security signatures, so not sure if it would ever be needed & we would always still have to pass the MCMS input, in that case should that be left alone and just error out if mcms input is not provided

KodeyThomas (approved):

simsonraj (approved):

PR #2165, opened 2026-07-01
fix: solana manual reg. - extend token pool LUT in token multisig flow

chris-de-leon-cll:
# Solana manual registration: extend token pool LUT with SPL multisig

## Problem

Manual registration can create an SPL multisig for customer mint authority but did not add it to the token pool LUT. This is an issue because Solana token pools need to resolve the mint-authority multisig from LUT-supplied remaining accounts. If it is missing, execution fails with `InvalidMultisig` and operators previously had to extend the LUT in a separate step.

## What this adds

New operation `ExtendTokenPoolLookupTable`, called from manual registration after multisig creation when `customerMintAuthorities` is set. The multisig pubkey is appended to the on-chain token pool LUT (from token admin registry, not datastore).

## Design choices

- **On-chain LUT address, not datastore** — The operation resolves which lookup table to extend from the token admin registry account on-chain (`LookupTable` field), not from a datastore ref. That matches what the router and token pool programs use at execution time. If the field is unset, it fails with an explicit message to run configure-for-transfers / set-pool first.

- **Extend only when multisig path is active** — LUT extension runs inside the existing `customerMintAuthorities` branch, immediately after multisig creation. Manual registration without customer mint authorities is unchanged. No new YAML surface beyond what operators already pass for multisig setup.

- **Deployer key signs the extend transaction** — Token pool LUTs are created and extended with the chain deployer key as the Solana address lookup table authority, including during set-pool in configure-for-transfers. Timelock may administer the token admin registry via MCMS, but that is separate from ALT authority. This follows the same signing model as set-pool and existing offramp LUT extend helpers—not an MCMS batch for the extend instruction.

- **Dedup and idempotency** — Before extending, the operation loads current LUT entries and only submits accounts not already present. Duplicate pubkeys within the input list are collapsed so a single extend instruction never repeats the same address. Re-running manual registration when the multisig is already in the LUT is a no-op rather than an error.

- **Slot warmup after extend** — After a successful extend, the operation waits for a slot change before returning, consistent with initial token pool LUT setup during set-pool. New LUT entries are not reliably usable in the same slot they are added.

- **Non-atomic multisig + extend** — Multisig creation and LUT extension are separate confirmed transactions. If extend fails after multisig creation, the multisig still exists on-chain—the same partial state as before this change. Retry extend or fix LUT state before customer cutover.

## Operational prerequisites

Requires an existing token pool LUT (from configure-for-transfers / set-pool). Typical order: configure connectivity first, then manual registration with `skipTokenPoolInit` if the pool is already initialized, then customer `setAuthority` to the multisig.

## Integration test adjustments

Existing manual registration subtest no longer passes `customerMintAuthorities` (runs before configure-for-transfers, so no LUT yet).

New subtest after configure-for-transfers asserts manual registration with `customerMintAuthorities` adds the multisig to the on-chain LUT.

### DIFF
--- a/chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
+++ b/chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
@@ -257,7 +257,7 @@ var ExtendTokenPoolLookupTable = operations.NewOperation(
 
 		var tokenAdminRegistry ccip_common.TokenAdminRegistry
 		if err := chain.GetAccountDataBorshInto(ctx, tokenAdminRegistryPDA, &tokenAdminRegistry); err != nil {
-			return sequences.OnChainOutput{}, fmt.Errorf("failed to get token admin registry at '%s': %w", tokenAdminRegistryPDA.String(), err)
+			return sequences.OnChainOutput{}, fmt.Errorf("failed to parse token admin registry account at '%s': %w", tokenAdminRegistryPDA.String(), err)
 		}
 
 		lookupTable := tokenAdminRegistry.LookupTable
--- a/chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
+++ b/chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
@@ -1,6 +1,7 @@
 package tokens
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/Masterminds/semver/v3"
@@ -8,7 +9,9 @@ import (
 	"github.com/gagliardetto/solana-go"
 	"github.com/gagliardetto/solana-go/rpc"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/deployment/utils"
+	"github.com/smartcontractkit/chainlink-ccip/chains/solana/gobindings/v0_1_1/ccip_common"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/common"
+	"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/state"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/utils/tokens"
 	tokenapi "github.com/smartcontractkit/chainlink-ccip/deployment/tokens"
 	common_utils "github.com/smartcontractkit/chainlink-ccip/deployment/utils"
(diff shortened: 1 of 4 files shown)


chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
```diff
@@ -220,6 +223,96 @@ var CreateTokenMultisig = operations.NewOperation(
...
+
+		lookupTable := tokenAdminRegistry.LookupTable
+		if lookupTable.IsZero() {
+			return sequences.OnChainOutput{}, fmt.Errorf(
+				"token pool lookup table is not set for token '%s' on router '%s'; run configure-for-transfers / set-pool first",
+				input.TokenMint.String(), input.Router.String(),
+			)
+		}
+
+		existingEntries, err := common.GetAddressLookupTable(ctx, chain.Client, lookupTable)
```
agusaldasoro: I know it's not part of this change, but I think that this improves the readability, as the method is actually retrieving the `Addresses` field from the `AddressLookupTableState`
```suggestion
		existingEntries, err := common.GetLookupTableEntries(ctx, chain.Client, lookupTable)
```

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
```diff
@@ -220,6 +223,96 @@ var CreateTokenMultisig = operations.NewOperation(
...
+		if err != nil {
+			return sequences.OnChainOutput{}, fmt.Errorf("failed to get token pool lookup table at '%s': %w", lookupTable.String(), err)
```
agusaldasoro:
```suggestion
			return sequences.OnChainOutput{}, fmt.Errorf("failed to parse addresses from lookup table state at '%s': %w", lookupTable.String(), err)
```

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
```diff
@@ -220,6 +223,96 @@ var CreateTokenMultisig = operations.NewOperation(
...
+		}
+
+		tokenAdminRegistryPDA, _, err := state.FindTokenAdminRegistryPDA(input.TokenMint, input.Router)
+		if err != nil {
+			return sequences.OnChainOutput{}, fmt.Errorf("failed to find token admin registry PDA: %w", err)
+		}
+
+		var tokenAdminRegistry ccip_common.TokenAdminRegistry
+		if err := chain.GetAccountDataBorshInto(ctx, tokenAdminRegistryPDA, &tokenAdminRegistry); err != nil {
+			return sequences.OnChainOutput{}, fmt.Errorf("failed to get token admin registry at '%s': %w", tokenAdminRegistryPDA.String(), err)
```
agusaldasoro:
```suggestion
			return sequences.OnChainOutput{}, fmt.Errorf("failed to parse token admin registry account at '%s': %w", tokenAdminRegistryPDA.String(), err)
```

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
```diff
@@ -220,6 +223,96 @@ var CreateTokenMultisig = operations.NewOperation(
...
+			)
+		}
+
+		existingEntries, err := common.GetAddressLookupTable(ctx, chain.Client, lookupTable)
+		if err != nil {
+			return sequences.OnChainOutput{}, fmt.Errorf("failed to get token pool lookup table at '%s': %w", lookupTable.String(), err)
+		}
+
+		seen := make(map[solana.PublicKey]bool, len(existingEntries))
+		for _, entry := range existingEntries {
```
agusaldasoro: I know that this already existed in code, but do we really need to check for avoid repeated entries? Is this just to make the changeset idempotent?

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
```diff
@@ -220,6 +223,96 @@ var CreateTokenMultisig = operations.NewOperation(
...
+			if acct.IsZero() {
+				return sequences.OnChainOutput{}, fmt.Errorf("account to add is zero")
+			}
+			if !seen[acct] {
+				toAdd = append(toAdd, acct)
+				seen[acct] = true
+			}
+		}
+
+		if len(toAdd) == 0 {
```
agusaldasoro, line 312: If we are doing this, we might need to add a param to skip this check. It was needed in the past.

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
```diff
@@ -220,6 +223,96 @@ var CreateTokenMultisig = operations.NewOperation(
...
+			b.Logger.Infof(
+				"all %d account(s) already present in token pool lookup table at '%s' - nothing to extend",
+				len(input.Accounts), lookupTable.String(),
+			)
+			return sequences.OnChainOutput{}, nil
+		}
+
+		b.Logger.Infof(
+			"extending token pool lookup table at '%s' with %d account(s)",
+			lookupTable.String(), len(toAdd),
```
agusaldasoro: should we also print the accounts here?

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
```diff
@@ -220,6 +223,96 @@ var CreateTokenMultisig = operations.NewOperation(
...
+		lookupTable := tokenAdminRegistry.LookupTable
+		if lookupTable.IsZero() {
+			return sequences.OnChainOutput{}, fmt.Errorf(
+				"token pool lookup table is not set for token '%s' on router '%s'; run configure-for-transfers / set-pool first",
+				input.TokenMint.String(), input.Router.String(),
+			)
+		}
+
+		existingEntries, err := common.GetAddressLookupTable(ctx, chain.Client, lookupTable)
+		if err != nil {
```
agusaldasoro, line 279: Should we check if the Authority from the `AddressLookupTableState` matches the deployer key?

agusaldasoro (dismissed):

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
chris-de-leon-cll: yep good call - added ✅

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
chris-de-leon-cll: yep it's mostly to make the operation idempotent - although if this is too much to look at then I can remove it

IIRC duplicate LUT entries are harmless but feel free to correct me if I'm wrong

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
chris-de-leon-cll: agreed - although renaming the existing method can be pretty disruptive (especially if it is being used in other repos)

to keep the diff small and maintain backwards compatibility, I added a new `GetAddressLookupTableState` function which returns the full LUT state and makes the code more readable

I also updated the existing `GetAddressLookupTable` such that it calls this new helper

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
chris-de-leon-cll: yep could help with debugging - added ✅

chains/solana/deployment/v1_6_0/operations/tokens/tokens.go
chris-de-leon-cll: sg - added ✅

agusaldasoro (approved):

tt-cll (approved):

PR #2169, opened 2026-07-02
deployment/v2_0_0: fix cross-family signer lookup

makramkd:
## Description

Related PR: https://github.com/smartcontractkit/chainlink-ccv/pull/1223

Fixes a production bug where committee verifier lane configuration failed for any lane whose source and destination chains are in different families. A Solana verifier NOP configuring a `solana -> evm` lane failed with:

```
NOP "solana-verifier-1" missing signer_address for family evm on committee "default" chain <selector>
```

Root cause: `signerAddressForNOPAlias` looks up a NOP's signer address keyed by the *destination* chain's family (`localFamily`), but JD only ever stores it keyed by the *NOP's own declared* chain family (from `chainConfig.Chain.Type`). Those coincided for every lane in production so far because every deployed NOP was EVM-only. Solana NOPs are the first to expose the mismatch.

The fix adds a translation fallback for when the exact-family lookup misses:
- `evm` and `solana` signer addresses are the same 20 bytes (secp256k1 -> keccak256), differing only in string format, so they translate freely in both directions.
- `aptos`/`stellar`/`canton` all store the same raw uncompressed public key hex, so they translate freely among themselves, and one-directionally into `evm`/`solana` (deriving an address from a public key is possible; the reverse is not, since it's a hash).
- To also support the reverse direction (e.g. `evm`/`solana` -> `canton`), JD's `OnchainSigningPubKey` field is now captured (`RawPubKeyByNOP`) and used as an additional translation source, once the paired chainlink-ccv change starts pushing it for every declared chain family, not just EVM.

The exact-match lookup path (static topology, then JD) is untouched and always takes precedence, so `evm <-> evm` lanes, the entire currently-deployed fleet, see byte-identical behavior; the new fallback code is unreachable for a NOP that already has a direct match.

## Testing

- Ran `go test ./v2_0_0/...` in the `deployment` module

### DIFF
--- a/deployment/v2_0_0/changesets/lane_signing_helpers.go
+++ b/deployment/v2_0_0/changesets/lane_signing_helpers.go
@@ -197,8 +197,8 @@ func signerAddressForNOPAlias(
 // identical bytes and differ only in string formatting, so they are freely
 // interconvertible in both directions.
 var addressClassFamilies = map[string]bool{
-	"evm":    true,
-	"solana": true,
+	chainsel.FamilyEVM:    true,
+	chainsel.FamilySolana: true,
 }
 
 // rawPubKeySourceFamily is a synthetic "family" tag used only as a translation source,
@@ -211,10 +211,10 @@ const rawPubKeySourceFamily = "xxx_notarealfamily_raw_pubkey"
 // whose signer identity is the full uncompressed secp256k1 public key, hex-encoded with
 // no per-family formatting differences. Members are interchangeable as-is.
 var rawPubKeyClassFamilies = map[string]bool{
-	"aptos":               true,
-	"stellar":             true,
-	"canton":              true,
-	rawPubKeySourceFamily: true,
+	chainsel.FamilyAptos:   true,
+	chainsel.FamilyStellar: true,
+	chainsel.FamilyCanton:  true,
+	rawPubKeySourceFamily:  true,
 }
 
 // translateFromKnownFamily tries to derive localFamily's signer address from whichever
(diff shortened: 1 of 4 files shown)


deployment/v2_0_0/changesets/lane_signing_helpers.go
```diff
@@ -158,12 +166,176 @@ func signerAddressForNOPAlias(
...
 	)
 }
 
+// addressClassFamilies are chain families whose signer identity is a 20-byte address
+// derived via secp256k1 -> keccak256 (EVM-style, e.g. ecrecover). Members encode the
+// identical bytes and differ only in string formatting, so they are freely
+// interconvertible in both directions.
+var addressClassFamilies = map[string]bool{
+	"evm":    true,
+	"solana": true,
```
jadepark-dev: nit: pretty sure it's already on your radar, just a heads up to use `chainsel.FamilyXXX`

carte7000 (dismissed):

deployment/v2_0_0/changesets/lane_signing_helpers.go
makramkd: Ah right good catch! let me switch it.

carte7000 (approved):

KodeyThomas (approved):

jadepark-dev (approved):

deployment/v2_0_0/changesets/lane_signing_helpers.go
```diff
@@ -158,12 +166,176 @@ func signerAddressForNOPAlias(
...
+	if len(b) != 20 {
+		return nil, fmt.Errorf("expected 20-byte address, got %d bytes", len(b))
+	}
+	return b, nil
+}
+
+// formatAddressForFamily renders a 20-byte address per family convention: EIP-55
+// checksummed with 0x for evm, lowercase without 0x for solana (matching the chainlink
+// node's prior art in its Solana OCR2 keyring).
+func formatAddressForFamily(addr []byte, family string) (string, error) {
```
tt-cll, line 328: ideally this is an interface method on an adapter so we don't need to add switches for every chain family in a centralized location

PR #2181, opened 2026-07-09
fix: allow migrate_chain_lanes_to_v2 to override prod-router OnRamp mappings

KodeyThomas:
## Problem

`MigrateChainLanesToV2` delegates to `ConfigureChainsForLanesFromTopology`, which hardcoded `AllowOnrampOverride: useTestRouter`. 

On the production router the `configure-chain-for-lanes` sequence therefore refuses to replace an existing Router→OnRamp mapping with a different OnRamp address. However as part of the migrate_chain_lanes_to_v2 this is something we'd like to do.

## Fix
- Add an `AllowOnrampOverride` field to `BuildLanesCrossFamilyConfig`.
- Thread it into the sequence input as `useTestRouter || allowOnrampOverride`, leaving router selection unchanged.
- Set it `true` in `MigrateChainLanesToV2` (a migration's whole purpose is to re-point the router to the new OnRamp).

`build_lanes_cross_family` leaves the flag `false`, so accidental prod-router overwrites stay blocked.

## Tests
- New `TestConfigureChainsForLanesFromTopology_AllowOnrampOverrideForcesOverrideOnProdRouter`: verifies the override is forced on the prod router without changing router selection.
- Existing `useTestRouter` true/false assertions still pass (test-router behaviour unchanged; `build_lanes` default stays `false`).

### DIFF
--- a/chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
+++ b/chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
@@ -149,6 +149,9 @@ func MigrateChainLanesToV2(
 				Lanes:      lanes,
 				MCMS:       cfg.MCMS,
 				TestRouter: cfg.TestRouter,
+				// A migration's whole purpose is to swap each lane's pre-2.0 OnRamp for the CCIP 2.0
+				// OnRamp on the (prod) router, so it must be allowed to overwrite the existing mapping.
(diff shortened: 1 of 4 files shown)


KodeyThomas:
Validated locally that both `TestRouter` and regular router paths work as expected

simsonraj (approved):

RensR (approved):

PR #2187, opened 2026-07-13
bump cldf, go-ethereum and /x/ packages

RensR:


chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -6477,7 +6477,7 @@ func TestCCIPRouter(t *testing.T) {
 			require.Equal(t, tokenPrice, tokenPriceAccount.Config.UsdPerToken.Value)
 
 			// This token isn't registered as a billing token.
-			require.True(t, common.IsClosedAccount(ctx, solanaGoClient, token2.Billing[config.EvmChainSelector], config.DefaultCommitment))
```
RensR, line 6480: Needed due to the required solana bump

deployment/utils/datastore/datastore_test.go
```diff
@@ -99,26 +99,30 @@ func TestFindAndFormatRef(t *testing.T) {
...
 
 func TestFindAndFormatFirstRef(t *testing.T) {
+	v := semver.MustParse("1.0.0")
 	// Setup datastore with multiple refs
 	ds := datastore.NewMemoryDataStore()
 	err := ds.Addresses().Add(datastore.AddressRef{
 		ChainSelector: 111,
 		Address:       "0x01",
 		Type:          datastore.ContractType("TestContract1"),
+		Version:       v,
```
RensR, line 109: Needed due to CLDF bump

KodeyThomas (dismissed):

chains/solana/contracts/tests/ccip/ccip_router_test.go
```diff
@@ -6498,7 +6498,7 @@ func TestCCIPRouter(t *testing.T) {
 				config.FqConfigPDA,
 			)
 			raw.Append(solana.Meta(wsol.fqBillingConfigPDA).WRITE())
-			raw.Append(solana.Meta(token2.Billing[config.EvmChainSelector]).WRITE())
+			raw.Append(solana.Meta(getFqTokenConfigPDA(token2.Mint)).WRITE())
```
toblich, line 6501: The proper method to use here is `getFqPerChainPerTokenConfigBillingPDA` instead, which calculates the PDA for that token _for that chain_ (where we would store fee overrides such as custom byte overheads or custom bps fees). The test works because, as the comment on line 6490 says, it's ignored altogether, but the proper PDA to use here would be the other one nonetheless.

For the other checks you've modified in this PR, your changes are correct, checking for the token config PDA (not per-chain) is a better way to guarantee that the token is not accepted for billing, so you're actually improving some old tests there :+1: So, just this line here would need to be changed.

For reference, because I know are naming here is not great:
- `getFqTokenConfigPDA(mint)` is equivalent to `token2.FeeTokenConfig` and references how we use that token in the protocol fees (using that token to pay us)
- `getFqPerChainPerTokenConfigBillingPDA(mint)` is equivalent to `token2.Billing[config.EvmChainSelector]` and references how we bill the user for transferring that token


```suggestion
			raw.Append(solana.Meta(getFqPerChainPerTokenConfigBillingPDA(token2.Mint)).WRITE())
```

toblich (dismissed):

toblich (dismissed):

0xsuryansh (dismissed):

makramkd (dismissed):

toblich (approved):

makramkd (approved):

0xsuryansh (approved):

PR #2192, opened 2026-07-15
feat: push signing key variants at JD registration; add family Signin…

stackman27:
Context: https://smartcontract-it.atlassian.net/browse/CCIP-11938


deployment/v2_0_0/offchain/shared/types.go
```diff
@@ -16,6 +16,7 @@ var ProtoChainTypeToFamily = map[nodev1.ChainType]string{
 	nodev1.ChainType_CHAIN_TYPE_SOLANA:   chainsel.FamilySolana,
 	nodev1.ChainType_CHAIN_TYPE_STARKNET: chainsel.FamilyStarknet,
 	nodev1.ChainType_CHAIN_TYPE_APTOS:    chainsel.FamilyAptos,
+	nodev1.ChainType_CHAIN_TYPE_STELLAR:  chainsel.FamilyStellar,
```
tt-cll, line 19: I don't like a global registration we need for each chain family. Definitely an anti-pattern and something we should try to move away from. Not your issue since this is currently "the standard" but something we should try to fix

tt-cll (approved):

carte7000 (approved):

PR #2193, opened 2026-07-15
[CCIP-12469] fix(commit): make observations non-blocking on stuck ops

makramkd:
## Problem

The commit plugin's `chainfee.Observation` was hanging indefinitely in production. Root cause: `asynclib.WaitForAllNoErrOperations` fans its operations out into goroutines and ends with an unconditional `wg.Wait()`. Despite setting a `timeout` on the context, however, it was _not_ being honored by the LOOPP code because other callers are calling with an unbounded context (CCIP launcher here: https://smartcontract-it.atlassian.net/browse/CCIP-12413 already fixed, but headreporter has the same issue, still not fixed https://smartcontract-it.atlassian.net/browse/CCIP-12468) which caused the plugin caller to wait on a `Lock()` that would never get unlocked.

## Solution

Replace the free function with a stateful `asynclib.Runner`:

- **Time-bounded barrier.** `WaitForAll` waits for completions *or* the timeout, whichever comes first, and proceeds with what finished. A stuck operation can no longer hold the round hostage.
- **In-flight guard.** An operation whose previous invocation is still running is skipped, not re-spawned. Goroutines are bounded to the number of distinct operations rather than accumulating one stuck goroutine per round.
- **Empty, never stale.** Results are per-round; a missing operation degrades to its zero value, and a late result from an abandoned goroutine is discarded. We prefer a visible missing price (which alerts) over silently posting a stale one.

Both commit observers (`chainfee`, `tokenprice`) now run through a reused `Runner`. This is defense-in-depth at the plugin boundary: the plugin can no longer be blocked by any downstream that ignores its context, independent of the chainlink-common connection fix and the launcher timeout fix in the companion PRs.

## Testing

- New unit tests: an operation that ignores its context still lets `WaitForAll` return within the timeout; a stuck operation is not re-spawned on the next round.
- Existing chainfee/tokenprice unit + e2e tests pass; configs that relied on the old "ignore the timeout and wait forever" behavior now set an explicit per-round timeout (production applies this via `ApplyDefaultsAndValidate`).

### DIFF
--- a/commit/chainfee/observation.go
+++ b/commit/chainfee/observation.go
@@ -36,29 +36,26 @@ func (p *processor) Observation(
 		p.obs.invalidateCaches(ctx, lggr)
 	}
 
-	var (
-		feeComponents     map[cciptypes.ChainSelector]types.ChainFeeComponents
-		nativeTokenPrices map[cciptypes.ChainSelector]cciptypes.BigInt
-		chainFeeUpdates   map[cciptypes.ChainSelector]Update
-		fChain            map[cciptypes.ChainSelector]int
-	)
-
 	operations := asynclib.AsyncNoErrOperationsMap{
-		"getChainsFeeComponents": func(ctx context.Context, l logger.Logger) {
-			feeComponents = p.obs.getChainsFeeComponents(ctx, l)
+		"getChainsFeeComponents": func(ctx context.Context, l logger.Logger) any {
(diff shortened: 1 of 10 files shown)


makramkd:
Verified on staging - commit plugin no longer blocks on a misconfigured Sui plugin.

KodeyThomas (approved):

RensR (approved):

PR #2195, opened 2026-07-15
feat: ConfigureTokenPool changeset (EVM v2.0)

futureproof1:
## What

Adds the `ConfigureTokenPool` changeset (`deployment/tokens/configure_token_pool.go`): a family-agnostic changeset that applies small, targeted, **idempotent** configuration changes to **existing** EVM v2.0.0+ token pools:

- `finalityConfig`
- `rateLimitAdmin`
- `feeAdmin`
- per-remote `tokenTransferFeeConfig` (partial)

Every field below `tokenPoolRef` is optional: absent = leave on-chain state untouched; explicit values (zero address) clear things. Every write is read-compare-write — matching on-chain state emits zero transactions (no wasted gas, no MCMS predecessor conflicts). Output is assembled via the shared `OutputBuilder`, so MCMS-owned pools produce timelock proposals and EOA-owned pools execute directly.

To fill the one capability gap, this adds an optional adapter interface `TokenPoolAdminAdapter` (+ `SetTokenPoolAdminsSequenceInput`) and a new EVM v2 sequence `SetTokenPoolAdmins`, wired on the v2 `TokenAdapter`. The sequence handles both `rateLimitAdmin` and `feeAdmin` in a single `setDynamicConfig` call, reading the current on-chain state first so unset fields are preserved.

## Scope change from review

Following review discussion and a consensus with Terry, **per-remote rate-limit bucket configuration was removed from this changeset's scope**:

- Dropped per-remote `rateLimits` (bucket values / scaling).
- Rationale: updating rate limits without the traditional 10% scaling is a misconfiguration foot-gun, and AI-generated YAMLs from historical runs risk bypassing the rate-limit safety guarantees other changesets enforce. Rate-limit value management stays in the dedicated rate-limit tooling.
- `rateLimitAdmin` was **retained** (CCIP-12132): the rate limit admin address is an administrative role update, not a rate-limit value update, and carries none of the foot-gun risk. It is managed alongside `feeAdmin` via `TokenPoolAdminAdapter` / `SetTokenPoolAdmins` / `SetTokenPoolAdminsSequenceInput`.

## Other review changes

- Duplicate-pool detection in `verify` now uses a struct key built from `datastore.AddressRefKey` (`ref.Key()`) instead of a `|`-delimited string. (`AddressRef` embeds a non-comparable `LabelSet`, so `.Key()` is required to make it map-key-safe.)
- `RateLimiterConfig.Equal` was migrated from a standalone package-level function (`rateLimiterConfigsEqual`) to a method on `RateLimiterConfig` in `product.go`.

## PR breakdown

This is **PR#1 of 3**:

- **PR#1 (this):** scaffolding + validation + apply orchestration, EVM v2.0.0+ support (finality, rate limit admin, fee admin, fee config).
- **PR#2:** EVM 1.5.x / 1.6.x (wire the existing pre-v2 fee path; capability gating is already in place).
- **PR#3:** Solana 1.6.x.

Verify is structural-only (no datastore/on-chain resolution): it checks selectors, rejects empty or duplicate entries, and validates remote fee-config fields. Version gating and adapter resolution happen in apply via `ResolveAdapterAndRefs`, which surfaces unresolvable refs and unsupported versions with clear errors before any writes execute. Because every write is read-compare-write (idempotent), surfacing these errors in apply rather than verify carries no consistency risk.

## Deliberate design choices

- **`ResolveAdapterAndRefs` is called unconditionally in apply**, even for entries that only touch `finalityConfig` or admin roles (no `Remotes`, so `fullTokenRef` is never consumed). This is intentional: a pool whose token cannot be resolved from the chain is not a pool this changeset should be writing to. The call acts as implicit pre-flight validation — if the pool is misconfigured or orphaned, apply fails fast before any on-chain write occurs. For EVM, the token ref is always resolvable on-chain (ERC20 symbol/decimals) regardless of datastore state, so this adds no datastore restriction.

## Deliberate divergences from sibling changesets

- **No counterpart / symmetry validation.** Each entry configures one pool's local view of a lane and never requires a counterpart section. Bidirectional use is supported (you may configure both sides in one changeset) but never required, and no symmetry is enforced — this is the changeset's core differentiator.
- **Fee config merges with current on-chain state, not defaults.** A partial fee update (`PartialTokenTransferFeeConfig.MergeWith` over the on-chain config) never resets unrelated fields when the lane is already enabled — intentionally different from the token-expansion flow, which merges with chain-agnostic defaults for a fresh lane.

## Testing

`integration-tests/deployment/configure_token_pool_test.go`:

- Verify-precondition table tests (duplicate/empty entries, structural violations — bad selectors, invalid remote selectors, malformed fee-config fields).
- Single-feature apply + idempotency for finality, fee admin, rate limit admin, and fee config (on-chain merge semantics).
- A combined multi-feature bidirectional apply, and an MCMS-owned-pool test (timelock proposal produced, executed, then a fresh-bundle re-apply emits zero proposals).

Local verification green: `go build` + `go vet` across the `deployment`, `chains/evm`, and `integration-tests` modules; `chains/evm` v2_0_0 sequences/tokens suite; the `deployment/tokens` package; and the `TestConfigureTokenPool*` suite. No `core ref` needed — the change is additive and Chainlink core compiles against it.

## Notes for reviewers

- `new(expr)` (e.g. `new("0x123...")`) is valid Go 1.26 — the `new` built-in now accepts a value argument and infers the type, returning a pointer to that value. It replaces the common `ptrTo[T any](v T) *T` helper pattern.
- `TokenTransferFeeConfig` on a `RemoteConfigUpdate` is optional (nil = skip fee config for that remote, matching how `FinalityConfig` is treated at the pool level). A non-nil but fully-unset `PartialTokenTransferFeeConfig{}` is rejected in verify since an all-zero merge would silently re-enable fees on a disabled lane.
- `isEnabled` is required whenever `TokenTransferFeeConfig` is provided. Without it, a partial update on a disabled lane would fall back to chain-agnostic defaults (which have `isEnabled=true`) and unintentionally enable the lane. Requiring the field makes the intent explicit and eliminates the ambiguity.

## Ticket

[TOOLING]: add changeset to update token pool configs
Reference: #1931


deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+	derivedTokenAddr, err := adapter.DeriveTokenAddress(e, sel, fullPoolRef)
+	if err != nil {
+		return nil, "", datastore.AddressRef{}, datastore.AddressRef{}, fmt.Errorf("failed to derive token address from pool %s: %w", fullPoolRef.Address, err)
+	}
+	fullTokenRef, err := ResolveTokenRef(e, reg, sel, datastore.AddressRef{ChainSelector: sel, Address: derivedTokenAddr})
+	if err != nil {
+		return nil, "", datastore.AddressRef{}, datastore.AddressRef{}, fmt.Errorf("failed to resolve token ref for chain selector %d: %w", sel, err)
+	}
+	return adapter, family, fullPoolRef, fullTokenRef, nil
+}
```
chris-de-leon-cll: Good news: this is already implemented and can be safely removed in favor of `ResolveAdapterAndRefs` https://github.com/smartcontractkit/chainlink-ccip/blob/65b95a0d0a80f30510fe5e2bd36ee931ea766992/deployment/tokens/token_expansion.go#L504-L536

```suggestion
```

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+				if pool.Version != nil {
+					fullPoolRef.Version = pool.Version
+				}
+				// PR#1 scope: EVM v2.0.0+ pools only. PR#2 adds EVM 1.5.x/1.6.x, PR#3 adds Solana.
+				if fullPoolRef.Version != nil && fullPoolRef.Version.LessThan(utils.Version_2_0_0) {
+					return fmt.Errorf(
+						"pool %s on chain selector %d has version %s: only v2.0.0+ pools are currently supported by this changeset",
+						fullPoolRef.Address, chainCfg.ChainSelector, fullPoolRef.Version,
+					)
+				}
```
chris-de-leon-cll: This can be removed bc ResolveAdapter already validates the version constraints

```suggestion
```

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+		// pre-v2 pool yields the scope error even when no pre-v2 adapter is registered.
+		for _, chainCfg := range cfg.Chains {
+			for _, pool := range chainCfg.Pools {
+				fullPoolRef, err := ResolveTokenPoolRef(e, registry, chainCfg.ChainSelector, pool.TokenPoolRef)
+				if err != nil {
+					return fmt.Errorf("chain selector %d: failed to resolve token pool ref: %w", chainCfg.ChainSelector, err)
+				}
+				if pool.Version != nil {
+					fullPoolRef.Version = pool.Version
+				}
```
chris-de-leon-cll: This can also be removed since `ResolveTokenPoolRef` handles it

```suggestion
```

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+	return func(e cldf.Environment, cfg ConfigureTokenPoolInput) error {
+		if len(cfg.Chains) == 0 {
+			return errors.New("input must contain at least one chain entry")
+		}
+		registry := GetTokenAdapterRegistry()
+
+		// First pass: purely structural checks (no datastore/on-chain resolution). Obvious
+		// input mistakes — bad selectors, empty updates, duplicate entries, malformed rate
+		// limits — must surface before any resolution error from a later pool would mask them.
+		seenPools := make(map[string]struct{})
```
chris-de-leon-cll: I'd recommend using a small struct as the key to avoid issues with using `|` as a delimiter

```suggestion
		type chainConfigKey struct {
    		poolRefr datastore.AddressRef
			selector uint64
		}
		seenPools := make(map[chainConfigKey]struct{})
```

Then later:

```go
key := chainConfigKey{poolRefr:pool.TokenPoolRef,selector:chainCfg.ChainSelector}
if _, dup := seenPools[key]; dup {
	return fmt.Errorf("duplicate pool entry: %+v", key)
}
```

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+		}
+	}
+	if family == chain_selectors.FamilyEVM {
+		if pool.RateLimitAdmin != nil && !common.IsHexAddress(*pool.RateLimitAdmin) {
+			return fmt.Errorf("invalid rateLimitAdmin address %q for pool on chain selector %d", *pool.RateLimitAdmin, chainSelector)
+		}
+		if pool.FeeAdmin != nil && !common.IsHexAddress(*pool.FeeAdmin) {
+			return fmt.Errorf("invalid feeAdmin address %q for pool on chain selector %d", *pool.FeeAdmin, chainSelector)
+		}
+	}
```
chris-de-leon-cll: For context, the top level changeset needs to be chain-agnostic, so anything that references a specific chain family should either be removed from here or pushed down to a lower level. In this case, I would remove it from here, and place it directly in the EVM sequence.

```suggestion
```

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+			Settings: map[string]map[uint64]*TokenTransferFeeConfig{
+				fullPoolRef.Address: {dst: &requestedConfig},
+			},
+		},
+	)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to execute set token transfer fee sequence for chain selector %d and remote chain selector %d: %w", src, dst, err)
+	}
+	return result.Output.BatchOps, result.ExecutionReports, nil
+}
```
chris-de-leon-cll: this looks oddly familiar 😉 https://github.com/smartcontractkit/chainlink-ccip/blob/65b95a0d0a80f30510fe5e2bd36ee931ea766992/deployment/tokens/configure_tokens_for_transfers.go#L442-L504

in fact, might I suggest: https://github.com/smartcontractkit/chainlink-ccip/blob/65b95a0d0a80f30510fe5e2bd36ee931ea766992/deployment/tokens/configure_tokens_for_transfers.go#L421-L427 (this handles all of v1.5.x, v1.6.x, and v.2.x automatically)

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+		bCap = zero
+	}
+	if aRate == nil {
+		aRate = zero
+	}
+	if bRate == nil {
+		bRate = zero
+	}
+	return aCap.Cmp(bCap) == 0 && aRate.Cmp(bRate) == 0
+}
```
chris-de-leon-cll: I'd recommend attaching this as a method on the `RateLimiterConfig` struct itself in `product.go`:

```go
// RateLimiterConfig specifies configuration for a rate limiter on a token pool.
type RateLimiterConfig struct {
	// IsEnabled specifies whether the rate limiter should be enabled.
	IsEnabled bool
	// Capacity is the maximum number of tokens that can be in a rate limiter bucket.
	Capacity *big.Int
	// Rate is the rate at which the rate limiter bucket refills, in tokens per second.
	Rate *big.Int
}

// Equal compares two RateLimiterConfigs treating nil big.Ints as zero.
func (a RateLimiterConfig) Equal(b RateLimiterConfig) bool {
	if a.IsEnabled != b.IsEnabled {
		return false
	}

	zero, aCap, bCap, aRate, bRate := big.NewInt(0), a.Capacity, b.Capacity, a.Rate, b.Rate
	if aCap == nil {
		aCap = zero
	}
	if bCap == nil {
		bCap = zero
	}
	if aRate == nil {
		aRate = zero
	}
	if bRate == nil {
		bRate = zero
	}

	return aCap.Cmp(bCap) == 0 && aRate.Cmp(bRate) == 0
}
```

Then call sites become more idiomatic and readable:

```go
if current.Outbound.Equal(outbound) && current.Inbound.Equal(inbound) {
	e.Logger.Infof("Skipping rate limit bucket (fastFinality=%t) for chain %d remote %d since the desired config is the same as the current on-chain config", in.FastFinality, selector, remoteSelector)
	continue
}
```

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,512 @@
...
+
+// PoolConfigUpdate describes a partial configuration update for a single token pool.
+// Every field other than TokenPoolRef is optional: absent fields leave on-chain state
+// untouched. To clear a value, provide it explicitly (e.g. the zero address).
+type PoolConfigUpdate struct {
+	// TokenPoolRef is a reference to the token pool in the datastore.
+	TokenPoolRef datastore.AddressRef `yaml:"tokenPoolRef" json:"tokenPoolRef"`
+	// Version overrides the pool version resolved from the datastore ref. Only needed when
+	// the datastore entry is missing or ambiguous.
+	Version *semver.Version `yaml:"version,omitempty" json:"version,omitempty"`
```
chris-de-leon-cll: I think we can safely remove this to make the YAML simpler

```suggestion
```

deployment/tokens/configure_token_pool.go
futureproof1: Hey @chris-de-leon-cll  - `ResolveAdapter` doesn't enforce a v2 floor — `newTokenAdapterID` is an exact version match, and devenv registers the pre-v2 adapters (v1.5.1, v1.6.0/v1.6.1). They embed `EVMPoolAdapter`, so they satisfy `RateLimitReaderAdapter` and the changeset proceeds on a pre-v2 pool rather than being rejected.

The risk is silent: `applyRateLimitsForRemote` passes localDecimals as both the local and remote arg to `GenerateTPRLConfigs`. For pre-1.6.1 pools `DoesPoolUseLocalDecimals` is false, so inbound should scale by the remote token's decimals — but this changeset never resolves those, so inbound gets scaled by local decimals instead. When the remote and local token decimals happen to match (e.g. both 18) it's coincidentally correct; when they differ, the inbound limits are silently wrong. That's the NOTE in `applyRateLimitsForRemote`.

So the gate is guarding correctness, not just error quality. I'd keep it as the PR#1 scope guard and drop it in PR#2 alongside the remote-decimals handling that makes pre-v2 actually safe.

deployment/tokens/configure_token_pool.go
futureproof1: @chris-de-leon-cll 
`datastore.AddressRef` isn't comparable — it embeds LabelSet, which holds a map[string]struct{}, and a struct with a map field can't be a map key. So `map[chainConfigKey]struct{}` is a compile error (invalid map key type), which is why I went with the string key. I can switch to a struct of just the comparable fields (selector + address + qualifier + type) if you'd prefer that over the |-delimited string — that compiles fine since it leaves LabelSet out.
What do you think?

chris-de-leon-cll: yep you're correct, my example should have used `AddressRefKey` instead - nice catch!

```suggestion
		type chainConfigKey struct {
			poolRefr datastore.AddressRefKey
			selector uint64
		}

		// First pass: purely structural checks (no datastore/on-chain resolution). Obvious
		// input mistakes — bad selectors, empty updates, duplicate entries, malformed rate
		// limits — must surface before any resolution error from a later pool would mask them.
		seenPools := make(map[chainConfigKey]struct{})
		for _, chainCfg := range cfg.Chains {
			if _, err := chain_selectors.GetSelectorFamily(chainCfg.ChainSelector); err != nil {
				return fmt.Errorf("invalid chain selector %d: %w", chainCfg.ChainSelector, err)
			}
			if len(chainCfg.Pools) == 0 {
				return fmt.Errorf("no pools provided for chain selector %d", chainCfg.ChainSelector)
			}
			for _, pool := range chainCfg.Pools {
				if err := validatePoolConfigUpdate(chainCfg.ChainSelector, pool); err != nil {
					return err
				}
				key := chainConfigKey{poolRefr: pool.TokenPoolRef.Key(), selector: chainCfg.ChainSelector}
				if _, dup := seenPools[key]; dup {
					return fmt.Errorf("duplicate pool entry for chain selector %d and ref %s", chainCfg.ChainSelector, datastore_utils.SprintRef(pool.TokenPoolRef))
				}
				seenPools[key] = struct{}{}
			}
		}
```

deployment/tokens/configure_token_pool.go
chris-de-leon-cll: @futureproof1 quick heads up - I met with Terry earlier today, and we came to the consensus that we should drop `rateLimit` as being part of the scope for this new changeset

While having an option to update rate limits without the traditional 10% scaling could be useful, the main gripe with it is that it opens up a real risk of being a common misconfiguration foot gun

Also, there have been instances in the past where folks use AI to generate YAMLs from historical runs, so if they're not careful, then operators may unintentionally bypass some of the rate limit safety guarantees that are present in other changesets and misconfigure things on-chain

With that said, feel free to remove any type of rate limit code from this PR - I'll update the ticket to reflect the new YAML and scope 🙏

chris-de-leon-cll:
@futureproof1 only remaining things for this PR are addressing the two follow-ups above

Once those are addressed, Terry and I can approve - overall nice work on this 💯 

When you're ready, feel free to proceed with the next PR!

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,265 @@
...
+func applyPoolConfigUpdate(
+	e cldf.Environment,
+	registry *TokenAdapterRegistry,
+	selector uint64,
+	pool PoolConfigUpdate,
+) ([]mcms_types.BatchOperation, []cldf_ops.Report[any, any], error) {
+	adapter, family, fullPoolRef, fullTokenRef, err := ResolveAdapterAndRefs(e, registry, selector, pool.TokenPoolRef, datastore.AddressRef{})
+	if err != nil {
+		return nil, nil, err
+	}
```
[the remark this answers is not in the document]
chris-de-leon-cll: this uses `ExecuteOperation` so results are cached

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,201 @@
...
+					}
+				}
+				key := chainConfigKey{poolRef: pool.TokenPoolRef.Key(), selector: chainCfg.ChainSelector}
+				if _, dup := seenPools[key]; dup {
+					return fmt.Errorf("duplicate pool entry for chain selector %d and ref %s", chainCfg.ChainSelector, datastore_utils.SprintRef(pool.TokenPoolRef))
+				}
+				seenPools[key] = struct{}{}
+			}
+		}
+		return nil
```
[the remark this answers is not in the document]
chris-de-leon-cll: intentionally dropped - already covered by apply()

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,209 @@
...
+	// Chains lists the per-chain pool configuration updates.
+	Chains []ConfigureTokenPoolPerChain `yaml:"chains" json:"chains"`
+	// MCMS configures the resulting proposal.
+	MCMS mcms.Input `yaml:"mcms,omitempty" json:"mcms"`
+}
+
+// ConfigureTokenPoolPerChain groups pool updates for a single chain.
+type ConfigureTokenPoolPerChain struct {
+	// ChainSelector identifies the chain on which the pools live.
+	ChainSelector uint64 `yaml:"selector,string" json:"selector,string"`
```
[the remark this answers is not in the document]
chris-de-leon-cll: this is harmless and keeps it consistent with the json tag

deployment/tokens/configure_token_pool.go
```diff
@@ -0,0 +1,214 @@
...
+					if remote.TokenTransferFeeConfig != nil {
+						if remote.TokenTransferFeeConfig.IsEmpty() {
+							return fmt.Errorf("remote entry %d for pool on chain selector %d has nothing to update", remote.RemoteChainSelector, chainCfg.ChainSelector)
+						}
+						if !remote.TokenTransferFeeConfig.IsEnabled.IsPresent() {
+							return fmt.Errorf("remote entry %d for pool on chain selector %d must specify isEnabled", remote.RemoteChainSelector, chainCfg.ChainSelector)
+						}
+					}
+				}
+				normalizedRef, err := deploy.TryNormalizeAddressRef(chainCfg.ChainSelector, pool.TokenPoolRef)
```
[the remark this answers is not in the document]
chris-de-leon-cll: this is okay - not necessarily a huge issue that's worth adding more code for

chris-de-leon-cll (dismissed):

chris-de-leon-cll (dismissed):

chris-de-leon-cll (approved):

tt-cll (approved):

PR #2211, opened 2026-07-28
feat: ConfigureTokenPool support for EVM 1.5.1/1.6.x pools

futureproof1:
## Summary

PR#2 of the `ConfigureTokenPool` changeset work. Extends the existing changeset (`deployment/tokens/configure_token_pool.go`, added in #2195 for EVM v2.0.0+ pools) to also work against EVM 1.5.1/1.6.0/1.6.1 token pools, with no changes to the changeset itself.

- Added `SetTokenPoolAdmins()` to the shared `EVMPoolAdapter` base (`chains/evm/deployment/v1_0_0/adapters/pool_adapter.go`), inherited by `v1_5_1.TokenAdapter`/`v1_6_1.TokenAdapter` via embedding. Supports idempotent `RateLimitAdmin` updates (read-before-write, no-op on match) and rejects `FeeAdmin` in-sequence before any on-chain read, since pre-v2 contracts have no fee-admin concept.
- Confirmed `tokenTransferFeeConfig` already dispatches correctly cross-version via the pre-existing `applyTokenTransferFeeConfig` helper — test-only coverage added, zero production changes.
- Added regression coverage confirming `finalityConfig` and pool-level `feeAdmin` continue to fail cleanly on pre-v2 pools.

Out of scope (per PR#1 review decision, carried forward): rate-limit *value* configuration (bucket scaling), and Solana v1.6.x support (tracked separately).

**Known follow-up (not in this PR):** extending coverage to 1.6.0/1.6.2 was attempted but blocked by a pre-existing gap in `chains/evm/deployment/v1_6_0/sequences/deploy_token_pool_contracts.go` — the test-deploy dispatch only supports `BurnMintTokenPool` at exactly version 1.6.1, even though the configure-side adapter registry already treats 1.6.0/1.6.1/1.6.2 identically. Fixing that is production code and out of scope here.

## Test plan

- [x] `cd chains/evm && go build ./...`
- [x] `cd integration-tests && go build ./...`
- [x] `TestConfigureTokenPool_Admins_PreV2` (v1_5_1, v1_6_1) — rate-limit-admin set, idempotency, combined-update rejection
- [x] `TestConfigureTokenPool_FeeConfig_PreV2` (v1_5_1, v1_6_1) — fee-quoter path, idempotency
- [x] `TestConfigureTokenPool_UnsupportedFields_PreV2` (v1_5_1, v1_6_1) — finalityConfig/feeAdmin rejection, independent subtests
- [x] All pre-existing `TestConfigureTokenPool*` tests still pass (no regressions)


chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
chris-de-leon-cll, line 1: oh very interesting - this function looks nearly identical to the one implemented in the last PR:

https://github.com/smartcontractkit/chainlink-ccip/blob/4f966ecacc230d5c15cf7b05933cd5f6d92b6a3a/chains/evm/deployment/v2_0_0/sequences/tokens/set_token_pool_admins.go#L1

Is it possible to delete the v2.0.0 file above in favor of the v1.0.0 implementation you have in this PR? I think the v2.0.0 token adapter should automatically inherit the `EVMPoolAdapter` method so you'll be able to keep all EVM versions consistent

futureproof1: @chris-de-leon-cll 
Yea, they're definitely related twins :-) , I understand that v2 would automatically inherit this method if we deleted its own version. We can join them, but it needs a small refactor rather than just deleting the v2 file.

The two pool generations don't have the same functions:
 - pre-2.0 pools have one admin: rateLimitAdmin
 - v2.0 pools have two: rateLimitAdmin and feeAdmin, and the contract sets them together in a single setDynamicConfig tx

The shared method in this PR only knows about the one function, and it deliberately errors if you pass a feeAdmin — because on the older contracts, that thing doesn't exist. So if v2 inherited it as-is, v2 pools would silently lose the ability to set a fee admin, and the feeAdmin tests from the last PR would start failing.

To unify properly, the shared method needs to stop hard-rejecting feeAdmin itself and instead hand the "set the admins" step down to each version's PoolOps — pre-2.0 rejects feeAdmin there, v2 sets both in one tx. That's a PoolOps interface change plus reworking both methods, not a file deletion.

Happy to do it — would you prefer that refactor in this PR, or as a fast-follow?

chris-de-leon-cll: Thx for catching this! I followed up in Slack with a plan to address this 🙏

futureproof1: thanks claude 😆

futureproof1: changes pushed, can you have another look @chris-de-leon-cll

chris-de-leon-cll (approved):

tt-cll (approved):

PR #2215, opened 2026-07-30
Now call curse sequence concurrently per selector

carte7000:
Now call curse sequence concurrently per selector


RensR (dismissed):

tt-cll (dismissed):
do we want to cap the number of threads we spawn? I feel like we've tried to parallelize in CLD before and had resource issues for certain large batches

tt-cll (approved):

flodesi (approved):

PR #2218, opened 2026-07-31
feat: Solana 1.6 pool support in ConfigureTokenPool

futureproof1:
## What

Makes `ConfigureTokenPool`'s `rateLimitAdmin` field work for Solana 1.6 BurnMint and LockRelease token pools, and adds integration coverage proving the unsupported fields reject loudly and the `tokenTransferFeeConfig` path works for Solana.

Follows #2211, which did the equivalent EVM work.

`SolanaAdapter` joins the existing chain-agnostic `TokenPoolAdminAdapter` seam by type assertion — no new interface. Because a Solana pool address is a *program ID* shared across many token mints, `SetTokenPoolAdminsSequenceInput` gains `TokenPoolRef` and `TokenRef`, which the changeset already resolved and previously discarded. Both adapters read the pool address from `TokenPoolRef.Address` — `PoolAddress` was redundant and dropped in commit 5 (see below). The EVM adapter changes are trivial: `input.PoolAddress` → `input.TokenPoolRef.Address`.

## Commits

1. **`fix: correct idempotency and error handling in Solana rate limit admin ops`** — two defects in `UpdateRateLimitAdmin{BurnMint,LockRelease}`, which are already live in the Solana pool deploy path:
   - **Swallowed read error.** `if err == nil { compare }` made a transient RPC or decode failure indistinguishable from "account absent", and both fell through to writing — degrading the no-op guarantee to "always write". Now only `rpc.ErrNotFound` falls through; every other error is returned.
   - **Bogus comparison in the fallback.** `if authority == input.NewOwner { return nil }` compared the *program upgrade authority* against the *desired rate limit admin*. When the deployer key was both — normal in tests and single-operator deployments — the op reported success having done nothing.
2. **`feat: support rateLimitAdmin on Solana 1.6 token pools in ConfigureTokenPool`** — widens the sequence input, passes the already-resolved refs at the call site, and implements `SetTokenPoolAdmins()` on `SolanaAdapter`. `feeAdmin` is rejected (Solana pools have no fee admin concept); a nil `rateLimitAdmin` is a no-op.
3. **`test: cover Solana rejection paths and fee config in ConfigureTokenPool`** — `feeAdmin` and `finalityConfig` rejection, plus the `tokenTransferFeeConfig` path, which routes Solana pools to the FeeQuoter because their version is below 2.0.0. That path was believed to work but nothing verified it; it does, and the test reads the config back on-chain through the same resolution path the changeset uses.
4. **`fix: address review findings on the Solana rate limit admin path`** — see below.
5. **`refactor: drop redundant PoolAddress from SetTokenPoolAdminsSequenceInput`** — `PoolAddress` always carried the same value as `TokenPoolRef.Address`. Removed. EVM adapter switched from `input.PoolAddress` to `input.TokenPoolRef.Address`.

## Two things reviewers should know

**Solana callers must pass the pool config PDA as `TokenPoolRef`, not the pool program ID.** The generic `ResolveTokenPoolRef` does an exact-match datastore lookup by address *before* the family-specific resolver runs, and a bare Solana pool program ID is exactly what the datastore registers — so the lookup short-circuits and the Solana PDA-normalizing resolver never runs, and token derivation then fails with "token derivation is only possible if a pool PDA is provided". This is inherent rather than a workaround: a program ID is shared across mints and so cannot identify a token. The requirement is now documented on `PoolConfigUpdate.TokenPoolRef`, where it previously lived only in a test comment.

**Widening `SetTokenPoolAdminsSequenceInput` changes its report hash.** Any durable or resumable report store will no longer match previously-recorded `SetTokenPoolAdmins` reports and will re-execute them. That is safe — the operation reads on-chain state and no-ops before writing — but it is worth knowing.

## Review fixes folded into commit 4

- The **same swallowed-error pattern** fixed in commit 1 still existed six lines below it: any `GetAuthority*` failure, not just "account absent", fell back to the program upgrade authority, so a transient read failure could submit `SetRateLimitAdmin` signed by the wrong authority. Now reuses the pool config read the op already performed — which also removes a redundant RPC round-trip per pool — and keeps the upgrade-authority fallback for the account-absent case only. That fallback is load-bearing: under MCMS the initialize op only queues a proposal, so the pool config account genuinely does not exist yet when this op runs.
- A **Solana idempotency assertion that verified nothing.** The re-apply reused the same operations bundle with byte-identical input, so `ExecuteSequence`'s report memoization returned the first apply's cached report without running the handler — the read-compare no-op path was never exercised. Now refreshes the reporter first, matching the idiom every pre-existing EVM test in that file already uses.
- Dropped no-op `,omitempty` tags from the two struct-typed input fields (`encoding/json` never omits struct values), wrapped dropped address-parse errors with `%w`, and asserted the third fee-config value (`MinFeeUSDCents`) actually landed on-chain.

## Deliberately not changed

`UpdateRateLimitAdmin{BurnMint,LockRelease}` keep their signatures — the input type is shared with the transfer-ownership and accept-ownership ops, and these are exported vars in a public module, so a new input type would be source-breaking downstream. That is why `NewOwner` carries the desired rate limit admin, which reads oddly but is intended. The ops also keep decoding pool state as `test_token_pool.State`, which has the same wire layout.

## Testing

- `TestConfigureTokenPool` — 12 subtests pass, including every pre-existing EVM test
- `TestTokensAndTokenPools` — passes; this is the regression harness for the operation changes, and it deploys Solana BurnMint and LockRelease pools through MCMS
- `go build` and `go vet` clean in `chains/solana`, `chains/solana/deployment`, `deployment`, `integration-tests`, and `chains/evm`

## Follow-ups, not blockers

- The sequence's cheap validation branches (`feeAdmin` rejection, nil-admin no-op, unknown selector) all run before the chain lookup and are trivially unit-testable without a Solana container.
- `TestConfigureTokenPool_FeeConfig_Solana` and `TestConfigureTokenPool_UnsupportedFields_Solana` each stand up their own Solana container; they are read-mostly and could share one environment.

### DIFF
--- a/chains/solana/deployment/v1_6_0/operations/token_pools/burnmint.go
+++ b/chains/solana/deployment/v1_6_0/operations/token_pools/burnmint.go
@@ -570,11 +570,13 @@ var UpdateRateLimitAdminBurnMint = operations.NewOperation(
 			b.Logger.Info("Rate limit admin already matches the desired value for burn mint token pool with token mint:", input.TokenMint.String())
 			return sequences.OnChainOutput{}, nil
 		}
-		authority, err := GetAuthorityBurnMint(chain, input.Program, input.TokenMint)
+		// err here is the error from the GetAccountDataBorshInto read above, which has already
+		// returned for every error except rpc.ErrNotFound — so err != nil at this point means
+		// exactly "account absent" (the pool config account may legitimately not exist yet:
+		// under MCMS the initialize op only queues a proposal, so this op must still build its
+		// instruction). Fall back to the program upgrade authority in that case.
+		authority := chainConfig.Config.Owner
 		if err != nil {
-			// The pool config account may legitimately not exist yet: under MCMS the initialize
-			// op only queues a proposal, so this op must still build its instruction. Fall back to
-			// the program upgrade authority in that case.
 			authority, err = utils.GetUpgradeAuthority(chain.Client, input.Program)
 			if err != nil {
 				return sequences.OnChainOutput{}, fmt.Errorf("failed to get upgrade authority for burn mint token pool: %w", err)
--- a/chains/solana/deployment/v1_6_0/operations/token_pools/burnmint.go
+++ b/chains/solana/deployment/v1_6_0/operations/token_pools/burnmint.go
@@ -3,9 +3,11 @@ package token_pools
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/gagliardetto/solana-go"
+	"github.com/gagliardetto/solana-go/rpc"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/deployment/utils"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/gobindings/v0_1_1/base_token_pool"
 	"github.com/smartcontractkit/chainlink-ccip/chains/solana/gobindings/v0_1_1/burnmint_token_pool"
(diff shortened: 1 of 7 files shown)


chains/solana/deployment/v1_6_0/operations/token_pools/burnmint.go
```diff
@@ -556,28 +558,29 @@ var UpdateRateLimitAdminBurnMint = operations.NewOperation(
...
 			if err != nil {
 				return sequences.OnChainOutput{}, fmt.Errorf("failed to get upgrade authority for burn mint token pool: %w", err)
 			}
-			// if we had to assume the authority, then the pool isn't initialized yet and therefore
-			// there won't be an authority set on-chain yet, so we can skip the update since the initializer
-			// will be able to set the correct authority during initialization
-			if authority == input.NewOwner {
-				b.Logger.Info("New owner is the same as the current owner for burn mint token pool with token mint:", input.TokenMint.String())
-				return sequences.OnChainOutput{}, nil
-			}
```
chris-de-leon-cll, line 580: [no action needed]

Nice catch here! Looks like this original code was buggy and can be safely removed. It seems like this check was asking: "Is the person who controls the program the same person we want to make the rate limit admin?" If yes, then skip. Don't set anything.

That's incorrect. The operation should ALWAYS set the rate limit admin. Whether the upgrade authority happens to be the same person or a different person is irrelevant.

deployment/tokens/product.go
chris-de-leon-cll, line 49: Ah looks like `PoolAddress` is redundant with `TokenPoolRef.Address` in `SetTokenPoolAdminsSequenceInput`, so we can safely remove it

**Rationale:**

- The only production call site (`configure_token_pool.go:185`) sets `PoolAddress: fullPoolRef.Address` — the same value that `TokenPoolRef.Address` already carries after resolution.
- EVM's `SetTokenPoolAdmins` (`chains/evm/deployment/v1_0_0/adapters/pool_adapter.go:210`) reads `input.PoolAddress`; switching to `input.TokenPoolRef.Address` is a one-line change.
- Solana never reads `input.PoolAddress` at all — it uses `input.TokenPoolRef.Type` and `input.TokenRef.Address`.

**Changes required:**

| File | Change |
|------|--------|
| `deployment/tokens/product.go` | Remove `PoolAddress` from `SetTokenPoolAdminsSequenceInput` |
| `chains/evm/deployment/v1_0_0/adapters/pool_adapter.go` | Line 210-213: `input.PoolAddress` → `input.TokenPoolRef.Address` |
| `deployment/tokens/configure_token_pool.go` | Line 187: remove `PoolAddress:` line |

chris-de-leon-cll (commented):
Overall looks good - just had one comment about simplifying the sequence input, so once that's resolved we should be good to merge!

deployment/tokens/product.go
futureproof1: `PoolAddress` is gone and both adapters now read `input.TokenPoolRef.Address`.

One other thing — Solana did read `input.PoolAddress`, at four pats in `chains/solana/deployment/v1_6_0/sequences/tokens.go`: `line :640` parsed it into the pool program ID passed to the operation, and `line :614`,`line :642` and `line :655`. So the removal touched that file too, otherwise it wouldn't compile.

chris-de-leon-cll (approved):

tt-cll (approved):

PR #2231, opened 2026-08-07
fix: wire OnRamp RMN via proxy, stop RMNProxy downgrade

RensR:
fix two nil derefs

### DIFF
--- a/chains/evm/deployment/v2_0_0/sequences/deploy_chain_contracts.go
+++ b/chains/evm/deployment/v2_0_0/sequences/deploy_chain_contracts.go
@@ -83,12 +83,6 @@ type MockReceiverParams struct {
 	Qualifier string
 }
 
-// RMNParams configures the RMN 2.1.0 deployment. Curse admins are not configurable here: the Ultra
-// Fast Curse RBACTimelock is resolved from ExistingAddresses and is always the curse admin.
-type RMNParams struct {
-	Version *semver.Version
-}
-
 type OffRampParams struct {
 	Version                   *semver.Version
 	GasForCallExactCheck      uint16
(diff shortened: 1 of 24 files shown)


chains/evm/deployment/v2_0_0/sequences/deploy_chain_contracts.go
```diff
@@ -183,27 +185,27 @@ var DeployChainContracts = cldf_ops.NewSequence(
...
+				CurseAdmins: input.ContractParams.RMN.CurseAdmins,
 			},
 		}, input.ExistingAddresses)
 		if err != nil {
 			return output, err
 		}
-		addresses = append(addresses, rmnRemoteRef)
-		ownableContracts = append(ownableContracts, rmnRemoteRef)
+		addresses = append(addresses, rmnRef)
+		ownableContracts = append(ownableContracts, rmnRef)
```
carte7000, line 202: When I was looking at this I had the impression that it would transfer ownership to CLLCCIP by default. More or less relevant in the immediate term but for new chains that would be a problem

chains/evm/deployment/v2_0_0/sequences/deploy_chain_contracts.go
```diff
@@ -83,9 +83,11 @@ type MockReceiverParams struct {
...
 }
 
-type RMNRemoteParams struct {
-	Version   *semver.Version
-	LegacyRMN common.Address
+// RMNParams configures the RMN 2.1.0 deployment. CurseAdmins are authorized to curse in addition
+// to the owner; the Ultra Fast Curse RBACTimelock is normally included here.
+type RMNParams struct {
+	Version     *semver.Version
+	CurseAdmins []common.Address
```
carte7000: Do we need to feed in the UFC MCM here? I was thinking we could make it a hard prerequisite like we did in the DeployAndActivateRMN sequence as there is no reason to deploy a net new chain with CLLCCIP, RMNMCMS and UFC MCMs

carte7000 (dismissed):

carte7000 (approved):

tt-cll (approved):

PR #2236, opened 2026-08-11
Introduce verifier helper for testing

RensR:


chains/evm/contracts/ccvs/CommitteeVerifier.sol
```diff
@@ -71,7 +71,7 @@ contract CommitteeVerifier is Ownable2StepMsgSender, ICrossChainVerifierV1, Sign
     address, // feeToken
     uint256, // feeTokenAmount
     bytes calldata // verifierArgs
-  ) external view returns (bytes memory verifierReturnData) {
+  ) external view virtual returns (bytes memory verifierReturnData) {
```
RensR, line 74: not strictly needed but no bytecode diff and should have been virtual in the first place

chains/evm/contracts/test/helpers/VerifierTestHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+    MessageV1Codec.MessageV1 calldata message,
+    bytes32, // messageId
+    bytes memory // verifierResults
+  ) external view virtual override {
+    _assertNotCursedByRMN(message.sourceChainSelector);
+    _onlyOffRamp(message.sourceChainSelector);
+
+    _assertNoSideEffects(message);
+
+    RemoteChainConfig storage chainConfig = _getRemoteChainConfig(message.destChainSelector);
```
matYang: `_getRemoteChainConfig(message.destChainSelector)` feels a bit off, more readable if the remote chain at the dest key'ed by sourceChainSelector?

chains/evm/contracts/test/helpers/VerifierTestHelper/VerifierTestHelper.config.t.sol
```diff
@@ -0,0 +1,74 @@
...
+      gasForVerification: 1,
+      payloadSizeBytes: 1
+    });
+
+    vm.prank(OWNER);
+    vm.expectRevert(VerifierTestHelper.MustUseAllowlist.selector);
+    s_verifier.applyRemoteChainConfigUpdates(configs);
+  }
+
+  function test_applyRemoteChainConfigUpdates_RevertWhen_RouterIsNotTestRouter() public {
```
0xsuryansh: I think we use the pattern `RevertWhen_<ErrorName>` ?

Although foundry guide says `test_FunctionName_RevertWhen_Condition` so this works too

0xsuryansh (dismissed):

chains/evm/contracts/test/helpers/VerifierTestHelper.sol
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  function _assertNoSideEffects(
+    MessageV1Codec.MessageV1 calldata message
+  ) internal view virtual {
+    if (message.data.length > 0 || message.ccipReceiveGasLimit != 0) {
+      revert MessageCannotHaveSideEffects();
+    }
+
+    if (message.tokenTransfer.length != 1) {
```
0xsuryansh, line 129: should we match to `sourceTokenAddress` to be more safe ?

Can skip as well because token pool is tied to prod router

RensR: Already doing it outside of this function :)

0xsuryansh (approved):

carte7000 (approved):

PR #2239, opened 2026-08-11
Add SuiExtraArgsV1

agusaldasoro:
Adds Sui extra args support to the Solana FeeQuoter and updates SVM/Sui payload size accounting.

This PR introduces Sui-specific extra args parsing and validation, including token receiver and receiver object IDs, and accounts for destination-side overhead when validating `max_data_bytes`. It also updates the SVM extra args path so token transfer overhead and destination byte overhead are included in expanded payload size checks. This matches one on one the behaviour for EVM.

Tests were updated to reflect the larger expanded payload size for SVM token transfers.

### DIFF
--- a/chains/solana/contracts/programs/fee-quoter/src/instructions/v1/messages.rs
+++ b/chains/solana/contracts/programs/fee-quoter/src/instructions/v1/messages.rs
@@ -1,8 +1,8 @@
 use anchor_lang::prelude::*;
 
 use ccip_common::v1::{
-    validate_aptos_address, validate_evm_address, validate_sui_address, validate_svm_address,
-    validate_tvm_address,
+    validate_aptos_address, validate_evm_address, validate_svm_address, validate_tvm_address,
+    SUI_PRECOMPILE_SPACE,
 };
 use ccip_common::{
     CommonCcipError, CHAIN_FAMILY_SELECTOR_APTOS, CHAIN_FAMILY_SELECTOR_EVM,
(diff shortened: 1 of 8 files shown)


chains/solana/contracts/programs/fee-quoter/src/instructions/v1/messages.rs
```diff
@@ -107,9 +138,42 @@ pub fn process_extra_args(
...
+                4,
+                FeeQuoterError::InvalidInputsMissingExtraArgs
+            );
+
+            let tag: [u8; 4] = extra_args[..4].try_into().unwrap();
+
+            let mut data = &extra_args[4..];
+            parse_and_validate_sui_extra_args(
+                dest_config,
+                tag.try_into().unwrap(),
```
toblich: Nitpick: looks like the types match so you should be able to remove the `.try_into`
```suggestion
                tag,
```

chains/solana/contracts/programs/fee-quoter/src/instructions/v1/messages.rs
```diff
@@ -107,9 +138,42 @@ pub fn process_extra_args(
...
+
+            if extra_args.is_empty() {
+                let args = SuiExtraArgsV1::default_config();
+
+                return Ok(ProcessedExtraArgs {
+                    bytes: args.serialize_with_tag(),
+                    gas_limit: args.gas_limit,
+                    allow_out_of_order_execution: args.allow_out_of_order_execution,
+                    token_receiver: None,
+                });
```
toblich: Should this check here if there is token transfer? It looks a bit odd that right now these two options are possible with this code:
- Empty extra args -> This will return successfully here with default values, regardless of a token transfer, and will set `token_receiver: None`
- Extra args containing just a tag (no data) -> it will set the default extra args at the beginning of `parse_and_validate_sui_extra_args`, but throw `InvalidTokenReceiver` because there is token transfer without a token receiver

agusaldasoro: Changed to return an error if the message has token transfer but missing extra args

agusaldasoro: @FelixFan1992 can you confirm this? Thank you!

FelixFan1992 (dismissed):
synced offline, no issues from my end

toblich (dismissed):

chains/solana/contracts/programs/fee-quoter/src/instructions/v1/messages.rs
```diff
@@ -107,9 +138,32 @@ pub fn process_extra_args(
...
+            require!(
+                !extra_args.is_empty(),
+                FeeQuoterError::InvalidInputsMissingExtraArgs
+            );
+
+            require_gte!(
+                extra_args.len(),
+                4,
+                FeeQuoterError::InvalidInputsMissingExtraArgs
+            );
```
toblich: Nitpick: The second check already covers the first condition (and with the same error type), so it's redundant IMHO
```suggestion
            require_gte!(
                extra_args.len(),
                4,
                FeeQuoterError::InvalidInputsMissingExtraArgs
            );
```

toblich (dismissed):

PR #2241, opened 2026-08-12
Add new OnrampUpgrader changesets

carte7000:
Add new OnrampUpgrader changesets

1. Add a new 3 phase upgrade Onramp Changeset
2. Make TransferToTimelock mockable
3. Generate mocks
4. Add Test verifier and OnrampUpgrader adapters
5. Add a new GetOffRampSourceOnRamps interface


deployment/v2_0_0/adapters/upgrade_onramp.go
```diff
@@ -0,0 +1,115 @@
...
+	// neither is an error: every lane must be classifiable before an upgrade starts.
+	ClassifyDestChains(e cldf.Environment, chainSelector uint64) (LaneClass, error)
+	// DestChainSelectors returns the remote chain selectors that the canonical
+	// OnRamp has lanes to. The implementation reads the OnRamp's dest chain
+	// configs on-chain in a single RPC call (e.g. getAllDestChainConfigs).
+	DestChainSelectors(e cldf.Environment, chainSelector uint64) ([]uint64, error)
+	// LegacyOnRampRef returns the datastore ref of the legacy-qualified OnRamp
+	// left behind by DeployNewOnRamp. It returns an error when no legacy OnRamp
+	// exists (e.g. cleanup already completed and the ref was removed).
+	LegacyOnRampRef(e cldf.Environment, chainSelector uint64) (datastore.AddressRef, error)
```
simsonraj, line 41: in this case legacyOnramp means the previous 2.0 onramp, not 1.6 or 1.5 no?

deployment/v2_0_0/changesets/upgrade_onramp.go
```diff
@@ -0,0 +1,644 @@
...
+
+		// DS merge at the end of the changeset is expected to upsert the new onramp ref and replace the legacy onramp
+		if addErr := ds.Addresses().Add(result.NewOnRampRef); addErr != nil {
+			return cldf.ChangesetOutput{}, fmt.Errorf("add new OnRamp ref to datastore: %w", addErr)
+		}
+		if addErr := ds.Addresses().Add(result.LegacyOnRampRef); addErr != nil {
+			return cldf.ChangesetOutput{}, fmt.Errorf("add legacy OnRamp ref to datastore: %w", addErr)
+		}
+
+		if !cfg.DisableTransferAndVerifierOwnership {
```
simsonraj: is it supposed to gate the verifier ownership? if so its not happening here?

deployment/utils/datastore/datastore.go
```diff
@@ -92,6 +92,44 @@ func findRef(ds datastore.DataStore, ref datastore.AddressRef) (datastore.Addres
 	return refs[0], 1, nil
 }
 
+// FindAndFormatCanonicalRef is like FindAndFormatRef but always includes
+// AddressRefByQualifier in the filter list, even when ref.Qualifier is empty.
+// This ensures no-qualifier lookups only match refs with an empty qualifier
+// (canonical refs), not refs with non-empty qualifiers like "legacy".
+func FindAndFormatCanonicalRef[T any](ds datastore.DataStore, ref datastore.AddressRef, chainSelector uint64, format FormatFn[T]) (T, error) {
```
simsonraj, line 99: this method is not called anywhere?

deployment/utils/datastore/datastore_canonical_test.go
```diff
@@ -0,0 +1,61 @@
...
+		Type:      "OnRamp",
+		Version:   semver.MustParse("2.0.0"),
+		Qualifier: "legacy",
+	}, 1, FullRef)
+	require.NoError(t, err)
+	assert.Equal(t, "0x2222", got.Address)
+	assert.Equal(t, "legacy", got.Qualifier)
+
+	// Regular FindAndFormatRef with no qualifier should fail (finds 2).
+	_, err = FindAndFormatRef(sealed, df.AddressRef{
```
simsonraj, line 55: shouldnt this be calling FindAndFormatCanonicalRef instead otherwise it may not fail or may be it fails always?

deployment/v2_0_0/changesets/upgrade_onramp.go
```diff
@@ -0,0 +1,644 @@
...
+		}
+
+		laneClass, err := upgrader.ClassifyDestChains(e, cfg.ChainSelector)
+		if err != nil {
+			return cldf.ChangesetOutput{}, fmt.Errorf("classify dest chains: %w", err)
+		}
+
+		// Pre-flight: removing the legacy OnRamp from remote OffRamp whitelists is only
+		// safe once every dest chain has been promoted to its final router (prod Router for
+		// the ProdRouter class, TestRouter for the TestRouter class).
```
simsonraj, line 498: TestRouter for the TestRouter class - should there be a check for this just like VerifyPromotedToProdRouter?

deployment/v2_0_0/adapters/upgrade_onramp.go
```diff
@@ -0,0 +1,115 @@
...
+// SetDynamicConfig) are executed directly. The returned BatchOps contain only
+// writes to timelock-owned contracts (if any).
+type OnRampUpgrader interface {
+	// VerifyOnrampRequireUpgrade returns no error if the canonical OnRamp is not yet upgraded to the new version, and an error otherwise
+	VerifyOnrampRequireUpgrade(e cldf.Environment, chainSelector uint64) error
+	DeployNewOnRamp(e cldf.Environment, chainSelector uint64) (OnRampUpgradeResult, error)
+	// ClassifyDestChains partitions the legacy OnRamp's dest chains by which router
+	// currently fronts them (prod Router vs TestRouter). A dest chain routed through
+	// neither is an error: every lane must be classifiable before an upgrade starts.
+	ClassifyDestChains(e cldf.Environment, chainSelector uint64) (LaneClass, error)
```
simsonraj, line 33: will this rightly classify new lanes if any added between the phase 1 & 3?

deployment/v2_0_0/adapters/upgrade_onramp.go
carte7000: Yes Legacy here refer to the previous 2.0.0 Onramp

deployment/v2_0_0/changesets/upgrade_onramp.go
carte7000: I use this mostly for testing in devenv where MCM is not deployed by default the expectation is that this should transfer ownership to MCMS

deployment/v2_0_0/changesets/upgrade_onramp.go
carte7000: Good point will add one

deployment/utils/datastore/datastore_canonical_test.go
carte7000: This is the difference between FindAndFormatCanonicalRef and FindAndFormatRef. The later fails when there is multiple ref for the same contract FindAndFormatCanonicalRef picks the one without qualifier

deployment/v2_0_0/adapters/upgrade_onramp.go
carte7000: This will not, this assumes that none are added. I added a dest selectors property to the input of the CS which filter the dest operations only for the selectors sent as input. In the case we have a lane that is added in between my assumption is that it will not be in the dest selectors scope so it should be untouched.

1. If we have lanes added in the middle it's worth noting we will have to redo the process for them.
2. As soon as phase 1 is executed the new onramp becomes the empty qualifier one in the datastore so if you add a new lane after I do phase 1 on a chain your new lane will use the new onramp. Just need to be sure that NOPs are running the temp job (pointing to the legacy onramp) before submitting new jobs

deployment/v2_0_0/changesets/upgrade_onramp.go
```diff
@@ -0,0 +1,671 @@
...
+		return changesetscore.NewOutputBuilder(e, mcmsReaderRegistry).
+			WithBatchOps(batchOps).
+			Build(cfg.MCMS)
+	}
+
+	return cldf.CreateChangeSet(apply, validate)
+}
+
+// Filter the destSelectors to only those in scope. Return an error if none are in scope.
+func scopeDestSelectors(destSelectors []uint64, scope []uint64) []uint64 {
```
simsonraj, line 430: Should this return error if nothing matched? seems like the onramp will still deploy, but wont whitelist on no remote offramps?

deployment/v2_0_0/changesets/upgrade_onramp.go
```diff
@@ -0,0 +1,671 @@
...
+	changesetscore "github.com/smartcontractkit/chainlink-ccip/deployment/utils/changesets"
+	"github.com/smartcontractkit/chainlink-ccip/deployment/utils/mcms"
+	"github.com/smartcontractkit/chainlink-ccip/deployment/v2_0_0/adapters"
+)
+
+// UpgradeOnrampConfig is the input for the OnRamp upgrade changesets.
+type UpgradeOnrampConfig struct {
+	ChainSelector                       uint64
+	DisableTransferAndVerifierOwnership bool
+	TestVerifierAllowedSenders          []string
```
simsonraj: seems unused?

deployment/v2_0_0/changesets/upgrade_onramp.go
simsonraj: Actually validated this with AI, seems like if the remote offramp is missed to whitelist at this stage then
`Phase3 then promotes it and every message is rejected at the destination.`

carte7000: That is correct but this assumes that the `DestSelectorsInScope` pass to phase 1 and phase 3 are different right? Even if we return an error if we send different `DestSelectorsInScope` it will have this effect

deployment/v2_0_0/changesets/upgrade_onramp.go
```diff
@@ -0,0 +1,670 @@
...
+		}
+
+		family, err := chainsel.GetSelectorFamily(cfg.ChainSelector)
+		if err != nil {
+			return cldf.ChangesetOutput{}, fmt.Errorf("get chain family: %w", err)
+		}
+		upgrader, ok := onrampUpgraderRegistry.Get(family)
+		if !ok {
+			return cldf.ChangesetOutput{}, fmt.Errorf("no OnRampUpgrader registered for family %q", family)
+		}
```
RensR, line 275: This is duplicated three times, can we extract it into a function?

deployment/v2_0_0/changesets/upgrade_onramp.go
```diff
@@ -0,0 +1,670 @@
...
+		family, err := chainsel.GetSelectorFamily(cfg.ChainSelector)
+		if err != nil {
+			return cldf.ChangesetOutput{}, fmt.Errorf("get chain family: %w", err)
+		}
+		upgrader, ok := onrampUpgrader.Get(family)
+		if !ok {
+			return cldf.ChangesetOutput{}, fmt.Errorf("no OnRampUpgrader registered for family %q", family)
+		}
+
+		if !cfg.DisableTransferAndVerifierOwnership {
```
RensR, line 349: This is duplicated as well, probably makes it more readable when extracted as well

deployment/v2_0_0/changesets/upgrade_onramp.go
```diff
@@ -0,0 +1,670 @@
...
+		operationsBundle := operations.NewBundle(func() context.Context { return context.Background() }, e.Logger, operations.NewMemoryReporter())
+		e.OperationsBundle = operationsBundle
+		promoteOps, err := upgrader.PromoteOnrampToTestRouter(e, cfg.ChainSelector, scopedDestSelectors)
+		if err != nil {
+			return cldf.ChangesetOutput{}, fmt.Errorf("stage on TestRouter: %w", err)
+		}
+		batchOps = append(batchOps, promoteOps...)
+
+		return changesetscore.NewOutputBuilder(e, mcmsReaderRegistry).
+			WithBatchOps(batchOps).
```
RensR: Can't we just pass promoteOps into this and remove all lines that include `batchOps`



```
		e.OperationsBundle = operations.NewBundle(func() context.Context { return context.Background() }, e.Logger, operations.NewMemoryReporter())
		promoteOps, err := upgrader.PromoteOnrampToTestRouter(e, cfg.ChainSelector, scopedDestSelectors)
		if err != nil {
			return cldf.ChangesetOutput{}, fmt.Errorf("stage on TestRouter: %w", err)
		}

		return changesetscore.NewOutputBuilder(e, mcmsReaderRegistry).
			WithBatchOps(promoteOps).
			Build(cfg.MCMS)
```

simsonraj (approved):

RensR (approved):

PR #2244, opened 2026-08-12
Auto deploy & configure TESTTR token setup during Lanes migrations

simsonraj:


chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
```diff
@@ -157,12 +167,186 @@ func MigrateChainLanesToV2(
...
+		if err != nil {
+			return deployment.ChangesetOutput{}, err
+		}
+		return mergeTestTokenOutput(e, laneOut, tokenOut)
 	}
 
 	return deployment.CreateChangeSet(apply, validate)
 }
 
+const testTokenSymbol = "TESTTR" // Note: a TEST token already exists on some chains behind the PROD Router
```
RensR: Could we not differentiate using a qualifier?

RensR: Or have one token configured on both?

chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
```diff
@@ -157,12 +167,186 @@ func MigrateChainLanesToV2(
...
+const testTokenDecimals = 18
```
RensR: These consts feel slightly out of place here

chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
```diff
@@ -157,12 +167,186 @@ func MigrateChainLanesToV2(
...
+			TokenRef: testTokenRef(sel),
+			TokenPoolRef: datastore.AddressRef{
+				ChainSelector: sel,
+				Qualifier:     testTokenSymbol,
+				Type:          datastore.ContractType(utils.BurnMintTokenPool),
+				Version:       utils.Version_2_0_0,
+			},
+			AllowedFinalityConfig: finality.Config{
+				WaitForFinality: false,
+				WaitForSafe:     false,
```
RensR: Might as well enable this so we don't have to later?

simsonraj: this depends on if we are using the TEST token for the prod router use case right? 
isnt it best to keep it false to test transfers quickly? why would we need to enable them?

HelloKashif (dismissed):

HelloKashif (approved):

chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
simsonraj: We will have to use TESTTR token due to having a single TAR for the chain

0xsuryansh (approved):

PR #2248, opened 2026-08-13
feat(reader): ccip reader and config poller metrics

makramkd:
This PR adds missing metric data for the CCIP reader and the config poller. The chain accessor is the next one up the stack. The document explaining the methodology and tracking this is in docs/metrics/reader-metrics.md - reviewers can read that first. The code implements the metrics marked as "High prio" in that doc.

Note that these abstractions _cannot_ expose chain-specific data. There will have to be a separate analysis (scheduled for after the plugin is instrumented) of the available metrics for e.g. EVM, Solana, etc. that can be used to complement these metrics here.

One fix this PR has: the config_poller_v2 was previously overwriting a potentially nonempty cache with empty data - this has been fixed to not do that, emit a metric + a warn log.

Still TODO:

- [x] Use the new metrics to run health checks on healthy and unhealthy dons: done on a fresh healthy DON, and for a fault-injected unhealthy DON which was fault-injected to return an empty msg range for MsgsBetweenSeqNums, see https://github.com/smartcontractkit/chainlink-ccip/pull/2248/commits/720a22e7b49dabfe0a44b9afcab13f2ef7d7ad41, reverted https://github.com/smartcontractkit/chainlink-ccip/pull/2248/commits/b8144f32d554df8ed37865fcbe12effbcc24f436)
- [x] Use the new metrics to triage a DON with an appropriate injected fault (fault in the same commit as above)

### DIFF
--- a/pkg/reader/config_poller_v2.go
+++ b/pkg/reader/config_poller_v2.go
@@ -3,6 +3,7 @@ package reader
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -12,6 +13,7 @@ import (
 	cciptypes "github.com/smartcontractkit/chainlink-common/pkg/types/ccipocr3"
 
 	"github.com/smartcontractkit/chainlink-ccip/pkg/logutil"
+	"github.com/smartcontractkit/chainlink-ccip/pkg/reader/rcmetrics"
 )
 
 // Analyzer-relevant log messages emitted by the config poller.
@@ -68,6 +70,8 @@ type configPollerV2 struct {
 	stopChan               chan struct{}
 	wg                     sync.WaitGroup
 	consecutiveFailedPolls atomic.Uint32
+
+	metrics rcmetrics.ConfigPollerMetrics
 }
 
 // chainCache represents the cache for a single chain.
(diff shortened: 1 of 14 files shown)


pkg/reader/rcmetrics/rcmetrics.go
```diff
@@ -0,0 +1,287 @@
...
+func (c *configPollerMetrics) RecordLastSuccess(chain string, epochSeconds int64) {
+	c.lastSuccessTimestamp.Record(context.Background(), epochSeconds, metric.WithAttributes(
+		attribute.String("chain", chain),
+	))
+}
+
+// AccessorMetrics instruments the chain accessor wrapper (per-operation reads).
+// Retained for the interface-based pattern even though the accessor wrapper is
+// not yet wired; the NoOp fallback keeps callers nil-free when it lands.
+type AccessorMetrics interface {
```
makramkd, line 216: Not used yet in this PR - will be used in a follow up (up the PR stack).

pkg/reader/config_poller_v2.go
```diff
@@ -453,10 +480,21 @@ func (c *configPollerV2) batchRefreshChainAndSourceConfigs(
...
+	// would time-stamp empty data as fresh and serve it to every consumer. Keep
+	// the previously-good cache and do NOT mark it fresh so the gap stays visible
+	// to RecordCacheAge instead of being masked.
 	cache.chainConfigMu.Lock()
-	cache.chainConfigData = chainConfigSnapshot
-	cache.chainConfigRefresh = time.Now()
+	if reflect.DeepEqual(chainConfigSnapshot, cciptypes.ChainConfigSnapshot{}) {
+		c.metrics.RecordOverwrittenEmpty("chain")
+		c.lggr.Warnw("chain config snapshot returned empty, keeping existing cached config",
+			logutil.FieldChain, chainSel)
```
makramkd, line 493: The only real "logic" change here - an empty config snapshot doesn't override a potentially non-empty one.

carte7000 (approved):

RensR (approved):

PR #2254, opened 2026-08-14
Don't exclude LOCK_RELEASE-mechanism tokens from migrate_chain_lanes_to_v2 symbol exclusion

HelloKashif:
## Summary
- `excludeLanesWithTokenSymbols` (e.g. `[USDC, LBTC, BTC.b]`) previously excluded a lane purely on a case-insensitive ERC20 `symbol()` match, which could incorrectly hold back a lane whose "USDC"-symbol token is not actually a genuine CCTP-backed pool.
- For every token whose symbol matches an excluded name, the resolver now also reads the pool's on-chain lock/burn mechanism via `TokenAdminRegistry.GetPool` + `USDCTokenPoolProxy.GetLockOrBurnMechanism`. If the mechanism is `LOCK_RELEASE`, the token is not treated as genuine USDC and the lane is not excluded on its account.
- If the mechanism lookup errors (pool doesn't implement `USDCTokenPoolProxy`, e.g. an LBTC/BTC.b pool) or returns any other mechanism (`CCTP_V1`/`CCTP_V2`/`CCTP_V2_WITH_CCV`), behavior falls back to the previous plain symbol-match exclusion.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./deployment/v2_0_0/changesets/...`
- [x] `go test ./deployment/v2_0_0/changesets/...` — 17/17 pass, including 2 new tests covering the LOCK_RELEASE bypass and the mechanism-lookup-failure fallback
- [x] Independent review pass confirmed enum value, control flow, and fallback behavior against the Solidity `LockOrBurnMechanism` enum

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
+++ b/chains/evm/deployment/v2_0_0/changesets/migrate_chain_lanes_to_v2.go
@@ -14,10 +14,16 @@ import (
 	"github.com/smartcontractkit/chainlink-evm/gethwrappers/shared/generated/initial/erc20"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
 	"github.com/smartcontractkit/chainlink-deployments-framework/deployment"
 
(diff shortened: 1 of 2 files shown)


HelloKashif:
Closing as this isn't needed anymore

PR #2258, opened 2026-08-15
fix: MigrateLockReleasePoolLiquidity should allow callers to switch between direct ERC20.transfer and lockbox.deposit()

chris-de-leon-cll:
# Liquidity migration: `deposit()` by default, direct transfer as a break-glass

## What this adds

This change reintroduces the lockbox `deposit()` path for liquidity migrations, reversing the earlier switch to a plain `erc20.transfer` into the lockbox (`#2240`). It also makes the funding step selectable and, in doing so, sharpens the ownership/routing model of the whole migration flow.

### Two funding paths, selected per migration

`MigrateLockReleasePoolLiquidityInput` gains a `UsePlainTransfer` flag. The default (`false`) uses the lockbox's `deposit()`; setting it `true` keeps the pre-existing direct `transfer()` behaviour as a break-glass option.

- **Deposit path (default).** Emits the lockbox `Deposit` event. This event is what treasury / security reviewers want for monitoring and off-chain accounting, and it future-proofs the flow against lockbox implementations that run internal accounting logic on entry (a plain `transfer` would silently bypass that logic).
- **Transfer path (break-glass).** A plain `safeTransfer` of the ERC20 directly to the lockbox. Permissionless, silent, but requires no authorized-caller ceremony. Kept available so there is an escape hatch if `deposit()` is ever unavailable (e.g. a token whose `transferFrom` misbehaves).

The funding step for both unsiloed and siloed migrations (each silo's lockbox and the shared unsiloed lockbox) is routed through a single `appendFundingOps` helper that branches on the flag.

### `DepositProposalOnly` for the funding step

The migration funding step uses a dedicated `DepositProposalOnly` write (proposal-only, `NoCallersAllowed`) rather than the generic `Deposit` op. `deposit` moves funds via `safeTransferFrom` out of the timelock and must run **after** the batched `approve` in the same MCMS batch; routing it on "is the *deployer* an authorized caller?" (as the generic `Deposit` does) can emit it as an out-of-order EOA transaction the moment the executor happens to be on the lockbox's allowlist — exactly what the migration's sequential unit tests model (where the executor stands in for the deployer, as there is no real MCMS instance in those tests). Forcing the deposit into the proposal keeps it atomic with the approve. The generic `Deposit` op is left untouched because it has other callers (the hybrid CCTP path); the proposal-only guarantee is provided by a dedicated variant, mirroring `ApproveProposalOnly` and `ApplyAuthorizedCallerUpdatesProposalOnly`.

> Note on reviewer feedback: in the real MCMS flow the deployer is never an authorized caller, so the generic `Deposit` op routes to the proposal anyway and production is unaffected — the real-MCMS migration path (covered by the integration suite) exercises the same deposit-via-proposal behavior.

### The deployer handles permissions; the timelock handles funds

The migration is split along this line:

- **Authorize** — `applyAuthorizedCallerUpdates` uses the framework's `OnlyOwner` routing, so the caller is whatever currently owns the lockbox. This makes the step ownership-dependent and automatic: an **EOA write by the deployer** when the lockbox is still deployer-owned (the seed step in token expansion), or an **MCMS-batched write** when the lockbox is already timelock-owned (a cleanup step post-activation). No ownership transfer ceremony is required to run the migration.
- **Move funds** — `withdraw`, `approve`, and now `deposit` are all proposal-only (`NoCallersAllowed`), so they always execute through the timelock as part of the MCMS batch. The timelock is the one that pulled the liquidity out of the old pool, so it must be the one that puts it into the lockbox; the deployer should never touch the funds.

Making `deposit` proposal-only (rather than routing-based like the original `Deposit` write) is a deliberate hardening: the deployer never holds the withdrawn funds and must not execute the deposit during proposal construction. `OnlyOwner`-routed authorization combined with a routing-based `deposit` would have let the deployer run the deposit out-of-order relative to the batched approve.

### No authorized-caller cleanup

The deposit path adds the timelock as an authorized caller and intentionally leaves it there — there is no post-migration removal step. This is safe because the timelock is governance and, after activation, owns the lockbox anyway; leaving it on the allowlist introduces no new privilege. It also removes the `usedLockboxes` tracking that the old cleanup ceremony needed: since we never undo the authorization, there is nothing to track.

### Idempotent authorization across tranches

Liquidity migration is performed in tranches (a partial amount per run), so the authorize step is made fully idempotent: before appending `applyAuthorizedCallerUpdates`, the funding path reads the lockbox's current authorized callers and **only issues the update if the timelock is not already authorized**. The underlying contract is state-idempotent (an `EnumerableSet`), but re-issuing the op would add a redundant MCMS batch entry and emit a spurious `AuthorizedCallerAdded` event on every subsequent tranche. The check-and-skip avoids both, while still re-adding the timelock if an external owner ever removed it.

### The lockbox's `remoteChainSelector` is passed through

The migration predecessor's config is preserved per path: the actual remote chain selector for siloed lockboxes, and `0` for the unsiloed lockbox. The lockbox contract ignores this parameter (it is an interface requirement), so the value is not behavioral — it is kept for interface consistency and to leave room for future lockbox versions that might use it.

### The flag is deliberately asymmetric across the two surfaces

`UsePlainTransfer` is exposed only on the standalone `MigrateLockReleasePoolLiquidity` changeset (`LockReleasePoolMigration.UsePlainTransfer`) — the intended surface for a liquidity migration / drain operation. It is intentionally **not** plumbed through the token-expansion seed flow (`TokenExpansion` → seed), which stays deposit-only and offers no direct-transfer override. The asymmetry is deliberate: the recommended path is uniformly enforced on the automated expansion tooling, while the escape hatch is reserved for the explicit, standalone operator who opts into it.

## Reverse-propagation for incremental (batched) migrations

This change also fixes the migration tooling's reverse-propagation so that an already-migrated, live pool's accept list is actually kept up to date when other nodes of a web migrate later — which is what makes **batched / sequential** migration of a fully-connected web (A, then B+C) correct.

### The problem it fixes

`autoMigrateRemoteChains` reverse-propagates a newly-migrated pool's address to each of its counterparts so the web stays connected. That worked for a counterpart that was a legacy v1 pool, but broke in the common batched case (two pools migrating in a single expansion run): one migrating chain's reverse-propagation would run the full configure sequence against a **same-batch peer's brand-new pool that had not gone live yet**, which **prematurely activated** that pool (registered/set the pool on its chain's registry). When that peer's own migration pass then ran, it saw itself as already the target pool and skipped auto-migration entirely — dropping its legacy remote config and its own reverse-propagation. The peer ended up disconnected, and the already-migrated pool never learned the new pool.

### Three coordinated changes

- **Skip reverse-propagation into a counterpart being migrated in the same batch.** A same-batch peer's not-yet-live pool is wired by that peer's own forward config; running the full configure sequence against it buys nothing and is what triggered the premature activation. Reverse-propagation now only reaches **non-migrating counterparts** (already-active v2 pools or legacy v1 pools) — exactly the places that need the new pool address added to stay reachable.
- **Reverse-propagation runs with extend semantics** (does not require `remoteChains` to enumerate every chain the destination's active pool supports). It adds a pool to an existing active pool, it does not replace the pool, so the upgrade-safety check that assumes a pool swap is not applicable.
- **Deterministic chain processing.** The expansion loop now iterates chains in sorted-selector order instead of relying on Go map iteration, removing order-dependent behavior from batched runs entirely.

### The intent

Reverse-propagation is only ever about *adding* the newly-migrated pool to a counterpart's accept list. It is deliberately kept minimal (no rate-limit/fee/ownership changes on the counterpart), and it deliberately avoids touching a peer that is itself being migrated in the same run — that peer's connectivity is its own forward config's job. The net effect is that every pool in a web ends up with a complete, consistent remote-pool list regardless of migration order or batching.


chains/evm/deployment/v2_0_0/operations/erc20_lock_box/erc20_lock_box_custom_operations.go
```diff
@@ -61,6 +61,29 @@ var Deposit = contract.NewWrite(contract.WriteParams[DepositArgs, *ERC20LockBoxC
 	},
 })
 
+// DepositProposalOnly is identical to Deposit but forces the operation into an MCMS proposal.
+var DepositProposalOnly = contract.NewWrite(contract.WriteParams[DepositArgs, *ERC20LockBoxContract]{
```
0xsuryansh, line 65: Do we need this ? Deployer key is never an authorised caller on the lockbox so the `Deposit` defined [above](https://github.com/smartcontractkit/chainlink-ccip/pull/2258/changes#diff-4bd8d14e2a43b28cad8955b5c8524caf2f5fbc53e14b4e253a5dbd2fb223d97fR36-R62) this one uses `IsAllowedCaller` and then it goes into the following block 


```
if allowed {
				// If the call has actually been sent, we need check the call error and confirm the transaction.
				_, confirmErr := deployment.ConfirmIfNoErrorWithABI(chain, tx, params.ContractABI, callErr)
				if confirmErr != nil {
					return WriteOutput{}, fmt.Errorf("failed to confirm %s tx against %s on %s with args %+v: %w", params.Name, input.Address, chain, input.Args, confirmErr)
				}
				execInfo = &ExecInfo{Hash: tx.Hash().Hex()}
				b.Logger.Debugw(fmt.Sprintf("Confirmed %s tx against %s on %s", params.Name, input.Address, chain), "hash", tx.Hash().Hex(), "args", input.Args)
			} else if callErr != nil {
				// If we didn't execute the transaction, but there was an error preparing it, return the error.
				return WriteOutput{}, fmt.Errorf("failed to prepare %s tx against %s on %s with args %+v: %w", params.Name, input.Address, chain, input.Args, callErr)
			} else {
				b.Logger.Debugw(fmt.Sprintf("Prepared %s tx against %s on %s", params.Name, input.Address, chain), "args", input.Args)
			}
```
where it prepares a transaction which is then added to proposal

chains/evm/deployment/v2_0_0/sequences/tokens/migrate_lock_release_pool_liquidity.go
```diff
@@ -513,6 +490,83 @@ func appendSetRebalancerAndWithdraw(
...
+			Spender: lockboxAddr,
+			Value:   amount,
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to approve lockbox %s to spend tokens: %w", lockboxAddr, err)
+	}
+	ops = append(ops, approveReport.Output)
+
+	depositReport, err := cldf_ops.ExecuteOperation(b, lockbox_ops.DepositProposalOnly, evmChain, evm_contract.FunctionInput[lockbox_ops.DepositArgs]{
```
0xsuryansh: can directly use deposit

chains/evm/deployment/v2_0_0/sequences/tokens/migrate_lock_release_pool_liquidity.go
```diff
@@ -513,6 +490,83 @@ func appendSetRebalancerAndWithdraw(
...
+		}
+		return append(ops, transferReport.Output), nil
+	}
+
+	// Deposit path: make the timelock an authorized caller so it can deposit, approve the lockbox,
+	// then deposit via the lockbox's deposit() function (emits the Deposit event). The framework
+	// routes the authorize step automatically based on lockbox ownership: EOA when deployer-owned
+	// or MCMS-batched when timelock-owned. The timelock remains an authorized caller afterward;
+	// this is harmless since the timelock is governance and typically owns the lockbox.
+	addAuthReport, err := cldf_ops.ExecuteOperation(b, lockbox_ops.ApplyAuthorizedCallerUpdates, evmChain, evm_contract.FunctionInput[lockbox_ops.AuthorizedCallerArgs]{
```
0xsuryansh: Should be idempotent.. 
First use `GetAllAuthorizedCallers` and if not then apply this update

This is needed because liquidity migration is done in tranches not all at once
It's always good to check before an op. 


Note for later: Digressing a but from this, but we should sync with Product to see if we need to remove MCMS as an allowed caller to Lockbox post-migration. Since MCMS isn't always the liquidity rebalancer, external owners might not want it retaining control over Lockbox funds. This doesn't need to block the current PR, may / may not be part of the changeset, but it's worth addressing as a follow-up operational task.

chris-de-leon-cll: good call - fixed

chains/evm/deployment/v2_0_0/operations/erc20_lock_box/erc20_lock_box_custom_operations.go
chris-de-leon-cll: I tried using the plain `Deposit` operation initially, but this caused a bunch of unit test failures due to the way that it's configured

For context, liquidity migrations expect timelock and the deployer to be distinct accounts, but the unit test harness makes the deployer key a stand-in for timelock to avoid deploying MCMS, which causes issues

Let me see if I can find a way to update the unit tests to make things more true to reality

chris-de-leon-cll: I did some more digging and it turns out that this is going to be a heavier lift than what I initially thought - most of the unit test harness would need to be re-written so that we have a more realistic prod set up. Right now the "deployer key imitates timelock" pattern is used everywhere throughout the existing unit test and migrating away from it can be its own self-contained change. For now, let's keep this as-is since it's harmless + unblocks things, then open a follow-up PR to update the test.

tt-cll (approved):

athegaul (approved):

0xsuryansh (approved):

PR #2262, opened 2026-08-18
fix(commit): add important metrics labels, update runbooks [CCIP-13069]

makramkd:
For reviewers: feel free to ignore the dashboard changes (>2k lines changed/added). Runbook changes are to refer to the labels that were added (dest/source chain labels).

A lot of the existing metrics were missing a critical label, which was the destination chain ID or selector. In order to make sure we don't miss anything, I defined a couple of helpers in a new package `metricsutil`, DestChainAttrs and SourceChainAttrs, to include all the chain labels we may use, which are chain family, ID, name, and selector. These are applied to the commit PromReporter and the CCIP reader and config poller metrics implementations where appropriate.

This chain information is fetched from chain-selectors; we fall back gracefully to "unknown" in the event the chain is not known by the library. This is something we could end up having a dashboard panel for or even alert on it if we want to be really strict, since we've had issues in the past where a chain is enabled but chain-selectors is not up to date.

Dashboards and runbooks were updated with this in mind.


commit/metrics/legacy_prom.go
makramkd, line 1: Moved the prom metrics to a separate file because these are probably unused and should be deleted in a followup.

commit/metrics/prom.go
makramkd, line 1: Pretty much all the changes to this file are adding source/dest chain attributes to all the metrics in a uniform way.

pkg/reader/rcmetrics/rcmetrics.go
makramkd, line 1: Same here, adding the source/dest chain attributes using the metricsutil package above.

pkg/reader/config_poller_v2.go
```diff
@@ -410,19 +410,13 @@ func (c *configPollerV2) refreshAllKnownChains() {
...
 	refreshFailed := false
-	now := time.Now()
 	for _, chain := range chainsToRefresh {
 		ctx, cancel := context.WithTimeout(context.Background(), bgRefreshTimeout)
-		chainLabel := rcmetrics.ChainLabel(chain)
 		c.lggr.Debugw("Issuing background refresh for known chain",
 			logutil.FieldChain, chain, logutil.FieldDestChain, c.destChainSelector)
 		if err := c.batchRefreshChainAndSourceConfigs(ctx, chain); err != nil {
 			refreshFailed = true
-			c.metrics.RecordPollFailure(chainLabel)
```
makramkd, line 421: Moved poll failure/success metrics to the method that actually does the refreshing below.

RensR (approved):

AmrMohamedRezk (approved):

PR #2267, opened 2026-08-19
docs(runbooks): absence of metrics and effect on commit plugin health

makramkd:
Update the runbooks and the dashboard to reflect what the _absence_ of some critical metrics means, and where to look when an absence is witnessed.

More context in this comment re: the reverts: https://github.com/smartcontractkit/chainlink-ccip/pull/2267#issuecomment-5352091480


RensR (dismissed):

tt-cll (dismissed):

bukata-sa (dismissed):

makramkd:
I noticed after testing this out on staging that we actually mark contracts as discovered even though the discovery processor is constantly failing 😞 . This is a separate bug in the plugins; for now I'm going to remove this metric because it is misleading, and will just update the runbooks to say that an absence of a lot of important metrics probably means the plugin is _not_ healthy.

RensR (approved):

0xsuryansh (approved):

PR #2273, opened 2026-08-20
feat(solana): implement TokenPoolMigrator for v1.6.x token pools

0xsuryansh:
Implements GetSupportedChains, GetRemoteToken, and GetRemotePools on the Solana SolanaAdapter (v1.6.0), so AutoMigrateRemoteChains can discover and reverse-propagate to an active Solana 1.6.x pool when migrating an EVM pool to v2.0.0. Previously only EVM adapters implemented TokenPoolMigrator, causing auto-migrate to fail with "does not support token pool migration" whenever a Solana chain was involved.

Since Solana pool programs have no instruction to enumerate configured remote chains, and identify a pool only by its shared program address (not a per-mint account), the mint and configured remotes are recovered by batch-probing candidate ChainConfig/pool-state PDAs via getMultipleAccounts instead of one RPC call per candidate.

### DIFF
--- a/chains/solana/deployment/v1_6_0/sequences/tokens.go
+++ b/chains/solana/deployment/v1_6_0/sequences/tokens.go
@@ -1,6 +1,7 @@
 package sequences
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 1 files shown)


0xsuryansh:
Correctly generates proposal here 

https://github.com/smartcontractkit/chainlink-deployments/pull/18195

image

chris-de-leon-cll:
Hmm did this PR genuinely unblock the proposal generation? Asking because I see some suspicious logs in each pipeline run 👀 in the CLD pipeline PR, it looks like you initially provided the wrong pool address (e.g. `B3rp2RHbuZeDeSSZLXww3EbaMr1TVtn9kF2a2FAobnxi`), which caused [this](https://github.com/smartcontractkit/chainlink-ccip/blob/7520b255725e639d2a71c23ef6235cb0d08d2e5c/deployment/tokens/configure_tokens_for_transfers.go#L224) check to pass by mistake:

```go
// deployment/tokens/configure_tokens_for_transfers.go#L214
if !bytes.Equal(activePool, targetPoolBytes) {
```

And eventually it reached [this](https://github.com/smartcontractkit/chainlink-ccip/blob/7080695cd687fbc2cf0c6fd9bebb58c7991f84ce/deployment/tokens/configure_tokens_for_transfers.go#L230-L234) code and showed a misleading error:

```go
// deployment/tokens/configure_tokens_for_transfers.go#L242-L248
legacyPoolMigrator, ok = legacyAdapter.(TokenPoolMigrator)
if !ok {
  return nil, nil, nil, fmt.Errorf("adapter for active pool version %s on chain selector %d does not support token pool migration", activePoolRef.Version, selector)
}
```

When you updated the address to the correct one, the bytes check correctly sees that the pools are equal on run 3 (the successful run):

```go
// deployment/tokens/configure_tokens_for_transfers.go#L214
if !bytes.Equal(activePool, targetPoolBytes) {
```

So it skipped over all the code changes in this PR and just logged a message:

```go
// deployment/tokens/configure_tokens_for_transfers.go#L266C1-L266C135
e.Logger.Infof("Active pool on chain selector %d is already the target pool, skipping auto-migration of remote chains", selector)
```

[This](https://github.com/smartcontractkit/chainlink-deployments/actions/runs/32425351494/job/96606971105#step:16:1086) log in run 3 (the successful pipeline run) confirms it:

```
{"level":"info","ts":1787266541.6140327,"caller":"tokens/configure_tokens_for_transfers.go:266","msg":"Active pool on chain selector 16423721717087811551 is already the target pool, skipping auto-migration of remote chains"}
```

With that said, it's probably better to hold off on this PR for now since Solana doesn't fully support upgrading their pools from v1 to v2 yet + it would avoid introducing more token pool migration code to debug 😅

0xsuryansh:
Yeah looked at the logs carefully in run 3, and this doesn't actually touches the code this PR adds, it was the address fixed in solana which correctly

  - Solana's entry: activePool == targetPoolBytes → returns before ever reaching the legacyAdapter.(TokenPoolMigrator) cast. That's exactly the L266 log.
  - EVM's entry: it does enter the legacy-migration branch (real v1.x→v2.0.0 swap), but legacyPoolMigrator there resolves to the EVM legacy adapter.. which already implemented TokenPoolMigrator before this PR- not the Solana one. The only Solana methods called in that path (DeriveTokenDecimals, ConfigureTokenForTransfersSequence) are pre-existing TokenAdapter methods, unrelated to this PR.

Dug into why the proposal still came out correct despite that: token_expansion.go has a separate, unconditional backfill step (L421-446) that fills in RemoteToken/RemotePool for any chain's {} remote-chain entry from its counterpart's resolved refs whenever both chains are in the same batch. That's what turned Solana's empty {16015286601757825753: {}} into the correct AppendRemotePoolAddresses.. **nothing to do with AutoMigrateRemoteChains or this PR.**

So: none of the pipeline run did not exercise GetSupportedChains/GetRemoteToken/GetRemotePools, and production doesn't need them either, the backfill mechanism already handles "existing Solana pool learns about new EVM pool" on its own. This PR only matters once Solana's own active pool needs to swap under AutoMigrateRemoteChains, which isn't a real case yet.

+1 on holding off will park it in draft and pick back up once there's an actual Solana pool-swap scenario to validate against 🙏

PR #2278, opened 2026-08-24
bump chainlink contracts package

RensR:
Bumps chainlink to 1.6.0 which cleans up a lot of code and it adds a 5.3 version of the single OZ 4.x dependency we use in CCIP (AuthorizedCallers). Since that would change bytecode, using it is not in scope of this PR.

The PR also stops importing the chainlink package dependencies, which keeps our package very clean.

### DIFF
--- a/chains/evm/package.json
+++ b/chains/evm/package.json
@@ -42,10 +42,10 @@
     "@openzeppelin/contracts-5.3.0": "npm:@openzeppelin/contracts@5.3.0"
   },
   "devDependencies": {
-    "@changesets/cli": "^2.31.0",
-    "@changesets/get-github-info": "^0.7.0",
-    "semver": "^7.8.4",
-    "solhint": "^6.2.2",
+    "@changesets/cli": "^3.0.1",
+    "@changesets/get-github-info": "^1.0.0",
+    "semver": "^7.8.5",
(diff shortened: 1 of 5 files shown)


chains/evm/package.json
```diff
@@ -32,21 +32,30 @@
     ".solhintignore-test"
   ],
   "engines": {
-    "node": ">=20",
+    "node": ">=22",
```
RensR, line 35: 20 is EOL, 22 is the lowest LTS

chains/evm/package.json
```diff
@@ -32,21 +32,30 @@
...
     "pnpm": ">=10"
   },
   "dependencies": {
     "@chainlink/ace": "1.0.0",
-    "@chainlink/contracts": "1.5.0",
-    "@openzeppelin/contracts-4.8.3": "npm:@openzeppelin/contracts@4.8.3",
+    "@chainlink/contracts": "1.6.0-beta.0",
+    "@openzeppelin/contracts-4.9.6": "npm:@openzeppelin/contracts@4.9.6",
```
RensR, line 41: The contract in chainlink/contracts we use has been bumped from 4.8.3 to 4.9.6 which means we have to import this more recent version. There is no bytecode diff.

PR #2284, opened 2026-08-25
CCIP-12654 - [TOOLING]: ERC20LockBox - applyAuthorizedCallerUpdates Support

futureproof1:
### Add UpdateLockboxAuthorizedCallers changeset

**What this does**

Adds a changeset that changes who is allowed to call a lockbox (ERC20LockBox, v2.0.0) on one or more chains.

Each ERC20LockBox keeps a set of "authorized callers" — the addresses permitted to move funds in and out of it. This changeset lets an operator add to or remove from that set, producing an MCMS timelock proposal for governance to sign.

**Why we need it**

We're migrating token pools from 1.6 to 2.0. As part of that, each chain gets a new pool and a new lockbox, and the old pool's authorization on the lockbox needs to be cleared once liquidity has moved across. Nothing currently does that — the existing migration sequences only ever add authorized callers, never remove them.

**How you use it**

Lockboxes are nested under the chain they live on, because one chain legitimately holds several (one per token, per silo group, or per remote chain):

```yaml
environment: prod_testnet
domain: ccv
changesets:
  - update_lockbox_authorized_callers:
      chainOverrides: [16015286601757825753]
      payload:
        mcms:
          description: "Remove old pool from USDC lockbox"
        cfg:
          input:
            - selector: 16015286601757825753
              lockboxes:
                - address: "0x314c8F8D921E7fC15F81093b7D17af5E99cdaABB"
                  removes:
                    - "0xB36DFBa076E940f3d400b3cE262344ecc3879992"
                - address: "0xdADe19cC917a83555619816a7f1fc0e7CcE80a19"
                  appends:
                    - "0x591dF28e56e76f25bf343c86e029Fc18dd48E43A"
```

Several lockboxes and several chains can go in one run. The `mcms` block is optional — the timelock qualifier and expiry are filled in automatically.

**Notes**

- **Addressed by address, not qualifier.** Qualifiers are a side-effect of whichever sequence deployed the lockbox — the pool qualifier, `<poolQualifier>-silo(<minRemoteSelector>)`, or `remoteChainSelector(<selector>)` for siloed USDC — so they aren't a stable addressing scheme.
- **No `version` field.** This is the `v2_0_0` package, so the target version is already pinned by the changeset the operator picked.
- **Idempotent.** It reads the current on-chain caller set first and drops anything already in the desired state, so re-running is safe and won't produce a pointless proposal. Both the read and the write use `WithForceExecute`, so a re-run after an out-of-band change sees current state rather than a cached report.
- **Atomic per chain.** A chain's writes are collected and turned into a single MCMS batch operation, so several lockboxes on one chain execute together or not at all. `NewBatchOperationFromWrites` rejects writes spanning chains, which is why it's called once per chain.
- **Reproducible ordering.** Batch operations follow input order, so the proposal's merkle root is stable for a given config.
- **Rejects contradictory configs** before touching a chain: the same address in both `appends` and `removes` (which never converges — the contract removes before adding, while the idempotency filter drops whichever side matches current state), a repeated chain selector, the same lockbox twice within a chain, a chain entry with no lockboxes, per-list duplicates, and zero addresses.
- **No datastore pre-check.** A wrong address fails on-chain when the read or write reverts. Gating on a datastore ref would force operators to hand-add refs just to proceed, and a hand-added ref with an unnormalized address is a landmine for later runs.

**Testing**

- 17 unit tests against simulated chains: happy path, idempotency (add / remove / mixed / partial filter), two lockboxes on one chain merging into a single batch operation, MCMS qualifier defaulting end-to-end, and every validation rule.
- Every guard is mutation-checked — disabled in turn, confirming the specific test fails, then restored. That's how `ForceExecutesWriteAfterOutOfBandChange` came about: the write's `WithForceExecute` turned out to be pinned by nothing, so removing it left all tests green. All 13 guards now fail with their guard disabled.
- Exercised end to end as a dry run against CCV prod_testnet — three lockboxes across sepolia and fuji, covering add, remove, add+remove, two chains in one run, a fully idempotent no-op, and a partial filter, plus 14 configs that must be rejected. Proposals decoded to the expected `applyAuthorizedCallerUpdates((address[],address[]))` calls, with per-chain batching as described above. No transactions were executed. (Run before the review fixes in the latest commit, which removed the `version` field and the datastore pre-check; the behaviour exercised is unaffected by both.)

### DIFF
--- a/chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
+++ b/chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
@@ -3,11 +3,9 @@ package changesets
 import (
 	"fmt"
 
-	"github.com/Masterminds/semver/v3"
 	"github.com/ethereum/go-ethereum/common"
 	chain_selectors "github.com/smartcontractkit/chain-selectors"
 	"github.com/smartcontractkit/chainlink-deployments-framework/chain/evm"
-	"github.com/smartcontractkit/chainlink-deployments-framework/datastore"
 	cldf_deployment "github.com/smartcontractkit/chainlink-deployments-framework/deployment"
 	cldf_ops "github.com/smartcontractkit/chainlink-deployments-framework/operations"
 	mcms_types "github.com/smartcontractkit/mcms/types"
@@ -16,7 +14,6 @@ import (
 	lbops "github.com/smartcontractkit/chainlink-ccip/chains/evm/deployment/v2_0_0/operations/erc20_lock_box"
 	common_utils "github.com/smartcontractkit/chainlink-ccip/deployment/utils"
 	cs_changesets "github.com/smartcontractkit/chainlink-ccip/deployment/utils/changesets"
-	datastore_utils "github.com/smartcontractkit/chainlink-ccip/deployment/utils/datastore"
 )
 
 // LockboxCallerUpdate is one lockbox's authorized-caller change. The chain it lives on comes
@@ -26,7 +23,9 @@ import (
 // side-effect of whichever sequence deployed the lockbox - the pool qualifier for a plain
 // lock-release pool, "<poolQualifier>-silo(<minRemoteSelector>)" per silo group, or
 // "remoteChainSelector(<selector>)" for siloed USDC - so they are not a stable addressing
-// scheme. Address is unambiguous, and verify confirms it really is a lockbox.
+// scheme. Address is unambiguous. A wrong address fails on-chain when the read or write
+// reverts, rather than being pre-checked against the datastore - datastore refs go missing
+// routinely, and blocking on one would force operators to hand-add refs just to proceed.
 type LockboxCallerUpdate struct {
 	Address common.Address   `json:"address"  yaml:"address"`
 	Appends []common.Address `json:"appends,omitempty"  yaml:"appends,omitempty"`
@@ -42,10 +41,9 @@ type ChainLockboxUpdate struct {
 }
 
 type UpdateLockboxAuthorizedCallersCfg struct {
-	// Version is cross-checked against the datastore ref for every Address, so a config
-	// written for one lockbox version cannot silently act on a newer one.
-	Version *semver.Version      `json:"version" yaml:"version"`
-	Input   []ChainLockboxUpdate `json:"input"   yaml:"input"`
+	// Version is not a field: this package is v2_0_0, so the target lockbox version is
+	// already pinned by the changeset the operator picked.
+	Input []ChainLockboxUpdate `json:"input" yaml:"input"`
 }
 
 var UpdateLockboxAuthorizedCallers = func(
@@ -64,9 +62,6 @@ func verifyUpdateLockboxAuthorizedCallers(
 	cfg := input.Cfg
 	if len(cfg.Input) == 0 {
 		return fmt.Errorf("at least one entry is required in input")
-	}
-	if cfg.Version == nil {
-		return fmt.Errorf("version is required so the target lockbox version can be cross-checked")
 	}
 
 	evmChains := e.BlockChains.EVMChains()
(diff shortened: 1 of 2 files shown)


chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
```diff
@@ -0,0 +1,299 @@
...
+	Selector uint64           `json:"selector" yaml:"selector"`
+	Address  common.Address   `json:"address"  yaml:"address"`
+	Appends  []common.Address `json:"appends,omitempty"  yaml:"appends,omitempty"`
+	Removes  []common.Address `json:"removes,omitempty"  yaml:"removes,omitempty"`
+}
+
+type UpdateLockboxAuthorizedCallersCfg struct {
+	// Version is cross-checked against the datastore ref for every Address, so a config
+	// written for one lockbox version cannot silently act on a newer one.
+	Version *semver.Version       `json:"version" yaml:"version"`
```
chris-de-leon-cll: I think this can be removed right? In this case, we're already in the `v2_0_0` folder, so the version is known

Ah wait, is this because you saw a `version` field in the JIRA ticket? If so, that'll be defined at the CLD level

The changeset input type can be minimal:

```go
type UpdateLockboxAuthorizedCallersCfg struct {
  Input []LockboxCallerUpdate `json:"input" yaml:"input"`
  MCMS  mcms.Input            `json:"mcms" yaml:"mcms"`
}
```

chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
```diff
@@ -0,0 +1,299 @@
...
+		return fmt.Errorf("at least one entry is required in input")
+	}
+	if cfg.Version == nil {
+		return fmt.Errorf("version is required so the target lockbox version can be cross-checked")
+	}
+
+	evmChains := e.BlockChains.EVMChains()
+	// Keyed on chain AND address, not chain alone: a chain legitimately holds several
+	// lockboxes (one per token, per silo group, or per remote chain), so several entries
+	// may share a selector. Only the same lockbox twice is a mistake.
```
chris-de-leon-cll: ah I keep forgetting this design detail - given this info, the proposed YAML input in the JIRA ticket should be updated otherwise we'll have to do hacky workarounds in the code to accommodate that use case

I've edited the YAML in the ticket so that it's more clear that a single chain can have many lockboxes. I've x-posted the new schema here for reference:

```yaml
environment: prod_testnet
domain: ccv
changesets:
  - update_lockbox_authorized_callers:
      chainOverrides: [ 11111111111111111111, 22222222222222222222, 33333333333333333333 ]
      payload:
        version: 2.0.0
        mcms:
          overridePreviousRoot: true
        input:
          - selector: 1111111111111111111
            lockboxes:
              - address: 0x1111111111111111111111111111111111111111
                appends:
                  - 0x1111111111111111111111111111111111111111
                  - 0x2222222222222222222222222222222222222222
                removes:
                  - 0x3333333333333333333333333333333333333333
                  - 0x4444444444444444444444444444444444444444
              - address: 0x2222222222222222222222222222222222222222              
                appends:
                  - 0x1111111111111111111111111111111111111111
                  - 0x2222222222222222222222222222222222222222
                removes:
                  - 0x3333333333333333333333333333333333333333
                  - 0x4444444444444444444444444444444444444444
```

Lmk if this makes sense! I think this should simplify a lot of the code now - for example instead of defining a custom `lockboxKey` struct you can most likely just maintain a set and check for duplicate addresses.

chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
```diff
@@ -0,0 +1,299 @@
...
+	return nil
+}
+
+func versionString(v *semver.Version) string {
+	if v == nil {
+		return "<nil>"
+	}
+
+	return v.String()
+}
```
chris-de-leon-cll: We can most likely remove this - for context, we encounter missing datastore entries on almost a daily basis, so this check will force whoever is running the changeset to add the missing ref manually just to unblock things. This can lead to several issues later (e.g. someone may forget to normalize an EVM address when they add the ref manually, which creates a landmine for future operatuibs)

I'd recommend one of the following:
- If the user provides the address of something other than a lockbox, then let the revert happen (preferred - keeps the code lean)
- Make this a soft check: if it resolves then no action needed, otherwise log a loud `warning` or `error`

chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	// Validated against a populated copy so ValidUntil stays optional while a value the
+	// operator did supply is still rejected if it is unusable. The qualifier needs no
+	// default here: EVMMCMSReader.GetTimelockRef and GetMCMSRef already fall back to
+	// CLLQualifier when it is empty.
+	mcmsInput := input.MCMS
+	if err := mcmsInput.PopulateDefaults(); err != nil {
+		return fmt.Errorf("failed to populate MCMS defaults: %w", err)
+	}
```
chris-de-leon-cll: I think this can be removed as well - if I'm not mistaken, the defaults won't get carried over into the `apply()` function anyways, and populating an MCMS struct with defaults just to validate it right after seems like a waste since the defaults are always valid

chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
```diff
@@ -0,0 +1,299 @@
...
+			}
+			if removes.Add(addr) {
+				return fmt.Errorf("duplicate address %s in removes for chain %d", addr, sel)
+			}
+			// An address in both lists never converges: the idempotency filter drops
+			// whichever side matches current on-chain state, so each apply flips it.
+			if appends.Has(addr) {
+				return fmt.Errorf(
+					"address %s is in both appends and removes for chain %d", addr, sel)
+			}
```
chris-de-leon-cll: nice validation check 💯  we should explicitly fail if someone is trying to remove and append the same address simultaneously

chains/evm/deployment/v2_0_0/changesets/update_lockbox_authorized_callers.go
```diff
@@ -0,0 +1,299 @@
...
+			// reads are not served from a cached report.
+			readBundle := cldf_ops.NewBundle(
+				e.GetContext, e.Logger, cldf_ops.NewMemoryReporter())
+			currentReport, err := cldf_ops.ExecuteOperation(
+				readBundle, lbops.GetAllAuthorizedCallers, chain,
+				contract.FunctionInput[struct{}]{
+					ChainSelector: chain.Selector,
+					Address:       lockboxAddr,
+					Args:          struct{}{},
+				})
```
chris-de-leon-cll: I'd recommend adding [WithForceExecute](https://docs.cld.cldev.sh/guides/changesets/operations-api#with-force-execute) to avoid any caching issues and removing the `readBundle`

chris-de-leon-cll (commented):
Nice work on this!! 🔥 I left a few comments below - lmk if you have any questions 🙏

chris-de-leon-cll (approved):
Incredible work 💯 really appreciate the in-depth test cases!

tt-cll (approved):

PR #2286, opened 2026-08-27
SmokeTest - WaitForCCIPFilterRegistration guard on the smoke test setup

vicentevieytes:
If you dispatch the devenv smoke test right after `ccip up` tests can fail due to the filters not being registered on the logpoller db yet. This happens because the filters are only registered when the plugin starts and the .

This PR adds a `WaitForCCIPFilterRegistration` check on the smoke test setup before dispatching the tests. Currently it only waits for the TON filters to be registered but it should be extended for other chains that find this issue.


duck-types (approved):
Maybe this should be part of the TestAdapter interface or the `ccip up` flow

archseer (approved):

toblich (approved):

ogtownsend (approved):

PR #2287, opened 2026-08-27
Migration changeset fixes

simsonraj:
- Allow 1.5 to 2.0 migration (validation fixes)
- Exclude sunset chains from migrating lanes

### DIFF
--- a/chains/evm/deployment/v2_0_0/adapters/lanemigrator.go
+++ b/chains/evm/deployment/v2_0_0/adapters/lanemigrator.go
@@ -23,6 +23,7 @@ import (
 	"github.com/smartcontractkit/chainlink-ccip/chains/evm/deployment/v1_0_0/operations/rmn_proxy"
 	mcms_seq "github.com/smartcontractkit/chainlink-ccip/chains/evm/deployment/v1_0_0/sequences"
 	routerops "github.com/smartcontractkit/chainlink-ccip/chains/evm/deployment/v1_2_0/operations/router"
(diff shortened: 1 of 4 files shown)


chains/evm/deployment/v2_0_0/adapters/lanemigrator.go
```diff
@@ -178,13 +179,19 @@ func verifyExistingLaneVersion(e deployment.Environment, evmChain evm.Chain, cha
...
-				"precondition failed for chain %d and remote chain %d: expected onRamp version on Router to be %s, but got version %s. ",
-				chainSelector, remoteChainSelector, onrampops_v160.Version.String(), onRampVersion.String(),
+				"precondition failed for chain %d and remote chain %d: expected onRamp version on Router to be %s or %s, but got version %s.",
+				chainSelector, remoteChainSelector, onrampops_v150.Version.String(), onrampops_v160.Version.String(), onRampVersion.String(),
 			)
 		}
 
+		// Only 1.6+ onRamps expose a FeeQuoter in their dynamic config; 1.5 onRamps
+		// use a PriceRegistry instead and share no fee quoter relationship to verify.
+		if !onRampVersion.Equal(onrampops_v160.Version) {
```
RensR, line 191: What are we even testing below? Just that a version is the expected one? Everything should be the same anyway right? Doesn't hurt but also doesn't add much security for quite a few lines of code

simsonraj: yea exactly! i just left it as it was doesnt hurt, but no value also

RensR (approved):

0xsuryansh (approved):

PR #2288, opened 2026-08-28
Add USDC proxy mechanism switch to lane migrator

0xsuryansh:
Adds USDCTokenPoolProxy handling to the EVM v2 LaneMigrateToNewVersionChangeset path.

- When migrating lanes to the prod router, the migrator now reads the current USDC lock/burn mechanism per remote chain and updates only `CCTP_V1` or `CCTP_V2` lanes to `CCTP_V2_WITH_CCV`. It skips unset, lock-release, and already-CCV lanes. 
- Also adds a TODO in `MigrateChainLanesToV2` for future prod-router support.

Changes
  - resolves USDCTokenPoolProxy from the datastore. No proxy ref means no-op.
  - Per remote lane, it reads `getLockOrBurnMechanism`
  - It updates only `CCTP_V1` or `CCTP_V2` to `CCTP_V2_WITH_CCV`
  - It skips `INVALID_MECHANISM`, `LOCK_RELEASE`, and already-CCV.
  - It preflights `getPools()` and errors if the `cctpV2PoolWithCCV` pool is not set before proposing/executing the update.
  - Add test

### DIFF
--- a/chains/evm/deployment/v2_0_0/sequences/cctp/configure_cctp_chain_for_lanes.go
+++ b/chains/evm/deployment/v2_0_0/sequences/cctp/configure_cctp_chain_for_lanes.go
@@ -36,10 +36,16 @@ import (
 )
 
 const (
-	mechanismCCTPV1        = "CCTP_V1"
-	mechanismCCTPV2        = "CCTP_V2"
-	mechanismLockRelease   = "LOCK_RELEASE"
-	mechanismCCTPV2WithCCV = "CCTP_V2_WITH_CCV"
+	MechanismCCTPV1        = "CCTP_V1"
+	MechanismCCTPV2        = "CCTP_V2"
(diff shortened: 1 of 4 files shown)


chains/evm/deployment/v2_0_0/sequences/cctp/configure_cctp_chain_for_lanes.go
```diff
@@ -921,19 +927,19 @@ func applyCCTPV1PoolSetDomainsWrites(b cldf_ops.Bundle, chain evm.Chain, poolAdd
 
 func convertMechanismToUint8(mechanism string) (uint8, error) {
```
RensR, line 928: out of scope here but maybe we should nuke the string mechanisms in the future

RensR (approved):

simsonraj (approved):

PR #2296, opened 2026-09-03
fix: token expansion - reduce RPC flakes and resolve edge case with reverse propagation

chris-de-leon-cll:
Two related fixes to the EVM token-pool setup tooling.

# Reduce RPC flakes (token role helpers)

## Purpose

The grant/revoke/has-admin helpers on BurnMintERC20 previously talked to the token contract through the raw go-ethereum bindings (`bind.CallOpts` + a bare `context.Context`), which bypassed the deployment framework's operation layer. That made reads and writes vulnerable to transient RPC failures (connection resets, skewed estimate views after a prior op in the same sequence, etc.), surfacing as hard errors instead of being retried. This routes them through the cldf operations framework with bounded retry so flakiness is absorbed by backoff rather than failing the changeset.

## What it adds

- `HasAdminRole` on the `Token` interface now takes `cldf_ops.Bundle` instead of `context.Context`, so token types issue reads/writes through the operations framework. All implementations (BurnMint, both WithDrip variants, ERC677, ERC20, TIP-20) are updated; TIP-20 keeps its direct binding call (no cldf operation exists for it) but sources its context from the Bundle.
- A `HasRole` read operation for BurnMintERC20 (`burn_mint_erc20:has-role`), so `hasDefaultAdminRoleBurnMintERC20` is a framework read rather than a raw binding call.
- A small generic `getRetryConfig[T]` helper that builds the retry config once and wires it into the operations it's used with, rather than duplicating the retry block per helper.

## Intentional design choices

- **Retry is bounded and scales with attempt.** Five attempts with backoff of `(attempt+1) × 2s` — a short, block-scale wait that lets the RPC node's view catch up when a change made earlier in the same sequence isn't yet visible to the estimate view.
- **Backoff is context-cancellation-aware.** The wait is a `select` between the bundle's context and `time.After`, deliberately avoiding `time.Sleep`, so a cancelled run unblocks immediately.
- **Retry applies only to operations that are idempotent by value.** Role grants/revokes and reads are safe to re-prepare. `transferTokensERC20` is intentionally left unwrapped — a transfer moves value and is not idempotent by value, so retrying after an unclear outcome risks a double transfer. That omission is documented at the call site.
- **The mutable `HasRole` read is force-executed to stay fresh.** The framework reuses previous successful reports within a bundle by default; the `HasRole` check bypasses that (`WithForceExecute`) because role membership can change and it gates an idempotency decision. The `GetDefaultAdminRole` read is constant, so it keeps the cache. Both still get the retry config.
- **Writes and permission checks share one execution path.** Threading the `Bundle` through `HasAdminRole` means the grant/revoke sequences and their role reads use the same execution and retry semantics as the rest of the deployment framework.

# Correct batched reverse-propagation skip

## Purpose

During an `AutoMigrateRemoteChains` batch, a migrated pool reverse-propagates its new pool into every counterpart so the web stays reachable in both directions. Reverse-propagation must not run at all into a counterpart in two cases; previously only one was handled, and the other caused the MCMS proposal to revert at execution.

## The two skip cases

1. **The counterpart is being replaced this batch** — its brand-new pool isn't live yet, so reverse-propagating would prematurely activate it and break that counterpart's own migration; its forward config wires its web instead. Skipped when the counterpart's target pool differs from its pre-batch active pool (or it has no active pool).
2. **The counterpart's own forward config already adds this migration's pool** — the peer is configured to connect to the migrating chain in this same batch, so its forward configuration adds the new pool on its side; reverse-propagating then adds it a second time and the on-chain `addRemotePool` reverts with `PoolAlreadyAdded`. Skipped when the migrating chain selector appears in the peer's `RemoteChains`.

Only the first was previously handled. The missed second case happened with a reusing peer that was itself forward-configured in the same batch (both sides of a connection end up in the changeset): the new pool was added twice — once by the peer's own forward config, once by this migration's reverse-propagation — and the batch reverted at execution.

Both decisions are driven by on-chain pool state and the peer's own configuration rather than any version- or family-specific interface, so they stay correct across EVM versions and Solana.

### DIFF
--- a/chains/evm/deployment/tokens/tokenimpl/helpers.go
+++ b/chains/evm/deployment/tokens/tokenimpl/helpers.go
@@ -1,12 +1,10 @@
 package tokenimpl
 
 import (
-	"context"
 	"fmt"
 	"math/big"
 	"time"
 
-	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 
 	"github.com/smartcontractkit/chainlink-ccip/chains/evm/deployment/v1_0_0/operations/burn_mint_erc20"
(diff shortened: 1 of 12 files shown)


chains/evm/deployment/tokens/tokenimpl/helpers.go
```diff
@@ -1,118 +1,198 @@
...
-		Args:          ccipAdmin.Hex(),
-	})
+	report, err := cldf_ops.ExecuteOperation(
+		b, burn_mint_erc20.SetCCIPAdmin, chain,
+		contract.FunctionInput[string]{
+			ChainSelector: chain.Selector,
+			Address:       token,
+			Args:          ccipAdmin.Hex(),
+		},
+		cldf_ops.WithRetryConfig(getRetryConfig[string](b, chain, token.Hex())),
```
huangzhen1997, line 170: Just curious, should we retry BurnMint here? is it idempotent ?

chris-de-leon-cll: it is indeed idempotent, so retrying it should be safe:

https://github.com/smartcontractkit/chainlink-evm/blob/2dca02f24e983863e72dfce3d715c716c23f6016/contracts/src/v0.8/shared/token/ERC20/BurnMintERC20.sol#L154-L166

if it does go through, then the result would be cached by `ExecuteOperation` anyways, so this retry config mostly helps in situations where submitting the initial tx fails, which can happen if the token is deployed, and we try to call methods on it too quickly (i.e. before the chain/RPC has had time to process the deployment)

huangzhen1997 (approved):

chains/evm/deployment/tokens/tokenimpl/helpers.go
huangzhen1997: Just realized it's set admin, overlooked and thought it's something else. Sounds good, thanks.

tt-cll (approved):

crispymangoes (approved):

PR #2302, opened 2026-09-08
[TOOLING] Add router support to the ConfigureTokenPool changeset

futureproof1:
## Summary

Adds a `Router *string` field to `ConfigureTokenPool` so a token pool's router can be updated declaratively. The value is a plain address in the chain family's own format (no datastore ref resolution), matching how `RateLimitAdmin` and `FeeAdmin` are already modeled. Follow-up to CCIP-12132.

Supported: **EVM only** — v1.5.1, v1.6.1, v2.0.0. Router is set via `setRouter` on pre-2.0 pools and via `setDynamicConfig` (with both admins) on 2.0+. Solana is deliberately **out of scope**: there is a single router program per chain that is upgraded in place, so repointing a pool at a different router is not meaningful — the Solana adapter rejects a non-nil `Router`.

## Notable: Solana router updates are unsupported (not just unimplemented)

A Solana `set_router` path is intentionally not added. Two things would otherwise need to be handled:
- the on-chain `AdminUpdateTokenPool` context requires a single signer that is both the pool owner and the token pool program's **upgrade authority**;
- token pool programs before **solana-v1.6.2** declare `AdminUpdateTokenPool.state` without `mut`, so `set_router` would succeed on-chain while silently discarding the write (fixed in the contracts in #2046).

Attempting router or fee-admin updates on Solana now fails fast with a clear error and no transaction/proposal.

## Other changes

- `Router` is validated in the shared EVM sequence at **apply time** (chain-agnostic changeset stays structural-only at verify): malformed hex is rejected and the family zero address is rejected (it always reverts on-chain — `ZeroAddressNotAllowed` on pre-2.0 `setRouter`, `ZeroAddressInvalid` on 2.0+ `setDynamicConfig`), rather than emitting a transaction/proposal that is certain to fail. Covers v1.5.1, v1.6.1 and v2.0.0 in one place.
- Renamed for clarity since a non-admin (router) was added:
  - interface `TokenPoolAdminAdapter` → `TokenPoolDynamicConfigAdapter`, method `SetTokenPoolAdmins` → `SetTokenPoolDynamicConfig`, input `SetTokenPoolAdminsSequenceInput` → `SetTokenPoolDynamicConfigSequenceInput`;
  - EVM `PoolOps.SetAdmins` → `SetDynamicPoolConfigs`.
- Added a v1.5.1 `SetRouter` write op (previously had no router-update path).
- Updates to a single pre-2.0 pool now emit a separate write per changed field, each guarded by a read-compare no-op.
- Removed an unused v1.5.1 `GetRouter` read op.

## Testing

- `deployment`: `go test ./...` passes (463 tests).
- EVM integration: `TestConfigureTokenPool_{VerifyPreconditions,Router,Router_PreV2,Admins,Admins_PreV2,MCMSOwnedPool,FeeConfig*}` pass. New coverage:
  - `TestConfigureTokenPool_Router` (v2.0.0) and `Router_PreV2` (v1.5.1 + v1.6.1): router set, admins preserved, idempotent re-apply sends no transaction, and malformed/zero router rejected before any transaction.
- Solana integration: `TestConfigureTokenPool_UnsupportedFields_Solana` asserts non-nil `Router` is rejected with "router is not supported".

## Follow-ups (out of scope)

- Solana `set_router` support (requires the writable/upgrade-authority handling above) if repointing a Solana pool ever becomes meaningful.
- Resolving `tokenPoolRef` and `Router` in the datastore at verify time (currently verify is structural-only; `Router` is a raw address by design).


chris-de-leon-cll:
Really appreciate the in-depth description! The findings presented there have actually changed my assessment of the ticket scope: after reviewing the code, I think we should drop Solana from this PR and simply hard error if someone tries to invoke v1.6 `set_router`.

Why: on Solana, the router semantics are very different from EVM. Most notably, there's currently only one router contract (which doubles as the token admin registry), and I am pretty sure that it's upgraded in place, so for v1.6 I don't think the concept of "setting the router" translates naturally to this chain family.

Once Solana is dropped from the PR, the diff size should be reduced by quite a bit! I'll update the ticket and focus the review on the EVM side, so that we'll be good to go there as well

deployment/tokens/configure_token_pool.go
```diff
@@ -51,6 +52,18 @@ type PoolConfigUpdate struct {
...
+	// with Type defaulting to the production Router contract type when unset (on EVM, set Type
+	// to the TestRouter contract type to target the test router; Solana has no TestRouter).
+	// Same semantics as TokenExpansionInputPerChain.RouterRef. Like TokenPoolRef, a datastore
+	// lookup that finds no match is reported at apply time, not by VerifyPreconditions.
+	//
+	// Solana: set_router requires the pool owner to also be the token pool program's upgrade
+	// authority, and only persists on programs >= solana-v1.6.2 (earlier releases accept the
+	// instruction without writing the state). The adapter checks both before sending or
+	// proposing anything and fails with a descriptive error.
+	RouterRef *datastore.AddressRef `yaml:"routerRef,omitempty" json:"routerRef,omitempty"`
```
chris-de-leon-cll: Let's follow the same pattern we did for `RateLimitAdmin` and `FeeAdmin` and do:

```suggestion
	Router *string `yaml:"router,omitempty" json:"router,omitempty"`
```

deployment/tokens/configure_token_pool.go
```diff
@@ -215,3 +250,66 @@ func configureTokenPoolApply() func(cldf.Environment, ConfigureTokenPoolInput) (
...
+		if b != 0 {
+			allZero = false
+			break
+		}
+	}
+	if allZero {
+		return "", fmt.Errorf("router address for chain selector %d must not be zero", selector)
+	}
+	return normalized, nil
+}
```
chris-de-leon-cll: This can most likely be removed once the router is typed as a `*string`

Also, generally speaking, let's be careful about checking for the zero address in the top level changeset since each chain has its own concept of it - ideally we should either introduce a new chain-agnostic interface for safely getting the zero address for a chain family OR let the underlying chain implementations deal with it

In fact, the code that checks for a zero address above is a bit buggy since Solana's zero address is `1111...` in base58 and on EVM it is `0x0000...`

chains/solana/deployment/v1_6_0/sequences/tokens.go
```diff
@@ -728,16 +734,42 @@ func (a *SolanaAdapter) SetTokenPoolAdmins() *cldf_ops.Sequence[tokenapi.SetToke
...
-			if err != nil {
-				return sequences.OnChainOutput{}, fmt.Errorf("failed to set rate limit admin on token pool %s on chain %d: %w", input.TokenPoolRef.Address, input.Selector, err)
+			var result sequences.OnChainOutput
+
+			if input.Router != nil {
+				// Router is validated (format + non-zero) by the changeset before it reaches the adapter.
+				newRouter, err := solana.PublicKeyFromBase58(*input.Router)
+				if err != nil {
+					return sequences.OnChainOutput{}, fmt.Errorf("invalid router address for chain %d: %s: %w", input.Selector, *input.Router, err)
+				}
```
chris-de-leon-cll: Solana code is out of scope, but flagging for informational purposes: here is where we should be checking for the zero address - we shouldn't be delegating to the top-level changeset

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
```diff
@@ -39,12 +39,13 @@ type PoolOps interface {
...
-	// admins in a single SetDynamicConfig write. Returns no writes when on-chain
-	// state already matches.
-	SetAdmins(b cldf_ops.Bundle, chain evm.Chain, poolAddr common.Address, rlAdmin, feeAdmin *common.Address) ([]evm_contract.WriteOutput, error)
+	// SetAdmins updates the admin roles and, when non-nil, the router on the pool. A nil
+	// pointer means "leave this value unchanged". Implementations own version-specific
+	// semantics: pre-2.0 pools reject a non-nil feeAdmin (no such concept on the
+	// contract) and set the router via setRouter, while v2.0+ sets router and both
+	// admins in a single SetDynamicConfig write. Returns no writes when on-chain state
+	// already matches.
+	SetAdmins(b cldf_ops.Bundle, chain evm.Chain, poolAddr common.Address, router, rlAdmin, feeAdmin *common.Address) ([]evm_contract.WriteOutput, error)
```
chris-de-leon-cll: The name of this function is starting to get a bit misleading now:

- Previous behavior: the rate limit admin and fee admin were the only things this modified (so the function name was appropriate)
- Now, it sets the admins AND the router (a non-admin)

It'd probably be good to do a rename in this PR to keep things more clear for future contributors

I think something like `SetDynamicPoolConfigs` would bring the naming in line with CCIP v2 and leave the door open for any additional dynamic configs that could get added in the future

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
```diff
@@ -199,16 +200,16 @@ func (a *EVMPoolAdapter) SetTokenPoolRateLimits() *cldf_ops.Sequence[tokensapi.T
...
-// capability lives in PoolOps.SetAdmins: pre-2.0 pools support only the rate limit
-// admin (a non-nil FeeAdmin is rejected there), while v2.0+ pools set both admins in
-// a single SetDynamicConfig write. No-op (zero BatchOps) when the desired values
-// already match on-chain state.
+// SetTokenPoolAdmins updates the router, rate limit admin, and fee admin on an EVM token pool.
+// Nil fields are left unchanged. Pre-2.0 pools only support the router (via setRouter) and the
+// rate limit admin (a non-nil FeeAdmin is rejected there), while v2.0+ pools set router and
+// both admins in a single SetDynamicConfig write. No-op (zero BatchOps) when the desired
+// values already match on-chain state.
 func (a *EVMPoolAdapter) SetTokenPoolAdmins() *cldf_ops.Sequence[tokensapi.SetTokenPoolAdminsSequenceInput, sequences.OnChainOutput, cldf_chain.BlockChains] {
```
chris-de-leon-cll: The naming here is also starting to get confusing 😅 mind if we change this

```go
type TokenPoolAdminAdapter interface {
	SetTokenPoolAdmins() *cldf_ops.Sequence[SetTokenPoolAdminsSequenceInput, sequences.OnChainOutput, cldf_chain.BlockChains]
}
```

To something like this?

```go
type TokenPoolDynamicConfigAdapter interface {
	SetTokenPoolDynamicConfig() *cldf_ops.Sequence[SetTokenPoolDynamicConfigSequenceInput, sequences.OnChainOutput, cldf_chain.BlockChains]
}
```

Notice the input rename as well: `SetTokenPoolAdminsSequenceInput` -> `SetTokenPoolDynamicConfigSequenceInput`

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
```diff
@@ -219,6 +220,15 @@ func (a *EVMPoolAdapter) SetTokenPoolAdmins() *cldf_ops.Sequence[tokensapi.SetTo
 			}
 			poolAddr := common.HexToAddress(input.TokenPoolRef.Address)
 
+			// Router is validated (format + non-zero) by the changeset before it reaches the adapter.
+			var router *common.Address
+			if input.Router != nil {
+				if !common.IsHexAddress(*input.Router) {
+					return sequences.OnChainOutput{}, fmt.Errorf("invalid router address for chain %d: %s", input.Selector, *input.Router)
+				}
```
chris-de-leon-cll, line 227: Let's also block the zero address (it'll revert on-chain):

https://github.com/smartcontractkit/chainlink-ccip/blob/c788a7e26273b5e6ea15f39ca82a2cad95a785e1/chains/evm/contracts/pools/TokenPool.sol#L235-L246

chris-de-leon-cll (commented):
The overall logic looks pretty solid 👏  main action items would be a few renaming updates to increase code readability for the future and removing the Solana stuff for now

deployment/tokens/configure_token_pool.go
futureproof1: @chris-de-leon-cll to replace `RouterRef *datastore.AddressRef` with `Router *string`, that drops datastore resolution, is that ok to do?

`deployment/tokens/token_expansion.go:141` declares `RouterRef *datastore.AddressRef` in the same package, while its `FeeAdmin` is a plain `string` — so the codebase already distinguishes 'admin = raw address' from 'router = datastore ref'.

chains/solana/deployment/v1_6_0/sequences/tokens.go
futureproof1: done

futureproof1:
Solana:

What I removed: the two Solana operations that performed the change, their shared helper, the version constants they needed, and their tests. What replaced it: the Solana code now returns a clear error if someone supplies a router, exactly like the existing error for a fee admin.

chains/solana/deployment/v1_6_0/sequences/tokens.go
futureproof1: No longer applies — Solana is out of this PR.

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
futureproof1: Renamed to SetDynamicPoolConfigs, as suggested.

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
futureproof1: All three renamed as suggested, plus two internal labels for consistency.

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
futureproof1: Done, in one place that covers all three EVM versions.
The check now lives in the shared EVM code, so 1.5.1, 1.6.1 and 2.0.0 are all covered by a single guard rather than three copies.

deployment/tokens/configure_token_pool.go
futureproof1: @chris-de-leon-cll i went ahead and did it in this commit https://github.com/smartcontractkit/chainlink-ccip/pull/2302/changes/6fd1f477040e093ec49593844c55b52aab5c90ec
let me know if its ok

chris-de-leon-cll: yep it's intentional to remove the router ref thank you for flagging - same reasoning as https://github.com/smartcontractkit/chainlink-ccip/pull/2284#discussion_r3856309707

chris-de-leon-cll (approved):

0xsuryansh (approved):

PR #2303, opened 2026-09-08
CCIP-13390: Validate lane prerequisites before configuring a v2 lane

athegaul:
Lane expansion now fails during validation instead of mid-apply when a lane chain is missing from the committee or executor pool chain_configs, when a contract address or committee signer cannot be resolved, or when the destination FeeQuoter has no gas price. It also refuses to lower an existing baseExecutionGasCost unless the new allowLoweringBaseExecutionGasCost flag is set, so a lane re-run cannot reset a hand-raised value back to the default.

### DIFF
--- a/chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
+++ b/chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
@@ -40,6 +40,11 @@ const NoExecutionAddress = "0xEBa517d200000000000000000000000000000000"
 // lane config supplies none, so sends on the lane would revert with NoGasPriceAvailable.
 var ErrNoDestGasPrice = errors.New("fee quoter has no gas price for destination chain")
 
+// ErrInvalidGasPrice means the lane config supplied a USDPerUnitGas that is not a usable
+// gas price. A non-positive value (e.g. 0) is not a price: writing it on-chain would either
+// be rejected or silently break the lane, so it is refused rather than ignored.
+var ErrInvalidGasPrice = errors.New("usdPerUnitGas must be a positive value")
+
 // ErrBaseExecutionGasCostLowered means the lane config would write a BaseExecutionGasCost
 // below the value already on the OnRamp without AllowLoweringBaseExecutionGasCost set.
 var ErrBaseExecutionGasCostLowered = errors.New("refusing to lower baseExecutionGasCost")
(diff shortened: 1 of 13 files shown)


chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
```diff
@@ -137,21 +146,31 @@ var ConfigureChainForLanes = cldf_ops.NewSequence(
...
+				ChainSelector: chain.Selector,
+				Address:       feeQuoterAddr,
+				Args:          remoteSelector,
+			})
+			if err != nil {
+				return seqtypes.OnChainOutput{}, fmt.Errorf("failed to get gas prices on FeeQuoter(%s) on chain %s: %w", feeQuoterAddr, chain, err)
+			}
+			onChainGasPrice := gasPriceReport.Output.Value
+			switch desired := remoteConfig.FeeQuoterDestChainConfig.USDPerUnitGas; {
+			case desired != nil:
```
simsonraj: if this BigInt is transformed/translated in an earlier layer such as CLD resolvers, for cases even if its 0, this will always be successful, the case might be even Yaml -> struct conversion, so the error branch is never evauluated

simsonraj: this can also be bypassed by deliberately setting zero value in the input file

chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
```diff
@@ -137,21 +146,31 @@ var ConfigureChainForLanes = cldf_ops.NewSequence(
...
+				if onChainGasPrice == nil || desired.Cmp(onChainGasPrice) != 0 {
```
simsonraj: onChainGasPrice can never be nil right? because of the binding?

chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
```diff
@@ -137,21 +146,31 @@ var ConfigureChainForLanes = cldf_ops.NewSequence(
...
+			gasPriceReport, err := cldf_ops.ExecuteOperation(b, fee_quoter.GetDestinationChainGasPrice, chain, contract.FunctionInput[uint64]{
+				ChainSelector: chain.Selector,
+				Address:       feeQuoterAddr,
+				Args:          remoteSelector,
+			})
+			if err != nil {
+				return seqtypes.OnChainOutput{}, fmt.Errorf("failed to get gas prices on FeeQuoter(%s) on chain %s: %w", feeQuoterAddr, chain, err)
+			}
+			onChainGasPrice := gasPriceReport.Output.Value
+			switch desired := remoteConfig.FeeQuoterDestChainConfig.USDPerUnitGas; {
```
simsonraj, line 169: also this can be a simple if else no?

chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
athegaul: What would be the minimum acceptable value that one should provide? Considering that 0 is valid but causes issues once it goes on-chain.

chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
athegaul: I kept this as a condition-based `switch` rather than `if-else` to keep the error case close to the lookup instead of nested `else`. The [Google Go Styleguide](https://google.github.io/styleguide/go/decisions#indent-error-flow) provides more details about it. I can convert it to `if-else` if you prefer it that way.

chains/evm/deployment/v2_0_0/sequences/configure_chain_for_lanes.go
athegaul: You are right, `onChainGasPrice` is never `nil` from the binding. I verified this and removed the `nil` check. Added a comment explaining that `Sign() == 0` is the real signal.

simsonraj (dismissed):

b-gopalswami (dismissed):

tt-cll (approved):

chris-de-leon-cll (approved):

PR #2305, opened 2026-09-09
Implement change and add test file

crispymangoes:


chains/evm/deployment/v2_0_0/sequences/tokens/migrate_lock_release_pool_liquidity.go
```diff
@@ -317,6 +354,30 @@ func migrateSiloedPool(
...
+	exactMode := isExactSiloMode(input)
+
+	// In exact mode, every siloed chain must be explicit - a chain silently falling back to zero
+	// migration is more likely to be an oversight than intent, so surface it before any write is
+	// emitted rather than migrating a partial set of silos.
+	if exactMode {
+		var missingSilos []uint64
+		for _, remoteChain := range supportedChains {
+			if !isSiloedByChain[remoteChain] {
+				continue
```
chris-de-leon-cll: might be worth it to fail loudly here - if the user provides a SiloExactAmount whose ChainSelector is not actually a siloed chain of the old pool, then the code will silently ignore it and leave no trace of this in the pipeline logs (which will make it harder to review the execution logs if we ever need to debug)

chains/evm/deployment/v2_0_0/sequences/tokens/migrate_lock_release_pool_liquidity.go
```diff
@@ -95,6 +117,21 @@ func validateMigrationInput(input tokens.MigrateLockReleasePoolLiquidityInput) e
 	if input.Amount != nil && input.Amount.Sign() <= 0 {
 		return fmt.Errorf("Amount must be positive")
 	}
```
chris-de-leon-cll: We may want to consider slightly adjusting the logic so that it looks like this:

```suggestion
	if input.Amount != nil && input.Amount.Sign() < 0 {
		return fmt.Errorf("Amount must be positive")
	}
```

Reason: suppose we have a siloed LockRelease pool on chain X, and it has two siloed chains, A and B, plus an unsiloed shared bucket:

| bucket          | balance |
|-----------------|---------|
| A               | 1000    |
| B               | 0       |
| unsiloed shared | 500     |

If the operator wants an exact migration of everything, they may try:

```
SiloExactAmounts: []SiloExactAmount{
    {ChainSelector: A, Amount: 1000},
    {ChainSelector: B, Amount: 0},   // B has nothing, migrate nothing
},
UnsiloedExactAmount: 500,
```

However, this would fail even though B naturally maps to "do nothing".

chains/evm/deployment/v2_0_0/sequences/tokens/migrate_lock_release_pool_liquidity.go
```diff
@@ -70,22 +70,44 @@ var MigrateLockReleasePoolLiquidity = cldf_ops.NewSequence(
...
+		}
 		return migrateUnsiloedPool(b, evmChain, input, oldPoolAddr, newPoolAddr, tokenAddr, timelockAddr)
 	},
 )
 
+// isExactSiloMode reports whether the input requests exact-amount migration (per-silo and/or
+// unsiloed shared bucket) rather than the legacy Amount/BasisPoints proportional mode.
+func isExactSiloMode(input tokens.MigrateLockReleasePoolLiquidityInput) bool {
+	return len(input.SiloExactAmounts) > 0 || input.UnsiloedExactAmount != nil
+}
```
chris-de-leon-cll, line 86: Small edge case: if an operator wants to migrate only the shared (unsiloed) bucket and leave the per-chain silos alone (e.g., they were already drained or a separate step handles them), then they might try:

```
MigrateLockReleasePoolLiquidityInput{
    UnsiloedExactAmount: big.NewInt(500), // only the shared bucket
    // SiloExactAmounts left empty on purpose
}
```

With the current impl:
- isExactSiloMode → true
- Missing-silo check → SiloExactAmounts is empty ⇒ error: "missing entries for siloed chains A B …"

So seems like there's no way to express "I just want to migrate unsiloed liquidity"? From what I understand, you'd also enumerate and migrate every silo as well, and since the per-silo loop migrates each listed silo, you can't scope an exact migration to only the shared bucket.

It seems like adding support for unsiloed only would be a more broader change, so it may be simpler to reject UnsiloedExactAmount alone unless SiloExactAmounts is also present - that way it's strictly a companion (no behavior change, just tighten isExactSiloMode/validation)

chris-de-leon-cll (commented):
Really nice work 💯  appreciate the thorough test cases! Only had a few minor comments

chains/evm/deployment/v2_0_0/sequences/tokens/migrate_lock_release_pool_liquidity.go
```diff
@@ -95,6 +117,21 @@ func validateMigrationInput(input tokens.MigrateLockReleasePoolLiquidityInput) e
...
+	if exactMode {
+		seen := make(map[uint64]bool, len(input.SiloExactAmounts))
+		for i, sa := range input.SiloExactAmounts {
+			if seen[sa.ChainSelector] {
+				return fmt.Errorf("duplicate ChainSelector %d in SiloExactAmounts", sa.ChainSelector)
+			}
+			seen[sa.ChainSelector] = true
+			if sa.Amount == nil || sa.Amount.Sign() <= 0 {
+				return fmt.Errorf("SiloExactAmounts[%d].Amount must be positive", i)
+			}
```
chris-de-leon-cll: may want to consider slightly adjusting the logic so that it looks like this:

```suggestion
			if sa.Amount == nil || sa.Amount.Sign() < 0 {
				return fmt.Errorf("SiloExactAmounts[%d].Amount must be positive", i)
			}
```

Reason: suppose we have a siloed LockRelease pool on chain X, and it has two siloed chains, A and B, plus an unsiloed shared bucket:

| bucket          | balance |
|-----------------|---------|
| A               | 1000    |
| B               | 0       |
| unsiloed shared | 500     |

If the operator wants an exact migration of everything, they may try:

```
SiloExactAmounts: []SiloExactAmount{
    {ChainSelector: A, Amount: 1000},
    {ChainSelector: B, Amount: 0},   // B has nothing, migrate nothing
},
UnsiloedExactAmount: 500,
```

However, this would fail even though B naturally maps to "do nothing", so making the check more lenient would help resolve this situation.

chris-de-leon-cll: @crispymangoes the other ones look resolved but seems like this one's still open 👀 was there an issue with implementing it or still in progress?

chris-de-leon-cll (approved):

Alkhara (approved):

PR #2307, opened 2026-09-10
CCIP-12738: Fix IsAllowedCaller as OnlyOwner for BurnMintERC677

b-gopalswami:


chris-de-leon-cll (approved):

chains/evm/deployment/v1_0_0/operations/burn_mint_erc677/burn_mint_erc677.go
```diff
@@ -111,15 +112,13 @@ func PrepareGrantMintAndBurnRoles(
...
-	// On-chain only the owner may call grantMintAndBurnRoles. Do not use OnlyOwner here:
-	// MCMS/timelock flows simulate with the deployer key while ownership is the timelock
-	IsAllowedCaller: contract.AllCallersAllowed[*burn_mint_erc677.BurnMintERC677, common.Address],
+	Name:            "burn_mint_erc677:grant-mint-and-burn-roles",
+	Version:         cciputils.Version_1_0_0,
+	Description:     "Grant mint and burn roles on BurnMintERC677 (owner-only on-chain)",
+	ContractType:    ContractType,
+	ContractABI:     burn_mint_erc677.BurnMintERC677ABI,
+	NewContract:     burn_mint_erc677.NewBurnMintERC677,
+	IsAllowedCaller: contract.OnlyOwner[*burn_mint_erc677.BurnMintERC677, common.Address],
```
simsonraj, line 121: the comment specifically says not to use onlyowner here

	// On-chain only the owner may call grantMintAndBurnRoles. Do not use OnlyOwner here:
	// MCMS/timelock flows simulate with the deployer key while ownership is the timelock

simsonraj (changes requested):

simsonraj (approved):

chains/evm/deployment/v1_0_0/operations/burn_mint_erc677/burn_mint_erc677.go
b-gopalswami: Yeah that is a wrong comment. For BNM677, we don't have granting admin role options. Onlyowner can provide burn/mint roles.
Whereas, if it BNMERC20 means, we can provide admin role to someone who can grant the access. Our implementation for those will be looking something like this.
https://github.com/smartcontractkit/chainlink-ccip/blob/e35d9898c782fdc046920051c70ef8e34627714c/chains/evm/deployment/v1_0_0/operations/burn_mint_erc20/burn_mint_erc20.go#L66-L85

	IsAllowedCaller: func(token *burn_mint_erc20.BurnMintERC20, opts *bind.CallOpts, caller common.Address, input RoleAssignment) (bool, error) {
		// Check if caller has the admin role for the role being granted
		roleAdmin, err := token.GetRoleAdmin(opts, input.Role)
		if err != nil {
			return false, err
		}
		return token.HasRole(opts, roleAdmin, caller)
	},

0xsuryansh (approved):

PR #2313, opened 2026-09-15
feat(cctp): fall back to Circle-defined CCTP addresses when config fields are empty

0xsuryansh:
Adds a chain-selector-indexed table of Circle-defined CCTP values and uses it  as a fallback for the `DeployCCTPChains` changeset. Callers can now leave the well-known Circle addresses unset; explicit values still take precedence.  

Fallback added for Create2Factory as well 

Addresses from 
- https://developers.circle.com/cctp/references/contract-addresses
- https://developers.circle.com/cctp/v1/evm-smart-contracts
- https://developers.circle.com/cctp/v1/solana-programs
- https://developers.circle.com/stablecoins/usdc-contract-addresses


deployment/v2_0_0/changesets/deploy_cctp_chains.go
```diff
@@ -59,8 +61,91 @@ func DeployCCTPChains(cctpChainRegistry *adapters.CCTPChainRegistry, mcmsRegistr
...
+	if len(refs) == 0 {
+		return ""
+	}
+	return refs[0].Address
+}
+
+func withCCTPChainDefaults(chainSel uint64, chainCfg CCTPChainConfig) CCTPChainConfig {
+	// Circle-defined addresses only apply to canonical USDC chains. Non-canonical
+	// chains use their own token and do not interact with Circle's contracts.
+	if chainCfg.USDCType != adapters.Canonical {
```
crispymangoes: This returns early before running the RemoteChains defaulting loop, so DomainIdentifier never gets defaulted for those chains. Since nothing later validates DomainIdentifier != 0, a chain with USDCType: NonCanonical and an omitted DomainIdentifier on a remote chain entry will silently deploy with Circle domain 0(ETH's domain), which is the wrong routing config with no error.

0xsuryansh: https://github.com/smartcontractkit/chainlink-ccip/pull/2313/commits/6569874aa52048ba688d146d0e755685dd0cab67

crispymangoes (approved):

deployment/v2_0_0/config/cctp.go
chris-de-leon-cll, line 1: Wondering if it makes more sense to have this live in `chainlink-deployments` instead? 👀 

Asking because (1) that's where most of the deployment configs already live (so it'd be great to keep it all in one place to reduce confusion) and (2) if we ever need to extend or override the defaults for whatever reason, then it would be easier to do that from CLD rather than here

Maybe we can have the changeset accept an optional `DefaultsPerChain` field? Then in CLD, we'll add a thin resolver which provides this default config to that field. If the field is not provided, then we'd require the more verbose format.

0xsuryansh: For explicitly overriding we can use the fields itself, why go to a file and override. 

If left empty then if fallbacks to the hardcoded values, we already keep other config like finality defaults in this repo, https://github.com/smartcontractkit/chainlink-ccip/blob/main/chains/evm/deployment/v2_0_0/sequences/cctp/finality_defaults.go

Extending this changeset expands our touchpoints across downstream dependencies (such as chainlink-ccv), requiring cross-repo updates and maintenance.

chris-de-leon-cll (approved):

PR #2315, opened 2026-09-16
Add test showing issue

crispymangoes:


chains/evm/deployment/tokens/tokenimpl/helpers.go
```diff
@@ -176,6 +177,122 @@ func setCCIPAdminBurnMintERC20(b cldf_ops.Bundle, chain evm.Chain, token, ccipAd
...
+			Args:          struct{}{},
+		},
+		cldf_ops.WithRetryConfig(getRetryConfig[struct{}](b, chain, token.Hex())),
+	)
+	if err != nil {
+		return common.Address{}, fmt.Errorf("failed to get pending default admin: %w", err)
+	}
+
+	return report.Output, nil
+}
```
chris-de-leon-cll: might make sense to directly inline these into `token_burn_mint_erc20_transparent.go` - the BnM ERC20 was a special case bc it's ABI was genuinely shared with the drip token

deployment/tokens/token_expansion.go
```diff
@@ -71,8 +71,15 @@ type DeployTokenInput struct {
...
-	ChainSelector     uint64
-	ExistingDataStore datastore.DataStore
+	ChainSelector uint64
+	// ExternalAdminIsTimelock is true when ExternalAdmin was left empty by the caller and was
+	// defaulted to the chain's timelock below, rather than explicitly provided. EVM adapters that
+	// support a 2-step admin transfer (e.g. upgradeable BurnMintERC20 variants using
+	// AccessControlDefaultAdminRulesUpgradeable) use this to decide whether the second step
+	// (acceptance) can be safely queued into an MCMS proposal for the timelock to execute itself,
+	// versus a customer-provided address, where acceptance must happen out-of-band.
+	ExternalAdminIsTimelock bool
```
chris-de-leon-cll: Any chance we could follow the same pattern that was done for `DeployTokenPoolInput` and instead replace this with:

```suggestion
	TimelockAddress string `yaml:"-" json:"-"`
```

Then later:

```go
// token_expansion:
deployTokenInput.ExternalAdmin = defaulted... 

// sequence (token.go):
if caps.UsesAsyncRoleManagement && input.ExternalAdmin == input.TimelockAddress {
    acceptWrites = ...
}
```

Right now one edge case is that if someone (for whatever reason) sets `ExternalAdmin = timelock address` then the flag would be false causing the accept step to be skipped.

chains/evm/deployment/tokens/tokenimpl/token_burn_mint_erc20_transparent.go
```diff
@@ -0,0 +1,282 @@
...
+
+	proxyRef, err := contract.MaybeDeployContract(b, burn_mint_erc20_transparent.DeployProxy, chain,
+		contract.DeployInput[burn_mint_erc20_transparent.ProxyConstructorArgs]{
+			TypeAndVersion: deployment.NewTypeAndVersion(burn_mint_erc20_transparent.ContractType, *utils.Version_1_0_0),
+			ChainSelector:  chain.Selector,
+			Qualifier:      &in.Symbol,
+			Args: burn_mint_erc20_transparent.ProxyConstructorArgs{
+				Logic:        implAddr,
+				InitialOwner: proxyAdmin,
+				Data:         []byte{},
```
chris-de-leon-cll: Would it be a heavy lift to add the initialize data here?

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
```diff
@@ -560,6 +560,38 @@ func (a *EVMPoolAdapter) TidyTokenRoles(
...
+	// would revert on-chain once the first one completes the transfer and clears the pending state.
+	if tokenCaps.UsesAsyncRoleManagement {
+		pending, err := tokenImpl.PendingAdminRoleTarget(b, chain, tokenAddr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check pending admin role target for token %q on chain %d: %w", tokenAddr.Hex(), input.ChainSelector, err)
+		}
+		if pending == timelockAddr {
+			b.Logger.Infof("token %q on chain %d already has a pending admin transfer to timelock %q; skipping redundant tidy", tokenAddr.Hex(), input.ChainSelector, timelockAddr.Hex())
+			return nil, nil
+		}
```
chris-de-leon-cll: If there's already a pending admin let's not clobber it:

```suggestion
		if pending == timelockAddr {
			b.Logger.Infof("token %q on chain %d already has a pending admin transfer to timelock %q; skipping redundant tidy", tokenAddr.Hex(), input.ChainSelector, timelockAddr.Hex())
			return nil, nil
		}
		
		// A transfer to a customer-chosen admin is in flight; don't displace it.
		if pending != (common.Address{}) {
		    b.Logger.Warnf("token %q on chain %d has a pending admin transfer to %q (not the CLL timelock %q); leaving it untouched and skipping tidy",
		        tokenAddr.Hex(), input.ChainSelector, pending.Hex(), timelockAddr.Hex())
		    return nil, nil
		}
```

chains/evm/deployment/tokens/tokenimpl/token_burn_mint_erc20_transparent.go
```diff
@@ -0,0 +1,282 @@
...
+			Args:          struct{}{},
+		},
+		operations.WithRetryConfig(getRetryConfig[struct{}](b, chain, token.Hex())),
+	)
+	if err != nil {
+		return common.Address{}, fmt.Errorf("failed to get pending default admin: %w", err)
+	}
+
+	return report.Output, nil
+}
```
chris-de-leon-cll: Ah seems like these still weren't inlined? Feel free to move the function bodies into the interface methods themselves so that the file is a bit easier to read through

chains/evm/deployment/v1_0_0/adapters/pool_adapter.go
```diff
@@ -560,6 +560,38 @@ func (a *EVMPoolAdapter) TidyTokenRoles(
...
+		grantWrites, err := tokenImpl.GrantAdminRole(b, chain, tokenAddr, timelockAddr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to grant timelock admin role for token %q on chain %d: %w", tokenAddr.Hex(), input.ChainSelector, err)
+		}
+		acceptWrites, err := tokenImpl.AcceptAdminRole(b, chain, tokenAddr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to prepare accept of timelock admin role for token %q on chain %d: %w", tokenAddr.Hex(), input.ChainSelector, err)
+		}
+		return append(grantWrites, acceptWrites...), nil
+	}
```
chris-de-leon-cll, line 603: I'd tweak this just a bit so it doesn't clobber any in-flight transfers and handles any edge case where the deployer somehow doesn't have the `DEFAULT_ADMIN_ROLE` anymore (can happen for the existing token route)

```suggestion
	if tokenCaps.UsesAsyncRoleManagement {
		pending, err := tokenImpl.PendingAdminRoleTarget(b, chain, tokenAddr)
		if err != nil {
			return nil, fmt.Errorf("failed to check pending admin role target for token %q on chain %d: %w", tokenAddr.Hex(), input.ChainSelector, err)
		}
		// Don't clobber a transfer that is already in flight: if it targets the timelock, tidy is
		// already satisfied; if it targets a customer-chosen admin, that explicit config wins.
		if pending != (common.Address{}) {
			b.Logger.Infof("token %q on chain %d already has a pending admin transfer to %q; skipping tidy", tokenAddr.Hex(), input.ChainSelector, pending.Hex())
			return nil, nil
		}

		// Nothing is pending. Only begin a transfer when the deployer still holds
		// DEFAULT_ADMIN_ROLE: BeginDefaultAdminTransfer is deployer-signed, so the deployer must be
		// the current admin. If it isn't (e.g. a customer admin already accepted), queueing a begin
		// would revert when the timelock executes it.
		deployerHasRole, err := tokenImpl.HasAdminRole(b, chain, tokenAddr, chain.DeployerKey.From)
		if err != nil {
			return nil, fmt.Errorf("failed to check deployer admin role for token %q on chain %d: %w", tokenAddr.Hex(), input.ChainSelector, err)
		}
		if !deployerHasRole {
			b.Logger.Infof("deployer does not hold DEFAULT_ADMIN_ROLE on token %q on chain %d; skipping tidy", tokenAddr.Hex(), input.ChainSelector)
			return nil, nil
		}

		grantWrites, err := tokenImpl.GrantAdminRole(b, chain, tokenAddr, timelockAddr)
		if err != nil {
			return nil, fmt.Errorf("failed to grant timelock admin role for token %q on chain %d: %w", tokenAddr.Hex(), input.ChainSelector, err)
		}
		acceptWrites, err := tokenImpl.AcceptAdminRole(b, chain, tokenAddr)
		if err != nil {
			return nil, fmt.Errorf("failed to prepare accept of timelock admin role for token %q on chain %d: %w", tokenAddr.Hex(), input.ChainSelector, err)
		}
		return append(grantWrites, acceptWrites...), nil
	}
```

chains/evm/deployment/v1_0_0/operations/burn_mint_erc20_transparent/burn_mint_erc20_transparent.go
```diff
@@ -0,0 +1,277 @@
...
+
+func validateInitializeArgs(args InitializeArgs) error {
+	if args.DefaultAdmin == (common.Address{}) {
+		return fmt.Errorf("defaultAdmin must not be the zero address")
+	}
+	if args.PreMint != nil && args.MaxSupply != nil && args.MaxSupply.Sign() > 0 && args.PreMint.Cmp(args.MaxSupply) > 0 {
+		return fmt.Errorf("preMint (%s) exceeds maxSupply (%s)", args.PreMint, args.MaxSupply)
+	}
+	return nil
+}
```
chris-de-leon-cll, line 162: [This token contract](https://github.com/smartcontractkit/chainlink-evm/blob/2a2b38e8b74479054db92ab911521d60278c3d78/contracts/src/v0.8/shared/token/ERC20/upgradeable/BurnMintERC20Transparent.sol#L92-L97) seems to have special rules around pre mint and supply - might be worth it to tighten the validation to match:

```go
	if args.PreMint != nil && args.PreMint.Sign() > 0 {
		if args.MaxSupply == nil || args.MaxSupply.Sign() == 0 {
			return fmt.Errorf("preMint requires a bounded maxSupply: preMint (%s) cannot be minted against unlimited supply", args.PreMint)
		}
		if args.PreMint.Cmp(args.MaxSupply) > 0 {
			return fmt.Errorf("preMint (%s) exceeds maxSupply (%s)", args.PreMint, args.MaxSupply)
		}
	}
```

deployment/tokens/token_expansion.go
```diff
@@ -286,6 +294,7 @@ func tokenExpansionApply() func(cldf.Environment, TokenExpansionInput) (cldf.Cha
 					if datastore_utils.IsAddressRefEmpty(timelockRef) {
 						e.Logger.Warnf("timelock ref is empty for chain selector %d - adapter is expected to provide fallbacks for ExternalAdmin and/or CCIPAdmin", selector)
 					} else {
+						deployTokenInput.TimelockAddress = timelockRef.Address
```
chris-de-leon-cll: I think I found a potential bug here - have a look at this code:

```go
if deployTokenInput.CCIPAdmin == "" && deployTokenInput.ExternalAdmin != "" {
    deployTokenInput.CCIPAdmin = deployTokenInput.ExternalAdmin
}
if deployTokenInput.CCIPAdmin == "" || deployTokenInput.ExternalAdmin == "" {
    ...
    deployTokenInput.TimelockAddress = timelockRef.Address
    ...
}
```

If the caller explicitly sets ExternalAdmin (even to the actual CLL timelock) and leaves CCIPAdmin empty, then CCIPAdmin is autofilled, and the second block is skipped so `TimelockAddress` stays ""

Then in `token.go`, the `input.TimelockAddress != ""` guard fails, so the accept is never queued. If we use the updated `TidyTokenRoles` implementation that I commented (which skips when pending != 0), then the transfer to the timelock stays pending forever and the deployer keeps DEFAULT_ADMIN_ROLE indefinitely, silently. It also makes the comment at token.go:114-116 ("also covers the case where the caller explicitly passes the timelock's own address as ExternalAdmin") inaccurate for the token-expansion path.

To fix this, let's always resolve the timelock address, so right after checking `deployTokenInput != nil` above, get the timelock ref and do `deployTokenInput.TimelockAddress = timelockRef.Address`.

Once resolved up front, the later if `CCIPAdmin == "" || ExternalAdmin == ""` block shouldn't call `GetTimelockRef` again — reuse the resolved ref for the ExternalAdmin/CCIPAdmin defaulting.

.github/workflows/test_smoke.yml
```diff
@@ -105,8 +105,9 @@ jobs:
         uses: actions/setup-go@v6 # v6
         with:
           cache: true
-          go-version-file: devenv/go.mod
```
chris-de-leon-cll, line 108: was this change intentional? 👀

chains/evm/deployment/v1_0_0/sequences/token_test.go
```diff
@@ -243,15 +245,191 @@ func TestTokenSupportsAdminRole(t *testing.T) {
...
+
+	for tt, usesAsync := range tokenTypes {
+		require.Equal(t, usesAsync, tokenimpl.Capabilities(tt).UsesAsyncRoleManagement, "Token type %s async role management mismatch", tt)
+	}
+}
+
+// TestEVMTokenDeployment_BurnMintERC20TransparentToken exercises the composite deploy (impl +
+// TransparentUpgradeableProxy + initialize) via the DeployToken sequence, using its own
+// assertions rather than the shared BurnMintERC20 test loop above: this token type has a
+// different capability set (no SupportsAdminRole yet - see the adapter's Capabilities doc) and
```
chris-de-leon-cll, line 283: [non-blocking]: I think the BnM transparent impl does have `SupportsAdminRole` set to `true`

```suggestion
// different capability set and
```

chris-de-leon-cll (approved):
Amazing work on this!! 💯

Thank you for really diving into the design and making sure that everything aligned with the existing toolings flows 🙌 

Only had a couple of non-blocking questions, but overall this looks clean as-is :)

.github/workflows/test_smoke.yml
crispymangoes: Yes there was a go toolchain version mismatch CI failure and while fixing it I had to make this change

Alkhara (approved):

b-gopalswami (approved):

matYang (approved):

PR #2317, opened 2026-09-18
feat(fastcurse): carry MCMS qualifier into CurseInput for explicit-caller families

JohnChangUK:
## Description

Adds `MCMSQualifier string` to `fastcurse.CurseInput`, populated from `cfg.MCMS.Qualifier` at both
`CurseInput` construction sites (`applyCurse`, `applyUncurse`); the global-curse path inherits it via
`formCurseConfigForGlobalCurse`, which already copies `cfg.MCMS`.

## Why

EVM's `RMN.curse(bytes16[])` takes no caller argument, `msg.sender` authorizes, so one payload is valid under whichever of the three timelocks runs it and the MCMS qualifier alone picks the route.

Stellar's `curse(caller, subjects)` requires the caller argument to equal the invoking address — the
executing timelock is baked into the payload at proposal-build time. Without the qualifier in the shared
input, the Stellar sequence cannot know which stack will execute.

## Notes

- Inert for every sender-authorized family (EVM/Solana/Sui): no adapter reads the field.
- No behavior change, no migration; empty on direct/no-MCMS runs.
- A capture-adapter test asserts both changesets set the field.


friedemannf (approved):

deployment/fastcurse/common.go
```diff
@@ -16,6 +16,12 @@ func IfSubjectEqual(s, other Subject) bool {
 type CurseInput struct {
 	Subjects      []Subject
 	ChainSelector uint64
+	// MCMSQualifier is the qualifier of the MCMS stack that will execute the resulting
+	// proposal. Families whose RMN curse entrypoint takes an explicit caller argument
+	// (Stellar: curse(caller, subjects), where the caller must be the invoking timelock)
+	// need it to resolve that timelock at proposal-build time. Empty on direct /
+	// no-MCMS runs. Families that authorize by transaction sender ignore it.
```
carte7000, line 23: @toblich I think this can be used for Solana UFC as well

carte7000 (approved):

PR #2318, opened 2026-09-18
feat: CCTP USDC tooling improvements

chris-de-leon-cll:
## Context

The CLOPS `token_expansion_usdc_cross_family` payload drives CCTP deployment through this repo's `DeployCCTPChains` changeset. Several inputs were either required but unread, silently materialised from a hardcoded value, or demanded from callers even though the environment already knows them. This change moves those responsibilities into the tooling so the payload schema stays minimal and the deployed configuration reflects the chain's actual state.

## What it adds

- **Optional token transfer fee configuration.** `RemoteCCTPChainConfig.TokenTransferFeeConfig` changes from a value to a pointer to `PartialTokenTransferFeeConfig`. Omitting it leaves the on-chain configuration untouched, while a partially specified entry is completed against the current on-chain configuration (or a chain-aware default). This mirrors how token expansion already models the same field on `tokens.RemoteChainConfig`.
- **On-chain token decimals resolution.** A new `RemoteCCTPChain.TokenDecimals` method is implemented per chain family — EVM reads ERC20 `decimals()`, Solana reads the SPL mint — and the deploy changeset resolves a token's decimals from its address when they are not supplied, rather than assuming a fixed value. The `CanonicalUSDCDecimals` constant is removed.
- **Token transfer fee application on the FeeQuoter for every CCTP lane.** `tokens.ApplyTokenTransferFeeConfigOnFeeQuoter` is a deps-based entry point (bundle / blockchains / datastore) to the FeeQuoter fee resolution, and the canonical, non-canonical, and Solana configure paths all use it so per-lane fee behavior is the same across USDC types and chain families. v2 pools additionally carry their own per-lane fee config, applied by the v2 configure sequence.
- **Chain-type-aware `tokenMessengerV2` requirement.** The verify gate requires a valid `TokenMessengerV2` only for canonical chains. Non-canonical chains never interact with Circle's contracts, so a placeholder address is no longer demanded of them.
- **Fee aggregator defaulted from the environment.** For canonical chains without an explicit `FeeAggregator`, the chain's deployer key is used (hex for EVM, base58 for Solana). `applyCCTPDefaults` and `withCCTPChainDefaults` receive `BlockChains` so they can read it.
- **Registered pool inferred from the deploy output.** `CCTPChainConfig.RegisteredPoolRef` is removed, and `RegisteredPoolRef` / `RemoteRegisteredPoolRefs` move from `ConfigureCCTPChainForLanesInput` to its deps. The changeset derives each chain's registered pool from the first address returned by that chain's CCTP deploy and threads it into the configure deps (local and per-remote). The deploy sequences order their output so the registered pool is first — `USDCTokenPoolProxy` on canonical home chains, the CCTP V1 `USDCTokenPool` on non-home canonical chains, `BurnMintWithLockReleaseFlagTokenPool` on non-canonical chains, and the pre-existing `CCTPTokenPool` on Solana. This replaces a caller-supplied `{type, version}` pair that could mismatch the pool actually deployed or reused (e.g. a non-canonical pool reused at either 1.5.1 or 1.6.1).

## Intentional design choices

- The fee configuration is modelled as pointer-to-partial rather than a value plus a separate "present" flag, so absence and partial presence are expressed by the type itself, consistent with the token-expansion surface.
- Decimals are resolved in the apply path; the verify path stays free of network access, following the existing pattern of resolving values at the point of use.
- Fee resolution stays in one place in the `tokens` package; the FeeQuoter application is surfaced to versioned sequences through a bundle/blockchains/datastore entry point rather than each versioned sequence re-implementing fee discovery or reconstructing an environment.
- Fee aggregation defaults only apply to canonical chains, and the environment's deployer key is treated as the source of truth for the fee recipient, letting the upstream payload drop its own deployer-key injection.
- Remote-chain interface additions are made on `RemoteCCTPChain` so every chain family that participates in CCTP configuration carries the same capabilities.
- The registered-pool ref is conveyed by a "first deploy address" convention rather than a new `RemoteCCTPChain` interface method: pool identity is derived where the pool is actually deployed, trading explicit interface surface for an implicit positional contract (enforced defensively by the deploy sequences ordering the registered pool first).

## Companion change

Solana lanes reuse an existing `CCTPTokenPool`, so that ref must already exist in the CCV datastore (the Solana deploy is a no-op and cannot produce it). The CCIP→CCV import allowlist in `chainlink-deployments` did not include it; `CCTPTokenPool 1.6.0`, `TokenMultisig 1.6.0`, and `USDCToken 1.6.0` were added there so the existing import command can backfill it.


0xsuryansh (dismissed):

chains/evm/deployment/v2_0_0/adapters/cctp_chain.go
```diff
@@ -90,6 +92,22 @@ func (c *CCTPChainAdapter) USDCType() adapters.USDCType {
 	return adapters.Canonical
 }
 
+// TokenDecimals returns the number of decimals of the token at the given address on the chain.
+func (c *CCTPChainAdapter) TokenDecimals(bundle operations.Bundle, ds datastore.DataStore, chains chain.BlockChains, selector uint64, token string) (uint8, error) {
```
0xsuryansh: Could have used the hardcoded value here as defined as constant since this is for CCTP and their standard is 6 decimals

chains/evm/deployment/v2_0_0/sequences/cctp/deploy_cctp_chain.go
```diff
@@ -259,6 +266,18 @@ var DeployCCTPChain = cldf_ops.NewSequence(
 		addresses = append(addresses, usdcTokenPoolProxyRef)
 		usdcTokenPoolProxyAddress := common.HexToAddress(usdcTokenPoolProxyRef.Address)
 
+		// The pool registered on the TokenAdminRegistry differs by role: home chains register the
+		// USDCTokenPoolProxy (which routes to the underlying pools), non-home chains register the
+		// CCTP V1 pool directly.
+		if isHomeChain || datastore_utils.IsAddressRefEmpty(cctpV1PoolRef) {
```
0xsuryansh: Every "canonical" EVM chains. should have `USDCTokenPoolProxy` as the registered pool.

0xsuryansh: USDCTokenPoolProxy support all CCTP versions, therefore EVM chains should support CCTPv1 and v2 at the same time, which is only possible through proxy

0xsuryansh (approved):

tt-cll (approved):

PR #2319, opened 2026-09-18
LombardVerifier v2.2.0: Add support for padded optional-message representation

0xsuryansh:


chains/evm/contracts/ccvs/LombardVerifier.sol
```diff
@@ -257,16 +267,17 @@ contract LombardVerifier is BaseVerifier, Ownable2StepMsgSender {
       revert MustTransferTokens();
     }
 
+    if (ccvData.length < PAYLOAD_START_INDEX) {
```
0xsuryansh, line 270: moved the minimum-length check before any slicing

chains/evm/contracts/ccvs/LombardVerifier.sol
```diff
@@ -96,20 +97,29 @@ contract LombardVerifier is BaseVerifier, Ownable2StepMsgSender {
     address localAdapter;
   }
 
-  string public constant typeAndVersion = "LombardVerifier 2.1.0";
+  string public constant typeAndVersion = "LombardVerifier 2.2.0";
```
archseer, line 100: Note: we also need to update the constant in https://github.com/smartcontractkit/chainlink-ccv/blob/96ca5224edc7295cea6df673432c46df36321ae8/verifier/pkg/token/lombard/consts.go#L11

0xsuryansh: Don't think we should update it since EVM<>EVM will be processed by this default tag. We can specify the v2.2.0 tag in chainlink-ccv-deploy for solana config

jadepark-dev: It seems this needs to be updated for deployment as well: https://github.com/smartcontractkit/chainlink-ccip/blob/fa1a268dac47cb17b8f3c6b4ff0a891beb8cd420/chains/evm/deployment/v2_0_0/verifier_tags/tags.go#L7

0xsuryansh: https://github.com/smartcontractkit/chainlink-ccip/pull/2319/commits/c3311bc7e7352d515d41ec03be50993e3769092f

crispymangoes (approved):

chris-de-leon-cll (approved):

matYang (approved):
conditionally approved assuming there is follow up fix for stale 2.1.0 references at places like https://github.com/smartcontractkit/chainlink-ccip/blob/c3311bc7e73/chains/evm/deployment/v2_0_0/sequences/lombard/deploy_lombard_chain.go#L25

0xsuryansh:
> conditionally approved assuming there is follow up fix for stale 2.1.0 references at places like https://github.com/smartcontractkit/chainlink-ccip/blob/c3311bc7e73/chains/evm/deployment/v2_0_0/sequences/lombard/deploy_lombard_chain.go#L25

Discussed on chat, I wanted to keep this small for auditors, also Lombard deploys the contract by themselves
and this new version of verifier will be deployed only for solana during launch So I'll update the tooling accordingly for testing on staging

PR #2321, opened 2026-09-21
Add support for LockRelease 1.5.0

Alkhara:
https://smartcontract-it.atlassian.net/browse/CCIP-13664


chains/evm/deployment/operations_gen_config.yaml
```diff
@@ -51,6 +61,31 @@ contracts:
...
+      - name: getRebalancer
+        access: public
+      - name: setRebalancer
+        access: owner
+      - name: transferLiquidity
+        access: owner
+      - name: provideLiquidity
+        access: public
+      - name: withdrawLiquidity
+        access: public
```
chris-de-leon-cll: feel free to correct me if I'm wrong, but do the access level needs to be `owner` for `provideLiquidity` / `withdrawLiquidity`?

at the moment, I don't think it's an issue since I don't see those ops being used, but would recommend either dropping them or changing to `access: owner`

integration-tests/deployment/token_expansion_migration_test.go
```diff
@@ -633,7 +633,7 @@ func runAutoMigrateUpgrade(t *testing.T, oldPoolVersion *semver.Version, opts *a
 		// That is a real behavioural difference from v1.5.1+: there is no window in which the old and
 		// new pool A are both accepted by pool B, so messages already in flight from old pool A are
 		// rejected with InvalidSourcePoolAddress. Asserted here so the cutover is pinned, not implied.
-		oldPoolB, err := bmtpapOps.NewBurnMintTokenPoolAndProxyContract(s.oldPoolAddrB, chainB.Client)
+		oldPoolB, err := tpapOps.NewTokenPoolAndProxyContract(s.oldPoolAddrB, chainB.Client)
```
chris-de-leon-cll: might be worth it to add a small test case for LnR as well if it's not too much of a heavy lift

I think there was a dedicated `token_expansion_v1_5_0_test.go` that could be extended

chris-de-leon-cll (approved):

